PR #6, opened 2019-07-15
Add user manager dependency and LDAP conf dependency

ratagupt:
Now bmcweb have dependency on the user manager where
bmcweb creates the user role map(user and its associated privilege)
when it starts, suppose if user manager starts after the bmcweb
then bmcweb would not be having the roles for the user and the authorization
will fail.
This will remain forever untill we restart the bmcweb.

This commit fixes this behaviour.

In one of the CI machine I have observed the following behaviour.

Jul 12 17:29:41 : Started Start bmcweb server.
Jul 12 17:30:08 : Started Phosphor User Manager.

TestedBy: Stopped the user manager service and LDAP conf service.
          restart the bmcweb and make sure that user manager and LDAP conf
          starts before the bmcweb.

Signed-off-by: Ratan Gupta <ratagupt@linux.vnet.ibm.com>
Change-Id: Ief61ea2171545d32d554a64036719126a3b128f1

### DIFF
--- a/bmcweb.service.in
+++ b/bmcweb.service.in
@@ -4,6 +4,12 @@ Description=Start bmcweb server
 Wants=network.target
 After=network.target
 
+Wants=xyz.openbmc_project.User.Manager.service
+After=xyz.openbmc_project.User.Manager.service
+
+Wants=xyz.openbmc_project.Ldap.Config.service
+After=xyz.openbmc_project.Ldap.Config.service
+
 [Service]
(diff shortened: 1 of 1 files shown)


ratagupt:
Gerrit link for the above pull request is: https://gerrit.openbmc-project.xyz/c/openbmc/bmcweb/+/23516

gtmills:
Looks good to me.

PR #12, opened 2019-07-21
Cert1

devenrao:
pulled in 3 patchsets


gtmills:
@devenrao Why merging into master?

PR #13, opened 2019-07-23
Mine

devenrao:
1) RSA supports only 2048 key bit length
2) Challenge password need to be empty as bmcweb does not support challenge password
3) CSR timeout has been changed to 10 seconds
4) If resource not found through "ResourceNotFound" error

### DIFF
--- a/redfish-core/lib/certificate_service.hpp
+++ b/redfish-core/lib/certificate_service.hpp
@@ -234,6 +234,7 @@ class CertificateActionGenerateCSR : public Node
     void doPost(crow::Response &res, const crow::Request &req,
                 const std::vector<std::string> &params) override
     {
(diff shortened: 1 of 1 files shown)


geissonator:
For future commits, please create a more descriptive branch name. "Mine" doesn't do a whole lot for me :)

geissonator:
I also just added this step to our process:
```Put in the links to the upstream code within a comment in your PR (even if it's not a 1x1 match, point the reviewers of the PR to the code so they can check overall status of the commit)```
Please add the upstream gerrit commits for reference in future PR's.

geissonator:
Relevant commits appear to be:
https://gerrit.openbmc-project.xyz/c/openbmc/bmcweb/+/19421
https://gerrit.openbmc-project.xyz/c/openbmc/bmcweb/+/21650

PR #23, opened 2019-08-07
Redfish: Add power metrics for openpower

msbarth:
The average and maximum consumed power aggregation history is read and
stored every 30 seconds within an openpower interface per power supply.
This pulls the entire aggregation history for each metric, over all
power supplies, and calculates a total average and maximum.

The total average consumed power is the result from adding the average
of all aggregation history entries per power supply. This is presented
as the `AverageConsumedWatts` in Redfish.

The total maximum consumed power is the result from adding the maximum
value determined from the aggregation history per power supply over the
same aggregation period. This is presented as the `MaxConsumedWatts` in
Redfish.

The `IntervalInMin` Redfish attribute is a single interval(in minutes)
at which all the Redfish power metrics were collected. Therefore, this
interval is set to the highest number of 30 second entries found of
aggregation history up to a 10 minute interval. (The openpower
aggregation history is collected and stored at the same time every 30
seconds, so this interval calculation should not affect the values
presented in Redfish for average and maximum power consumed.)

Tested:
    Used BMCWEB_LOG_DEBUG to verify number of method calls
    Used BMCWEB_LOG_DEBUG to verify data returned

  "PowerControl": [
    {
      "@odata.id": "/redfish/v1/Chassis/chassis/Power#/PowerControl/0",
      "@odata.type": "#Power.v1_0_0.PowerControl",
      "MemberId": "0",
      "Name": "Chassis Power Control",
      "PowerConsumedWatts": 268.0,
      "PowerLimit": {
        "LimitInWatts": 600.0
      },
      "PowerMetrics": {
        "AverageConsumedWatts": 173,
        "IntervalInMin": 10,
        "MaxConsumedWatts": 179
      },
      "Status": {
        "Health": "OK",
        "State": "Enabled"
      }
    }
  ]

Signed-off-by: Matthew Barth <msbarth@us.ibm.com>
Change-Id: I23e88ef77999ff4da005e5430160f196a9574b46

### DIFF
--- a/redfish-core/lib/power.hpp
+++ b/redfish-core/lib/power.hpp
@@ -19,8 +19,199 @@
 #include "node.hpp"
 #include "sensors.hpp"
 
+#include <regex>
+
 namespace redfish
 {
+
+struct PowerMetrics : std::enable_shared_from_this<PowerMetrics>
+{
+    PowerMetrics(const std::shared_ptr<SensorsAsyncResp>& asyncResp) :
+        asyncResp(asyncResp)
+    {
+    }
+
+    ~PowerMetrics()
+    {
+        nlohmann::json& pc = asyncResp->res.jsonValue["PowerControl"];
+        if (pc.empty())
+        {
+            // Add mandatory properties for PowerControl
+            pc.push_back(
+                {{"@odata.type", "#Power.v1_0_0.PowerControl"},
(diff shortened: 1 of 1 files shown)


msbarth:
https://gerrit.openbmc-project.xyz/24002

PR #41, opened 2019-09-25
SW474357: Remove not working Server Identify LED

gtmills:
The IndicatorLED at /redfish/v1/Systems/system does not work and
has never worked.

The problem is it calls xyz.openbmc_project.LED.Controller.identify
which IBM does not use.

https://github.com/openbmc/bmcweb/blob/944ffaf9ccae2a084cd650370facfb3f76701186/redfish-core/lib/systems.hpp#L650

With OP940 closing in and this not being a needed Redfish Property,
remove for now.

https://github.com/ibm-openbmc/dev/issues/1202 tracks fixing
this. Will leave the defect open but defer out of OP940 if
this workaround is accepted.

Tested: Loaded bmcweb on a Witherspoon. System Resource looks
        good as do others.

### DIFF
--- a/redfish-core/lib/systems.hpp
+++ b/redfish-core/lib/systems.hpp
@@ -539,119 +539,6 @@ void getComputerSystem(std::shared_ptr<AsyncResp> aResp)
 }
 
 /**
- * @brief Retrieves identify led group properties over dbus
- *
- * @param[in] aResp     Shared pointer for generating response message.
- * @param[in] callback  Callback for process retrieved data.
- *
- * @return None.
- */
-template <typename CallbackFunc>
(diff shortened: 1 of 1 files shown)


vishwabmc:
There are couple issues here:

1/. The code is directly referencing the physical LED as opposed to Group. Need to see why
2/. IBM ID led is named : `front-id` and `rear-id` and not "identify" --> This is why it would never work

gtmills:
@vishwabmc This needs to be wrote in a way that works for both Intel and ourselves. 
A GetSubTree(xyz.openbmc_Project.Led.Group) call to the Mapper, find the "enclosure_identify", then a GetAll call to that service.

PR #43, opened 2019-09-27
Change to reset the password

asmithakarun:
The REST call in this commit is needed for backward compatibility.
In earlier webserver(phosphor-rest-server), this API was available,
but not implemented in bmcweb.

So, this commit implements this API in bmcweb that enables resetting
the root user password.

Tested By:

curl -k -H "X-Auth-Token: $bmc_token" -d '{"Password":"0penBmc123"}' -X PATCH https://${bmc}/xyz/openbmc_project/user/root/action/SetPassword

Signed-off-by: asmithakarun <asmitk01@in.ibm.com>
Change-Id: I149d2dd9446ade976e57d447060150b24fb09228

### DIFF
--- a/include/openbmc_dbus_rest.hpp
+++ b/include/openbmc_dbus_rest.hpp
@@ -2125,6 +2125,48 @@ template <typename... Middlewares> void requestRoutes(Crow<Middlewares...> &app)
                 handleDBusUrl(req, res, objectPath);
             });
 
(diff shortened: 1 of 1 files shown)


gtmills:
An operator should not be able to change the root password. 
         `.requires({"ConfigureComponents", "ConfigureManager"})`

ratagupt:
@gtmills : Configure Manager is for admins, It is aligned with other API.

PR #46, opened 2019-09-27
Allow cookie to pass XSRF TOKEN

gtmills:
https://github.com/openbmc/bmcweb/blob/master/include/token_authorization_middleware.hpp#L413
Adds the XSRF-TOKEN to the Cookie.

If XSRF TOKEN is not in the header look in the Cookie.

Curl PUT/PATCH/POST will now work with a cookie. E.g.
export bmc=xx.xx.xx.xx

curl -c cjar -b cjar -k -H "Content-Type: application/json" -X POST \
https://${bmc}/login -d "{\"data\": [ \"root\", \"0penBmc\" ] }"

curl -b cjar -k -H "Content-Type: application/json" -X PUT -d \
'{"data":"xyz.openbmc_project.State.Host.Transition.Reboot"}' \
https://${bmc}/xyz/openbmc_project/state/host0/attr/RequestedHostTransition

This is needed because OP940 has customers (XCAT) which require
the OP930 behavior of not passing the XSRF TOKEN as part of
the header. This is more secure than just not requiring
the XSRF TOKEN.

Tested:
curl -b cjar -k -H 'Content-Type: application/json' -X POST -d \
'{"data":[]}' https://$bmc/xyz/openbmc_project/logging/action/DeleteAll
{
  "data": null,
  "message": "200 OK",
  "status": "ok"

### DIFF
--- a/include/token_authorization_middleware.hpp
+++ b/include/token_authorization_middleware.hpp
@@ -209,6 +209,26 @@ class Middleware
         if (req.method() != "GET"_method)
         {
             std::string_view csrf = req.getHeaderValue("X-XSRF-TOKEN");
+
+            // If XSRF TOKEN was not passed in the header, look in the cookie
+            // for the XSRF TOKEN. Required for backwards compatibility.
+            if (csrf.empty())
+            {
+                auto startIndexXsrf = cookieValue.find("XSRF-TOKEN");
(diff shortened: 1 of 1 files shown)


gtmills:
This did not work for xcat.

PR #49, opened 2019-10-03
In the previous pull request, the password was being displayed in the logs. This has the changes to fix it.

asmithakarun:
(DO NOT MERGE)
Change to reset the password

The REST call in this commit is needed for backward compatibility.
In earlier webserver(phosphor-rest-server), this API was available,
but not implemented in bmcweb.

So, this commit implements this API in bmcweb that enables resetting
the root user password.

Tested By:

curl -k -H "X-Auth-Token: $bmc_token" -d '{"data":["0penBmc123"]}'
-X POST https://${bmc}/xyz/openbmc_project/user/root/action/SetPassword

Signed-off-by: asmithakarun <asmitk01@in.ibm.com>
Change-Id: I149d2dd9446ade976e57d447060150b24fb09228

### DIFF
--- a/include/logging_middleware.hpp
+++ b/include/logging_middleware.hpp
@@ -124,6 +124,8 @@ class Middleware
             // requests that contain passwords.
             if ((req.url ==
                  "/xyz/openbmc_project/user/ldap/action/CreateConfig") ||
+                (req.url ==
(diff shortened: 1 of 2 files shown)


ratagupt:
Seems Joseph has to make his change for the Expire password.

PR #54, opened 2019-11-06
Op940.01 expired password

joseph-reynolds:
The first commit (d9d7571) is in review here: https://gerrit.openbmc-project.xyz/c/openbmc/bmcweb/+/25146
This first set of changes was tested in an internal test driver.
Note that bmcweb master (file include/sessions.hpp) has diverged from the OP940 branch: In master, the userRole has been removed from the UserSession object and is looked up for every operation.  For OP940, the userRole is still present in the UserSession.

The second two commits are bug fixes for the SetPassword API and to allow basic auth (syntax like GET https://root:0penBMC@${BMC}/xyz/...) to proceed when the password is expired.  They have not been reviewed.

### DIFF
--- a/include/openbmc_dbus_rest.hpp
+++ b/include/openbmc_dbus_rest.hpp
@@ -1480,7 +1480,8 @@ void handleActionSetPassword(std::shared_ptr<InProgressActionData> transaction)
         return;
     }
 
-    if (!pamUpdatePassword("root", pwd))
+    bool gotPamAuthtokError = false;
+    if (!pamUpdatePassword("root", pwd, gotPamAuthtokError))
     {
         BMCWEB_LOG_ERROR << "pamUpdatePassword Failed";
         transaction->setErrorStatus("Password Not Modified");
@@ -2214,48 +2215,6 @@ template <typename... Middlewares> void requestRoutes(Crow<Middlewares...> &app)
                 }
                 handleDBusUrl(req, res, objectPath);
             });
-
-    BMCWEB_ROUTE(app, "/xyz/openbmc_project/user/root/action/SetPassword")
-        .requires({"ConfigureComponents", "ConfigureManager"})
-        .methods(
-            "PATCH"_method)([](const crow::Request &req, crow::Response &res) {
-            nlohmann::json data = nlohmann::json::parse(std::move(req.body));
-            const std::string *pwd;
-            for (const auto &item : data.items())
-            {
-                if (item.key() == "Password")
-                {
-                    pwd = item.value().get_ptr<const std::string *>();
-                }
-            }
-
-            if (!pwd || pwd->empty())
-            {
-                BMCWEB_LOG_ERROR << "Password Empty!";
-                res.jsonValue = {{"message", "400 Bad Request"},
-                                 {"status", "Error"},
-                                 {"data", "Password Not Updated"}};
-                res.end();
-                return;
-            }
-
-            if (!pamUpdatePassword("root", *pwd))
-            {
-                BMCWEB_LOG_ERROR << "pamUpdatePassword Failed";
-                res.jsonValue = {{"message", "500 Internal Server Error"},
-                                 {"status", "Error"},
-                                 {"data", "Password Not Updated"}};
-                res.end();
-                return;
-            }
-
-            BMCWEB_LOG_DEBUG << "pamUpdatePassword Successful";
-            res.jsonValue = {{"message", "204 Resource Updated Successfully"},
-                             {"status", "ok"},
-                             {"data", "Password Updated"}};
-            res.end();
-            return;
-        });
 
     BMCWEB_ROUTE(app, "/org/<path>")
         .requires({"Login"})
--- a/include/openbmc_dbus_rest.hpp
+++ b/include/openbmc_dbus_rest.hpp
@@ -2165,11 +2165,53 @@ template <typename... Middlewares> void requestRoutes(Crow<Middlewares...> &app)
         });
 
     BMCWEB_ROUTE(app, "/xyz/<path>")
-        .requires({"ConfigureComponents", "ConfigureManager"})
+        .requires({{"ConfigureComponents", "ConfigureManager"},
+                   {"ConfigureSelf"}})
         .methods("PUT"_method, "POST"_method, "DELETE"_method)(
             [](const crow::Request &req, crow::Response &res,
                const std::string &path) {
                 std::string objectPath = "/xyz/" + path;
+                // This is too permissive.  The ConfigureSelf privilege
+                // applies only to the root user POSTing a new password.
+                if ((req.method() != "POST"_method) ||
+                    (objectPath !=
+                         "/xyz/openbmc_project/user/root/action/SetPassword") ||
+                    (req.session->username != "root"))
+                {
+                    // Check privileges again without ConfigureSelf.
+                    // Hack alert!  This duplicates code from
+                    // isAllowedWithoutConfigureSelf which is not yet included.
+                    redfish::Privileges effectiveUserPrivileges;
+                    if ((req.session != nullptr) &&
+                        (!req.session->isConfigureSelfOnly))
+                    {
+                        const std::string& userRole =
+                            crow::persistent_data::UserRoleMap::getInstance().
+                                getUserRole(req.session->username);
+                        effectiveUserPrivileges =
+                            redfish::getUserPrivileges(userRole);
+                        effectiveUserPrivileges.
+                            resetSinglePrivilege("ConfigureSelf");
+                    }
+                    redfish::Privileges requiredPrivileges =
+                        {"ConfigureComponents", "ConfigureManager"};
+                    if (!effectiveUserPrivileges.
+                        isSupersetOf(requiredPrivileges))
+                    {
+                        std::string desc = crow::msg::forbiddenURIDesc;
+                        desc.append(req.url);
+                        crow::setErrorResponse(
+                            res, boost::beast::http::status::forbidden,
+                            crow::msg::forbiddenMsg, desc);
+                        if ((req.session != nullptr) &&
+                            (req.session->isConfigureSelfOnly))
+                        {
+                            crow::setPasswordChangeRequired(res, "root");
+                        }
+                        res.end();
+                        return;
+                    }
+                }
                 handleDBusUrl(req, res, objectPath);
             });
 
--- a/include/openbmc_dbus_rest.hpp
+++ b/include/openbmc_dbus_rest.hpp
@@ -23,42 +23,17 @@
 #include <dbus_utility.hpp>
 #include <filesystem>
 #include <fstream>
+#include <json_msg_utility.hpp>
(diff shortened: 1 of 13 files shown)


joseph-reynolds:
Here is a handy file-by-file review guide:

crow/routing.hpp:
Set userPrivileges to ConfigureSelf if the session.isConfigureSelfOnly flag is set.
If failing with http status 403 forbidden, give an extra message if the session authority is limited to CondfigureSelf only.  The message is a JSON Field in the http response, and its content varies by API.

include/json_msg_utility.hpp
Moved out of obmc_dbus_rest.hpp so it can be used by routing.hpp.
Added setPasswordChangeRequired().

include/obmc_dbus_rest.hpp:
Moved setErrorResponse() to separate header file so it can be used by routing.hpp/
NOTE: function handleActionSetPassword() is in base OP940 
Added special authority checking for URI=/xyz/openbmc_project/user/root/action/SetPassword so it can invoke the SetPassword API when isConfigureSelfOnly.

include/pam_authenticate.hpp:
Articulate the PAM_NEW_AUTHTOK_REQD error as passwordChangeRequired.

include/sessions.hpp:
Added new bool UserSession.isConfigureSelfOnly.  When true, this session should effectively have only the ConfigureSelf Privilege.

include/token_authorization_middleware.hpp:
Enhance URI=/login and basic auth to set session.isConfigureSelfOnly appropriately.  For /login, when isConfigureSelfOnly==true, also add a new JSON field in the http response body (crow::setPasswordChangeRequired).  

redfish-core/include/error_messages.hpp:
Add new Redfish messages.

redfish-core/include/node.hpp:
Add new authority checking function (isAllowedWithoutConfigureSelf) need to implement operation-to-privilege overrides.

redfish-core/include/privileges.hpp:
Break out function isOperationAllowedWithPrivileges() so it can be used by node.hpp.

redfish-core/lib/account_service.hpp:
Implement privilege overrides related to ConfigureSelf.

redfish-core/lib/redfish_sessions.hpp:
Implement privilege overrides related to ConfigureSelf.
Add the passwordChangeRequired message to the Session object when appropriate.  Note that this message is also sent in routing.hpp when the authority check fails with 403 forbidden.  This is a different use case: The login succeeded but the resulting session has limited authority, so this message is sent to let the client know that fact.

redfish-core/src/error_messages.cpp:
Add new Redfish messages.

joseph-reynolds:
The fourth commit (better password change failure message) enhances the error message returned from PATCH Password into /redfish/v1/AccountService/Accounts/{account} ;
Why is this needed? This returns HTTP status code 400 (was 500) so the webui can give a better response.

include/openbmc_dbus_rest.hpp
```diff
@@ -2215,48 +2216,6 @@ template <typename... Middlewares> void requestRoutes(Crow<Middlewares...> &app)
                 handleDBusUrl(req, res, objectPath);
             });
 
-    BMCWEB_ROUTE(app, "/xyz/openbmc_project/user/root/action/SetPassword")
```
gtmills: Doesn't xcat need this?

gtmills:
Can we push the other 3 upstream?

include/openbmc_dbus_rest.hpp
joseph-reynolds: Yes, it does.  This is dead code, meaning there is no way for control to flow into it.  I explained in the commit message that this API is handled by a special case in the route handler for /xyz/<path>.  That was added in base OP940 and remains enabled with this set of changes (with an accompanying the authority fix in this PR).

joseph-reynolds:
I push another commit to fix a problem reported in the gerrit review:

This fixes a bug where if a session was created with the PasswordChangeRequired handling, and then bmcweb was restarted, the recovered session would lose its special PasswordChangeRequired restrictions.

Tested: yes, sessions with PasswordChangeRequired continue to have that message and accompanying restictions after bmcweb restarts.

include/openbmc_dbus_rest.hpp
ratagupt: > This is dead code, meaning there is no way for control to flow into it

this API is handled by a special case in the route handler for /xyz/, Do you mean to say that we don't need this?
Why? I am missing something.

include/sessions.hpp
```diff
@@ -362,6 +362,10 @@ struct UserSession
             {
                 userSession->userRole = *thisValue;
             }
+            else if (element.key() == "is_configure_self_only")
+            {
+                userSession->isConfigureSelfOnly = (*thisValue == "true");
```
ratagupt: what is the default value for userSession->isConfigureSelfOnly? I have not seen the code for initializing the default value.

redfish-core/lib/account_service.hpp
```diff
@@ -1638,10 +1639,22 @@ class ManagerAccount : public Node
...
+            bool gotPamAuthtokError = false;
+            if (!pamUpdatePassword(username, *password, gotPamAuthtokError))
             {
                 BMCWEB_LOG_ERROR << "pamUpdatePassword Failed";
-                messages::internalError(asyncResp->res);
+                if (gotPamAuthtokError)
+                {
+                    messages::accountNotModified(asyncResp->res);
+                    // See https://redfishforum.com/thread/246/message-send-patch-password-failure
+                    asyncResp->res.jsonValue["error"]["Oem"]["IBM"] = "The "
```
ratagupt: Did we make any changes in the schema(error_registry) to have error/Oem.IBM?

include/sessions.hpp
joseph-reynolds: The route /xyz/openbmc_project/user/root/action/SetPassword is handled by the /xyz/<path> handler.  Here is how I verified that: I added debug statements to both handlers, and found the /xyz/<path> was the only handler being invoked.  I tried changing the order these routes were declared in, that is, moving /xyz/openbmc_project/user/root/action/SetPassword ahead of /xyz/<path> in openbmc_dbus_rest.hpp; that had no effect.  Then finally I deleted the /xyz/openbmc_project/user/root/action/SetPassword handler and tested again.  The SetPassword API works correctly.

Here is the control flow when a client invokes POST /xyz/openbmc_project/user/root/action/SetPassword:
- Handler for BMCWEB_ROUTE(app, "/xyz/<path>").methods(..., "POST"_method, ...) gets control
- This handler performs a custom privilege check which was added by this PR.
- calls handleDBusUrl()
- calls handleAction()
- if (methodName == "SetPassword") calls handleActionSetPassword()
which was previously added to OP940.00 (the base release): https://github.com/ibm-openbmc/bmcweb/blame/OP940/include/openbmc_dbus_rest.hpp#L1493

joseph-reynolds: Field UserSession.isConfigureSelfOnly gets its value when the UserSession is constructed.  there are two ways a UserSession is constructed.

1. UserSesion objects are created and added to the SessionStore by SessionStore.generateUserSession() which requires the caller to pass in the configureSelfOnly argument.  There are 3 such call sites:
- include/token_authorization_middleware.hpp - performBasicAuth()
- include/token_authorization_middleware.hpp - handler for POST /login
- redfish-core/lib/redfish_sessions.hpp - SessionCollection.doPost (the handler for /redfish/v1/SessionService/Sessions/)
In each case, the caller has gotten the correct information from pamAuthenticateUser to know if the password is expired.  That value is passed in as the configureSelfOnly argument which then becomes the value of the isConfigureSelfOnly field.

2. UserSession objects are created and added to the SessionStore by UserSession::fromJson().  The value is set if the JSON is_configure_self_only field is present.
The UserSession serializer is in include/sessions.hpp - adl_serializer::to_json.  BMCWeb beginning with OP940.01 will serialize the isConfigureSelfOnly field as the is_configure_self_only member.  Previous versions will not have this field.
If the is_configure_self_only field is not present, UserSession.isConfigureSelfOnly will be created with an indeterminate value.  Although it is likely nobody would be using stored OP940.00 session after an upgrade to OP940.01, I'll change the code to default the value to false. <-- bug fix.

redfish-core/lib/account_service.hpp
joseph-reynolds: I am unaware of any schema(error_registry).  I 'm not sure what you are asking about.

I see functions to add the AccountNotModified message ... but the changes in this PR are to add a *optional* Oem.IBM field to the AccountNotModified message, so I'm not sure where they would go, so I just put them into the code.  Is that what you were asking about?

I asked the Redfish forum the right way to do this here: https://redfishforum.com/thread/246/message-send-patch-password-failure

joseph-reynolds:
I pushed a new commit with the bug fix that defaults isConfigureSelfOnly=false when the JSON is_configure_self_only field is missing.   The commit message has the details.

include/openbmc_dbus_rest.hpp
ratagupt: @joseph-reynolds I got it, It seems we don't need this.

include/sessions.hpp
ratagupt: >  I'll change the code to default the value to false. <-- bug fix.
Thanks for the detailed explanation, so you will fix the userSession->isConfigureSelfOnly to false.

include/openbmc_dbus_rest.hpp
ratagupt: It would have been good if we have different commit for reverting some PR.

redfish-core/lib/account_service.hpp
ratagupt: Each standard message is having the registry schema, Here we are adding the oem.
So we need to change the schema. https://redfish.dmtf.org/registries/Base.1.5.0.json

ratagupt:
@joseph-reynolds https://github.com/ibm-openbmc/bmcweb/pull/54/commits/d9d757155c37849aa9d351b44d5c0e3be5ed0421 I will re-review this commit on Gerrit.

rfrandse:
Joseph P. Reynolds  2 minutes ago
It looks like Ratan’s main issues have been addressed.  There may be a few nits.

Joseph P. Reynolds  2 minutes ago
I would be comfortable merging this PR and building a driver.

Joseph P. Reynolds  2 minutes ago
We can add fixes later.

Merging based on these comments.

redfish-core/lib/account_service.hpp
joseph-reynolds: The way I understand the "Oem" property is this:
According to Redfish document DSP2046 (https://www.dmtf.org/sites/default/files/standards/documents/DSP2046_2019.1_0.pdf) the "Oem" property is already defined for all Redfish schemas.  Therefore, we can use it without having to update any Message Registry schemas.

joseph-reynolds: It might be better to change this code to use the AccountNotModified message Resolution field (see DSP2046 Message https://www.dmtf.org/sites/default/files/standards/documents/DSP2046_2019.2.pdf).  The resolution text could be something like:
PATCH the Password property again with a value that meets the password requirements.

joseph-reynolds:
I'll add any remaining fixes as new PRs.

PR #55, opened 2019-12-06
bmcweb: better message for SetPassword failure

joseph-reynolds:
Enhances API /xyz/openbmc_project/user/root/action/SetPassword to
give a better error message when the new password is not accepted.

Fixes SW482129


joseph-reynolds:
This PR should be applied to the tip of the OP940 branches.  Ideally beginning with OP940.01.

joseph-reynolds:
The changes here are not going into OP940.01.   They are included in PR57 for OP940 which is the base for OP940.10.  Closing this PR.

PR #59, opened 2020-02-04
Redfish: Make PATCH NTP Servers operate correctly

raviteja-b:
The Managers NTP node did not get implemented correctly.  The first
level parsing did not search for "NTP" as the base entry, nor did it
correctly attempt to acquire the NTPServers, NTPEnabled sub-nodes.

Removed the data response as it was returning a string vector instead
of a json::array. The original code caused a seg fault. Instead of
returning a carbon copy of the values sent, the return is a 204 code,
indicating success and returning no data.

Change-Id: Ifb82855e1cd143e3cf2cb8979531b01b27d32234
Signed-off-by: Johnathan Mantey <johnathanx.mantey@intel.com>

### DIFF
--- a/redfish-core/lib/network_protocol.hpp
+++ b/redfish-core/lib/network_protocol.hpp
@@ -353,7 +353,6 @@ class NetworkProtocol : public Node
                     messages::internalError(asyncResp->res);
                     return;
                 }
-                messages::success(asyncResp->res);
             },
             "xyz.openbmc_project.Network",
             "/xyz/openbmc_project/network/config",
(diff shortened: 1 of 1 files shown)


gtmills:
@raviteja-b Can you add the defect number and what defect this is fixing (i.e. why we should pull this into OP940)?

gtmills:
https://gerrit.openbmc-project.xyz/c/openbmc/bmcweb/+/24855 is the upstream

ratagupt:
As this commit is already in master, we should pull this in OP940 also, This is correct implementation for patching the NTP.

gtmills:
@raviteja-b @ratagupt Is there a defect number for this?

ratagupt:
No, There was no defect number for this.

asmithakarun:
[VMI IP Deleted Event not sent - SW515334](https://trello.com/c/7WLksOsV/1-vmi-ip-deleted-event-not-sent-sw515334)

PR #60, opened 2020-08-07
Allow certificate upload for "not yet valid" case

ojayanth:
Currently bmcweb rejects the certificates  with  validity
starts after the current  time also replaces with self signed
certificate. This change will allow to install the certificate
with "notBefore" date after current time.

### DIFF
--- a/include/ssl_key_handler.hpp
+++ b/include/ssl_key_handler.hpp
@@ -102,9 +102,11 @@ inline bool validateCertificate(const std::string &filePath,
         unsigned int retErrCode = X509_STORE_CTX_get_error(storeCtx);
         X509_STORE_CTX_free(storeCtx);
         X509_LOOKUP_free(x509Lookup);
-        if (isTrustChainError(retErrCode))
+        if (isTrustChainError(retErrCode) ||
+           (retErrCode == X509_V_ERR_CERT_NOT_YET_VALID))
         {
-            BMCWEB_LOG_DEBUG << "Ignoring Trust Chain error";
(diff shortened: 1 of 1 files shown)


ojayanth:
Not ready for merge.

rahulmah:
Not valid certificate is getting uploaded and I can see warning in browser for the not yet certificate after uplaod. So fix is working fine.  All certificate automated certificate tests are also passing with this build.

include/ssl_key_handler.hpp
```diff
@@ -102,9 +102,11 @@ inline bool validateCertificate(const std::string &filePath,
         unsigned int retErrCode = X509_STORE_CTX_get_error(storeCtx);
         X509_STORE_CTX_free(storeCtx);
         X509_LOOKUP_free(x509Lookup);
-        if (isTrustChainError(retErrCode))
+        if (isTrustChainError(retErrCode) ||
+           (retErrCode == X509_V_ERR_CERT_NOT_YET_VALID))
```
gtmills, line 106: Are we going to push this upsteam ?
https://github.com/openbmc/bmcweb/blob/c94d37490b0090a2a4a0b6bad51cd8ab1339051e/include/ssl_key_handler.hpp#L82

ojayanth:
@rahulmah test shows latest BMCweb doesn't complain this.  940 and Master looks like lot of chnages, need to investigate why the validation part is skipping in BMCweb side , need some expert look from BMCweb front.

include/ssl_key_handler.hpp
gtmills: Master is fixed with a different solution

PR #64, opened 2021-08-05
Revert "Add concurrent KVM sessions support"

LuluTHSu:
This reverts commit 73926aebe40bc5d126ffec9ca34844322a5669b9.

Reason for revert: IBM decided to not support multi-session of KVM on
OP940.

Signed-off-by: LuluTHSu <Lulu_Su@wistron.com>

### DIFF
--- a/include/kvm_websocket.hpp
+++ b/include/kvm_websocket.hpp
@@ -4,6 +4,7 @@
 #include <sys/socket.h>
 
 #include <boost/container/flat_map.hpp>
+#include <boost/container/flat_set.hpp>
 #include <webserver_common.hpp>
 
 namespace crow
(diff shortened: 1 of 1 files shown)


LuluTHSu:
Summary: As OP940 no longer supports KVM multi-session, only one KVM session is allowed.

When user2 tries to use KVM, cannot connect (reasonably),
**but the connection of user1's KVM will be disconnected (as shown below).
Is this situation allowed/reasonable?**

user1's KVM webpage::
image
image

anoo1 (approved):

anoo1 (changes requested):
This would affect all platforms, including witherspoon. It probably should be done as a patch file to openbmc/openbmc so that the patch file can be appended to mowgli only.

anoo1:
> This would affect all platforms, including witherspoon. It probably should be done as a patch file to openbmc/openbmc so that the patch file can be appended to mowgli only.

To create the patch file, from your bmcweb repo where this commit was created:

`git format-patch -1 8ed337d4d1b1911606b9d42eb462607bc2f4cf93`

That command will create a file with a filename similar to this:

`0001-Revert-Add-concurrent-KVM-sessions-support.patch`

Copy that patch to your openbmc/openbmc repo:

create this directory:
`meta-ibm/recipes-phosphor/interfaces/bmcweb/`

and copy the file to that new directory.

Modify the bbappend file `meta-ibm/recipes-phosphor/interfaces/bmcweb_%.bbappend` with the following (ensure the file name matches the name of the patch file):

```
FILESEXTRAPATHS_prepend_mowgli := "${THISDIR}/${PN}:"
SRC_URI_append_mowgli= " file://0001-Revert-Add-concurrent-KVM-sessions-support.patch
```

Build and ensure the patch is applied to bmcweb.

LuluTHSu:
I have followed your suggestion to push it to [openbmc](https://github.com/ibm-openbmc/openbmc/pull/160).
Also build and make sure to apply the patch to bmcweb.

anoo1:
> Do you have any suggestions for modifying this condition? thank you very much.

Don't know much about KVM, maybe @eddiejames or @gtmills could help.

gtmills:
That bug above scares me. 
The revert itself looks fine to except as @anoo1 this goes into Witherspoon as well.

gtmills:
Since we have https://github.com/ibm-openbmc/openbmc/pull/160/files I think we can abandon?

LuluTHSu:
> That bug above scares me.
> The revert itself looks fine to me except as @anoo1 this goes into Witherspoon as well.

Agree, will use bbappend to add this revert patch to [openbmc](https://github.com/ibm-openbmc/openbmc/pull/160/files).

> Since we have https://github.com/ibm-openbmc/openbmc/pull/160/files I think we can abandon?

No, the above bug is a description of what happens when it's revert to bmcweb only allows one KVM session.
So this should be fixed, otherwise the same situation will happen in mowgli...

clairefish:
Hi @gtmills @Erich-Hauptli

We’re still wondering if the 4 sessions solution could be kept in the bmcweb patch on Mowgli then just decrease the maxsession to 1 only with the warning message.

Advantage:
1.  This method is aligned to op940 all projects, it’s good for maintain.
2. This is as new as the master code, it will be easier to get any fix/update soon.
3. No refresh/F5 issue and bmc crash issue in this new structure.
4. We tested/verified more after found bmc loading and bmc crash issue that based on 4 sessions solution, it will has lower risk.

That’s our suggestions for your reference, we think this method has lower risk and the refresh side effect can be fixed soon.

If you still prefer to use the revert version (1 session base solution), the refresh/F5 issue may need your support.
May I know what's your considerations? 
Any suggestion for refresh/F5 problem?
Not sure if this is working as design.....

Erich-Hauptli:
Claire,

I want to proceed with the single session solution and work to refine the handoff of the session between users (F5/Refresh issue) in the next service pack.  We saw performance issues with the multi session setup, thus I don't view that as a viable solution.  I also don't feel that prior systems used KVM enough to declare this was functional previously.  I think we had bugs that we're only now exposing with the significant use that the HMC system uses the KVM interface.  My expectation would be that MiHawk or Witherspoon would see these issues if we put them under similar stress.

gtmills:
> We’re still wondering if the 4 sessions solution could be kept in the bmcweb patch on Mowgli then just decrease the maxsession to 1 only with the warning message.

I would suggest the current solution since we ran with it in op940 / upstream for a long time. 
We only recently pulled the 4 session solution into op940. https://github.com/ibm-openbmc/bmcweb/commit/73926aebe40bc5d126ffec9ca34844322a5669b9

>  maxsession to 1 only with the warning message

Is something that has never been tested.

clairefish:
Hi Gunnar,

The 4 sessions solution was uploaded to the master branch on Sep 24, 2019
https://gerrit.openbmc-project.xyz/c/openbmc/bmcweb/+/25231/4#message-58dca8cf963b34852c41f1fe44614e24ad8fec2b

We tested 2 code bases solutions yesterday and sent the result by mail.
You could check the mail for final result.
Thank you very much.

LuluTHSu:
Hi @anoo1 ,
Thanks for your help and suggestions.

Hi @Erich-Hauptli , @gtmills ,
After we analyzed and discussed the pros and cons with 2 code bases, glad we finally reached a consensus to use the code base that based on multi-session and limited the maxSessions number to 1. 
The [patch](https://github.com/ibm-openbmc/openbmc/pull/170) has been republished, thanks for all your support.

PR #68, opened 2021-08-16
Fix error code for SNMP subscription

sunharis:
This commit fixes defect SW531983

Tested by:
Create SNMP subscriber with invalid IP address
Client should receive 400 response code

Change-Id: I373a467ddf8e9996204d707cdf1fdd752dfa7e8c

### DIFF
--- a/redfish-core/lib/event_service.hpp
+++ b/redfish-core/lib/event_service.hpp
@@ -191,7 +191,8 @@ inline void
                     const std::string& dbusSNMPid) {
             if (ec)
             {
-                messages::internalError(asyncResp->res);
+                BMCWEB_LOG_ERROR << "Create SNMP subscriber failed. " << ec;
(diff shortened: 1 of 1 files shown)


geissonator:
add to approvelist

gtmills:
But this covers any rc with xyz.openbmc_project.Network.SNMP as Redfish invalidObject... How is this any better? 
I think we need to have xyz.openbmc_project.Network.SNMP return different return codes depending on the fail.

sunharis:
Not merged - needs changes at  xyz.openbmc_project.Network.SNMP - which is of less priority now

sunharis:
@gtmills Created https://github.com/ibm-openbmc/dev/issues/3389 to implement the error codes at the smp repo. Can we get this in for the GA1 - so that we can at least avoid getting 500 internal server error?

PR #70, opened 2021-08-17
Restrict patching of read-only bios attributes

asmithakarun:
Since the P10 system has BMC to manage the BIOS attributes,
and bmcweb does not know the attribute values of the BIOS,
it should allow the host to be able to set BIOS readOnly
attributes, while the redfish-doPATCH operation should not
be allowed.

With this change, when a redfish command to update a bios
attribute (which is read-only), is fired, bmcweb will send
the error back to the client.

Tested By:

Updation of read-only attr:

-X PATCH https://${bmc}/redfish/v1/Systems/system/Bios/Settings -d '{"Attributes":{"hb_debug_console":"Disabled"}}'
{
  "hb_debug_console@Message.ExtendedInfo": [
    {
      "Message": "The property hb_debug_console is a read only property and cannot be assigned a value.",
      "MessageArgs": [
        "hb_debug_console"
      ],
      "MessageId": "Base.1.8.1.PropertyNotWritable",
      "MessageSeverity": "Warning",
      "Resolution": "Remove the property from the request body and resubmit the request if the operation failed."
    }
  ]
}

### DIFF
--- a/redfish-core/lib/bios.hpp
+++ b/redfish-core/lib/bios.hpp
@@ -407,7 +407,8 @@ inline void requestRoutesBiosSettings(App& app)
                         return;
                     }
 
-                    boost::container::flat_map<std::string, std::string>
+                    boost::container::flat_map<std::string,
+                                               std::pair<bool, std::string>>
                         biosAttrsType;
                     const BiosBaseTableType* baseBiosTable =
                         std::get_if<BiosBaseTableType>(&retBiosTable);
@@ -423,7 +424,9 @@ inline void requestRoutesBiosSettings(App& app)
                     {
                         biosAttrsType.try_emplace(
                             item.first,
-                            std::get<biosBaseAttrType>(item.second));
+                            std::make_pair(
+                                std::get<biosBaseReadonlyStatus>(item.second),
+                                std::get<biosBaseAttrType>(item.second)));
                     }
 
                     PendingAttributesType pendingAttributes;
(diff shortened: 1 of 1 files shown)


geissonator:
add to approvelist

redfish-core/lib/bios.hpp
```diff
@@ -445,7 +448,16 @@ inline void requestRoutesBiosSettings(App& app)
                             return;
                         }
 
-                        std::string biosAttrType = (*it).second;
+                        bool biosAttrReadOnlyStatus = ((*it).second).first;
+                        if (biosAttrReadOnlyStatus == true)
+                        {
+                            BMCWEB_LOG_ERROR
+                                << "Attribute Type is ReadOnly. Patch failed!";
+                            messages::internalError(asyncResp->res);
```
gtmills, line 456: I don't think this should be internal Error, how about PropertyNotWritable ?
https://redfish.dmtf.org/registries/Base.1.10.0.json

sunharis (changes requested):
Please return the error : PropertyNotWritable

redfish-core/lib/bios.hpp
asmithakarun: done

gtmills:
Upstream for this?

asmithakarun:
> Please return the error : PropertyNotWritable

Done

geissonator:
CI fail is due to clang format:
```
-                            messages::propertyNotWritable(asyncResp->res, attrName);
+                            messages::propertyNotWritable(asyncResp->res,
+                                                          attrName);
```

gtmills (changes requested):
CI fail

gtmills:
Upstream for this?

gtmills (approved):
Seems good to me. Can we get this upstream?

asmithakarun:
This is the upstream change:
https://gerrit.openbmc-project.xyz/c/openbmc/bmcweb/+/38702

I will push this change to the upstream Gerrit commit.
Seems the parent commits doesn't contain the changes for node class migration..

sunharis (approved):

PR #76, opened 2021-08-27
Added new LogServices for Hardware Isolation (aka Guard) feature

RameshIyyar:
- The patches which are added for HardwareIsolation service will be downstream for now since we did not model all subunits (which can be isolated by system applications in error path) into Redfish so the community won't agree to push into upstream so pushing in GHE.

Reference:
- Design document:
https://github.com/openbmc/docs/blob/master/designs/guard-on-bmc.md

- DBus interfaces:
https://gerrit.openbmc-project.xyz/q/topic:%2522hardware-isolation%2522+(status:open+AND+-is:wip+OR+status:merged+AND+project:openbmc/phosphor-dbus-interfaces)


redfish-core/lib/log_services.hpp
```diff
@@ -1085,6 +1116,40 @@ inline void requestRoutesSystemLogServiceCollection(App& app)
...
+                                return;
+                            }
+                        }
+                    },
+                    "xyz.openbmc_project.ObjectMapper",
+                    "/xyz/openbmc_project/object_mapper",
+                    "xyz.openbmc_project.ObjectMapper", "GetSubTreePaths", "/",
+                    0,
+                    std::array<const char*, 1>{
+                        "xyz.openbmc_project.HardwareIsolation.Create"});
```
gtmills, line 1152: Seems like an odd interface for a log service? 
xyz.openbmc_project.HardwareIsolation.Log 
xyz.openbmc_project.HardwareIsolation.Event

RameshIyyar: I think, we need to return `"/redfish/v1/Systems/system/LogServices/HardwareIsolation"` in the LogServiceCollection when query like `curl -k -H "X-Auth-Token: $bmc_token" -X GET \
https://${bmc}/redfish/v1/Systems/system/LogServices`  based on the backend support so using the **xyz.openbmc_project.HardwareIsolation.Create** (hosted in the HardwareIsolation root D-Bus path) to decide whether the backend support is present or not since without Create API we cannot create the guard record in the D-Bus. 

The one you mentioned needs to add in D-Bus YAML to host in the HardwareIsolation root D-Bus path but, I am seeing any value to just decide the backend support for the bmcweb use case when we have other D-Bus interfaces.

RameshIyyar:
jenkins run tests please

RameshIyyar:
jenkins run tests please

geissonator:
add to approvelist

redfish-core/lib/log_services.hpp
```diff
@@ -3411,6 +3436,610 @@ inline void getSystemHardwareIsolationLogService(
...
+                                     << dbusObjPath.str;
+                    return;
+                }
+
+                std::string uriIdObj;
+                for (auto& object : objects)
+                {
+                    if (dbusObjPath.str.find(object) == std::string::npos)
+                    {
+                        continue;
```
sunharis, line 3552: Instead of using continue, you can have the != check and break once the object is found

redfish-core/lib/log_services.hpp
```diff
@@ -3411,6 +3436,610 @@ inline void getSystemHardwareIsolationLogService(
...
+                << redfishUri
+                << " of the given DBus object path: " << dbusObjPath.str;
+            return;
+        }
+
+        crow::connections::systemBus->async_method_call(
+            [asyncResp, entryJsonIdx, dbusObjPath, uriIdPos, uriIdPattern,
+             parentRedfishUri,
+             redfishUri](const boost::system::error_code ec,
+                         const std::vector<std::string>& objects) mutable {
```
sunharis, line 3536: what is mutable here?

redfish-core/lib/log_services.hpp
```diff
@@ -3411,6 +3436,610 @@ inline void getSystemHardwareIsolationLogService(
...
+        auto parentRedfishUri = redfishUri.substr(0, uriIdPos - 1);
+        auto parentRedfishUriIt =
+            std::find_if(redfishUriList.begin(), redfishUriList.end(),
+                         [parentRedfishUri](const auto& ele) {
+                             return parentRedfishUri == std::get<1>(ele);
+                         });
+
+        if (parentRedfishUriIt == redfishUriList.end())
+        {
+            BMCWEB_LOG_ERROR
```
sunharis, line 3524: This is not actually a failure right ? When the url list is over, this is used to end the recursive calls

redfish-core/lib/log_services.hpp
```diff
@@ -3411,6 +3436,610 @@ inline void getSystemHardwareIsolationLogService(
...
+/**
+ * @brief API used to fill the respective redfish uri (which will get by using
+ *        the given dbus object path) into "OriginOfCondition" property of
+ *        LogEntry schema.
+ *
+ * @param[in] asyncResp - The redfish response to return.
+ * @param[in] entryJsonIdx - The json entry index to add isolated hardware
+ *                            details in the appropriate entry json object.
+ * @param[in] dbusObjPath - The DBus object path which represents redfishUri.
+ * @param[in] redfishUri - The redfish uri to fill with its "Id".
```
sunharis, line 3605: redfishUri - The redfish uri to fill with its "Id". ===> this seems to be an iterator

redfish-core/lib/log_services.hpp
```diff
@@ -3411,6 +3436,610 @@ inline void getSystemHardwareIsolationLogService(
...
+        BMCWEB_LOG_ERROR
+            << "Failed to get Redfish URI for the given DBus object path: "
+            << dbusObjPath.str;
+        return;
+    }
+
+    crow::connections::systemBus->async_method_call(
+        [asyncResp, entryJsonIdx, dbusObjPath,
+         redfishUriIt](const boost::system::error_code ec,
+                       const GetObjectType& objType) mutable {
```
sunharis, line 3632: what is mutable here?

redfish-core/lib/log_services.hpp
```diff
@@ -4040,6 +4046,114 @@ inline void getSystemHardwareIsolationLogEntryCollection(
...
+                << "] when tried to get the HardwareIsolation managed objects";
+            messages::internalError(asyncResp->res);
+            return;
+        }
+
+        bool entryIsPresent = false;
+        for (auto dbusObjIt = mgtObjs.begin(); dbusObjIt != mgtObjs.end();
+             dbusObjIt++)
+        {
+            if (dbusObjIt->first != entryObjPath)
```
sunharis, line 4088: Its good if you use == operator and avoid using 'continue'. This comment applies to many other places in this PR

redfish-core/lib/log_services.hpp
```diff
@@ -4040,6 +4046,114 @@ inline void getSystemHardwareIsolationLogEntryCollection(
...
+                                    const boost::system::error_code ec,
+                                    const GetObjectType& objType) {
+        if (ec || objType.empty())
+        {
+            BMCWEB_LOG_ERROR << "DBUS response error [" << ec.value() << " : "
+                             << ec.message()
+                             << "] when tried to get the HardwareIsolation "
+                                "dbus name the given object path: "
+                             << entryObjPath.str;
+            if (ec.value() == EBADR)
```
sunharis, line 4116: What is EBADR ?

redfish-core/lib/log_services.hpp
```diff
@@ -4040,6 +4046,114 @@ inline void getSystemHardwareIsolationLogEntryCollection(
...
+
+        if (objType.size() > 1)
+        {
+            BMCWEB_LOG_ERROR << "More than one dbus service implemented "
+                                "the HardwareIsolation service";
+            messages::internalError(asyncResp->res);
+            return;
+        }
+
+        if (objType[0].first.empty())
```
sunharis, line 4135: This seems duplicate of check at line 4109

redfish-core/lib/log_services.hpp
```diff
@@ -4154,6 +4154,85 @@ inline void getSystemHardwareIsolationLogEntryById(
...
+ * @brief API Used to deisolate the given HardwareIsolation entry.
+ *
+ * @param[in] req - The HardwareIsolation redfish request (unused now).
+ * @param[in] asyncResp - The redfish response to return.
+ * @param[in] entryId - The entry id of HardwareIsolation entries to deisolate.
+ *
+ * @return The redfish response in the given buffer.
+ */
+inline void deleteSystemHardwareIsolationLogEntryById(
+    const crow::Request& /* req */,
```
sunharis, line 4167: what is the need to pass this 'req' if its not used ? Similar comment in multiple places

redfish-core/lib/log_services.hpp
```diff
@@ -4154,6 +4154,85 @@ inline void getSystemHardwareIsolationLogEntryById(
...
+
+            if (objType.size() > 1)
+            {
+                BMCWEB_LOG_ERROR << "More than one dbus service implemented "
+                                    "the HardwareIsolation service";
+                messages::internalError(asyncResp->res);
+                return;
+            }
+
+            if (objType[0].first.empty())
```
sunharis, line 4206: duplicate check?
Its in many places. Please check and remove

redfish-core/lib/log_services.hpp
```diff
@@ -4154,6 +4154,85 @@ inline void getSystemHardwareIsolationLogEntryById(
...
+                                               entryId);
+                }
+                else
+                {
+                    messages::internalError(asyncResp->res);
+                }
+                return;
+            }
+
+            if (objType.size() > 1)
```
sunharis, line 4198: For delete operation, need not check how many objects are there etc. You can just get the specific object (entryId) which is asked for delete, and remove the same from the dbus

redfish-core/lib/log_services.hpp
```diff
@@ -4233,6 +4233,71 @@ inline void deleteSystemHardwareIsolationLogEntryById(
         entryObjPath.str, hwIsolationEntryIfaces);
 }
 
+/**
+ * @brief API Used to deisolate the all HardwareIsolation entries.
```
sunharis, line 4237: What is the context of deisolate  ? POST command or DELETE command ? can you please use different terms here ? to clearly differentiate between Delete and DeleteAll action?

Commit message has no difference in the output for "Get the isolated hardware entry after deisolation."

sunharis (changes requested):

redfish-core/lib/log_services.hpp
```diff
@@ -51,6 +53,32 @@ constexpr char const* crashdumpOnDemandInterface =
 constexpr char const* crashdumpTelemetryInterface =
     "com.intel.crashdump.Telemetry";
 
+constexpr auto objMapperSerName = "xyz.openbmc_project.ObjectMapper";
```
gtmills, line 56: We don't do this in any other mappers calls, lets not do this here

redfish-core/lib/log_services.hpp
```diff
@@ -15,6 +15,8 @@
 */
 #pragma once
 
+#include "memory.hpp"
```
gtmills, line 18: Why include these? If we need to include these, lets move that function to a util

redfish-core/lib/log_services.hpp
gtmills: Yes, we want to check here

redfish-core/lib/log_services.hpp
gtmills: This looks correct. good job adding this.

redfish-core/lib/log_services.hpp
gtmills: I guess we are using them below for 

constexpr std::array<const char*, 2> processorInterfaces = {
    "xyz.openbmc_project.Inventory.Item.Cpu",
    "xyz.openbmc_project.Inventory.Item.Accelerator"}; 
    
    Will we ever have a  "xyz.openbmc_project.Inventory.Item.Accelerator" ?

redfish-core/lib/log_services.hpp
```diff
@@ -3415,6 +3440,609 @@ inline void getSystemHardwareIsolationLogService(
...
+                        BMCWEB_LOG_ERROR << "DBUS response error ["
+                                         << ec.value() << " : " << ec.message()
+                                         << "] when tried to get the "
+                                            "PrettyName of isolated hareware: "
+                                         << dbusObjPath.str;
+                        messages::internalError(asyncResp->res);
+                        return;
+                    }
+
+                    if (auto prettyNameVal =
```
gtmills, line 3760: need to check nullptr here

gtmills (changes requested):

redfish-core/lib/log_services.hpp
```diff
@@ -3346,4 +3415,931 @@ inline void requestRoutesPostCodesEntry(App& app)
...
+    {
+        // Reduce by "1" from uriIdPos to get the redfish uri of uriIdPattern
+        // For example,
+        // "/redfish/v1/Systems/system/Processors/<str>/SubProcessors/core0"
+        //                                       ^
+        auto parentRedfishUri = redfishUri.substr(0, uriIdPos - 1);
+        auto parentRedfishUriIt =
+            std::find_if(redfishUriList.begin(), redfishUriList.end(),
+                         [parentRedfishUri](const auto& ele) {
+                             return parentRedfishUri == std::get<1>(ele);
```
gtmills, line 3523: This throws common error 5

redfish-core/lib/log_services.hpp
```diff
@@ -3346,4 +3415,931 @@ inline void requestRoutesPostCodesEntry(App& app)
...
+                     * "/redfish/v1/Systems/system/Processors/<str>/ \
+                     * SubProcessors/core0"
+                     */
+                    fillRedfishUriWithIds(asyncResp, entryJsonIdx, dbusObjPath,
+                                          redfishUri);
+                }
+            }
+        },
+        objMapperSerName, objMapperObjPath, objMapperIface, "GetObject",
+        dbusObjPath.str, std::get<0>(*redfishUriIt));
```
gtmills, line 3670: common error 5

redfish-core/lib/log_services.hpp
```diff
@@ -3346,4 +3415,931 @@ inline void requestRoutesPostCodesEntry(App& app)
...
+                        {
+                            entryJson["Severity"] = "Warning";
+                        }
+                    }
+                }
+                else if (property.first == "Resolved")
+                {
+                    if (auto resolved = std::get_if<bool>(&property.second))
+                    {
+                        entryJson["Resolved"] = *resolved;
```
gtmills, line 3882: check nullptr here

redfish-core/lib/log_services.hpp
```diff
@@ -3346,4 +3415,931 @@ inline void requestRoutesPostCodesEntry(App& app)
...
+            }
+        }
+        else if (interface.first == "xyz.openbmc_project."
+                                    "Time.EpochTime")
+        {
+            for (auto& property : interface.second)
+            {
+                if (property.first == "Elapsed")
+                {
+                    if (auto elapsdTime =
```
gtmills, line 3894: don't use auto here and please check nullptr

redfish-core/lib/log_services.hpp
```diff
@@ -3346,4 +3415,931 @@ inline void requestRoutesPostCodesEntry(App& app)
...
+            for (auto& property : interface.second)
+            {
+                if (property.first == "Associations")
+                {
+                    if (auto associations =
+                            std::get_if<AssociationsValType>(&property.second))
+                    {
+                        for (auto& assoc : *associations)
+                        {
+                            if (std::get<0>(assoc) == "isolated_hw")
```
gtmills, line 3915: common error 5

redfish-core/lib/log_services.hpp
```diff
@@ -3346,4 +3415,931 @@ inline void requestRoutesPostCodesEntry(App& app)
...
+                            {
+                                fillIsolatedHwDetailsByObjPath(
+                                    asyncResp,
+                                    sdbusplus::message::object_path(
+                                        std::get<2>(assoc)),
```
gtmills, line 3920: common error 5

redfish-core/lib/log_services.hpp
```diff
@@ -3346,4 +3415,931 @@ inline void requestRoutesPostCodesEntry(App& app)
...
+                                    entryJsonIdx);
+                            }
+                            else if (std::get<0>(assoc) ==
```
gtmills, line 3923: common error 5

redfish-core/lib/log_services.hpp
```diff
@@ -3346,4 +3415,931 @@ inline void requestRoutesPostCodesEntry(App& app)
...
+                                     "isolated_hw_errorlog")
+                            {
+                                sdbusplus::message::object_path errPath =
+                                    std::get<2>(assoc);
```
gtmills, line 3927: common error 5

redfish-core/lib/log_services.hpp
```diff
@@ -3346,4 +3415,931 @@ inline void requestRoutesPostCodesEntry(App& app)
...
+                }
+                else
+                {
+                    fillRedfishUriWithIds(asyncResp, entryJsonIdx, dbusObjPath,
+                                          redfishUri);
+                }
+            },
+            objMapperSerName, objMapperObjPath, objMapperIface,
+            "GetSubTreePaths", "/xyz/openbmc_project/inventory", 0,
+            std::get<0>(*parentRedfishUriIt));
```
gtmills, line 3595: common error 5

redfish-core/lib/log_services.hpp
```diff
@@ -3346,4 +3415,931 @@ inline void requestRoutesPostCodesEntry(App& app)
...
+    {
+        if (interface.first == "xyz.openbmc_project."
+                               "HardwareIsolation.Entry")
+        {
+            entryIsPresent = true;
+            for (auto& property : interface.second)
+            {
+                if (property.first == "Severity")
+                {
+                    if (auto severity =
```
gtmills, line 3858: don't use auto here and please check nullptr

redfish-core/lib/log_services.hpp
```diff
@@ -3346,4 +3415,931 @@ inline void requestRoutesPostCodesEntry(App& app)
...
+                                               "HardwareIsolation."
+                                               "Entry.Type.Manual"))
+                        {
+                            entryJson["Severity"] = "Warning";
+                        }
+                    }
+                }
+                else if (property.first == "Resolved")
+                {
+                    if (auto resolved = std::get_if<bool>(&property.second))
```
gtmills, line 3880: don't use auto here

redfish-core/lib/log_services.hpp
```diff
@@ -1085,6 +1116,44 @@ inline void requestRoutesSystemLogServiceCollection(App& app)
...
+                        {
+                            BMCWEB_LOG_ERROR
+                                << "D-Bus response error [" << ec.value()
+                                << " : " << ec.message() << "] "
+                                << "when tried to check the HardwareIsolation "
+                                << "service support";
+                            return;
+                        }
+
+                        for (auto& pathStr : subtreePath)
```
gtmills, line 1133: const auto& pathStr ?

redfish-core/lib/log_services.hpp
RameshIyyar: Can I know the reason why we don't want to do it? FYI, I am using this constexpr in many places so, if we want to change in future (agree very rare) it will be easy to change in one place.

Any advantage if we hardcoding in all places?

redfish-core/lib/log_services.hpp
RameshIyyar: > I guess we are using them below for

constexpr std::array<const char*, 2> processorInterfaces = {
"xyz.openbmc_project.Inventory.Item.Cpu",
"xyz.openbmc_project.Inventory.Item.Accelerator"};

Yes, right. 

>  Will we ever have a  "xyz.openbmc_project.Inventory.Item.Accelerator" ?

I am not sure about the IBM system on this - It is defined in the existing processor.hpp so it will behave as like early

redfish-core/lib/log_services.hpp
RameshIyyar: Agree on the common error 5 but, here I am using for tuple, not for the vector. 
I think tuple won't throw an exception. FYR https://en.cppreference.com/w/cpp/utility/tuple/get - Even that bmcweb common error did not mention for the `tuple` - I am missing anything on this?

redfish-core/lib/log_services.hpp
RameshIyyar: Agree on the common error 5 but, here I am using for tuple, not for the vector. 
I think tuple won't throw an exception. FYR https://en.cppreference.com/w/cpp/utility/tuple/get - Even that bmcweb common error did not mention for the `tuple` - I am missing anything on this?

redfish-core/lib/log_services.hpp
RameshIyyar: Agree on the common error 5 but, here I am using for tuple, not for the vector. 
I think tuple won't throw an exception. FYR https://en.cppreference.com/w/cpp/utility/tuple/get - Even that bmcweb common error did not mention for the `tuple` - I am missing anything on this?

redfish-core/lib/log_services.hpp
RameshIyyar: Not required, since the implicit conversion will happen that will return true for non-null pointer and false for a null pointer.

redfish-core/lib/log_services.hpp
RameshIyyar: > don't use auto here

Can I know the reason "why we cannot use auto here"?

> please check nullptr

Not required, since the implicit conversion will happen that will return true for non-null pointer and false for a null pointer.

redfish-core/lib/log_services.hpp
RameshIyyar: > don't use auto here

Can I know the reason "why we cannot use auto here"?

redfish-core/lib/log_services.hpp
RameshIyyar: > check nullptr here

Not required, since the implicit conversion will happen that will return true for non-null pointer and false for a null pointer in the if block itself.

redfish-core/lib/log_services.hpp
RameshIyyar: > don't use auto here

Can I know the reason "why we cannot use auto here"?

> please check nullptr

Not required, since the implicit conversion will happen that will return true for non-null pointer and false for a null pointer.

redfish-core/lib/log_services.hpp
RameshIyyar: Agree on the common error 5 but, here I am using for tuple, not for the vector. 
I think tuple won't throw an exception. FYR https://en.cppreference.com/w/cpp/utility/tuple/get - Even that bmcweb common error did not mention for the `tuple` - I am missing anything on this?

redfish-core/lib/log_services.hpp
RameshIyyar: Agree on the common error 5 but, here I am using for tuple, not for the vector. 
I think tuple won't throw an exception. FYR https://en.cppreference.com/w/cpp/utility/tuple/get - Even that bmcweb common error did not mention for the `tuple` - I am missing anything on this?

redfish-core/lib/log_services.hpp
RameshIyyar: Agree on the common error 5 but, here I am using for tuple, not for the vector. 
I think tuple won't throw an exception. FYR https://en.cppreference.com/w/cpp/utility/tuple/get - Even that bmcweb common error did not mention for the `tuple` - I am missing anything on this?

redfish-core/lib/log_services.hpp
RameshIyyar: Agree on the common error 5 but, here I am using for tuple, not for the vector. 
I think tuple won't throw an exception. FYR https://en.cppreference.com/w/cpp/utility/tuple/get - Even that bmcweb common error did not mention for the `tuple` - I am missing anything on this?

redfish-core/lib/log_services.hpp
gtmills: > Can I know the reason "why we cannot use auto here"? 

Because it is not what we do upstream in bmcweb 

Please follow https://github.com/openbmc/bmcweb/blob/8b6a35f0f99ed38293f7d134511d6181c88c5a59/redfish-core/lib/log_services.hpp#L486

This should be similar code

redfish-core/lib/log_services.hpp
gtmills: It is easier to verify the call. 
I want to match upstream as much as possible with this match

redfish-core/lib/log_services.hpp
gtmills: Can we follow https://github.com/openbmc/bmcweb/blob/8b6a35f0f99ed38293f7d134511d6181c88c5a59/redfish-core/lib/log_services.hpp#L1397 as close as we can here

redfish-core/lib/log_services.hpp
gtmills: To follow upstream as close as possible

redfish-core/lib/log_services.hpp
```diff
@@ -1085,6 +1108,44 @@ inline void requestRoutesSystemLogServiceCollection(App& app)
                     "/xyz/openbmc_project/object_mapper",
                     "xyz.openbmc_project.ObjectMapper", "GetSubTreePaths", "/",
                     0, std::array<const char*, 1>{postCodeIface});
+
```
gtmills, line 1111: Could add a comment here on what this is doing. I think this is putting 

"/redfish/v1/Systems/system/LogServices/HardwareIsolation" out if we can find an HardwareIsolation log

redfish-core/lib/log_services.hpp
gtmills: bmcweb typically looks for an "Item.Inventory" interface for inventory resource. 
It typically looks for like a "Logging" interface for Log resources. 
This just seems odd to look for "xyz.openbmc_project.HardwareIsolation.Create"

redfish-core/lib/log_services.hpp
```diff
@@ -3346,4 +3407,983 @@ inline void requestRoutesPostCodesEntry(App& app)
...
+ */
+inline void
+    fillRedfishUriWithIds(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
+                          const size_t entryJsonIdx,
+                          const sdbusplus::message::object_path& dbusObjPath,
+                          std::string& redfishUri)
+{
+
+    std::string uriIdPattern{"<str>"};
+    if (auto uriIdPos = redfishUri.rfind(uriIdPattern);
```
gtmills, line 3504: I would prefer not auto here..  like we do other places in bmcweb https://github.com/openbmc/bmcweb/search?q=rfind
std::size_t

redfish-core/lib/log_services.hpp
```diff
@@ -3346,4 +3407,983 @@ inline void requestRoutesPostCodesEntry(App& app)
...
+                    BMCWEB_LOG_ERROR << "Failed to get parent DBus path for "
+                                        "identified parent Redfish URI: "
+                                     << parentRedfishUri
+                                     << " of the given DBus object path: "
+                                     << dbusObjPath.str;
+                    return;
+                }
+
+                std::string uriIdObj;
+                for (auto& object : objects)
```
gtmills, line 3544: const ?

redfish-core/lib/log_services.hpp
gtmills: We don't have recursive functions in bmcweb. 
Can we just do this as a for loop?

redfish-core/lib/log_services.hpp
```diff
@@ -3346,4 +3407,983 @@ inline void requestRoutesPostCodesEntry(App& app)
...
+ * @param[in] asyncResp - The redfish response to return.
+ * @param[in] dbusObjPath - The DBus object path which represents redfishUri.
+ * @param[in] entryJsonIdx - The json entry index to add isolated hardware
+ *                            details in the appropriate entry json object.
+ *
+ * @return The redfish response with "Message" property of LogEntry schema
+ *         if success else nothing in redfish response.
+ */
+
+inline void getPrettyNameByDbusObjPath(
```
gtmills, line 3703: This is the way upstream is headed: https://gerrit.openbmc-project.xyz/c/openbmc/bmcweb/+/43967/18/redfish-core/include/utils/name_utils.hpp 
Anyway we can pull that in and use that?

redfish-core/lib/log_services.hpp
```diff
@@ -3346,4 +3407,983 @@ inline void requestRoutesPostCodesEntry(App& app)
...
+                else if (property.first == "Resolved")
+                {
+                    const bool* resolved = std::get_if<bool>(&property.second);
+                    if (resolved == nullptr)
+                    {
+                        BMCWEB_LOG_ERROR
+                            << "Failed to get the Resolved"
+                            << "from object: " << dbusObjIt->first.str;
+                        messages::internalError(asyncResp->res);
+                        break;
```
gtmills, line 3904: Why not return for these?

redfish-core/lib/log_services.hpp
RameshIyyar: I will update after tested the pointed code - Here, I am playing with a json object.

RameshIyyar:
Closing this PR and pushed the patches in below new PRs since conflicts are huge after rebased the bmcweb.

- https://github.com/ibm-openbmc/bmcweb/pull/93
- https://github.com/ibm-openbmc/bmcweb/pull/94

PR #78, opened 2021-08-31
HMC Lock Management : Remove lock json file

sunharis:
BMC till now was saving the locks for HMC at BMC filesystem
"/var/lib/bmcweb/ibm-management-console/locks/ibm_mc_persistent_lock_data.json"

This commit removed this file, as there is no usecase for HMC-eBMC
interface to persist the lock record in a file and retrive the same
after a BMC reboot

HMC will need to re-acquire the locks after every BMC reboot. The
lock records will be only in-memory

Tested by:
 1. With the patch, verified the lock records are not saved to the file
 2. Removed the lock file and verified that its not created again for the
 next acquireLock API
 3. All other lock APIs are working fine (GET and DELETE)

Change-Id: Ib81634ef5bc1dbb6c906fabad8d95675e7d6a4ea

### DIFF
--- a/include/ibm/locks.hpp
+++ b/include/ibm/locks.hpp
@@ -36,28 +36,13 @@ using RcAcquireLock = std::pair<bool, std::variant<Rc, std::pair<bool, int>>>;
 using RcReleaseLockApi = std::pair<bool, std::variant<bool, RcRelaseLock>>;
 using SessionFlags = std::pair<SType, SType>;
 using ListOfSessionIds = std::vector<std::string>;
-static constexpr const char* fileName =
-    "/var/lib/bmcweb/ibm-management-console/locks/"
-    "ibm_mc_persistent_lock_data.json";
 
 class Lock
 {
     uint32_t transactionId;
     boost::container::flat_map<uint32_t, LockRequests> lockTable;
 
-    /*
-     * This API implements the logic to load the locks that are present in the
-     * json file into the lock table.
-     */
-    void loadLocks();
-
   protected:
-    /*
-     * This API implements the logic to persist the locks that are contained in
-     * the lock table into a json file.
-     */
-    void saveLocks();
-
     /*
      * This function implements the logic for validating an incoming
      * lock request/requests.
@@ -124,9 +109,6 @@ class Lock
 
     Lock()
     {
-        // Remove the persistent file
-        std::filesystem::remove(fileName);
-        lockTable.clear();
         transactionId = lockTable.empty() ? 0 : prev(lockTable.end())->first;
     }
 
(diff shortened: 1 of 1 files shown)


gtmills:
How does the HMC know the BMC rebooted and it should re-acquire locks ?
e.g. the BMC rebooted on its own or by another client, how does the HMC know this?

gtmills:
Will we push upstream? https://github.com/openbmc/bmcweb/blob/master/include/ibm/locks.hpp#L40

gtmills:
Code seems fine. Still looking for answers on my questions

sunharis:
https://gerrit.openbmc-project.xyz/c/openbmc/bmcweb/+/46443 is the upstream commit

sunharis:
> How does the HMC know the BMC rebooted and it should re-acquire locks ?
HMC will check if they have the lock on that resource before starting the operation. Otherwise they will acquire the locks again.

> e.g. the BMC rebooted on its own or by another client, how does the HMC know this?
Any BMC reboot is notified to HMC via state change events. They will have the logic to find that out at their use-cases

gtmills (approved):
Seems reasonable

include/ibm/locks.hpp
```diff
@@ -317,7 +253,7 @@ inline RcAcquireLock Lock::acquireLock(const LockRequests& lockRequestStructure)
 
     if (status)
     {
-        BMCWEB_LOG_DEBUG << "There is a conflict within itself";
+        BMCWEB_LOG_ERROR << "There is a conflict within itself";
```
manojkiraneda, line 256: this change is out of scope for this PR.why was this made ERROR ?

include/ibm/locks.hpp
```diff
@@ -345,18 +278,15 @@ inline void Lock::releaseLock(const ListOfTransactionIds& refRids)
 
         else
         {
-            BMCWEB_LOG_DEBUG << "Removing the locks from the lock table "
+            BMCWEB_LOG_ERROR << "Removing the locks from the lock table "
```
manojkiraneda, line 281: same here, this does not meet the scope of this change

include/ibm/locks.hpp
```diff
@@ -124,8 +109,6 @@ class Lock
 
     Lock()
     {
-        // Remove the persistent file
-        std::filesystem::remove(fileName);
         lockTable.clear();
```
manojkiraneda, line 112: this is not required as well , because we are removing the persistent file , the lockable will be already gone when we restart bmcweb or reboot bmc.

include/ibm/locks.hpp
```diff
@@ -124,8 +109,6 @@ class Lock
...
         transactionId = lockTable.empty() ? 0 : prev(lockTable.end())->first;
```
manojkiraneda, line 113: this is where I am confused, if we remove the persistency the transaction id's will be starting from 0 every time bmcweb restarts or when BMC takes a reboot, so there is a possibility for HMC to see the same transaction ID for different locks across reboot. Are they okay with that ?

manojkiraneda (commented):
Overall changes looks good, but have some questions.

include/ibm/locks.hpp
sunharis: This change is to log a trace if there is a lock conflict

include/ibm/locks.hpp
sunharis: This change is to log a trace if there is a lock removal. We dont have the json file now, and there is nothing to track the lock activities at BMC. So moved these to the error log level to get them at production driver

include/ibm/locks.hpp
sunharis: I don't see  lockTable.clear(); in upstream.

include/ibm/locks.hpp
sunharis: They had mentioned that they will work good when they re-acquire the locks. I have asked them again to re-confirm. Lets wait for their answer and i will update here

sunharis: Seems like this concern is OK for HMC team.

manojkiraneda: sounds good

manojkiraneda (approved):

PR #80, opened 2021-09-02
Modify VMI IP configuration to use new hypervisor network app

asmithakarun:
Today what we have in the downstream is:

BMCweb updates the bios-settings manager for the VMI-related BIOS attribute which is not acceptable by the community.

Agreed design:
=> There would be a backend Dbus service that will host the object for the hypervisor and will interact with the BIOS settings manager.
=> bmcweb will directly update those DBus objects hosted by new hypervisor network app
=> By doing so we are making bmcweb agnostic to the BIOS attribute which would be vendor-specific attributes.

The upstream commit : https://gerrit.openbmc-project.xyz/c/openbmc/bmcweb/+/45063


gtmills:
Can we get some reviews upstream?

asmithakarun:
Creating a fresh PR with few more changes

PR #82, opened 2021-09-03
bmcweb: Add support for Correctable Event Log

sbteeks:
This commit allows users of any privilege to access Event Log entries
with 'Hidden' property set to false. The Log entries with 'Hidden'
property set to true are only accessed by users with admin role.

This is downstream only change


geissonator:
add to approvelist

gtmills:
Please run clang-format and the validator 

```
                                 sensors::node::thermal
                             ? "Temperatures"
-                            : sensorsAsyncResp->chassisSubNode ==
-                                      sensors::node::power
-                                  ? "Voltages"
-                                  : "Sensors");
+                        : sensorsAsyncResp->chassisSubNode ==
+                                sensors::node::power
+                            ? "Voltages"
+                            : "Sensors");
                     return;
```

redfish-core/lib/log_services.hpp
```diff
@@ -1801,6 +2109,154 @@ inline void requestRoutesDBusEventLogEntry(App& app)
...
+                                                               "EventLogEntry",
+                                                               entryID);
+                                    return;
+                                }
+                            }
+                            else if (propertyMap.first == "Timestamp")
+                            {
+                                const uint64_t* millisTimeStamp =
+                                    std::get_if<uint64_t>(&propertyMap.second);
+                                if (millisTimeStamp != nullptr)
```
gtmills, line 2176: These are typically inverted 

                                 if (millisTimeStamp == nullptr)
                                 {
                                 internalError

redfish-core/lib/log_services.hpp
```diff
@@ -1801,6 +2109,154 @@ inline void requestRoutesDBusEventLogEntry(App& app)
...
+                                {
+                                    updateTimestamp =
+                                        crow::utility::getTimestamp(
+                                            *millisTimeStamp);
+                                }
+                            }
+                            else if (propertyMap.first == "Severity")
+                            {
+                                severity = std::get_if<std::string>(
```
gtmills, line 2195: Need to check for nullput

redfish-core/lib/log_services.hpp
gtmills: and I see that below please follow

redfish-core/lib/log_services.hpp
```diff
@@ -1123,6 +1135,43 @@ inline void requestRoutesEventLogService(App& app)
...
+                redfishDateTimeOffset.second;
+
+            asyncResp->res.jsonValue["Entries"] = {
+                {"@odata.id", "/redfish/v1/Systems/system/LogServices/"
+                              "CorrectableEventLog/Entries"}};
+            asyncResp->res.jsonValue["Actions"]["#LogService.ClearLog"] = {
+
+                {"target",
+                 "/redfish/v1/Systems/system/LogServices/CorrectableEventLog/"
+                 "Actions/LogService.ClearLog"}};
```
gtmills, line 1171: But where is this action implemented ? 
Right now it is a dead action

redfish-core/lib/log_services.hpp
```diff
@@ -1413,6 +1462,34 @@ inline void requestRoutesJournalEventLogEntry(App& app)
             });
 }
 
+static bool isHiddenEventLogEntry(std::string& path)
+{
+    constexpr auto loggingService = "xyz.openbmc_project.Logging";
```
gtmills, line 1467: You can't combine this with the ?

                "xyz.openbmc_project.Logging", "/xyz/openbmc_project/logging",
                "org.freedesktop.DBus.ObjectManager", "GetManagedObjects");

redfish-core/lib/log_services.hpp
```diff
@@ -1413,6 +1462,34 @@ inline void requestRoutesJournalEventLogEntry(App& app)
...
+    constexpr auto loggingPelInterface = "org.open_power.Logging.PEL.Entry";
+    constexpr auto loggingPelProperty = "Hidden";
+
+    constexpr auto interface = "org.freedesktop.DBus.Properties";
+
+    auto bus = sdbusplus::bus::new_default();
+    bool propValue = false;
+    try
```
gtmills, line 1475: Why the try / catch? 
Let's follow https://github.com/openbmc/bmcweb/blob/83fd8e19867f43f3a175d1ede65ad8636da553b3/redfish-core/lib/systems.hpp#L673

gtmills (changes requested):

redfish-core/lib/log_services.hpp
sbteeks: It is possible to combine, but doing so will lead to walk through all properties corresponding to  xyz.openbm_project.Logging.Entry - which is not necessary when the Hidden property is true. Also, there are other properties under org.open_power.Logging.PEL.Entry interface. We don't need to look through those props as well. Having a separate function specifically to check only the Hidden property is cleaner in my opinion.  But it is desired to use GetManagedObject, I can make that change.

redfish-core/lib/log_services.hpp
sbteeks: I needed the try block since the org.open_power.Logging.PEL.Entry interface and 'Hidden' property weren't there in older LogEvent entries - that is the event entries created before the Hidden property feature was added. It was coming out with (internal?) error, so just added the try...catch block to avoid such error and also to let those event log entries printed. 
I can remove this if it is assumed that all event log entries will have Hidden proprty...

gtmills: Lets do something like


```
{
    BMCWEB_LOG_DEBUG << "Get host information.";
        [aResp](const boost::system::error_code ec,
                const std::variant<std::string>& hostState) {
            if (ec)
            {
                BMCWEB_LOG_DEBUG << "DBUS response error " << ec;
                return;
            }
            
            ...
            

        },
        "xyz.openbmc_project.Logging", "/xyz/openbmc_project/state/host0",
        "org.freedesktop.DBus.Properties", "Get",
        "org.open_power.Logging.PEL.Entry", "Hidden");
}
```

Look at like https://github.com/openbmc/bmcweb/blob/83fd8e19867f43f3a175d1ede65ad8636da553b3/redfish-core/lib/systems.hpp#L673

redfish-core/lib/log_services.hpp
gtmills: It is desired to only have 1 dbus call

PR #75, opened 2021-08-26
Move Inventory.Item.Board from Chassis to Assembly

patelabhishek9893:
A newer way of thinking in Redfish is for a simple rack server to have
one Redfish Chassis and things like boards modeled as Redfish
Assemblies under the Chassis.

This commit move Board under assembly. Moreover, if the association path
has "chassis/motherboard" or "/board/<Panel info>" this commit is able
to retrieve both information under assembly

busctl call xyz.openbmc_project.ObjectMapper \
  /xyz/openbmc_project/object_mapper xyz.openbmc_project.ObjectMapper \
  GetSubTreePaths sias /xyz/openbmc_project/ 0 1 \
  xyz.openbmc_project.Inventory.Item.Board

 as 2 "/xyz/openbmc_project/inventory/system/board/Blyth_Panel"
      "/xyz/openbmc_project/inventory/system/board/Nisqually_Backplane"

Tested: manually tested on system
  curl -k -H "X-Auth-Token: $token" -X GET
  https://${bmc}/redfish/v1/Chassis/chassis/Assembly

{
  "@odata.id": "/redfish/v1/Chassis/chassis/Assembly",
  "@odata.type": "#Assembly.v1_3_0.Assembly",
  "Assemblies": [
  {
    "@odata.id": "/redfish/v1/Chassis/chassis/Assembly#/Assemblies/0",
    "@odata.type": "#Assembly.v1_3_0.AssemblyData",
    "LocationIndicatorActive": null,
    "MemberId": "0",
    "Name": "Blyth_Panel"
  },
  {
    "@odata.id": "/redfish/v1/Chassis/chassis/Assembly#/Assemblies/1",
    "@odata.type": "#Assembly.v1_3_0.AssemblyData",
    "LocationIndicatorActive": null,
    "MemberId": "1",
    "Name": "Nisqually_Backplane"
  },
  {
    "@odata.id": "/redfish/v1/Chassis/chassis/Assembly#/Assemblies/2",
    "@odata.type": "#Assembly.v1_3_0.AssemblyData",
    "Location": {
      "PartLocation": {
        "ServiceLabel": "U78DA.ND0.WZS004K-D0"
       }
    },
    "LocationIndicatorActive": false,

    ...

  },

  ...
  ...
  ...

  ],
  "Assemblies@odata.count": 11,
  "Id": "Assembly",
  "Name": "Assembly Collection"
}

Email sent to openbmc list:
[...]

PR #85, opened 2021-09-08
remove Item.Board from Power Redfish_util sensors

patelabhishek9893:
When we tried to Move Inventory.Item.Board from Chassis to Assembly

as you see in this commit:- https://github.com/ibm-openbmc/bmcweb/pull/75

this commit brack Redfish validator and we below error
- 1 failProp errors in /redfish/v1/Systems/system
- 2 failProp errors in /redfish/v1/Managers/bmc
- 1 problemResource errors in /redfish/v1/Chassis/Blyth_Panel

This error message we get.
ERROR - Links.Chassis[0]: GET of resource at URI /redfish/v1/Chassis/Blyth_Panel returned HTTP 404. Check URI.

This error we get because we tried to move Item.Board from chassis to assembly that might be affected other place.
https://github.ibm.com/Abhishek-Patel/openbmc/commit/761a771cd99bc428bce6bcd526adad3f7c8b3b3a

To fix that, we remove Item.Board from all below files, So they unable to retrieve association path

- redfish-core/lib/power.hpp
- redfish-core/lib/redfish_util.hpp
- redfish-core/lib/sensors.hpp

Tested: Manually tested on the system and Run Redfish validator. Found no errors.

### DIFF
--- a/redfish-core/lib/power.hpp
+++ b/redfish-core/lib/power.hpp
@@ -314,8 +314,7 @@ inline void requestRoutesPower(App& app)
                     "/xyz/openbmc_project/object_mapper",
                     "xyz.openbmc_project.ObjectMapper", "GetSubTreePaths",
                     "/xyz/openbmc_project/inventory", 0,
-                    std::array<const char*, 2>{
-                        "xyz.openbmc_project.Inventory.Item.Board",
+                    std::array<const char*, 1>{
                         "xyz.openbmc_project.Inventory.Item.Chassis"});
             });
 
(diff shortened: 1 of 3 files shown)


gtmills:
@patelabhishek9893 Where is the upstream for this? 
Does the validator pass? 

And then you need some more context. 
Why is this needed? What validator error is this fixing? 
What commit / PR before removed some of this?

gtmills (changes requested):
Need more than 1 sentence PR message. Also want to upstream in review first.

gtmills:
@patelabhishek9893 Can you add the upstream commit for this?

gtmills:
> @patelabhishek9893 Can you add the upstream commit for this? 

https://gerrit.openbmc-project.xyz/c/openbmc/bmcweb/+/45938 is upstream for this

gtmills (approved):

PR #94, opened 2021-09-14
Added new LogServices for Hardware Isolation (aka Guard) feature

RameshIyyar:
Created this PR to fix a bunch of conflicts which are got after rebased the upstream bmcweb into ibm-openbmc/bmcweb/1020

**Note:** In this PR, addressed all the comments (few are clarification) which are posted in https://github.com/ibm-openbmc/bmcweb/pull/76 - Please refer [here](https://github.com/ibm-openbmc/bmcweb/pull/76) to get to know the comments history.

### DIFF
--- a/redfish-core/include/redfish.hpp
+++ b/redfish-core/include/redfish.hpp
@@ -224,6 +224,10 @@ class RedfishService
         requestRoutesMetricReportDefinition(app);
         requestRoutesMetricReportCollection(app);
         requestRoutesMetricReport(app);
+
+#ifdef BMCWEB_ENABLE_HW_ISOLATION
+        requestRoutesSystemHardwareIsolationLogService(app);
+#endif
     }
 };
 
--- a/redfish-core/lib/log_services.hpp
+++ b/redfish-core/lib/log_services.hpp
@@ -4279,6 +4279,73 @@ inline void deleteSystemHardwareIsolationLogEntryById(
 }
 
 /**
+ * @brief API Used to deisolate the all HardwareIsolation entries.
+ *
+ * @param[in] req - The HardwareIsolation redfish request (unused now).
+ * @param[in] asyncResp - The redfish response to return.
+ *
+ * @return The redfish response in the given buffer.
+ */
+inline void postSystemHardwareIsolationLogServiceClearLog(
+    const crow::Request& /* req */,
+    const std::shared_ptr<bmcweb::AsyncResp>& asyncResp)
+{
+    // Get the DBus name of HardwareIsolation service
+    crow::connections::systemBus->async_method_call(
+        [asyncResp](const boost::system::error_code ec,
+                    const GetObjectType& objType) {
+            if (ec || objType.empty())
+            {
+                BMCWEB_LOG_ERROR << "DBUS response error [" << ec.value()
+                                 << " : " << ec.message()
+                                 << "] when tried to get the HardwareIsolation "
+                                    "dbus name";
+                messages::internalError(asyncResp->res);
+                return;
+            }
+
+            if (objType.size() > 1)
+            {
+                BMCWEB_LOG_ERROR << "More than one dbus service implemented "
+                                    "the HardwareIsolation service";
+                messages::internalError(asyncResp->res);
+                return;
+            }
+
+            if (objType[0].first.empty())
+            {
+                BMCWEB_LOG_ERROR << "The retrieved HardwareIsolation dbus "
+                                    "name is empty";
+                messages::internalError(asyncResp->res);
+                return;
+            }
+
+            // Delete all HardwareIsolation entries
+            crow::connections::systemBus->async_method_call(
+                [asyncResp](const boost::system::error_code ec) {
+                    if (ec)
+                    {
+                        BMCWEB_LOG_ERROR
+                            << "DBUS response error [" << ec.value() << " : "
+                            << ec.message()
+                            << "] when tried to delete all HardwareIsolation "
+                               "entries";
+                        messages::internalError(asyncResp->res);
+                        return;
+                    }
+                    messages::success(asyncResp->res);
+                },
+                objType[0].first, "/xyz/openbmc_project/hardware_isolation",
+                "xyz.openbmc_project.Collection.DeleteAll", "DeleteAll");
+        },
+        "xyz.openbmc_project.ObjectMapper",
+        "/xyz/openbmc_project/object_mapper",
+        "xyz.openbmc_project.ObjectMapper", "GetObject",
+        "/xyz/openbmc_project/hardware_isolation",
+        std::array<const char*, 1>{"xyz.openbmc_project.Collection.DeleteAll"});
+}
+
+/**
  * @brief API used to route the handler for HardwareIsolation Redfish
  *        LogServices URI
  *
@@ -4311,6 +4378,13 @@ inline void requestRoutesSystemHardwareIsolationLogService(App& app)
         .privileges(redfish::privileges::deleteLogEntry)
         .methods(boost::beast::http::verb::delete_)(
             deleteSystemHardwareIsolationLogEntryById);
+
+    BMCWEB_ROUTE(app,
+                 "/redfish/v1/Systems/system/LogServices/HardwareIsolation/"
+                 "Actions/LogService.ClearLog/")
+        .privileges(redfish::privileges::postLogService)
+        .methods(boost::beast::http::verb::post)(
+            postSystemHardwareIsolationLogServiceClearLog);
 }
 #endif // BMCWEB_ENABLE_HW_ISOLATION
 
--- a/redfish-core/lib/log_services.hpp
+++ b/redfish-core/lib/log_services.hpp
@@ -52,6 +52,11 @@ constexpr char const* crashdumpTelemetryInterface =
     "com.intel.crashdump.Telemetry";
 
 #ifdef BMCWEB_ENABLE_HW_ISOLATION
+constexpr std::array<const char*, 3> hwIsolationEntryIfaces = {
+    "xyz.openbmc_project.HardwareIsolation.Entry",
+    "xyz.openbmc_project.Association.Definitions",
+    "xyz.openbmc_project.Time.EpochTime"};
+
 using RedfishResourceDBusInterfaces = std::string;
 using RedfishResourceCollectionUri = std::string;
 using RedfishUriListType = std::unordered_map<RedfishResourceDBusInterfaces,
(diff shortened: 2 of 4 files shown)


RameshIyyar:
CI failed due to dependency, this PR is co-req with https://github.com/ibm-openbmc/bmcweb/pull/93

RameshIyyar:
jenkins run tests please

redfish-core/lib/log_services.hpp
```diff
@@ -3407,4 +3470,965 @@ inline void requestRoutesPostCodesEntry(App& app)
...
+
+    std::string uriIdPattern{"<str>"};
+    std::size_t uriIdPos = redfishUri.rfind(uriIdPattern);
+    if (uriIdPos != std::string::npos)
+    {
+        // Reduce by "1" from uriIdPos to get the redfish uri of uriIdPattern
+        // For example,
+        // "/redfish/v1/Systems/system/Processors/<str>/SubProcessors/core0"
+        //                                       ^
+        auto parentRedfishUri = redfishUri.substr(0, uriIdPos - 1);
```
gtmills, line 3574: std::string here would fit more with upstream bmcweb style

redfish-core/lib/processor.hpp
```diff
@@ -46,6 +46,19 @@ constexpr std::array<const char*, 2> processorInterfaces = {
...
+constexpr auto processorRedfishUri = "/redfish/v1/Systems/system/Processors";
+
+// Interfaces which imply a D-Bus object represents a Processor Core
+constexpr std::array<const char*, 1> procCoreInterfaces = {
+    "xyz.openbmc_project.Inventory.Item.CpuCore"};
+
+// Redfish URI which imply a Processor Core resource represents a
+// "procCoreInterfaces"
+constexpr auto processorCoreRedfishUri =
+    "/redfish/v1/Systems/system/Processors/<str>/SubProcessors";
```
gtmills, line 60: But this is wrong for Core interface

gtmills: This is a processor collection

redfish-core/lib/memory.hpp
```diff
@@ -45,6 +45,9 @@ using MapperGetSubTreeResponse =
 constexpr std::array<const char*, 1> dimmInterfaces = {
     "xyz.openbmc_project.Inventory.Item.Dimm"};
 
+// Redfish URI which imply a Memory resource represents a "dimmInterfaces"
+constexpr auto memoryRedfishUri = "/redfish/v1/Systems/system/Memory";
```
gtmills, line 49: This is the memory collection

redfish-core/lib/log_services.hpp
```diff
@@ -51,6 +53,24 @@ constexpr char const* crashdumpOnDemandInterface =
...
 
+constexpr std::array<const char*, 3> hwIsolationEntryIfaces = {
+    "xyz.openbmc_project.HardwareIsolation.Entry",
+    "xyz.openbmc_project.Association.Definitions",
+    "xyz.openbmc_project.Time.EpochTime"};
+
+using RedfishUriListType =
+    std::vector<std::tuple<std::vector<std::string>, std::string>>;
+
+const RedfishUriListType redfishUriList = {
```
gtmills, line 64: I think you are better off coding this here and just containing this to this file..

redfish-core/lib/log_services.hpp
```diff
@@ -1147,6 +1170,46 @@ inline void requestRoutesSystemLogServiceCollection(App& app)
...
+                        {
+                            if (pathStr.find("hardware_isolation") !=
+                                std::string::npos)
+                            {
+                                nlohmann::json& logServiceArrayLocal =
+                                    asyncResp->res.jsonValue["Members"];
+                                logServiceArrayLocal.push_back(
+                                    {{"@odata.id",
+                                      "/redfish/v1/Systems/system/"
+                                      "LogServices/HardwareIsolation"}});
```
gtmills, line 1199: Even when the link isn't there, you can still get to the path... :/ 
line 3481 / 4402
I'm not sure this makes sense, I think it would be better to just always show it and when HardwareIsolation is empty, just show an list

redfish-core/lib/log_services.hpp
```diff
@@ -3407,4 +3470,965 @@ inline void requestRoutesPostCodesEntry(App& app)
...
+ * @param[in] req - The HardwareIsolation redfish request (unused now).
+ * @param[in] asyncResp - The redfish response to return.
+ *
+ * @return The redfish response in the given buffer.
+ */
+inline void getSystemHardwareIsolationLogService(
+    const crow::Request& /* req */,
+    const std::shared_ptr<bmcweb::AsyncResp>& asyncResp)
+{
+    asyncResp->res.jsonValue["@odata.id"] =
```
gtmills, line 3485: You can still get here even when you don't find any 

                    std::array<const char*, 1>{
                        "xyz.openbmc_project.HardwareIsolation.Create"});

redfish-core/lib/log_services.hpp
```diff
@@ -3407,4 +3470,965 @@ inline void requestRoutesPostCodesEntry(App& app)
...
+            if (objType[0].first.empty())
+            {
+                BMCWEB_LOG_ERROR << "The retrieved dbus name is empty for the "
+                                    "given dbus object: "
+                                 << dbusObjPath.str;
+                messages::internalError(asyncResp->res);
+                return;
+            }
+
+            if (entryJsonIdx > 0)
```
gtmills, line 3804: Why this if / else?

redfish-core/lib/log_services.hpp
```diff
@@ -3407,4 +3470,965 @@ inline void requestRoutesPostCodesEntry(App& app)
...
+inline void fillSystemHardwareIsolationLogEntry(
+    const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
+    const size_t entryJsonIdx, GetManagedObjectsType::const_iterator& dbusObjIt)
+{
+    nlohmann::json& entryJson =
+        (entryJsonIdx > 0
+             ? asyncResp->res.jsonValue["Members"][entryJsonIdx - 1]
+             : asyncResp->res.jsonValue);
+
+    bool entryIsPresent{false};
```
gtmills, line 3894: nitpick: I know this is the same but I have never seen this in upstream bmcweb 
All I see is 
https://github.com/openbmc/bmcweb/blob/97128e926c8253f02cf25d2ca9881cadc426df2c/include/sessions.hpp#L49

redfish-core/lib/log_services.hpp
```diff
@@ -3407,4 +3470,965 @@ inline void requestRoutesPostCodesEntry(App& app)
...
+                << ec.message()
+                << "] when tried to get the HardwareIsolation managed objects";
+            messages::internalError(asyncResp->res);
+            return;
+        }
+
+        nlohmann::json& entriesArray = asyncResp->res.jsonValue["Members"];
+        if (!entriesArray.size())
+        {
+            entriesArray = nlohmann::json::array();
```
gtmills, line 4064: Can you move this out of the if like 
https://github.com/openbmc/bmcweb/blob/9a128eb3fc3008f802974360e98de6495ed1a66b/redfish-core/include/utils/telemetry_utils.hpp#L41

gtmills: I don't think you need the if

RameshIyyar: Done

redfish-core/lib/log_services.hpp
RameshIyyar: Checking backend support and if not supported then returning "404" error.

RameshIyyar:
@gtmills please review

redfish-core/lib/log_services.hpp
gtmills: I don't think PostCode did this correctly :( 
Can we just show this always, I think that is the simplest solution?

redfish-core/lib/processor.hpp
gtmills: How about we just name this 
processorCoreCollectionUri ?
processorCoreCollectionRedfishUri ?
Or processorCoreParentUri ? 
coreParentUri?

redfish-core/lib/processor.hpp
```diff
@@ -46,6 +46,19 @@ constexpr std::array<const char*, 2> processorInterfaces = {
     "xyz.openbmc_project.Inventory.Item.Cpu",
     "xyz.openbmc_project.Inventory.Item.Accelerator"};
 
+// Redfish URI which imply a Processor resource represents a
+// "processorInterfaces"
+constexpr auto processorRedfishUri = "/redfish/v1/Systems/system/Processors";
```
gtmills, line 51: Again this is the ProcessorCollection.. Can we name it accordingly?

redfish-core/lib/memory.hpp
gtmills: memoryCollectionRedfishUri  ?
memoryParentUri ?

redfish-core/lib/log_services.hpp
gtmills: I think as is (e.g. processorInterfaces.begin(), processorInterfaces.end()), ) makes this harder to understand. 
I think just hardcoding these makes this easier to carry downstream (touches less files)
I don't think these will change in the future and if they do this code might need to be adjusted anyway

redfish-core/lib/log_services.hpp
RameshIyyar: What happens if we don't have `HardwareIsolation` feature backend support in the system? still, we should show/allow the URI (/redfish/v1/Systems/system/LogServices/HardwareIsolation)?

redfish-core/lib/log_services.hpp
```diff
@@ -3407,4 +3469,1001 @@ inline void requestRoutesPostCodesEntry(App& app)
...
+    if (uriIdPos != std::string::npos)
+    {
+        // Reduce by "1" from uriIdPos to get the redfish uri of uriIdPattern
+        // For example,
+        // "/redfish/v1/Systems/system/Processors/<str>/SubProcessors/core0"
+        //                                       ^
+        std::string parentRedfishUri = redfishUri.substr(0, uriIdPos - 1);
+        auto parentRedfishUriIt =
+            std::find_if(redfishUriList.begin(), redfishUriList.end(),
+                         [parentRedfishUri](const auto& ele) {
```
gtmills, line 3615: &parentRedfishUri

redfish-core/lib/log_services.hpp
```diff
@@ -3407,4 +3469,1001 @@ inline void requestRoutesPostCodesEntry(App& app)
...
+ *         given dbus object path to fill the respective uri Id in the
+ *         given redfishUri.
+ *       - This function will run recursively until filling all Ids in given
+ *         given redfishUri.
+ */
+inline void
+    fillRedfishUriWithIds(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
+                          const size_t entryJsonIdx,
+                          const sdbusplus::message::object_path& dbusObjPath,
+                          std::string& redfishUri)
```
gtmills, line 3601: const std::string& redfishUri ?

redfish-core/lib/log_services.hpp
```diff
@@ -3407,4 +3469,1001 @@ inline void requestRoutesPostCodesEntry(App& app)
...
+                redfishUri.replace(
+                    uriIdPos, uriIdPattern.length(),
+                    getIsolatedHwItemId(
+                        sdbusplus::message::object_path(uriIdObj)));
+
+                if (redfishUri.find(uriIdPattern) == std::string::npos)
+                {
+                    nlohmann::json& entryJson =
+                        (entryJsonIdx > 0
+                             ? asyncResp->res
```
gtmills, line 3673: Typically don't use the ternary operator in bmcweb

redfish-core/lib/log_services.hpp
```diff
@@ -3407,4 +3469,1001 @@ inline void requestRoutesPostCodesEntry(App& app)
...
+                                   .jsonValue["Members"][entryJsonIdx - 1]
+                             : asyncResp->res.jsonValue);
+
+                    entryJson["Links"]["OriginOfCondition"] = {
+                        {"@odata.id", redfishUri}};
+                }
+                else
+                {
+                    fillRedfishUriWithIds(asyncResp, entryJsonIdx, dbusObjPath,
```
gtmills, line 3682: Why do we need recursive, something I have not seen anywhere else in bmcweb? Why not just a for loop ?

redfish-core/lib/log_services.hpp
```diff
@@ -1147,6 +1169,46 @@ inline void requestRoutesSystemLogServiceCollection(App& app)
...
+                // backend, if it is available then, add into "LogServices".
+                crow::connections::systemBus->async_method_call(
+                    [asyncResp](const boost::system::error_code ec,
+                                const std::vector<std::string>& subtreePath) {
+                        if (ec)
+                        {
+                            BMCWEB_LOG_ERROR
+                                << "D-Bus response error [" << ec.value()
+                                << " : " << ec.message() << "] "
+                                << "when tried to check the HardwareIsolation "
```
raviteja-b, line 1183: return http error?

redfish-core/lib/log_services.hpp
```diff
@@ -3407,4 +3469,1001 @@ inline void requestRoutesPostCodesEntry(App& app)
...
+                             return parentRedfishUri == std::get<1>(ele);
+                         });
+
+        if (parentRedfishUriIt == redfishUriList.end())
+        {
+            BMCWEB_LOG_ERROR
+                << "Failed to get parent Redfish URI DBus interface for the "
+                   "identified Redfish URI: "
+                << redfishUri
+                << " of the given DBus object path: " << dbusObjPath.str;
```
raviteja-b, line 3625: please see if we can return http error in here and other error cases as well

redfish-core/lib/log_services.hpp
sunharis: How many times does this recursive call loops ?

redfish-core/include/redfish.hpp
```diff
@@ -224,6 +224,8 @@ class RedfishService
         requestRoutesMetricReportDefinition(app);
         requestRoutesMetricReportCollection(app);
         requestRoutesMetricReport(app);
+
+        requestRoutesSystemHardwareIsolationLogService(app);
```
sunharis, line 228: Can this go under the flag BMCWEB_ENABLE_REDFISH_DBUS_LOG_ENTRIES

redfish-core/lib/log_services.hpp
RameshIyyar: > How many times does this recursive call loops ?

I would day in the worst case, 
- `4 (Max recursive call) * N (No of guard records)`
  - if the user firing the `/redfish/v1/Systems/system/LogServices/HardwareIsolation/Entries`
- `4 * 1`
  - if the user firing the `/redfish/v1/Systems/system/LogServices/HardwareIsolation/Entries/n`

redfish-core/lib/log_services.hpp
RameshIyyar: What is the advantage? it is an iterator (std::find_if returns iterator), not the vector object, right?

RameshIyyar: Using the iterator.

redfish-core/lib/log_services.hpp
RameshIyyar: This function is more available i.e removed recursion function.

redfish-core/lib/log_services.hpp
RameshIyyar: Done - Using the for loop and removed the recursion function.

redfish-core/lib/log_services.hpp
RameshIyyar: Done - changed as requested.

redfish-core/include/redfish.hpp
sunharis: Fine. We can skip the compiler flag

sunharis (approved):

redfish-core/lib/log_services.hpp
RameshIyyar: This code removed as part of compiler flag changes :)

raviteja-b (approved):

rfrandse:
Common practice discussion for downstream PR's 
https://github.com/ibm-openbmc/dev/issues/3341

PR #77, opened 2021-08-27
Enabled the Hardware Isolation (aka Guard) feature for Core and Memory

RameshIyyar:
- The first [patch](https://github.ibm.com/openbmc/bmcweb/commit/b1bc622a3a9d3f189ff366f23d0d60cd3b14a53b) is picked from  https://gerrit.openbmc-project.xyz/c/openbmc/bmcweb/+/38570/37 which is pushed by IPS folks to the model core in Redfish.

PR #96, opened 2021-09-14
Workaround to handle DCM (Dual-Chip Module) package for Redfish

RameshIyyar:
Created this PR to address the below comments.

- Enabled the Hardware Isolation (aka Guard) feature for Core and Memory #77 (comment)

**Note:** 
- This workaround (Refer commit message for more details) patch is required to enable the manual guard feature. 
- This PR is created on top of https://github.com/ibm-openbmc/bmcweb/pull/95 so, please ignore the first commit in this PR.

### DIFF
--- a/redfish-core/lib/processor.hpp
+++ b/redfish-core/lib/processor.hpp
@@ -48,6 +48,55 @@ constexpr std::array<const char*, 2> processorInterfaces = {
     "xyz.openbmc_project.Inventory.Item.Accelerator"};
 
 /**
+ * @brief Workaround to handle DCM (Dual-Chip Module) package for Redfish
+ *
+ * Make sure processor modeled as dual chip module ("dcmN-cpuN"),
+ * If yes then, replace Redfish processor id as "dcmN/cpuN" and check with
+ * given object path because Redfish does not support chip module concept.
+ *
+ * @param[in] processorId - The Redfish processor Id
+ * @param[in] objectPath  - The D-Bus object path that contain the processor
+ *                          segment
+ *
+ * @return true if matched with the given object path else false.
+ *
+ * @note Inventory modeled as "dcmN/cpuN" to support DCM so wherever using
+ *       Redfish processor id as "dcmN-cpuN" then this function (it support
+ *       both SCM and DCM) can be used for the inventory processor object
+ *       path validation.
+ */
+inline bool
+    isProcObjectMatched(const std::string& processorId,
+                        const sdbusplus::message::object_path& objectPath)
+{
+    bool isMatched = false;
+    if (processorId.find("dcm") != std::string::npos)
+    {
+        std::size_t delimiterPos = processorId.find('-');
+        if (delimiterPos != std::string::npos)
+        {
+            std::string procParent = processorId.substr(0, delimiterPos);
+            std::string procId =
+                processorId.substr(delimiterPos + 1, processorId.length());
+
+            if ((objectPath.parent_path().filename() == procParent) &&
+                (objectPath.filename() == procId))
+            {
+                isMatched = true;
+            }
+        }
+    }
+    else
+    {
+        if (objectPath.filename() == processorId)
+        {
+            isMatched = true;
+        }
+    }
+    return isMatched;
+}
+
+/**
  * @brief Fill out uuid info of a processor by
  * requesting data from the given D-Bus object.
  *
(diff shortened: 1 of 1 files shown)


RameshIyyar:
> * This PR is created on top of [Implement SubProcessors for processor core #95](https://github.com/ibm-openbmc/bmcweb/pull/95) so, please ignore the first commit in this PR.

Rebased this PR since #95 was merged

redfish-core/lib/processor.hpp
```diff
@@ -47,6 +47,51 @@ constexpr std::array<const char*, 2> processorInterfaces = {
...
+ */
+inline bool
+    isProcObjectMatched(const std::string& processorId,
+                        const sdbusplus::message::object_path& objectPath)
+{
+    bool isMatched = false;
+    if (processorId.find("dcm") != std::string::npos)
+    {
+        std::string procId{processorId};
+        if (auto pos = procId.find('-'); pos != std::string::npos)
```
gtmills, line 76: Would this be separated on the line above to match bmcweb style Also to use std::size_t

like

https://github.com/openbmc/bmcweb/blob/97a056ac02e0f246a24d1a4353a3d24b109ccf2e/include/authorization.hpp#L57

redfish-core/lib/processor.hpp
```diff
@@ -47,6 +47,51 @@ constexpr std::array<const char*, 2> processorInterfaces = {
...
+        {
+            procId.replace(pos, 1, "/");
+            if (objectPath.str.find(procId) != std::string::npos)
```
gtmills, line 79: Imprecise matching 
Common error 12. Can you use precise matching?

gtmills:
Looks okay. Couple minor things.

RameshIyyar:
> Looks okay. Couple minor things.

Addressed those comments, please take a look.

redfish-core/lib/processor.hpp
```diff
@@ -47,6 +47,55 @@ constexpr std::array<const char*, 2> processorInterfaces = {
...
+{
+    bool isMatched = false;
+    if (processorId.find("dcm") != std::string::npos)
+    {
+        std::size_t delimiterPos = processorId.find('-');
+        if (delimiterPos != std::string::npos)
+        {
+            std::string procParent{processorId.substr(0, delimiterPos)};
+            std::string procId{
+                processorId.substr(delimiterPos + 1, processorId.length())};
```
gtmills, line 80: Don't we need to check the size of processorId ?

https://github.com/openbmc/bmcweb/blob/master/COMMON_ERRORS.md#4-walking-off-the-end-of-a-string

Also, I know this is the same but I don't see "std::string procId{" anywhere else in bmcweb..

https://github.com/openbmc/bmcweb/search?q=substr

    std::string procId = processorId.substr(delimiterPos + 1, processorId.length())

RameshIyyar:
@gtmills please review

redfish-core/lib/processor.hpp
gtmills: > Not required, are you seeing any issue with that?

What if your processorId looks like "dcmN-" ? 
We have these checks not for the good case but for the bad case so bmcweb doesn't throw an exception 
Still think we should add a check 

ttps://github.com/openbmc/bmcweb/blob/master/COMMON_ERRORS.md#4-walking-off-the-end-of-a-string is this exact case

RameshIyyar: > What if your processorId looks like "dcmN-" ?

**procId** will be empty in that case and it won't match in below `if` and bmcweb will return below errors:
```
if ((objectPath.parent_path().filename() == procParent) &&
    (objectPath.filename() == procId))
{   
    isMatched = true;
}
```

Error:
```
curl -k -H "X-Auth-Token: $bmc_token" -X GET https://${bmc}/redfish/v1/Systems/system/Processors/dcm0-
{
  "error": {
    "@Message.ExtendedInfo": [
      {
        "Message": "The requested resource of type Processor named dcm0- was not found.",
        "MessageArgs": [
          "Processor",
          "dcm0-"
        ],
        "MessageId": "Base.1.8.1.ResourceNotFound",
        "Resolution": "Provide a valid resource identifier and resubmit the request."
      }
    ],
    "code": "Base.1.8.1.ResourceNotFound",
    "message": "The requested resource of type Processor named dcm0- was not found."
  }
}
```

gtmills (approved):

redfish-core/lib/processor.hpp
gtmills: Still think we need to check here, but not sure I have the energy to fight here

PR #103, opened 2021-09-16
Trigger panel lamp test

PriyangaRamasamy:
Whenever the lamp test request is sent to
the LedManager, panel lamp test needs to be handled.

Panel lamp test is nothing but validating the
back leds present in the lcd panel.

This can be done by calling the "TriggerPanelLampTest"
dbus api which is hosted by the panel service.

When the LampTest request is set to true in bmcweb,
along with other led tests(initiated by led manager),
"TriggerPanelLampTest" dbus method is called with input
set to true, which in turn sends the panel lamp test command
to the panel micro code which takes care of validating the
panel leds.

When the LampTest request is set to false in bmcweb,
led manager stops the lamp test abruptly. Similarly
the lamp test is made to stop in panel by calling
"TriggerPanelLampTest" dbus method with input set to
false, which in turn executes panel function 01 which
is the default display in panel.

Link to "TriggerPanelLampTest" dbus api : https://rchgit01.rchland.ibm.com/gerrit1/c/122940/1

Test:
Tested on rainier.

1.
curl -k -H "X-Auth-Token: $bmc_token" -X PATCH https://${bmc}/redfish/v1/Systems/system -d '{"Oem":{"IBM":{"LampTest": true}}}’

Journal Log:
Sep 15 17:30:35 rain71bmc bmcweb[1499]: (2021-09-15 17:30:35) [DEBUG "lamp_test.hpp":101] Set lamp test status.
Sep 15 17:30:35 rain71bmc bmcweb[1499]: (2021-09-15 17:30:35) [DEBUG "lamp_test.hpp":131] GetObjectType: xyz.openbmc_project.LED.GroupManager
Sep 15 17:30:35 rain71bmc phosphor-ledmanager[421]: Failed to set Asserted property
Sep 15 17:30:35 rain71bmc ibm-panel[1720]: Panel lamp test initiated.

2.
curl -k -H "X-Auth-Token: $bmc_token" -X PATCH https://${bmc}/redfish/v1/Systems/system -d '{"Oem":{"IBM":{"LampTest": false}}}’

Journal Log:
Sep 15 17:31:01 rain71bmc bmcweb[1499]: (2021-09-15 17:31:01) [DEBUG "lamp_test.hpp":101] Set lamp test status.
Sep 15 17:31:01 rain71bmc bmcweb[1499]: (2021-09-15 17:31:01) [DEBUG "lamp_test.hpp":131] GetObjectType: xyz.openbmc_project.LED.GroupManager
Sep 15 17:31:01 rain71bmc phosphor-ledmanager[421]: Failed to set Asserted property
Sep 15 17:31:01 rain71bmc ibm-panel[1720]: [1B blob data]
Sep 15 17:31:01 rain71bmc ibm-panel[1720]: L1 : 01     N    PVM
Sep 15 17:31:01 rain71bmc ibm-panel[1720]: L2 :             T

Signed-off-by: Priyanga Ramasamy <priyanga24@in.ibm.com>

### DIFF
--- a/redfish-core/lib/oem/ibm/lamp_test.hpp
+++ b/redfish-core/lib/oem/ibm/lamp_test.hpp
@@ -131,7 +131,7 @@ inline void setLampTestState(const std::shared_ptr<bmcweb::AsyncResp>& aResp,
             BMCWEB_LOG_DEBUG << "GetObjectType: " << service;
 
             crow::connections::systemBus->async_method_call(
-                [aResp](const boost::system::error_code ec) {
+                [aResp, state](const boost::system::error_code ec) {
                     if (ec)
                     {
                         BMCWEB_LOG_DEBUG
@@ -139,6 +139,18 @@ inline void setLampTestState(const std::shared_ptr<bmcweb::AsyncResp>& aResp,
                         messages::internalError(aResp->res);
                         return;
                     }
+                    crow::connections::systemBus->async_method_call(
+                        [aResp](const boost::system::error_code ec) {
+                            if (ec)
+                            {
+                                BMCWEB_LOG_DEBUG << "Panel Lamp test failed."
+                                                 << ec;
+                                messages::internalError(aResp->res);
+                                return;
+                            }
+                        },
+                        "com.ibm.PanelApp", "/com/ibm/panel_app",
+                        "com.ibm.panel", "TriggerPanelLampTest", bool(state));
                 },
                 service, "/xyz/openbmc_project/led/groups/lamp_test",
                 "org.freedesktop.DBus.Properties", "Set",


gtmills:
bmcweb already does a d-bus call on a lamp test, why can't this application listen for that? 

```
                service, "/xyz/openbmc_project/led/groups/lamp_test",
                "org.freedesktop.DBus.Properties", "Set",
                "xyz.openbmc_project.Led.Group", "Asserted",
```

redfish-core/lib/oem/ibm/lamp_test.hpp
```diff
@@ -131,14 +131,26 @@ inline void setLampTestState(const std::shared_ptr<bmcweb::AsyncResp>& aResp,
...
+                                BMCWEB_LOG_DEBUG << "Panel Lamp test failed."
+                                                 << ec;
+                                messages::internalError(aResp->res);
+                                return;
+                            }
+                        },
+                        "com.ibm.PanelApp", "/com/ibm/panel_app",
+                        "com.ibm.panel", "TriggerPanelLampTest", bool(state));
                 },
                 service, "/xyz/openbmc_project/led/groups/lamp_test",
```
gtmills, line 155: Why can't this application listen for this ?

gtmills (changes requested):

redfish-core/lib/oem/ibm/lamp_test.hpp
PriyangaRamasamy: Lamp Test api present in led manager does not do the lamp testing for panel lcd blocks (where there are back leds present for each of the panel lcd block). For that to happen a lamp test command needs to be sent to the panel's micro code. "TriggerPanelLampTest" dbus method does the latter.

so whenever there is a lamp test request initiated (from bmcweb/from panel) both - led manager's lamp test and panel lamp test command needs to be executed. thats why we have added this portion of panel lamp testing.

santoshpuranik: @gtmills : Let me elaborate a little - We came to this design after discussions with Vishwa and @lkammath , here's the rationale:

There is a use-case where one needs to restart lamp test. The LED manager handles a repeated "set" to the asserted property to restart the lamp test timer it runs. In such as case, property changed signal does not work because the property goes from true to true.

There were two options:
* Have the LED manager make a call to the panel app (this would be downstream code)
* Add another D-Bus call to the panel app from code in bmcweb that is anyway downstream only (this PR).

We chose the latter.

gtmills:
Let's discuss this in a design call

redfish-core/lib/oem/ibm/lamp_test.hpp
gtmills: I don't like this. I added this to our list to revisit for P11. 
I feel the Lamp Test Interface doesn't make a lot of sense as a bool. A true to true reset the timer. 
That said, approved, as we discussed, a lot of code wrote against it already.

gtmills (approved):

gtmills:
From the design call, this can go. We added to our list of things to revisit for P11

PR #106, opened 2021-09-16
Make fewer copies

geissonator:
There are apparently places where we make complete copies of the request
structure, including the body data within it.  Recently,
59b98b2222fddbea3d6f678d9e94006521f0c381 came along and made the size of
the Request structure include the body payload.  This is good for object
ownership, but bad for code that wants to make a copy.

This commit tries to do something less insane, and construct the Payload
object (which only really includes the http headers, uri and method)
before jumping into callback hell, thus meaning that the Payload object
is copied 4 times instead of the Request object.

Tested:
Code compiles.  Deleted the boost::beast::http::message copy
constructor, and observed that new code now compiles without needing a
copy.

Signed-off-by: Ed Tanous <edtanous@google.com>
Change-Id: Ie81bade61fb1f6f392b163f98c84540e09a26690

Upstream Commit: https://gerrit.openbmc-project.xyz/c/openbmc/bmcweb/+/46956

### DIFF
--- a/redfish-core/lib/update_service.hpp
+++ b/redfish-core/lib/update_service.hpp
@@ -66,7 +66,7 @@ inline static void activateImage(const std::string& objPath,
 static void
     softwareInterfaceAdded(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
                            sdbusplus::message::message& m,
-                           const crow::Request& req)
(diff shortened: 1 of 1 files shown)


geissonator:
I've confirmed that this fixed the issue.

PR #107, opened 2021-09-17
Redfish: Implement HealthRollup in chassis schema

ChicagoDuan:
The health status of the current CPUs and powersupplies will not rollup
to chassis. This submission implements rollup of the health status of
CPUs and powersupplies to chassis.

Tested:
We need add the xyz.openbmc_project.Inventory.Item.Global interface to the chassis.
I create a critical association of powersupply1 using the same method as phosphor-logging:
https://github.com/openbmc/phosphor-logging/blob/76198a2ea2e3deb14078e0d3dd3932b657ddd9b8/extensions/openpower-pels/service_indicators.cpp#L83

curl -k -H "X-Auth-Token: $token" -X GET   https://${bmc}/redfish/v1/Chassis/chassis
"Status": {
    "Health": "OK",
    "HealthRollup": "Critical",
    "State": "Enabled"
  },

### DIFF
--- a/redfish-core/lib/chassis.hpp
+++ b/redfish-core/lib/chassis.hpp
@@ -240,7 +240,7 @@ inline void requestRoutesChassis(App& app)
                                     return;
                                 }
                                 health->inventory = std::move(*data);
-                                constexpr const std::array<const char*, 4>
+                                constexpr const std::array<const char*, 13>
                                     inventoryForChassis = {
                                         "xyz.openbmc_project.Inventory.Item."
                                         "Dimm",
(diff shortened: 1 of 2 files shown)


ChicagoDuan:
Hi @patelabhishek9893 I have rebased it. Now there is no conflict. Please review this commit. Thanks

patelabhishek9893 (approved):

redfish-core/lib/health.hpp
```diff
@@ -126,7 +126,6 @@ struct HealthPopulate : std::enable_shared_from_this<HealthPopulate>
             if (boost::starts_with(path.str, globalInventoryPath) &&
                 boost::ends_with(path.str, "critical"))
             {
-                health = "Critical";
```
gtmills, line 129: Should you also remove health everywhere in this file?

ChicagoDuan: I don't think so. Other resources also use this file. They need to set health here.
https://github.com/ibm-openbmc/bmcweb/search?q=health-%3Epopulate

gtmills: I thought we said "health use operationalStatus" "healthrollup, use this design" ?

ChicagoDuan: > I thought we said "health use operationalStatus" "healthrollup, use this design" ?

For powersupply, fan, CPU and DIMM, it is right now.
But there are some old codes that still use the original design. For example, the health of storage, task and Ethernet. Do you want me to change all these old code into this design?

redfish-core/lib/chassis.hpp
```diff
@@ -240,6 +240,35 @@ inline void requestRoutesChassis(App& app)
...
                                 health->inventory = std::move(*data);
+                                constexpr const std::array<const char*, 4>
+                                    inventoryForChassis = {
+                                        "xyz.openbmc_project.Inventory.Item."
+                                        "Dimm",
+                                        "xyz.openbmc_project.Inventory.Item."
+                                        "Cpu",
+                                        "xyz.openbmc_project.Inventory.Item."
+                                        "PowerSupply",
+                                        "xyz.openbmc_project.Inventory.Item."
```
lkammath, line 251: As discussed please add for fan as well

ChicagoDuan: > As discussed please add for fan as well

Done.

lkammath:
Testing looks good on this PR. @ChicagoDuan @gtmills

geissonator:
jenkins run tests please

redfish-core/lib/health.hpp
gtmills: The problem with this is sometimes we have health being set by both this function and the operational status in the route
E.g.
https://github.com/openbmc/bmcweb/blob/11a2f0f0186983d5939351c276c60b054c059c5a/redfish-core/lib/memory.hpp#L450
https://github.com/openbmc/bmcweb/blob/11a2f0f0186983d5939351c276c60b054c059c5a/redfish-core/lib/memory.hpp#L482

Might need to do something better later here

gtmills (approved):

PR #117, opened 2021-09-27
Redfish-Validator Fix: Remove Bios/Settings url

sunharis:
Redfish validator fails for /redfish/v1/Systems/system/Bios/Settings
This commit fixes this by removing the refernce to this url while
performing GET on redfish/v1//Systems/system/Bios

Tested by: Run validator - pass

Change-Id: I0684cd069a07c9faeed48330a48c68f9052e9be8

### DIFF
--- a/redfish-core/lib/bios.hpp
+++ b/redfish-core/lib/bios.hpp
@@ -161,11 +161,6 @@ inline void requestRoutesBiosService(App& app)
                 // Get the ActiveSoftwareImage and SoftwareImages
                 fw_util::populateFirmwareInformation(
                     asyncResp, fw_util::biosPurpose, "", true);
-                asyncResp->res.jsonValue["@Redfish.Settings"] = {
-                    {"@odata.type", "#Settings.v1_3_0.Settings"},
-                    {"SettingsObject",
-                     {{"@odata.id",
-                       "/redfish/v1/Systems/system/Bios/Settings"}}}};
                 asyncResp->res.jsonValue["AttributeRegistry"] =
                     "BiosAttributeRegistry";
                 asyncResp->res.jsonValue["Attributes"] = {};


redfish-core/lib/bios.hpp
```diff
@@ -161,11 +161,6 @@ inline void requestRoutesBiosService(App& app)
                 // Get the ActiveSoftwareImage and SoftwareImages
                 fw_util::populateFirmwareInformation(
                     asyncResp, fw_util::biosPurpose, "", true);
-                asyncResp->res.jsonValue["@Redfish.Settings"] = {
-                    {"@odata.type", "#Settings.v1_3_0.Settings"},
-                    {"SettingsObject",
-                     {{"@odata.id",
-                       "/redfish/v1/Systems/system/Bios/Settings"}}}};
                 asyncResp->res.jsonValue["AttributeRegistry"] =
```
raviteja-b, line 164: need to remove BMCWEB_ROUTE registered for Get support  right?

  BMCWEB_ROUTE(app, "/redfish/v1/Systems/system/Bios/Settings")
        .privileges({{"ConfigureManager"}})
        .methods(boost::beast::http::verb::get)

raviteja-b (changes requested):

redfish-core/lib/bios.hpp
gtmills: Yeah, we should remove the GET and PATCH as well.

sunharis: GET is needed if a client is interested to see the Pending attributes on Bios - So i am not that sure if this can be removed.
PATCH is needed for a client to set the Bios attributes on the system - This must not be removed

redfish-core/lib/bios.hpp
```diff
@@ -161,11 +161,6 @@ inline void requestRoutesBiosService(App& app)
                 // Get the ActiveSoftwareImage and SoftwareImages
                 fw_util::populateFirmwareInformation(
                     asyncResp, fw_util::biosPurpose, "", true);
-                asyncResp->res.jsonValue["@Redfish.Settings"] = {
```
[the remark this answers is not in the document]
sunharis: ERROR - Unable to gather property info for URI /redfish/v1/Systems/system/Bios/Settings: AttributeError("'NoneType' object has no attribute 'get'")

sunharis:
Any comments @gtmills @asmithakarun ?

sunharis:
As per https://www.dmtf.org/sites/default/files/Redfish_School-BIOS_Configuration.pdf - the parameter "@Redfish.Settings" at the Bios response seems to be a valid one. This PR is not needed. The validator fix should be something else other than removing this parameter from the response. Closing this PR

PR #118, opened 2021-09-27
VMI Cert: Increase wait timeout for client certificate

raviteja-b:


rfrandse:
add to approvelist

gtmills:
Is this what we decided in the design meeting? 
Why 60 secs? 
That seems crazy long

include/ibm/management_console_rest.hpp
```diff
@@ -879,10 +882,11 @@ void handleCsrRequest(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
                 }
 
                 timeout->cancel();
+                asyncResp->res.addHeader("Retry-After", "60");
```
gtmills, line 885: I'm good with this, but we are saying. Service is unavailable, wait 60 secs and try again?

raviteja-b: yeah this is to ask client to retry after 60 secs

This is workaround for time being for HMC.
we are yet to decide on long term solution

raviteja-b:
Right now sometimes vmi is taking around  52 seconds for signing, we need some details from VMI team
for taking decision on long term solution.

this fix is to unblock  HMC for now

sunharis (approved):
This change is after the Host-BMC design discussion. Phyp/VMI are working on the root cause fix. The timeout can be moved back to 5 sec - after their fix is available.
Meanwhile, HMC is working on handling the 503 error code and will be retrying the operation instead of moving to incomplete state.

include/ibm/management_console_rest.hpp
```diff
@@ -833,7 +835,7 @@ void handleCsrRequest(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
         std::make_shared<boost::asio::steady_timer>(
             crow::connections::systemBus->get_io_context());
 
-    timeout->expires_after(std::chrono::seconds(5));
+    timeout->expires_after(std::chrono::seconds(60));
```
gtmills, line 838: Why can't we do 58 like we talked about?

gtmills:
Do we have an issue to address this later?

raviteja-b:
we are going to use this defect for fix https://w3.rchland.ibm.com/projects/bestquest/?defect=SW528890

gtmills:
Can we track with a ibm-openbmc issue too?

raviteja-b:
sure will create one, can  you please approve now,  lets get this into today's driver
HMC badly blocked on this

PR #122, opened 2021-09-30
Added all Redfish schemas

patelabhishek9893:
This commit allows downstream eBMC to list out all the Redfish schemas for all pulls. 

change:
 remove `include_list` from scripts/update_schemas.py file, which builds schemas file for all schemas. 

Why:
So now all pulls have all the schemas.
That allows the user(dev) to run the validator offline and eliminate any network glitch, which is the Redfish recommended way.

This is a downstream commit only

### DIFF
--- a/scripts/update_schemas.py
+++ b/scripts/update_schemas.py
@@ -11,113 +11,6 @@ import xml.etree.ElementTree as ET
 
 VERSION = "DSP8010_2021.2"
 
-# To use a new schema, add to list and rerun tool
-include_list = [
-    'AccountService',
-    'ActionInfo',
-    'Assembly',
-    'AttributeRegistry',
-    'Bios',
-    'Certificate',
-    'CertificateCollection',
-    'CertificateLocations',
-    'CertificateService',
-    'Chassis',
-    'ChassisCollection',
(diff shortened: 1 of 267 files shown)


patelabhishek9893:
This should go in after 2021.2 @gtmills

scripts/update_schemas.py
```diff
@@ -295,7 +178,7 @@
     basename = schema + "." + version + ".json"
     zip_filepath = os.path.join(VERSION, VERSION, "json-schema", basename)
     schemadir = os.path.join(json_schema_path, schema)
-    os.makedirs(schemadir)
+    os.makedirs(schemadir, exist_ok=True)
```
gtmills, line 181: Why add this?

patelabhishek9893: without that it now overwriting that folder (we delete that file), because the folder is still there even we delete all files.

patelabhishek9893: I was getting errors, now I tried that without `exist_ok=True` that error is gone that mean we are deleting that folder now

so let me remove that.

gtmills (approved):

geissonator (approved):

PR #128, opened 2021-10-04
Added Hypervisor Serial Socket

patelabhishek9893:
This commit adds `wss://${window.location.host}/console1` WebSocket
into bmcweb. it retrieves 'PHYP debug shell' info and displays
WebSocket data into WebUI using SOL (Serial over LAN) console.

'console1' read data from UNIX domain socket '@obmc-console.hypervisor'
and display it in WebUI.

Privilege set to {"ConfigureComponents", "ConfigureManager"}.
That mean only Administrator and Operator can retrieve console data.

Tested:
  configure UNIX domain socket '@obmc-console.hypervisor' with
  'wss://${window.location.host}/console0' (which currently use for
  retrieve 'host serial console data on SOL'). The test is successful,
  able to receive 'PHYP debug shell' data in SOL on WebUI.

Signed-off-by: Abhishek Patel <Abhishek.Patel@ibm.com>
Change-Id: Icece6ceddd0eeab8db7ebf1e5e863c22a59ae45f

### DIFF
--- a/include/obmc_console.hpp
+++ b/include/obmc_console.hpp
@@ -153,5 +153,45 @@ inline void requestRoutes(App& app)
             doWrite();
         });
 }
+
+inline void requestRoutesHypervisor(App& app)
+{
+    BMCWEB_ROUTE(app, "/console1")
+        .privileges({{"ConfigureComponents", "ConfigureManager"}})
+        .websocket()
+        .onopen([](crow::websocket::Connection& conn,
+                   const std::shared_ptr<bmcweb::AsyncResp>&) {
+            BMCWEB_LOG_DEBUG << "Connection " << &conn << " opened";
+
+            sessions.insert(&conn);
+            if (hostSocket == nullptr)
+            {
+                const std::string consoleName("\0obmc-console.hypervisor", 24);
+                boost::asio::local::stream_protocol::endpoint ep(consoleName);
+
+                hostSocket = std::make_unique<
+                    boost::asio::local::stream_protocol::socket>(
+                    conn.getIoContext());
+                hostSocket->async_connect(ep, connectHandler);
+            }
+        })
+        .onclose([](crow::websocket::Connection& conn,
+                    [[maybe_unused]] const std::string& reason) {
+            BMCWEB_LOG_INFO << "Closing websocket. Reason: " << reason;
+
+            sessions.erase(&conn);
+            if (sessions.empty())
+            {
+                hostSocket = nullptr;
+                inputBuffer.clear();
+                inputBuffer.shrink_to_fit();
+            }
+        })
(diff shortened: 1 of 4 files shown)


gtmills:
I opened https://github.com/ibm-openbmc/dev/issues/3349 to track moving this to the correct permissions

gtmills:
For upstream we need to do something like 

```

We need to come up with a better idea than hardcoding configurable names (obmc-console.guests) into bmcweb. In my view the socket name should be an implementation detail between obmc-console-server and obmc-console-client.

This may take the form of adding some DBus support to obmc-console and providing an interface to acquire a connected socket e.g. via a method call.

```

For downstream and the coming deadline, I think this is good enough.

gtmills:
https://github.ibm.com/openbmc/openbmc/pull/1416 needs to go first

meson_options.txt
```diff
@@ -15,6 +15,7 @@ option('rest', type : 'feature', value : 'enabled', description : '''Enable Phos
 option('event-subscription', type : 'feature', value : 'enabled', description: 'Enable Event Subscription through websocket')
 option('redfish', type : 'feature',value : 'enabled', description: 'Enable Redfish APIs.  Paths are under \'/redfish/v1/\'. See https://github.com/openbmc/bmcweb/blob/master/DEVELOPING.md#redfish.')
 option('host-serial-socket', type : 'feature', value : 'enabled', description : 'Enable host serial console WebSocket. Path is \'/console0\'.  See https://github.com/openbmc/docs/blob/master/console.md.')
+option('hypervisor-serial-socket', type : 'feature', value : 'enabled', description : 'Enable hypervisor serial console WebSocket. Path is \'/console1\'.')
```
sbteeks, line 18: Should this be enabled by default for all platforms?

sbteeks (changes requested):

meson_options.txt
gtmills: Upstream should be disabled, downstream since we only build for tacoma and rainier this is probably okay but it would be better if we disable this and only enable for those systems

gtmills:
This is disabled by default. This can go first.

include/obmc_console.hpp
```diff
@@ -153,5 +153,45 @@ inline void requestRoutes(App& app)
             doWrite();
         });
 }
+
+inline void requestRouteshypervisor(App& app)
```
gtmills, line 157: This should be requestRoutesHypervisor

gtmills (approved):

joseph-reynolds (approved):
Looks good to me, merge it!  I'll do the authority change later.

gtmills:
@rfrandse This can go

PR #138, opened 2021-10-08
1020 100821 add error traces to bmcweb

geissonator:
Per a SoS meeting, enabling the error traces in bmcweb to better be able to debug issues. This is not uptreamable as the discussion there has continued to be that by default, no logging is enabled.

### DIFF
--- a/http/http_client.hpp
+++ b/http/http_client.hpp
@@ -510,6 +510,8 @@ class HttpClient : public std::enable_shared_from_this<HttpClient>
     {
         if ((state == ConnState::suspended) || (state == ConnState::terminated))
         {
+            BMCWEB_LOG_ERROR
+                << "sendData: ConnState is suspended or terminated";
             return;
         }
         requestDataQueue.push_back(data);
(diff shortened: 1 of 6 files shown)


gtmills:
So we want to enable BMCWEB_LOG_ERROR tracing? 
Why can't we push this upstream? Make an option that works ?
Maybe https://github.com/ibm-openbmc/bmcweb/blob/master/meson_options.txt#L23 needs to be a enum? 

It looks like several of these should you changed to BMCWEB_LOG_DEBUB, why can't that also be upstream?

gtmills:
From @mzipse we need this now. 
@gtmills's squad will know driving this in upstream. 
I opened https://github.com/ibm-openbmc/dev/issues/3387 to track this work

gtmills (approved):
Although I would prefer we drive this upstream, I understand why we need this now.

mzipse:
The decision has been made to pull this change in for now.  This is handy for lab debug now.  However we still need to look at the right way to do this upstream and for GA1.  @sunharis will be opening a github issue to track.  In addition, I see Gunnar opened https://github.com/ibm-openbmc/dev/issues/3387 as well.

PR #150, opened 2021-10-20
Remove VLAN support from Redfish interface

sunharis:
IBM systems does not support vlan configurations on the BMC
network. The upstream bmcweb has CRUD support for VLANs

Tested by:
 1. Verified VLAN resource is not listed while GET on the eth interface
 2. Other operations on eth interface works fine

Change-Id: I82bfe435f893327edb85a8be34bc4ca26f61cded


sunharis:
> @sunharis Do we have other repos we need to remove this from?

This activity needs to be done by IPMI as well - if they support the vlan configuration. There is no plan to remove the backend - systemd support as that is complicated and impact could be on the regular interface settings.
We need this back on P11 - where we support vlans on the switches.

PR #152, opened 2021-10-21
Redfish : Look for OperationalStatus for the memory Health

ChicagoDuan:

Tested: Validator passes.
{
  "Status": {
    "Health": "Critical",
    "State": "Enabled"
  }
}

### DIFF
--- a/redfish-core/lib/memory.hpp
+++ b/redfish-core/lib/memory.hpp
@@ -15,8 +15,6 @@
 */
 #pragma once
 
-#include "health.hpp"
-
 #include <app.hpp>
 #include <boost/container/flat_map.hpp>
 #include <boost/format.hpp>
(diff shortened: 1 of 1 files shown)


redfish-core/lib/memory.hpp
```diff
@@ -460,10 +458,6 @@ inline void getDimmDataByService(std::shared_ptr<bmcweb::AsyncResp> aResp,
                                  const std::string& service,
                                  const std::string& objPath)
 {
-    auto health = std::make_shared<HealthPopulate>(aResp);
```
RameshIyyar, line 463: HealthRollup is not required for the DIMM in the OpenPOWER platforms like upstream?

gtmills:
@ChicagoDuan Please for next time, use a branch that describes the feature. E.g. Memory-look-for-OperationalStatus or use-operationalstatus-for-memory-health or combination like that.
dzj1020 doesn't describe the feature

gtmills:
This seems reasonable to me but want @lkammath to sign off. 
This is what we do for other resources

PR #154, opened 2021-10-28
OBMC_SHELL web-base_shell

patelabhishek9893:

Path of WebSocket:
    wss://{hostname}/bmc-console

Implementation:

So when bmcweb WebSocket gets a request for a new connection, it creates a child process of bmcweb using "/bin/sh" via calling bp::Child, which is similar to call exec("/bin/sh").

bmcweb is communicating with that child process using async_pipe. So the shell can execute all commands like ls, cd, etc.

Here shell WebSocket only retrieves individual charactor at a time.

WebSocket gets 'l', 's', and '\n' three char separately for the ls command. So bmcweb buffer them and pass to the child's process when it gets '\n'. In return child process return the output of that command. Which bmcweb return to the WebSocket.

Authorize user for Shell :

Only service users can access this interface, which means only managers can access this interface.

Test:

The essential operation can perform on this xterm console, Ex: ls, cd, grep, etc.

We can also execute the interactive command, like "top", but it doesn't perform well. Refresh time is slow, and the kill signal does not work using a terminal.

### DIFF
--- a/http/obmc_shell.hpp
+++ b/http/obmc_shell.hpp
@@ -1,0 +1,245 @@
+#pragma once
+
+#include <pty.h>
+#include <pwd.h>
+#include <termios.h>
+
+#include <app.hpp>
+#include <boost/asio/posix/stream_descriptor.hpp>
+#include <boost/process/io.hpp>
+#include <websocket.hpp>
+
+#include <map>
+
+namespace crow
+{
+
+namespace obmc_shell
+{
+
+class Handler : public std::enable_shared_from_this<Handler>
+{
+  public:
+    Handler(crow::websocket::Connection* conn) :
+        session(conn), streamFileDescriptor(conn->getIoContext()),
+        doingWrite(false)
+    {}
+
+    ~Handler() = default;
+
+    void doClose()
+    {
+        streamFileDescriptor.close();
+
+        // boost::process::child::terminate uses SIGKILL, need to send SIGTERM
+        int rc = kill(pid, SIGKILL);
+        session = nullptr;
+        if (rc)
+        {
+            return;
+        }
+        waitpid(pid, nullptr, 0);
+    }
+
+    void connect()
+    {
+        pid = forkpty(&ttyFileDescriptor, nullptr, nullptr, nullptr);
+
+        if (pid == -1)
+        {
+            // ERROR
+            if (session)
+            {
+                session->close("Error creating ssh child process");
+            }
+            return;
+        }
+        if (pid == 0)
+        {
+            // CHILD
+
+            // sets the effective user ID
+            // as service user
+            bool isUidSet = false;
+            if (auto userName = session->getUserName(); !userName.empty())
+            {
+                if (struct passwd* pw = getpwnam(userName.c_str());
+                    pw != nullptr)
+                {
+                    int uidr = setuid(pw->pw_uid);
+                    if (!uidr)
+                    {
+                        isUidSet = true;
+                    }
+                    else
+                    {
+                        BMCWEB_LOG_ERROR << "sets service user failed Error: "
+                                         << uidr;
+                    }
+                }
+                else
+                {
+                    BMCWEB_LOG_ERROR << "getpwnam return null passwd";
+                }
+            }
+            else
+            {
+                BMCWEB_LOG_ERROR << "userName is empty";
+            }
+
+            // close connection unable to set
+            // setuid()
+            if (!isUidSet)
+            {
+                session->close("sets service user failed");
+                return;
+            }
+
+            // create /bin/sh chiled process
+            execl("/bin/sh", "/bin/sh", nullptr);
+            return;
+        }
+        if (pid > 0)
+        {
+            // PARENT
+
+            // for io operation assing file discriptor
+            // to boost stream_descriptor
+            streamFileDescriptor.assign(ttyFileDescriptor);
+            doWrite();
+            doRead();
+        }
+    }
+
+    void doWrite()
+    {
+        if (!session)
+        {
+            BMCWEB_LOG_DEBUG << "session is closed";
+            return;
+        }
+        if (doingWrite)
+        {
+            BMCWEB_LOG_DEBUG << "Already writing.  Bailing out";
+            return;
+        }
+
+        if (inputBuffer.empty())
+        {
+            BMCWEB_LOG_DEBUG << "inputBuffer empty.  Bailing out";
+            return;
+        }
+
+        doingWrite = true;
+        streamFileDescriptor.async_write_some(
+            boost::asio::buffer(inputBuffer.data(), inputBuffer.size()),
+            [this, self(shared_from_this())](boost::beast::error_code ec,
+                                             std::size_t bytesWritten) {
+                BMCWEB_LOG_DEBUG << "Wrote " << bytesWritten << "bytes";
+                doingWrite = false;
+                inputBuffer.erase(0, bytesWritten);
+                if (ec == boost::asio::error::eof)
+                {
+                    session->close("ssh socket port closed");
+                    return;
+                }
+                if (ec)
+                {
+                    session->close("Error in writing to processSSH port");
+                    BMCWEB_LOG_ERROR << "Error in ssh socket write " << ec;
+                    return;
+                }
+                doWrite();
+            });
+    }
+
+    void doRead()
+    {
+        if (!session)
+        {
+            BMCWEB_LOG_DEBUG << "session is closed";
+            return;
+        }
+        streamFileDescriptor.async_read_some(
+            boost::asio::buffer(outputBuffer.data(), outputBuffer.size()),
+            [this, self(shared_from_this())](
+                const boost::system::error_code& ec, std::size_t bytesRead) {
+                BMCWEB_LOG_DEBUG << "Read done.  Read " << bytesRead
+                                 << " bytes";
+                if (!session)
+                {
+                    BMCWEB_LOG_DEBUG << "session is closed";
+                    return;
+                }
+                if (ec)
+                {
+                    BMCWEB_LOG_ERROR << "Couldn't read from ssh port: " << ec;
+                    session->close("Error in connecting to ssh port");
+                    return;
+                }
+                std::string_view payload(outputBuffer.data(), bytesRead);
+                session->sendBinary(payload);
+                doRead();
+            });
+    }
+
+    // this has to public
+    std::string inputBuffer;
+
+  private:
+    crow::websocket::Connection* session;
+    boost::asio::posix::stream_descriptor streamFileDescriptor;
+    bool doingWrite;
+    int ttyFileDescriptor;
+    pid_t pid;
+
+    std::array<char, 4096> outputBuffer;
+};
+
+static std::map<crow::websocket::Connection*, std::shared_ptr<Handler>>
+    mapHandler;
+
+inline void requestRoutes(App& app)
+{
+    BMCWEB_ROUTE(app, "/bmc-console")
+        .privileges({{"ConfigureManager"}})
+        .websocket()
+        .onopen([](crow::websocket::Connection& conn,
+                   const std::shared_ptr<bmcweb::AsyncResp>&) {
+            BMCWEB_LOG_DEBUG << "Connection " << &conn << " opened";
+            if (auto it = mapHandler.find(&conn); it == mapHandler.end())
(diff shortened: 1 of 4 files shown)


http/obmc_ssh.hpp
```diff
@@ -0,0 +1,221 @@
...
+                    session->close("Error in writing to processSSH port");
+                    BMCWEB_LOG_ERROR << "Error in ssh socket write " << ec;
+                    return;
+                }
+                doWrite();
+            });
+    }
+
+    void doRead()
+    {
```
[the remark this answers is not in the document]
patelabhishek9893: If the client passes "rm -rf /," should we put any thought into stopping that?
Excellent question, but it looks like only the manager can access this, so I think it can perform the rm command.

But still, I am not sure other folks need to answer this question.

http/obmc_ssh.hpp
```diff
@@ -0,0 +1,228 @@
...
+
+static std::shared_ptr<Handler> handler;
+
+inline void requestRoutes(App& app)
+{
+    BMCWEB_ROUTE(app, "/bmc-console")
+        .privileges({{"ConfigureManager"}})
+        .websocket()
+        .onopen([](crow::websocket::Connection& conn,
+                   const std::shared_ptr<bmcweb::AsyncResp>&) {
```
gtmills, line 173: Is this used?

http/obmc_ssh.hpp
```diff
@@ -0,0 +1,228 @@
...
+                return;
+            }
+
+            session = &conn;
+
+            handler = std::make_shared<Handler>(conn.getIoContext());
+            handler->connect();
+        })
+        .onclose([](crow::websocket::Connection& conn,
+                    const std::string& /*reason*/) {
```
gtmills, line 194: Should we log something here? 
Make this easier to debug?
https://github.com/openbmc/bmcweb/blob/9a128eb3fc3008f802974360e98de6495ed1a66b/include/nbd_proxy.hpp#L423

http/obmc_ssh.hpp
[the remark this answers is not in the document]
amboar: Do not filter commands.

amboar (changes requested):
The description seems confused with respect to the implementation.

SSH is a network protocol. It's not involved here. This patch is forking a shell and plumbing the IO through a websocket over TLS. There's not SSH involved. Please reword the description and rename the header so it's clear that SSH is not involved.

http/obmc_ssh.hpp
patelabhishek9893: no

gtmills:
jenkins run tests please

http/obmc_shell.hpp
```diff
@@ -0,0 +1,199 @@
...
+                    mapHandler.erase(it);
+                }
+            })
+        .onmessage([]([[maybe_unused]] crow::websocket::Connection& conn,
+                      const std::string& data, [[maybe_unused]] bool isBinary) {
+            if (auto it = mapHandler.find(&conn); it != mapHandler.end())
+            {
+                it->second->inputBuffer += data;
+                it->second->doWrite();
+            }
```
aahmed-2, line 194: Add an error message here? If we have received a onmessage call then connection should definitely exist, if it doesn't maybe the connection was removed.

http/obmc_shell.hpp
```diff
@@ -0,0 +1,199 @@
...
+            // ERROR
+            if (session)
+            {
+                session->close("Error creat child processSSH");
+            }
+            return;
+        }
+        if (pid == 0)
+        {
+            // CHILD
```
joseph-reynolds, line 58: At this point in the code flow (AFTER the child process is created and BEFORE the shell is exec'd) the process should change away from the "root" user to the "service" user.
If this is not done, the shell will run as the root user.  The correct behavior is for the shell to run as the user who invoked the URI, that is, the "service" user.
A testcase for this is to create a shell with this new function and invoke the command `echo $UID` to learn which user is running.  0 means root.  (Note you can see user ids in the /etc/passwd file.)  The "service" user has UID 1001.
I think this should be fixed before this is merged.

http/obmc_shell.hpp
```diff
@@ -0,0 +1,199 @@
...
+    void connect()
+    {
+        pid = forkpty(&ttyFileDescriptor, nullptr, nullptr, nullptr);
+
+        if (pid == -1)
+        {
+            // ERROR
+            if (session)
+            {
+                session->close("Error creat child processSSH");
```
aahmed-2, line 52: nitpick. maybe "Error creating ssh child process."

http/obmc_shell.hpp
```diff
@@ -0,0 +1,199 @@
...
+    std::string inputBuffer;
+};
+
+static std::map<crow::websocket::Connection*, std::shared_ptr<Handler>>
+    mapHandler;
+
+inline void requestRoutes(App& app)
+{
+    BMCWEB_ROUTE(app, "/bmc-console")
+        .privileges({{"ConfigureManager"}})
```
joseph-reynolds, line 163: The privilege is correct for now, but will need to change to the OemIBMPerformService privilege when that is available.

http/obmc_shell.hpp
patelabhishek9893: code updated and tested

http/obmc_shell.hpp
```diff
@@ -0,0 +1,214 @@
...
+        .onopen([](crow::websocket::Connection& conn,
+                   const std::shared_ptr<bmcweb::AsyncResp>&) {
+            BMCWEB_LOG_DEBUG << "Connection " << &conn << " opened";
+            if (auto it = mapHandler.find(&conn); it == mapHandler.end())
+            {
+                auto insertData =
+                    mapHandler.emplace(&conn, std::make_shared<Handler>(&conn));
+                if (std::get<bool>(insertData))
+                {
+                    std::get<0>(insertData)->second->connect();
```
joseph-reynolds, line 185: At this point, you should pass the username of the authenticated user into the connect() method.
This is needed so the new command shell will run as the user who initiated the connection.
You can get the username from `conn->getUserName()`.  Reference: https://github.com/openbmc/bmcweb/blob/master/http/websocket.hpp

http/obmc_shell.hpp
```diff
@@ -0,0 +1,214 @@
...
+            }
+            return;
+        }
+        if (pid == 0)
+        {
+            // CHILD
+
+            // sets the effective user ID
+            // as service user
+            int uidr = setuid(1001);
```
joseph-reynolds, line 62: This is a step in the right direction.  We should get the UID from the username.  You can use the `getpwnam` function to do this.  For example, as described here: https://stackoverflow.com/questions/39157675/how-to-get-linux-user-id-by-user-name  - but with "service" instead of "root".

Also, do not hard-code the name "service".  Instead please pass it into this method as described in the comment below.

http/obmc_shell.hpp
```diff
@@ -0,0 +1,242 @@
...
+                    session->close("Error in connecting to ssh port");
+                    return;
+                }
+                std::string_view payload(outputBuffer.data(), bytesRead);
+                session->sendBinary(payload);
+                doRead();
+            });
+    }
+
+    crow::websocket::Connection* session;
```
gtmills, line 186: Some of these shouldn't be protected or private?
E.g. https://github.com/openbmc/bmcweb/blob/9062d478d4dc89598e215e1538ba8fbb8db2cf10/include/kvm_websocket.hpp#L66

http/obmc_shell.hpp
```diff
@@ -0,0 +1,242 @@
...
+                                 << reason << "')";
+                if (auto it = mapHandler.find(&conn); it != mapHandler.end())
+                {
+                    it->second->doClose();
+                    mapHandler.erase(it);
+                }
+            })
+        .onmessage([]([[maybe_unused]] crow::websocket::Connection& conn,
+                      const std::string& data, [[maybe_unused]] bool isBinary) {
+            if (auto it = mapHandler.find(&conn); it != mapHandler.end())
```
gtmills, line 229: What about a size check like https://github.com/openbmc/bmcweb/blob/9062d478d4dc89598e215e1538ba8fbb8db2cf10/include/kvm_websocket.hpp#L44

http/obmc_shell.hpp
```diff
@@ -0,0 +1,242 @@
...
+
+namespace crow
+{
+
+namespace obmc_shell
+{
+
+class Handler : public std::enable_shared_from_this<Handler>
+{
+  public:
```
gtmills, line 22: Do you need a max sessions? 4 ? 10 ?

http/obmc_shell.hpp
```diff
@@ -0,0 +1,242 @@
...
+                    pw != nullptr)
+                {
+                    int uidr = setuid(pw->pw_uid);
+                    if (!uidr)
+                    {
+                        isUidSet = true;
+                    }
+                    else
+                    {
+                        BMCWEB_LOG_ERROR << "sets service user failed Error: "
```
gtmills, line 76: Don't you want to return in these error cases and not keep going?

http/obmc_shell.hpp
patelabhishek9893: for input buffer, we are using string so there is no fixed size I am using
I think we are not needed if I am not wrong

http/obmc_shell.hpp
patelabhishek9893: right now there is no limit defined. 
do we need a limit?

http/obmc_shell.hpp
patelabhishek9893: isUidSet will not set true so it will retun after the closing session 
` if (!isUidSet)
            {
                session->close("sets service user failed");
                return;
            }
`

http/obmc_shell.hpp
```diff
@@ -0,0 +1,245 @@
...
+    void doClose()
+    {
+        streamFileDescriptor.close();
+
+        // boost::process::child::terminate uses SIGKILL, need to send SIGTERM
+        int rc = kill(pid, SIGKILL);
+        session = nullptr;
+        if (rc)
+        {
+            return;
```
aahmed-2, line 39: If kill returns a non-zero value, should we just return? If fail to kill process, then will we leave a dangling process?

Reasons why kill will return non-zero value.

ERRORS         top
       EINVAL An invalid signal was specified.

       EPERM  The calling process does not have permission to send the
              signal to any of the target processes.

       ESRCH  The target process or process group does not exist.  Note
              that an existing process might be a zombie, a process that
              has terminated execution, but has not yet been wait(2)ed
              for.

http/obmc_shell.hpp
```diff
@@ -0,0 +1,245 @@
...
+    ~Handler() = default;
+
+    void doClose()
+    {
+        streamFileDescriptor.close();
+
+        // boost::process::child::terminate uses SIGKILL, need to send SIGTERM
+        int rc = kill(pid, SIGKILL);
+        session = nullptr;
+        if (rc)
```
aahmed-2, line 37: If kill returns a non-zero value, should we just return? If fail to kill process, then will we leave a dangling process I think. Maybe a try, catch block (but maybe that is too much work for this commit)

Reasons why kill will return non-zero value.

   ERRORS
   EINVAL An invalid signal was specified.

   EPERM  The calling process does not have permission to send the
          signal to any of the target processes.

   ESRCH  The target process or process group does not exist.  Note
          that an existing process might be a zombie, a process that
          has terminated execution, but has not yet been wait(2)ed
          for.

patelabhishek9893: EINVAL : when I was sending SIGTERM that time it becomes a zombi process, the only solution to fix this us reboot server. SIGKILL kills the process, so I don't think we are sending the wrong signal.

 EPERM : bmcweb run as the root user so I don't think so this is a problem.
 
 ESRCH : no idea about this but if this happened then yes we might have a zombie process. but I can think what is a solution for that.
 
 try and catch: for system-level call does try-catch work? because they do not throw errors they return errors.

gtmills:
@joseph-reynolds Could you look at this, approve this?

joseph-reynolds (approved):
Seems okay to me, modulo a subsequent change to allow only the service user to access this.  (And that is on my TODO list.)

aahmed-2 (approved):

gtmills (approved):
I am going to default mostly to Ali and Joseph here.

PR #158, opened 2021-11-04
Redfish: Remove powersupply asset error log

ChicagoDuan:
According to the discussion with @lkammath , if the powersupply does not have asset information, no error need be
reported because the mex power supply does not have asset information
on dbus.

### DIFF
--- a/redfish-core/lib/power_supply.hpp
+++ b/redfish-core/lib/power_supply.hpp
@@ -19,8 +19,6 @@ inline void
                         propertiesList) {
             if (ec)
             {
-                BMCWEB_LOG_ERROR << "Can't get PowerSupply asset!";
-                messages::internalError(asyncResp->res);
                 return;
             }
             for (const std::pair<std::string, std::variant<std::string>>&


gtmills:
Have we made this upstream?

gtmills:
jenkins run tests please

lkammath:
@gtmills do you think this should make it upstream since all these are optional as per redfish schema

redfish-core/lib/power_supply.hpp
```diff
@@ -19,8 +19,6 @@ inline void
                         propertiesList) {
             if (ec)
             {
-                BMCWEB_LOG_ERROR << "Can't get PowerSupply asset!";
```
gtmills, line 22: Can we check the ec here?
If not found, don't log an error

gtmills: Do we need to do this other places as well like LocaitonCode ?

lkammath:
@gtmills We have pushed  https://github.com/ibm-openbmc/bmcweb/pull/166 . This can be closed once #166 goes in

PR #160, opened 2021-11-10
Remove stale lock table entries of invalid sessionid

asmithakarun:
Locks are cleaned up when there is a graceful logout /
session expiration due to timeout. But there are possibilities
of leaving stale lock entries in the BMC, if the session is
abruptly terminated.

This commit is to check for the existence of the session
whenever there is a lock conflict. If the session is not active,
remove the lock record considering that as a stale record and
return no error. If the session is active, return the lock
conflict to the client.

It fixes the following defect:
https://w3.rchland.ibm.com/projects/bestquest/?verb=view&id=SW538023&table=defect

Change-Id: I1b97895cbdc4e602e00313bfd3b59e3b3445bd59

### DIFF
--- a/include/ibm/management_console_rest.hpp
+++ b/include/ibm/management_console_rest.hpp
@@ -499,7 +499,7 @@ inline void
 
     if (varAcquireLock.first)
     {
-        // Either validity failure of there is a conflict with itself
+        // Either validity failure or there is a conflict with itself
 
         auto validityStatus =
             std::get<std::pair<bool, int>>(varAcquireLock.second);
@@ -533,14 +533,47 @@ inline void
             asyncResp->res.jsonValue["TransactionID"] = var;
             return;
         }
-        BMCWEB_LOG_DEBUG << "There is a conflict with the lock table";
-        asyncResp->res.result(boost::beast::http::status::conflict);
+
+        // Check if the session is active. If active, send back the lock
+        // conflict, else clear the stale entry in the lock table.
         auto var =
             std::get<std::pair<uint32_t, LockRequest>>(conflictStatus.second);
+        std::string sessionId = std::get<0>(var.second);
+        auto session =
+            persistent_data::SessionStore::getInstance().getSessionByUid(
+                sessionId);
+
+        if (session == nullptr)
+        {
+            BMCWEB_LOG_ERROR << "Releasing lock of the session id(stale): "
+                             << sessionId;
+            // clear the stale entry
+            crow::ibm_mc_lock::Lock::getInstance().releaseLock(sessionId);
+
+            // create the new lock record for the session
+            varAcquireLock =
+                crow::ibm_mc_lock::Lock::getInstance().acquireLock(t);
+            conflictStatus =
+                std::get<crow::ibm_mc_lock::Rc>(varAcquireLock.second);
+            if (!conflictStatus.first)
+            {
+                asyncResp->res.result(boost::beast::http::status::ok);
+
+                auto var = std::get<uint32_t>(conflictStatus.second);
+                nlohmann::json returnJson;
+                returnJson["id"] = var;
+                asyncResp->res.jsonValue["TransactionID"] = var;
+                return;
+            }
+        }
+
+        BMCWEB_LOG_ERROR << "There is a conflict with the lock table";
+
+        asyncResp->res.result(boost::beast::http::status::conflict);
         nlohmann::json returnJson, segments;
         nlohmann::json myarray = nlohmann::json::array();
         returnJson["TransactionID"] = var.first;
-        returnJson["SessionID"] = std::get<0>(var.second);
+        returnJson["SessionID"] = sessionId;
         returnJson["HMCID"] = std::get<1>(var.second);
         returnJson["LockType"] = std::get<2>(var.second);
         returnJson["ResourceID"] = std::get<3>(var.second);


include/ibm/management_console_rest.hpp
```diff
@@ -533,14 +533,32 @@ inline void
...
         auto var =
             std::get<std::pair<uint32_t, LockRequest>>(conflictStatus.second);
+        std::string sessionId = std::get<0>(var.second);
+        auto session =
+            persistent_data::SessionStore::getInstance().getSessionByUid(
+                sessionId);
+
+        if (session == nullptr)
+        {
+            BMCWEB_LOG_DEBUG << "Releasing lock of the session id(stale): "
```
sunharis, line 548: Please change this to LOG_ERROR so that this gets printed in the journal

include/ibm/management_console_rest.hpp
```diff
@@ -533,14 +533,32 @@ inline void
...
+                             << sessionId;
+            // clear the stale entry
+            crow::ibm_mc_lock::Lock::getInstance().releaseLock(sessionId);
+            return;
```
sunharis, line 552: After releasing this stale lock - before returning the response you should continue processing the lock request - to actually set the lock in the lock table for this current command. 
By calling crow::ibm_mc_lock::Lock::getInstance().acquireLock(t) again, it will create a successful lock record for this new session

include/ibm/management_console_rest.hpp
```diff
@@ -533,14 +533,32 @@ inline void
...
+        }
+
+        BMCWEB_LOG_DEBUG << "There is a conflict with the lock table";
```
sunharis, line 555: Please convert this to ERROR log

sunharis (changes requested):

include/ibm/management_console_rest.hpp
asmithakarun: done

include/ibm/management_console_rest.hpp
asmithakarun: Done

include/ibm/management_console_rest.hpp
asmithakarun: Done

include/ibm/management_console_rest.hpp
```diff
@@ -533,14 +533,45 @@ inline void
...
+            varAcquireLock =
+                crow::ibm_mc_lock::Lock::getInstance().acquireLock(t);
+            conflictStatus =
+                std::get<crow::ibm_mc_lock::Rc>(varAcquireLock.second);
+            if (!conflictStatus.first)
+            {
+                auto var = std::get<uint32_t>(conflictStatus.second);
+                nlohmann::json returnJson;
+                returnJson["id"] = var;
+                asyncResp->res.jsonValue["TransactionID"] = var;
```
sunharis, line 563: asyncResp->res.result(boost::beast::http::status::ok); should be set

sunharis (changes requested):

include/ibm/management_console_rest.hpp
asmithakarun: Done

include/ibm/management_console_rest.hpp
```diff
@@ -533,14 +533,47 @@ inline void
...
+                crow::ibm_mc_lock::Lock::getInstance().acquireLock(t);
+            conflictStatus =
+                std::get<crow::ibm_mc_lock::Rc>(varAcquireLock.second);
+            if (!conflictStatus.first)
+            {
+                asyncResp->res.result(boost::beast::http::status::ok);
+
+                auto var = std::get<uint32_t>(conflictStatus.second);
+                nlohmann::json returnJson;
+                returnJson["id"] = var;
```
manojkiraneda, line 564: why was this needed ?

sunharis (approved):

include/ibm/management_console_rest.hpp
```diff
@@ -533,14 +533,47 @@ inline void
...
-        asyncResp->res.result(boost::beast::http::status::conflict);
+
+        // Check if the session is active. If active, send back the lock
+        // conflict, else clear the stale entry in the lock table.
         auto var =
             std::get<std::pair<uint32_t, LockRequest>>(conflictStatus.second);
+        std::string sessionId = std::get<0>(var.second);
+        auto session =
+            persistent_data::SessionStore::getInstance().getSessionByUid(
+                sessionId);
```
manojkiraneda, line 544: do we wanna do this for every conflicting request ?

manojkiraneda: cant we do it just once in the constructor, so we remove stale entries in one shot when bmcweb comes up ? rather than checking them every time ?

manojkiraneda (commented):
How was this tested ? can we get some testing via redfish or via unit tests ?

include/ibm/management_console_rest.hpp
sunharis: With the several defects data - what we saw is the locks are going stale while the bmcweb/bmc is running. Somehow the session from HMC is getting vanished with-out a proper cleanup path hitting at bmc. This is leaving stale locks. 
So we need a way to check if the conflicting lock is associated with any stale session OR and active session

include/ibm/management_console_rest.hpp
asmithakarun: so the following actually would have been reported as a lock conflict earlier
Now what it does is, removes the invalid session stale entry then creates a lock record for the current session and returns the json


curl -k -H "X-Auth-Token:$bmc_token" -X POST - "Content-type: application/json" -d '{ "Request" :[ { "LockType":"Write", "SegmentFlags": [ {"LockFlag":"DontLock","SegmentLength":1}, {"LockFlag":"DontLock","SegmentLength":2}, {"LockFlag":"LockAll","SegmentLength":1}], "ResourceID": 288241371267989504 } ]}' https://${bmc}/ibm/v1/HMC/LockService/Actions/LockService.AcquireLock
{
  "TransactionID": 2
}

manojkiraneda: no i was asking ,
```  nlohmann::json returnJson;
     returnJson["id"] = var;
```
why was this needed ?

include/ibm/management_console_rest.hpp
manojkiraneda: hmm, so this is ideally a bug in bmcweb where the sessions are removed in a path that we don't know (not logout/timeout) - this is essentially a workaround of that bug.

PR #162, opened 2021-11-12
1020 reduce error traces in bmcweb

geissonator:
One commit in this PR is upstream at https://gerrit.openbmc-project.xyz/c/openbmc/bmcweb/+/48795/

The other two commits are for code that we only have in downstream.

### DIFF
--- a/redfish-core/lib/chassis.hpp
+++ b/redfish-core/lib/chassis.hpp
@@ -136,8 +136,8 @@ inline void getPhysicalSecurityData(std::shared_ptr<bmcweb::AsyncResp> aResp)
             {
                 // do not add err msg in redfish response, because this is not
                 //     mandatory property
-                BMCWEB_LOG_ERROR << "DBUS error: no matched iface " << ec
-                                 << "\n";
+                BMCWEB_LOG_INFO << "DBUS error: no matched iface " << ec
+                                << "\n";
                 return;
             }
             // Iterate over all retrieved ObjectPaths.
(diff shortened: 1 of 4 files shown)


redfish-core/lib/chassis.hpp
```diff
@@ -136,8 +136,8 @@ inline void getPhysicalSecurityData(std::shared_ptr<bmcweb::AsyncResp> aResp)
             {
                 // do not add err msg in redfish response, because this is not
                 //     mandatory property
-                BMCWEB_LOG_ERROR << "DBUS error: no matched iface " << ec
-                                 << "\n";
+                BMCWEB_LOG_INFO << "DBUS error: no matched iface " << ec
```
gtmills, line 139: Since we don't have /xyz/openbmc_project/Intrusion it is hitting here

gtmills: I don't think I see this one upstream.. Can we upstream this one?

gtmills (approved):
Looks good, please just keep pushing those upstream

redfish-core/lib/chassis.hpp
geissonator: Hmm, I have this upstream at https://gerrit.openbmc-project.xyz/c/openbmc/bmcweb/+/48795/2/redfish-core/lib/chassis.hpp. Is that what you're asking or more about the intrusion sensor function?

gtmills: Got it. 
Nah, I am good. I missed it.

manojkiraneda (approved):

PR #171, opened 2021-11-19
Enabled deconfiguration reason support to the DIMM and Core and Few fixes

RameshIyyar:
This is downstream changes (no Gerrit links for the added patches)

Below are Changes included in this PR:
- **LogServices: HardwareIsolation: Fixed the "AdditionalDataURI"**
  - In the IBM platform system, recently the error entries are listed in the two different Redfish URI as mentioned below.
    - /redfish/v1/Systems/system/LogServices/EventLog/Entries/N
    - /redfish/v1/Systems/system/LogServices/CELog/Entries/N
  - But, the HardwareIsolation LogEntry "AdditionalDataURI" property (which is used to point the associated error log entry) using the old link i.e "EventLog" so if the user tried to access the hidden error log through that property then, that won't work so fixed that issue in this PR.
- **LogServices: HardwareIsolation: Fixed the LogEntryCollection**
  - Recently, the hardware isolation D-Bus service is modified to host another type of D-Bus object i.e Logging::Event so the HardwareIsolation LogEntryCollection returns all types of the object (aka resource or entry) instead of returning only the HardwareIsolation entries.

- **redfish-core: Memory & Core: Enabled the deconfiguration reason**
  - **Requirement**: In the IBM system platforms, we need to tell the reason and additional information to the external system user if any of DIMM or Core is not functional or whether is isolated from the system boot or is pending to isolate in the system.

### DIFF
--- a/redfish-core/include/registries/openbmc_message_registry.hpp
+++ b/redfish-core/include/registries/openbmc_message_registry.hpp
@@ -29,7 +29,7 @@ const Header header = {
     "0.2.0",
     "OpenBMC",
 };
-constexpr std::array<MessageEntry, 189> registry = {
+constexpr std::array<MessageEntry, 190> registry = {
     MessageEntry{
         "ADDDCCorrectable",
         {
@@ -2430,6 +2430,17 @@ constexpr std::array<MessageEntry, 189> registry = {
             "%1 Voltage Regulator Overheated.",
             "Critical",
             "Critical",
+            1,
+            {"string"},
+            "None.",
+        }},
+    MessageEntry{
+        "HardwareIsolationReason",
+        {
+            "Indicates the condition that affects the health of this resource.",
+            "The reason for the resource isolation: %1",
+            "OK",
+            "OK",
             1,
             {"string"},
             "None.",
--- a/redfish-core/include/registries/openbmc_message_registry.hpp
+++ b/redfish-core/include/registries/openbmc_message_registry.hpp
@@ -563,6 +563,17 @@ constexpr std::array<MessageEntry, 190> registry = {
             {"string"},
             "None.",
         }},
+    MessageEntry{
+        "HardwareIsolationReason",
+        {
+            "Indicates the condition that affects the health of this resource.",
+            "The reason for the resource isolation: %1",
+            "OK",
+            "OK",
+            1,
+            {"string"},
+            "None.",
+        }},
     MessageEntry{"InvalidLoginAttempted",
                  {
                      "Indicates that a login was attempted on the specified "
@@ -2430,17 +2441,6 @@ constexpr std::array<MessageEntry, 190> registry = {
             "%1 Voltage Regulator Overheated.",
             "Critical",
             "Critical",
-            1,
-            {"string"},
-            "None.",
-        }},
-    MessageEntry{
-        "HardwareIsolationReason",
-        {
-            "Indicates the condition that affects the health of this resource.",
-            "The reason for the resource isolation: %1",
-            "OK",
-            "OK",
             1,
             {"string"},
             "None.",
--- a/redfish-core/include/utils/hw_isolation.hpp
+++ b/redfish-core/include/utils/hw_isolation.hpp
@@ -1,5 +1,7 @@
 #pragma once
 
+#include <utils/error_log_utils.hpp>
+
 #include <string>
 #include <vector>
 
@@ -7,6 +9,11 @@ namespace redfish
 {
 namespace hw_isolation_utils
 {
+
+using AssociationsValType =
+    std::vector<std::tuple<std::string, std::string, std::string>>;
+using HwStausEventPropertiesType = boost::container::flat_map<
+    std::string, std::variant<std::string, uint64_t, AssociationsValType>>;
 
 /**
  * @brief API used to isolate the given resource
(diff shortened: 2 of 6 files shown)


RameshIyyar:
Ran Redfish validator - passed

redfish-core/include/registries/openbmc_message_registry.hpp
```diff
@@ -2434,6 +2434,17 @@ constexpr std::array<MessageEntry, 189> registry = {
             {"string"},
             "None.",
         }},
+    MessageEntry{
+        "HardwareIsolationReason",
```
gtmills, line 2438: Can we push this part upstream? 
Aren't these alphabetical ?

RameshIyyar: > Can we push this part upstream?

I think no since our guard feature is downstream. If you are ok, I can push into upstream (no reference for the backend code in the upstream). 

> Aren't these alphabetical ?

Ah, it is my mistake (did not realize that message registry is defined in the alphabetical order) I will fix it

RameshIyyar: >> Aren't these alphabetical ?

> Ah, it is my mistake (did not realize that message registry is defined in the alphabetical order) I will fix it

Done - Added the new commit `Fixup` (it can be merged into the previous commit without commit message) as per the latest PR process.

redfish-core/include/utils/hw_isolation.hpp
```diff
@@ -8,6 +10,11 @@ namespace redfish
 namespace hw_isolation_utils
 {
 
+using AssociationsValType =
```
asmithakarun, line 13: Should we put it in line 345, if we are using these only in the scope of that method?

asmithakarun (approved):
Other than that single comment, LGTM

redfish-core/include/utils/hw_isolation.hpp
RameshIyyar: Ack

gtmills:
Can you add more to the PR msg?

redfish-core/lib/log_services.hpp
```diff
@@ -4790,10 +4791,22 @@ inline void fillSystemHardwareIsolationLogEntry(
...
                             sdbusplus::message::object_path errPath =
                                 std::get<2>(assoc);
-                            entryJson["AdditionalDataURI"] =
-                                "/redfish/v1/Systems/system/"
-                                "LogServices/EventLog/Entries/" +
-                                errPath.filename() + "/attachment";
+
+                            if (entryJsonIdx > 0)
+                            {
+                                auto errorLogPropPath = "/Members"_json_pointer;
```
gtmills, line 4797: Nitpick: Typically this isn't auto

gtmills: Can you add a comment or 2?

redfish-core/include/utils/hw_isolation.hpp
```diff
@@ -335,5 +342,297 @@ inline void
...
+                                        messages::internalError(aResp->res);
+                                        break;
+                                    }
+                                }
+                            }
+                        },
+                        objType[0].first, hwStatusEventObj,
+                        "org.freedesktop.DBus.Properties", "GetAll", "");
+                },
+                "xyz.openbmc_project.ObjectMapper",
```
gtmills, line 627: Why 2 mapper calls here?

redfish-core/include/utils/hw_isolation.hpp
```diff
@@ -335,5 +342,297 @@ inline void
...
+                                    {
+                                        BMCWEB_LOG_ERROR
+                                            << "Failed to get the Severity "
+                                            << "from object: "
+                                            << hwStatusEventObj;
+                                        messages::internalError(aResp->res);
+                                        break;
+                                    }
+
+                                    if (*severity ==
```
gtmills, line 591: typically lines 589 - 618 are a separate function upstream

redfish-core/include/utils/hw_isolation.hpp
```diff
@@ -335,5 +342,297 @@ inline void
...
+                                    {
+                                        BMCWEB_LOG_ERROR
+                                            << "Failed to get the Associations "
+                                            << "from object: "
+                                            << hwStatusEventObj;
+                                        messages::internalError(aResp->res);
+                                        break;
+                                    }
+
+                                    for (auto& assoc : *associations)
```
gtmills, line 484: const ?

redfish-core/include/utils/hw_isolation.hpp
```diff
@@ -335,5 +342,297 @@ inline void
...
+                                    {
+                                        if (std::get<0>(assoc) == "error_log")
```
gtmills, line 486: Doesn't this throw?

redfish-core/include/utils/hw_isolation.hpp
```diff
@@ -335,5 +337,304 @@ inline void
...
+
+                    if (objType[0].first.empty())
+                    {
+                        BMCWEB_LOG_ERROR << "The retrieved hardware status "
+                                         << "event object dbus name is empty";
+                        messages::internalError(aResp->res);
+                        return;
+                    }
+
+                    using AssociationsValType = std::vector<
```
gtmills, line 432: Don't we already have this defined somewhere you can use ?

gtmills: E.g. see https://gerrit.openbmc-project.xyz/c/openbmc/bmcweb/+/47776/8/redfish-core/lib/systems.hpp

RameshIyyar:
> Can you add more to the PR msg? I am really just looking for the commit msgs to be copied to the PR msg and summarized

Thanks for pointing that, added now

redfish-core/include/utils/hw_isolation.hpp
gtmills: > Do you want to change that upstream and use it here?

Sure if you don't mind

redfish-core/include/utils/hw_isolation.hpp
gtmills: Typically upstream, for example: https://gerrit.openbmc-project.xyz/c/openbmc/bmcweb/+/43761/15/redfish-core/lib/systems.hpp

gtmills: I like our downstream commits to match upstream as much as possible for easier maintainability and one day I hope a lot of these downstream patches do go upstream

RameshIyyar: Hmm got it, I think I have addressed that [here](https://github.com/ibm-openbmc/bmcweb/pull/171/commits/56de3b4047b4648989473ddf5188737dd0a96384#diff-2b15c834283a368213b30e0c0e8e913e313e61bd8060d94a79d434b99f1e127eR353) but, with some minor changes to align with hardware isolation status use case i.e
- Returning flag if unsupported severity because we are throwing an internal error if any of attribute set is failed.
- Passing object path include in trace for better debug.
- Passing property path so that we can extend future while adding more than one condition.

redfish-core/include/utils/hw_isolation.hpp
RameshIyyar: @gtmills I just looked to add in the upstream but, looks nowhere using the "**Associations**" property and using only **endpoints** in the upstream. 

downstream code:
```
-bash-4.2$ git grep -in "\"Associations\""
redfish-core/include/utils/hw_isolation.hpp:523:                                if (property.first == "Associations")
redfish-core/lib/assembly.hpp:479:        "xyz.openbmc_project.Association.Definitions", "Associations");
redfish-core/lib/log_services.hpp:4768:                if (property.first == "Associations")
```
upstream code:
```
-bash-4.2$ git grep -in "\"Associations\""
-bash-4.2$ 
```

I am not sure without use in the upstream we can add so let me know if we can add.  Now marking as resolved and let me know still you have concerns.

redfish-core/include/utils/hw_isolation.hpp
RameshIyyar: Assuming this is clarified so marking as resolved.

redfish-core/include/utils/hw_isolation.hpp
RameshIyyar: Assuming this is clarified so marking as resolved.

redfish-core/include/utils/hw_isolation.hpp
```diff
@@ -335,5 +337,338 @@ inline void
...
+                                        std::get_if<std::string>(
+                                            &property.second);
+                                    if (severity == nullptr)
+                                    {
+                                        BMCWEB_LOG_ERROR
+                                            << "Failed to get the Severity "
+                                            << "from object: "
+                                            << hwStatusEventObj;
+                                        messages::internalError(aResp->res);
+                                        break;
```
gtmills, line 642: I think should be return which is more common in bmcweb, set an internalError and return

redfish-core/include/utils/hw_isolation.hpp
```diff
@@ -335,5 +337,338 @@ inline void
...
+
+                                    // we have only one condition
+                                    nlohmann::json_pointer severityPropPath =
+                                        "/Status/Conditions/0/Severity"_json_pointer;
+                                    if (!setSeverity(aResp, hwStatusEventObj,
+                                                     severityPropPath,
+                                                     *severity))
+                                    {
+                                        // Failed to set the severity
+                                        break;
```
gtmills, line 653: I think should be return which is more common in bmcweb, set an internalError and return

redfish-core/include/utils/hw_isolation.hpp
gtmills: How about return anytime we have an internalError ?

RameshIyyar: That's my expectation too but. The last time I got a comment from somebody as opposite way i.e. use break instead of return.
I will set an internal error and return it.

redfish-core/include/utils/hw_isolation.hpp
RameshIyyar: Done

redfish-core/include/utils/hw_isolation.hpp
RameshIyyar: Done

RameshIyyar (commented):
@gtmills addressed the comment, please take a look.

gtmills (approved):
Good enough. Ramesh will own any defects for this.

PR #166, opened 2021-11-16
PowerSubsystem and PowerSupply schema Refresh for MEX support

lkammath:
This PR has latest refresh commit 
https://gerrit.openbmc-project.xyz/c/openbmc/bmcweb/+/40326

1) PowerSubsystem and PowerSupply Schema updates for MEX chassis
2) Skip erroring out part based on EC

PR #174, opened 2021-11-25
Fix for not showing PowerSupplies schema under chassis

lkammath:
Signed-off-by: Lakshminarayana R. Kammath <lkammath@in.ibm.com>
#166 introduced regression which caused to not show PS schema under chassis because all was moved under macro BMCWEB_NEW_POWERSUBSYSTEM_THERMALSUBSYSTEM. 

needed to build bmcweb with BMCWEB_NEW_POWERSUBSYSTEM_THERMALSUBSYSTEM and rainier is not doing same so, moving back to original place

### DIFF
--- a/redfish-core/include/redfish.hpp
+++ b/redfish-core/include/redfish.hpp
@@ -86,9 +86,6 @@ class RedfishService
 #ifdef BMCWEB_NEW_POWERSUBSYSTEM_THERMALSUBSYSTEM
         requestRoutesThermalSubsystem(app);
         requestRoutesThermalMetrics(app);
-        requestRoutesPowerSubsystem(app);
-        requestRoutesPowerSupplyCollection(app);
-        requestRoutesPowerSupply(app);
 #endif
         requestRoutesManagerCollection(app);
         requestRoutesFan(app);
@@ -97,6 +94,9 @@ class RedfishService
         requestRoutesManagerResetAction(app);
         requestRoutesManagerResetActionInfo(app);
         requestRoutesManagerResetToDefaultsAction(app);
+        requestRoutesPowerSubsystem(app);
+        requestRoutesPowerSupplyCollection(app);
+        requestRoutesPowerSupply(app);
         requestRoutesPCIeSlots(app);
         requestRoutesEnvironmentMetrics(app);
         requestRoutesChassisCollection(app);


ChicagoDuan:
> needed to build bmcweb with BMCWEB_NEW_POWERSUBSYSTEM_THERMALSUBSYSTEM and rainier is not doing same so, moving back to original place

Hi @lkammath  I think PowerSupplies schemas belong to new power subsystem. we need to build bmcweb with BMCWEB_NEW_POWERSUBSYSTEM_THERMALSUBSYSTEM if we want to use the new power subsystem.

Why rainier can't build bmcweb with BMCWEB_NEW_POWERSUBSYSTEM_THERMALSUBSYSTEM ?

lkammath:
> > needed to build bmcweb with BMCWEB_NEW_POWERSUBSYSTEM_THERMALSUBSYSTEM and rainier is not doing same so, moving back to original place
> 
> Hi @lkammath I think PowerSupplies schemas belong to new power subsystem. we need to build bmcweb with BMCWEB_NEW_POWERSUBSYSTEM_THERMALSUBSYSTEM if we want to use the new power subsystem.
> 
> Why rainier can't build bmcweb with BMCWEB_NEW_POWERSUBSYSTEM_THERMALSUBSYSTEM ?

I checked the recipe. It is currently not used. I will have to hear from @gtmills about why? Since it is not used till this point i am just reverting back it via this PR

lkammath:
jenkins test run please

redfish-core/include/redfish.hpp
```diff
@@ -86,9 +86,6 @@ class RedfishService
 #ifdef BMCWEB_NEW_POWERSUBSYSTEM_THERMALSUBSYSTEM
```
gtmills, line 86: We need to enable this then

gtmills (changes requested):
We need changes to openbmc/openbmc not here

gtmills:
Don't we already enable this here: https://github.ibm.com/openbmc/openbmc/blob/1020-ghe/meta-ibm/recipes-phosphor/interfaces/bmcweb_%25.bbappend#L5 

What am I missing?

lkammath:
> Don't we already enable this here: https://github.ibm.com/openbmc/openbmc/blob/1020-ghe/meta-ibm/recipes-phosphor/interfaces/bmcweb_%25.bbappend#L5
> 
> What am I missing?

Sorry, I think i was using very old BMC levels. I will close this

PR #178, opened 2021-11-30
bmcweb: CElog entry not received by HMC

sbteeks:
Event log entries are split into EventLog and CELog entries
depending on Hidden property value. The event log entry paths
are different for EventLog and CELog. Current implementation
emits events with EventLog path irrespective of Hidden property
value. This commit provides log entry paths correctly for both
EventLog and CELog depending on the Hidden property value.

Testing:
- Injected INFO and ERROR events using busctl command, and verified
  the paths in journal log.
Example:
bmcweb[3689]:(2021-11-30 16:41:59)[DEBUG "event_dbus_monitor.hpp":369]
EventLog path:
/redfish/v1/Systems/system/LogServices/EventLog/Entries/659
bmcweb[3689]:(2021-11-30 16:42:25)[DEBUG "event_dbus_monitor.hpp":362]
CELog path: /redfish/v1/Systems/system/LogServices/CELog/Entries/660

- HMC team verified that CELog events are received.
Example:
2021-11-29 21:58:01.337 [TID-313c83] :: GOT CELog EVENT :
RESOURCE_CREATED : MTMS :9105-22A*RA00029 EVENT FOR
/redfish/v1/Systems/system/LogServices/CELog/Entries/458

Signed-off-by: Shantappa Teekappanavar <sbteeks@yahoo.com>

### DIFF
--- a/include/event_dbus_monitor.hpp
+++ b/include/event_dbus_monitor.hpp
@@ -301,6 +301,8 @@ void eventLogCreatedSignal(sdbusplus::message::message& msg)
 {
     BMCWEB_LOG_DEBUG << "Event Log Created - match fired";
 
+    constexpr auto pelEntryInterface = "org.open_power.Logging.PEL.Entry";
+
     if (msg.is_method_error())
     {
         BMCWEB_LOG_ERROR << "Event Log Created signal error";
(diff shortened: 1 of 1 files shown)


include/event_dbus_monitor.hpp
```diff
@@ -297,6 +297,35 @@ void registerBootProgressChangeSignal()
...
+            {
+                BMCWEB_LOG_ERROR << "Hidden Property not found: " << ec;
+                return;
+            }
+            bool hiddenPropVal = *hiddenPropValPtr;
+            callback(hiddenPropVal);
+        };
+
+    // Get the Hidden Property
+    crow::connections::systemBus->async_method_call(
```
spinler, line 322: I would like to avoid a call back into phosphor-logging for every new error log created by just reading the PEL interface out of the interfacesAdded signal contents, however I'm not seeing that interface in that signal so I will investigate.

sbteeks: You mean we should read Hidden property directly from PEL interface - correct? @gtmills didn't like reading the property from the interface when I was providing support for CELog. He suggested to use async/callback. @gtmills any thoughts?

gtmills: > I would like to avoid a call back into phosphor-logging for every new error log created by just reading the PEL interface out of the interfacesAdded signal contents, however I'm not seeing that interface in that signal so I will investigate.

+10. InterfaceAdded looks to just have the interface name though. 
I think this is as good as we are going to get. :/ 
"Hidden" being a property instead of just a separate collection of logs has caused some really complicated code. 
I think we really need to revisit this later. 
I think either the same Redfish log service for both, and the GUI can filter, or different D-Bus collections.

gtmills (approved):
Like Matt, I would like to avoid calling into phosphor-logging for every new error log created but I don't see a better solution.

include/event_dbus_monitor.hpp
spinler: actually, give me a bit more time to debug before merging this

spinler: OK, I got the interfacesAdded signal working:
```
signal time=1638306787.955473 sender=:1.636 -> destination=(null destination) serial=74 path=/xyz/openbmc_project/logging; interface=org.freedesktop.DBus.ObjectManager; member=InterfacesAdded                   
   object path "/xyz/openbmc_project/logging/entry/5"
   array [
      dict entry(
         string "org.open_power.Logging.PEL.Entry"
         array [                           
            dict entry(
               string "Hidden"
               variant                   boolean false
            )
            dict entry(                                                                                                                                                                                           
               string "Subsystem"           
               variant                   string "BMC Firmware"
            )    
            dict entry(
               string "ManagementSystemAck"
               variant                   boolean false
            )
         ]
      )
   ]  
```

spinler (changes requested):
Would like to see this done without the extra get property call.  may be a few days before I get my changes in.

include/event_dbus_monitor.hpp
gtmills: Then sounds like we don't need an extra call here. Thanks Matt.

gtmills (changes requested):
Sounds like we can do without an extra call

spinler:
https://gerrit.openbmc-project.xyz/c/openbmc/phosphor-logging/+/49184 is the PEL fix.

spinler (approved):

gtmills (approved):

PR #180, opened 2021-12-01
Multipart parser

sunharis:
As part of HMC performance improvement, this commit implements
multiple configfile upload with a single http request

This commit uses the upstream MIME parsing algorithm.

Upstream commit: https://gerrit.openbmc-project.xyz/c/openbmc/bmcweb/+/34974
The 2nd commit is a downstream only based on https://gerrit.openbmc-project.xyz/c/openbmc/bmcweb/+/32174/

Tested by:
    1. Multiple file upload
    curl -k -H "X-Auth-Token: $bmc_token" -H "Content-Type: multipart/form-data"
    -X POST https://${bmc}/ibm/v1/Host/ConfigFiles -F "testfile1=@<path>;type=application/octet-stream"
    -F "testfile2=@<path>;type=application/octet-stream"
    2. Tested the single file upload for sanity

### DIFF
--- a/include/ibm/management_console_rest.hpp
+++ b/include/ibm/management_console_rest.hpp
@@ -34,6 +34,8 @@ constexpr const char* methodNotAllowedMsg = "Method Not Allowed";
 constexpr const char* resourceNotFoundMsg = "Resource Not Found";
 constexpr const char* contentNotAcceptableMsg = "Content Not Acceptable";
 constexpr const char* internalServerError = "Internal Server Error";
+constexpr const char* internalFileSystemError = "Internal FileSystem Error";
+constexpr const char* badRequestMsg = "Bad Request";
 constexpr const char* propertyMissing = "Required Property Missing";
 constexpr uint32_t maxCSRLength = 4096;
 std::filesystem::path rootCertPath = "/var/lib/ibm/bmcweb/RootCert";
@@ -89,7 +91,7 @@ inline void saveConfigFile(const std::string& data, const std::string& fileID,
     {
         asyncResp->res.result(
             boost::beast::http::status::internal_server_error);
-        asyncResp->res.jsonValue["Description"] = internalServerError;
+        asyncResp->res.jsonValue["Description"] = internalFileSystemError;
         BMCWEB_LOG_DEBUG << "handleIbmPut: Failed to find if file exists. ec : "
                          << ec;
         return;
@@ -103,7 +105,7 @@ inline void saveConfigFile(const std::string& data, const std::string& fileID,
         {
             asyncResp->res.result(
                 boost::beast::http::status::internal_server_error);
-            asyncResp->res.jsonValue["Description"] = internalServerError;
+            asyncResp->res.jsonValue["Description"] = internalFileSystemError;
             BMCWEB_LOG_DEBUG << "handleIbmPut: Failed to find file size. ec : "
                              << ec;
             return;
@@ -216,7 +218,7 @@ inline void
     {
         asyncResp->res.result(
             boost::beast::http::status::internal_server_error);
-        asyncResp->res.jsonValue["Description"] = internalServerError;
+        asyncResp->res.jsonValue["Description"] = internalFileSystemError;
         BMCWEB_LOG_DEBUG << "handleIbmPut: Failed to prepare save-area "
                             "directory iterator. ec : "
                          << ec;
@@ -231,7 +233,8 @@ inline void
             {
                 asyncResp->res.result(
                     boost::beast::http::status::internal_server_error);
-                asyncResp->res.jsonValue["Description"] = internalServerError;
+                asyncResp->res.jsonValue["Description"] =
+                    internalFileSystemError;
                 BMCWEB_LOG_DEBUG << "handleIbmPut: Failed to find save-area "
                                     "directory . ec : "
                                  << ec;
@@ -242,7 +245,8 @@ inline void
             {
                 asyncResp->res.result(
                     boost::beast::http::status::internal_server_error);
-                asyncResp->res.jsonValue["Description"] = internalServerError;
+                asyncResp->res.jsonValue["Description"] =
+                    internalFileSystemError;
                 BMCWEB_LOG_DEBUG << "handleIbmPut: Failed to find save-area "
                                     "file size inside the directory . ec : "
                                  << ec;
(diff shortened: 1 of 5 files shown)


sunharis:
This change could potentially fix the performance issue https://w3.rchland.ibm.com/projects/bestquest/?defect=SW538927

gtmills:
I updated the PR msg to add links. For next time, please do that. 
My concern with  https://gerrit.openbmc-project.xyz/c/openbmc/bmcweb/+/34974, is it still has unaddressed valid comments. 
I don't see any of the 7 that are a show stopper but they would help maintainability. Could these be addressed? @lxwinspur

include/ibm/management_console_rest.hpp
```diff
@@ -1,9 +1,10 @@
 #pragma once
```
gtmills, line 1: Could you add some more detail to the commit msg about exactly what you are doing here? 
Where it is from, etc.

include/ibm/management_console_rest.hpp
```diff
@@ -1075,6 +1160,14 @@ inline void requestRoutes(App& app)
                 handleFileUrl(req, asyncResp, fileName);
             });
 
+    BMCWEB_ROUTE(app, "/ibm/v1/Host/ConfigFiles")
```
gtmills, line 1163: Why before were we posting to /ibm/v1/Host/ConfigFiles/<str> ?

gtmills: So Code update is doing multipart at 

    "MultipartHttpPushUri": "/redfish/v1/UpdateService/upload",
    
    Does it make sense for us to follow?

gtmills: Should we advertise this in the GET somehow like Code update?

include/ibm/management_console_rest.hpp
```diff
@@ -238,14 +172,165 @@ inline void handleFilePut(const crow::Request& req,
...
+
+    std::filesystem::path loc(configFilePath);
+
+    // Get the current size of the savearea directory
+    std::filesystem::recursive_directory_iterator iter(loc, ec);
+    if (ec)
+    {
+        asyncResp->res.result(
+            boost::beast::http::status::internal_server_error);
+        asyncResp->res.jsonValue["Description"] = internalServerError;
```
gtmills, line 219: We could add more here.. These Descriptions are all the same. Up to you

include/ibm/management_console_rest.hpp
```diff
@@ -238,14 +172,165 @@ inline void handleFilePut(const crow::Request& req,
...
+        std::string fileName;
+        for (const FormPart& formpart : parser.mime_fields)
+        {
+            boost::beast::http::fields::const_iterator it =
+                formpart.fields.find("Content-Disposition");
+            if (it == formpart.fields.end())
+            {
+                BMCWEB_LOG_ERROR << "Couldn't find Content-Disposition";
+                asyncResp->res.result(
+                    boost::beast::http::status::internal_server_error);
```
gtmills, line 281: Sure seems like not an internal error and instead a bad request

include/ibm/management_console_rest.hpp
```diff
@@ -238,14 +172,165 @@ inline void handleFilePut(const crow::Request& req,
...
+    {
+        BMCWEB_LOG_DEBUG << "This is a multipart upload";
+        MultipartParser parser(req, ec);
+        if (ec)
+        {
+            // handle error
+            BMCWEB_LOG_ERROR << "MIME parse failed, ec : " << ec.value()
+                             << " , ec message : " << ec.message();
+            asyncResp->res.result(
+                boost::beast::http::status::internal_server_error);
```
gtmills, line 267: Really an internal error and not a bad request?

gtmills:
Comments upstream, returning internal error at times when I don't think we should but overall I am good with the direction

include/ibm/management_console_rest.hpp
sunharis: There is already a GET and PATCH  command implemented on this URL for the save area file upload and download purpose. This is a new POST command to upload multiple files at one shot. 
Code update is the redfish path. This /ibm/v1 is a REST way of implementing the routes.

lxwinspur:
> I updated the PR msg to add links. For next time, please do that. My concern with https://gerrit.openbmc-project.xyz/c/openbmc/bmcweb/+/34974, is it still has unaddressed valid comments. I don't see any of the 7 that are a show stopper but they would help maintainability. Could these be addressed? @lxwinspur

@gtmills Okay

lxwinspur:
> Comments upstream, returning internal error at times when I don't think we should but overall I am good with the direction

Let me update upstream

include/ibm/management_console_rest.hpp
gtmills: I would prefer we match code update here...
"MultipartHttpPushUri": "/redfish/v1/UpdateService/upload",

include/ut/multipart_test.cpp
```diff
@@ -0,0 +1,62 @@
...
+                 "-----------------------------d74496d66958873e\r\n"
+                 "Content-Disposition: form-data; name=\"Test2\"\r\n"
+                 "123456\r\n"
+                 "-----------------------------d74496d66958873e--\r\n";
+
+    std::error_code ec;
+    crow::Request reqIn(req);
+    MultipartParser parser(reqIn, ec);
+    ASSERT_EQ(ec.value(), static_cast<int>(ParserError::ERROR_EMPTY_HEADER));
+}
```
gtmills, line 62: There is comment upstream about more tests, I think that is valid

include/ut/multipart_test.cpp
```diff
@@ -0,0 +1,62 @@
...
+    ASSERT_EQ(parser.mime_fields[0].fields.at("Content-Disposition"),
+              "form-data; name=\"Test1\"");
+    ASSERT_EQ(parser.mime_fields[0].content, "{\"Key1\": 112233}");
+
+    ASSERT_EQ(parser.mime_fields[1].fields.at("Content-Disposition"),
+              "form-data; name=\"Test2\"");
+    ASSERT_EQ(parser.mime_fields[1].content, "123456");
+}
+
+TEST(MultipartTest, TestBadMultipartParser)
```
gtmills, line 42: Valid comment here about the name

PR #181, opened 2021-12-01
PCIeDevice Schema and PCIeSlot to PCIeDevice link

SunnySrivastava1984:
This PR implement changes to populate PCIeDevice schema with desired properties and populate link between PCIeSlot and PCIeDevice schema.

### DIFF
--- a/redfish-core/lib/pcie.hpp
+++ b/redfish-core/lib/pcie.hpp
@@ -16,25 +16,39 @@
 
 #pragma once
 
+#include "pcie_slots.hpp"
+
 #include <app.hpp>
 #include <boost/system/linux_error.hpp>
 #include <registries/privilege_registry.hpp>
 
+#include <map>
+
 namespace redfish
 {
 
-static constexpr char const* pcieService = "xyz.openbmc_project.PCIe";
-static constexpr char const* pciePath = "/xyz/openbmc_project/PCIe";
+static constexpr char const* pcieRootPath = "/xyz/openbmc_project/inventory";
 static constexpr char const* pcieDeviceInterface =
-    "xyz.openbmc_project.PCIe.Device";
+    "xyz.openbmc_project.Inventory.Item.PCIeDevice";
+
+using PCIeDevice = std::string;
+using PCIeDevicePath = std::string;
+using ServiceName = std::string;
+using MapperGetSubTreeResponse = std::vector<
+    std::pair<std::string,
+              std::vector<std::pair<std::string, std::vector<std::string>>>>>;
+
+// will be used when we need to fetch single PCIeDevice properties.
+std::map<PCIeDevice, std::pair<ServiceName, PCIeDevicePath>>
+    mapOfPcieDevicePaths;
 
 static inline void
     getPCIeDeviceList(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
                       const std::string& name)
 {
-    auto getPCIeMapCallback = [asyncResp, name](
-                                  const boost::system::error_code ec,
-                                  std::vector<std::string>& pcieDevicePaths) {
+    auto getPCIeMapCallback = [asyncResp,
+                               name](const boost::system::error_code ec,
+                                     const MapperGetSubTreeResponse& subTree) {
         if (ec)
         {
             BMCWEB_LOG_DEBUG << "no PCIe device paths found ec: "
@@ -42,32 +56,39 @@ static inline void
             // Not an error, system just doesn't have PCIe info
             return;
         }
+
         nlohmann::json& pcieDeviceList = asyncResp->res.jsonValue[name];
         pcieDeviceList = nlohmann::json::array();
-        for (const std::string& pcieDevicePath : pcieDevicePaths)
-        {
-            size_t devStart = pcieDevicePath.rfind('/');
-            if (devStart == std::string::npos)
-            {
-                continue;
-            }
 
-            std::string devName = pcieDevicePath.substr(devStart + 1);
-            if (devName.empty())
+        for (const auto& [pcieDevicePath, serviceMap] : subTree)
+        {
+            for (const auto& [serviceName, interfaceList] : serviceMap)
             {
-                continue;
+                std::string devName =
+                    sdbusplus::message::object_path(pcieDevicePath).filename();
+
+                if (devName.empty())
+                {
+                    BMCWEB_LOG_DEBUG << "Invalid Name";
+                    continue;
+                }
+
+                mapOfPcieDevicePaths.emplace(
+                    devName, std::make_pair(serviceName, pcieDevicePath));
+                pcieDeviceList.push_back(
+                    {{"@odata.id",
+                      "/redfish/v1/Systems/system/PCIeDevices/" + devName}});
             }
-            pcieDeviceList.push_back(
-                {{"@odata.id",
-                  "/redfish/v1/Systems/system/PCIeDevices/" + devName}});
         }
+
         asyncResp->res.jsonValue[name + "@odata.count"] = pcieDeviceList.size();
     };
     crow::connections::systemBus->async_method_call(
         std::move(getPCIeMapCallback), "xyz.openbmc_project.ObjectMapper",
         "/xyz/openbmc_project/object_mapper",
-        "xyz.openbmc_project.ObjectMapper", "GetSubTreePaths",
-        std::string(pciePath) + "/", 1, std::array<std::string, 0>());
+        "xyz.openbmc_project.ObjectMapper", "GetSubTree",
+        std::string(pcieRootPath), 0,
+        std::array<const char*, 1>{pcieDeviceInterface});
 }
 
 inline void requestRoutesSystemPCIeDeviceCollection(App& app)
@@ -79,9 +100,7 @@ inline void requestRoutesSystemPCIeDeviceCollection(App& app)
         .privileges(redfish::privileges::getPCIeDeviceCollection)
         .methods(boost::beast::http::verb::get)(
             [](const crow::Request&,
-               const std::shared_ptr<bmcweb::AsyncResp>& asyncResp)
-
-            {
+               const std::shared_ptr<bmcweb::AsyncResp>& asyncResp) {
                 asyncResp->res.jsonValue = {
                     {"@odata.type",
                      "#PCIeDeviceCollection.PCIeDeviceCollection"},
@@ -90,6 +109,7 @@ inline void requestRoutesSystemPCIeDeviceCollection(App& app)
                     {"Description", "Collection of PCIe Devices"},
                     {"Members", nlohmann::json::array()},
                     {"Members@odata.count", 0}};
+
                 getPCIeDeviceList(asyncResp, "Members");
             });
 }
@@ -150,17 +170,41 @@ inline void requestRoutesSystemPCIeDevice(App& app)
                         asyncResp->res.jsonValue["DeviceType"] = *property;
                     }
 
+                    if (std::string* property = std::get_if<std::string>(
+                            &pcieDevProperties["GenerationInUse"]);
+                        property)
+                    {
+                        std::string generation = analysisGeneration(*property);
+                        if (!generation.empty())
+                        {
+                            asyncResp->res
+                                .jsonValue["PCIeInterface"]["PCIeType"] =
+                                generation;
+                        }
+                    }
+
                     asyncResp->res.jsonValue["PCIeFunctions"] = {
                         {"@odata.id",
                          "/redfish/v1/Systems/system/PCIeDevices/" + device +
                              "/PCIeFunctions"}};
                 };
-                std::string escapedPath = std::string(pciePath) + "/" + device;
-                dbus::utility::escapePathForDbus(escapedPath);
-                crow::connections::systemBus->async_method_call(
-                    std::move(getPCIeDeviceCallback), pcieService, escapedPath,
-                    "org.freedesktop.DBus.Properties", "GetAll",
-                    pcieDeviceInterface);
+
+                auto it = mapOfPcieDevicePaths.find(device);
+                if (it != mapOfPcieDevicePaths.end())
+                {
+                    std::string& serviceName = std::get<0>(it->second);
+                    std::string& path = std::get<1>(it->second);
+                    crow::connections::systemBus->async_method_call(
+                        std::move(getPCIeDeviceCallback), serviceName, path,
+                        "org.freedesktop.DBus.Properties", "GetAll",
+                        std::string(pcieDeviceInterface));
+                }
+                else
+                {
+                    messages::resourceNotFound(asyncResp->res, "PCIeDevice",
+                                               device);
+                    return;
+                }
             });
 }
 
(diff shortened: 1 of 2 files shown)


gtmills:
Validator pass? I would think you would need 2021.3 which just merged a few hours ago

SunnySrivastava1984:
I executed validator and could not see any errors in these schemas. Also all the properties were already there that is being used here. Any specific reason I would need 2021.3?

redfish-core/lib/pcie_slots.hpp
```diff
@@ -90,6 +90,55 @@ inline std::string analysisSlotType(const std::string& slotType)
...
+            }
+            if (subtree.size() == 0)
+            {
+                BMCWEB_LOG_ERROR
+                    << "Can't find PCIeDevice D-Bus object for given slot";
+                return;
+            }
+
+            // There should be only one path under the slot.
+            const std::string& pcieDevciePath = std::get<0>(subtree[0]);
```
gtmills, line 120: Doesn't std::get throw?

gtmills:
> Also all the properties were already there that is being used here. Any specific reason I would need 2021.3?

I thought this link was only added in 2021.3 but I think I am wrong so ignore.

redfish-core/lib/pcie.hpp
```diff
@@ -90,10 +101,52 @@ inline void requestRoutesSystemPCIeDeviceCollection(App& app)
...
+        return;
+    }
+
+    crow::connections::systemBus->async_method_call(
+        [asyncResp, slotPath](const boost::system::error_code ec,
+                              const std::variant<std::string>& property) {
+            if (ec)
+            {
+                BMCWEB_LOG_DEBUG << "DBUS response error";
+                messages::internalError(asyncResp->res);
```
bradbishop, line 127: Not doing any error handling here.  Is this normal for bmcweb?

redfish-core/lib/pcie.hpp
```diff
@@ -90,10 +101,52 @@ inline void requestRoutesSystemPCIeDeviceCollection(App& app)
...
+        messages::internalError(asyncResp->res);
+        return;
+    }
+
+    crow::connections::systemBus->async_method_call(
+        [asyncResp, slotPath](const boost::system::error_code ec,
+                              const std::variant<std::string>& property) {
+            if (ec)
+            {
+                BMCWEB_LOG_DEBUG << "DBUS response error";
```
bradbishop, line 126: the value of this printf is dubious without the errno.  Also, I think it might be expected in some cases like if the pcie device is concurrently maintained at the same time as the api is being queried.

redfish-core/lib/pcie.hpp
```diff
@@ -90,10 +101,52 @@ inline void requestRoutesSystemPCIeDeviceCollection(App& app)
...
+
+            std::string generation = analysisGeneration(*value);
+            if (!generation.empty())
+            {
+                asyncResp->res.jsonValue["PCIeInterface"]["PCIeType"] =
+                    generation;
+            }
+        },
+        std::string(pcieService), slotPath, "org.freedesktop.DBus.Properties",
+        "Get", "xyz.openbmc_project.Inventory.Item.PCIeSlot", "Generation");
```
bradbishop, line 147: This should be on the Device interface, not the slot.  The DBus property is GenerationInUse.

redfish-core/lib/pcie.hpp
```diff
@@ -155,12 +208,24 @@ inline void requestRoutesSystemPCIeDevice(App& app)
...
                              "/PCIeFunctions"}};
                 };
-                std::string escapedPath = std::string(pciePath) + "/" + device;
-                dbus::utility::escapePathForDbus(escapedPath);
-                crow::connections::systemBus->async_method_call(
-                    std::move(getPCIeDeviceCallback), pcieService, escapedPath,
-                    "org.freedesktop.DBus.Properties", "GetAll",
-                    pcieDeviceInterface);
+
+                const auto& it = mapOfPcieDevicePaths.find(device);
```
bradbishop, line 212: binding an r-value to a reference doesn't make much sense here.  I'm assuming you got the "can't bind non-const reference to r-value" compiler error and thus made it const - instead just:

`auto it = mapOfPcieDevicePaths.find(device)`

iterators are just pointers so tricks like this aren't saving any resources.

If you really wanted an iterator that can't be used to change the thing being pointed at you can't use auto because the container itself isn't const - instead you need something like:

`decltype(mapOfPcieDevicePaths)::const_iterator it =  mapOfPcieDevicePaths.find(device)`

but even that feels a bit unusual.

redfish-core/lib/pcie.hpp
```diff
@@ -155,12 +208,24 @@ inline void requestRoutesSystemPCIeDevice(App& app)
...
+                if (it != mapOfPcieDevicePaths.end())
+                {
+                    std::string path = it->second;
```
bradbishop, line 215: You could avoid a copy here:

`std::string& path = it->second`

redfish-core/lib/pcie.hpp
```diff
@@ -369,12 +444,22 @@ inline void requestRoutesSystemPCIeFunction(App& app)
...
                 }
             };
-            std::string escapedPath = std::string(pciePath) + "/" + device;
-            dbus::utility::escapePathForDbus(escapedPath);
-            crow::connections::systemBus->async_method_call(
-                std::move(getPCIeDeviceCallback), pcieService, escapedPath,
-                "org.freedesktop.DBus.Properties", "GetAll",
-                pcieDeviceInterface);
+
+            const auto& it = mapOfPcieDevicePaths.find(device);
```
bradbishop, line 448: same comment here as on line 212

redfish-core/lib/pcie.hpp
```diff
@@ -238,12 +303,22 @@ inline void requestRoutesSystemPCIeFunctionCollection(App& app)
...
                         pcieFunctionList.size();
                 };
-                std::string escapedPath = std::string(pciePath) + "/" + device;
-                dbus::utility::escapePathForDbus(escapedPath);
-                crow::connections::systemBus->async_method_call(
-                    std::move(getPCIeDeviceCallback), pcieService, escapedPath,
-                    "org.freedesktop.DBus.Properties", "GetAll",
-                    pcieDeviceInterface);
+
+                const auto& it = mapOfPcieDevicePaths.find(device);
```
bradbishop, line 307: same comment here as on line 212

redfish-core/lib/pcie.hpp
```diff
@@ -238,12 +303,22 @@ inline void requestRoutesSystemPCIeFunctionCollection(App& app)
                     asyncResp->res.jsonValue["PCIeFunctions@odata.count"] =
                         pcieFunctionList.size();
                 };
-                std::string escapedPath = std::string(pciePath) + "/" + device;
-                dbus::utility::escapePathForDbus(escapedPath);
-                crow::connections::systemBus->async_method_call(
-                    std::move(getPCIeDeviceCallback), pcieService, escapedPath,
-                    "org.freedesktop.DBus.Properties", "GetAll",
-                    pcieDeviceInterface);
+
```
bradbishop, line 306: I'm curious why the path escaping was removed?

redfish-core/lib/pcie.hpp
```diff
@@ -369,12 +444,22 @@ inline void requestRoutesSystemPCIeFunction(App& app)
...
-            dbus::utility::escapePathForDbus(escapedPath);
-            crow::connections::systemBus->async_method_call(
-                std::move(getPCIeDeviceCallback), pcieService, escapedPath,
-                "org.freedesktop.DBus.Properties", "GetAll",
-                pcieDeviceInterface);
+
+            const auto& it = mapOfPcieDevicePaths.find(device);
+            if (it != mapOfPcieDevicePaths.end())
+            {
+                std::string path = it->second;
```
bradbishop, line 451: reference again

redfish-core/lib/pcie_slots.hpp
```diff
@@ -324,6 +373,8 @@ inline void getPCIeSlots(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
                                             "Get", locationInterface,
                                             "LocationCode");
                                 }
+                                getLinkedPcieDevice(asyncResp, pcieSlotPath,
```
bradbishop, line 376: A function with "get" in the name should return something shouldn't it?  addLinkedPCIeDevices?  (devices is plural here because the redfish spec (and the pcie spec) allows multiple devices in one slot ((it's fine if we only support a single device for the moment))).

redfish-core/lib/pcie_slots.hpp
```diff
@@ -90,6 +90,55 @@ inline std::string analysisSlotType(const std::string& slotType)
...
+                subtree) {
+            if (ec)
+            {
+                BMCWEB_LOG_ERROR << "D-Bus response error on GetSubTree " << ec;
+                messages::internalError(asyncResp->res);
+                return;
+            }
+            if (subtree.size() == 0)
+            {
+                BMCWEB_LOG_ERROR
```
bradbishop, line 114: this isn't an error is it?  you can have an empty slot?

redfish-core/lib/pcie_slots.hpp
```diff
@@ -90,6 +90,55 @@ inline std::string analysisSlotType(const std::string& slotType)
...
+                    << "Can't find PCIeDevice D-Bus object for given slot";
+                return;
+            }
+
+            // There should be only one path under the slot.
```
bradbishop, line 119: technically this is true on Rainier, but in general it is not and there is really no value in enforcing it is there?  Update the comment to say "Assuming one device per slot for the moment" so someone looking at it doesn't get confused and think that multiple pcie devices in pcie slots is illegal from a pcie or redfish point of view.

redfish-core/lib/pcie_slots.hpp
```diff
@@ -90,6 +90,55 @@ inline std::string analysisSlotType(const std::string& slotType)
...
+            const std::string& pcieDevciePath = std::get<0>(subtree[0]);
+            const std::string pcieDevice =
+                sdbusplus::message::object_path(pcieDevciePath).filename();
```
bradbishop, line 122: same question here as in the other patch in this PR about use of this upstream.  would this be allowed?  it seems inefficient given that we are starting with a string....

SunnySrivastava1984: Same as above. There was a comment to use wrapper apis as far as possible for any operation on path.

redfish-core/lib/pcie_slots.hpp
SunnySrivastava1984: Fixed. Changed to Debug

redfish-core/lib/pcie_slots.hpp
SunnySrivastava1984: updated

redfish-core/lib/pcie_slots.hpp
```diff
@@ -90,6 +90,55 @@ inline std::string analysisSlotType(const std::string& slotType)
...
+                return;
+            }
+
+            asyncResp->res.jsonValue["Slots"][index]["Links"]["PCIeDevice"] = {
+                {{"@odata.id",
+                  "/redfish/v1/Systems/system/PCIeDevices/" + pcieDevice}}};
+        },
+        "xyz.openbmc_project.ObjectMapper",
+        "/xyz/openbmc_project/object_mapper",
+        "xyz.openbmc_project.ObjectMapper", "GetSubTree", slotPath, int32_t(0),
```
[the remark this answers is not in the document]
SunnySrivastava1984: Done

redfish-core/lib/pcie_slots.hpp
SunnySrivastava1984: Updated.

redfish-core/lib/pcie.hpp
```diff
@@ -42,21 +53,22 @@ static inline void
             // Not an error, system just doesn't have PCIe info
             return;
         }
+
```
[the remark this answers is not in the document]
bradbishop: 👍

redfish-core/lib/pcie.hpp
```diff
@@ -42,21 +53,22 @@ static inline void
...
-        for (const std::string& pcieDevicePath : pcieDevicePaths)
+
+        for (const auto& pcieDevicePath : pcieDevicePaths)
         {
-            size_t devStart = pcieDevicePath.rfind('/');
-            if (devStart == std::string::npos)
-            {
-                continue;
-            }
+            std::string devName =
```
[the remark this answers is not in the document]
bradbishop: Ok, sounds good.  I suggest that in the future, for the sake of your reviewers, make unrelated changes like this in a different commit, with a commit message that explains why you made the change, much as you have in the review comments here.  That avoids this entire back and forth.

redfish-core/lib/pcie.hpp
```diff
@@ -42,21 +53,22 @@ static inline void
...
+                sdbusplus::message::object_path(pcieDevicePath).filename();
 
-            std::string devName = pcieDevicePath.substr(devStart + 1);
             if (devName.empty())
             {
+                BMCWEB_LOG_DEBUG << "Invalid Name";
                 continue;
             }
+
+            mapOfPcieDevicePaths.emplace(devName, pcieDevicePath);
```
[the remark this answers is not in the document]
bradbishop: 👍

redfish-core/lib/pcie.hpp
```diff
@@ -79,9 +92,7 @@ inline void requestRoutesSystemPCIeDeviceCollection(App& app)
         .privileges(redfish::privileges::getPCIeDeviceCollection)
         .methods(boost::beast::http::verb::get)(
             [](const crow::Request&,
-               const std::shared_ptr<bmcweb::AsyncResp>& asyncResp)
-
-            {
+               const std::shared_ptr<bmcweb::AsyncResp>& asyncResp) {
```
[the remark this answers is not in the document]
bradbishop: ok - so that means a clang-format error slipped in a previous patch.  what I do in that situation is run clang-format before I make any changes, put that in a commit of its own, and then make the changes.  Not doing this just irritates your reviewers (maybe some reviewers don't mind but a lot of them do :-))

redfish-core/lib/pcie.hpp
```diff
@@ -42,21 +53,22 @@ static inline void
...
-            {
-                continue;
-            }
+            std::string devName =
+                sdbusplus::message::object_path(pcieDevicePath).filename();
 
-            std::string devName = pcieDevicePath.substr(devStart + 1);
             if (devName.empty())
             {
+                BMCWEB_LOG_DEBUG << "Invalid Name";
```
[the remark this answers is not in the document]
bradbishop: >  just to be double sure that we don't end up filling out map 

well..this doesn't do that though, it just makes noise in the journal.  maybe an assert is what you are looking for if you want to catch bugs?

redfish-core/lib/pcie.hpp
bradbishop: I just meant that every error is being handled in the same way - with a 5xx response.  If ec was ENOENT for example, would a http 404 response make more sense?

redfish-core/lib/pcie.hpp
bradbishop: 👍

redfish-core/lib/pcie_slots.hpp
bradbishop: 👍

redfish-core/lib/pcie_slots.hpp
bradbishop: Even demoted to debug I don't really think tracing is appropriate here.  There are plenty of other ways to figure out if a slot is empty or not.

redfish-core/lib/pcie.hpp
bradbishop: ok.  I probably would have left the escaping in since it doesn't hurt anything and provides broader compatibility with applications like entity-manager....but we probably won't be supporting that with this fork of bmcweb so I guess it probably doesn't matter.

bradbishop (approved):
Thanks Sunny.

gtmills (approved):

gtmills:
Defaulting to Brad here.

PR #184, opened 2021-12-03
HMC Lock management improvements

sunharis:
This commit optimizes the release lock code and adds
some traces to give more data for lock conflict scenarios

Tested by:
 1. With dual HMC, tested create & activate LPARs
 2. Triggered lock conflict by starting create LPAR on both HMCs
    at the same time

Change-Id: I13f642be2f98059ee3401d35a767863b6d78d225

### DIFF
--- a/include/ibm/locks.hpp
+++ b/include/ibm/locks.hpp
@@ -209,7 +209,15 @@ inline RcGetLockList Lock::getLockList(const ListOfSessionIds& listSessionId)
 inline RcReleaseLockApi Lock::releaseLock(const ListOfTransactionIds& p,
                                           const SessionFlags& ids)
 {
-    // Check if the locks are owned by the
+    bool status = validateRids(p);
+
+    if (!status)
+    {
+        // Validation of rids failed
+        BMCWEB_LOG_ERROR << "releaseLock: Contains invalid request id";
+        return std::make_pair(false, status);
+    }
+    // Validation passed, check if all the locks are owned by the
     // requesting HMC
     auto status2 = isItMyLock(p, ids);
     if (!status2.first)
@@ -333,7 +341,8 @@ inline bool Lock::validateRids(const ListOfTransactionIds& refRids)
         }
         else
         {
-            BMCWEB_LOG_DEBUG << "At least 1 inValid Request id";
+            BMCWEB_LOG_ERROR << "validateRids: At least 1 inValid Request id: "
+                             << id;
             return false;
         }
     }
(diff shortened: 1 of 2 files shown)


gtmills:
Can you explain the optimization a bit more?

gtmills:
Need to pass CI also

sunharis:
> Can you explain the optimization a bit more?

I have removed a function call during the release lock execution and moved the logic into the caller. This avoided one for loop.

sunharis:
jenkins run CI please

include/ibm/locks.hpp
```diff
@@ -209,23 +209,12 @@ inline RcGetLockList Lock::getLockList(const ListOfSessionIds& listSessionId)
...
-    if (!status)
-    {
-        // Validation of rids failed
-        BMCWEB_LOG_DEBUG << "Not a Valid request id";
-        return std::make_pair(false, status);
-    }
-    // Validation passed, check if all the locks are owned by the
+    // Check if the locks are owned by the
     // requesting HMC
     auto status2 = isItMyLock(p, ids);
```
gtmills, line 214: nitpick: this could be "status" now

include/ibm/locks.hpp
```diff
@@ -209,23 +209,12 @@ inline RcGetLockList Lock::getLockList(const ListOfSessionIds& listSessionId)
 inline RcReleaseLockApi Lock::releaseLock(const ListOfTransactionIds& p,
                                           const SessionFlags& ids)
 {
-
-    bool status = validateRids(p);
```
gtmills, line 213: What does isItMyLock do if the lock doesn't exist? 
I would think validating the locks is still a valid operation?

sunharis: @gtmills I re-added the validateRids() function in my last patch set. The function isItMyLock() will return false if the lock (transaction id) is not owned by that session id

gtmills: Ack

include/ibm/locks.hpp
```diff
@@ -209,23 +209,20 @@ inline RcGetLockList Lock::getLockList(const ListOfSessionIds& listSessionId)
...
     // Validation passed, check if all the locks are owned by the
     // requesting HMC
     auto status2 = isItMyLock(p, ids);
-    if (status2.first)
+    if (!status2.first)
     {
-        // The current hmc owns all the locks, so we can release
-        // them
-        releaseLock(p);
+        return std::make_pair(false, status2);
```
gtmills, line 225: This will release this lock?

sunharis: I have actually negated the if condition logic at line 223. If the status from the isItMyLock at 222 is false , i am returning false at line 225. Otherwise will return true as the release lock status at line 227

gtmills (approved):

PR #185, opened 2021-12-06
Modify LicenseService as per DMTF released schemas

raviteja-b:
This fixes redfish license service validator errors

Tested By:
GET  
Ran redfish service validator

### DIFF
--- a/redfish-core/lib/license_service.hpp
+++ b/redfish-core/lib/license_service.hpp
@@ -265,6 +265,60 @@ inline void requestRoutesLicenseEntryCollection(App& app)
         });
 }
 
+inline void translateLicenseTypeDbusToRedfish(
+    const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
+    const std::string& licenseType)
+{
+    if (licenseType == "com.ibm.License.Entry.LicenseEntry.Type.Purchased")
+    {
+        asyncResp->res.jsonValue["LicenseType"] = "Production";
+    }
+    else if (licenseType == "com.ibm.License.Entry.LicenseEntry.Type.Prototype")
+    {
+        asyncResp->res.jsonValue["LicenseType"] = "Prototype";
+    }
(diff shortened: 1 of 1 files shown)


sunharis (approved):
Code looks good to me. @gtmills please check if you are fine with the mappings of backend values to redfish and if fine, this is good to go

redfish-core/lib/license_service.hpp
```diff
@@ -265,6 +265,46 @@ inline void requestRoutesLicenseEntryCollection(App& app)
...
+    }
+    if (licenseType == "com.ibm.License.Entry.LicenseEntry.Type.Prototype")
+    {
+        return "Prototype";
+    }
+    if (licenseType == "com.ibm.License.Entry.LicenseEntry.Type.Trial")
+    {
+        return "Trial";
+    }
+    return "";
```
gtmills, line 283: This would be a validator error if this is empty 
LicenseType="" would be a validator error 

else { 
// Any other values would be invalid 
BMCWEB_LOG_DEBUG << "PowerMode value was not valid: " << modeValue; 
}

raviteja-b: it won't be error if value is empty, validator just throws a warning which can be ignored

gtmills: Are you sure? I thought since it was an enum, it would throw an error if empty

gtmills: Another choice would be 
             std::string licenseType = translateLicenseTypeDbusToRedfish(*value);
             if (!licenseType .empty())
             
             like 
             https://github.com/ibm-openbmc/bmcweb/pull/181/files

raviteja-b: yeah I observed only warning for empty string with validator.
anyways I have modified as you suggested.
please have a look

gtmills (approved):

PR #186, opened 2021-12-06
Deduplicate url parsing code

gtmills:
Upstream is https://gerrit.openbmc-project.xyz/c/openbmc/bmcweb/+/47123/5 
This has been merged upstream, no problems have been observed upstream. 
This was suggested by @sunharis to pick up for the general performance improvement. 
I am not sure if we will see any real benefit here but it can't hurt and this commit is more correct.

The 2nd commit is making changes to include/ut/multipart_test.cpp to support this, something we recently pulled in. 
The 3rd commit is removing  several lines around query_params

Tested: GUI works including requests like powering on 

Orginal commit msg: 

In the current model, URLs get parsed twice, once to satisfy the
authenticate method, and once again later to satisfy the handle() call.
This commit deduplicates the parsing.  This is wasteful, and as of the
previous commit, unnecessary.

Specifically, it moves the actual parsing into the request object, and
adds a target() method to explicitly set a url.  This deduplicates the
code that was in http_connection, and centralizes it in request, where
it should really belong.

Tested:
curl --insecure "https://192.168.7.2/redfish/v1"
Returns the redfish v1 resource

curl --insecure "https://192.168.7.2/redfish/v1/Systems"
Returns 401 unauthorized

curl --insecure --user root:0penBmc "https://192.168.7.2/redfish/v1/Systems"
returns the SystemsCollection

Signed-off-by: Ed Tanous <edtanous@google.com>
Change-Id: Ie7ee2d9a9a51bf21c03793b35730e7a0ca82623a

### DIFF
--- a/http/http_connection.hpp
+++ b/http/http_connection.hpp
@@ -318,8 +318,13 @@ class Connection :
     void handle()
     {
         cancelDeadlineTimer();
-
-        crow::Request& thisReq = req.emplace(parser->release());
+        std::error_code reqEc;
+        crow::Request& thisReq = req.emplace(parser->release(), reqEc);
+        if (reqEc)
+        {
+            BMCWEB_LOG_DEBUG << "Request failed to construct" << reqEc;
+            return;
+        }
         thisReq.session = userSession;
 
         // Fetch the client IP address
@@ -341,18 +346,6 @@ class Connection :
                         << "." << thisReq.version() % 10 << ' '
                         << thisReq.methodString() << " " << thisReq.target()
                         << " " << thisReq.ipAddress;
-
-        boost::urls::error_code ec;
-        req->urlView = boost::urls::parse_relative_ref(
-            boost::urls::string_view(req->target().data(),
-                                     req->target().size()),
-            ec);
-        if (ec)
-        {
-            return;
-        }
-        req->url = std::string_view(req->urlView.encoded_path().data(),
-                                    req->urlView.encoded_path().size());
 
         res.setCompleteRequestHandler(nullptr);
         res.isAliveHelper = [this]() -> bool { return isAlive(); };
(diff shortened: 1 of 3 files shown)


sunharis (changes requested):
The file include/ut/multipart_test.cpp needs a change to update the crow::Request constructor while pulling this commit

http/http_connection.hpp
```diff
@@ -579,21 +576,6 @@ class Connection :
...
-                req->urlView =
-                    boost::urls::parse_relative_ref(uriStringView, uriEc);
-                if (uriEc)
-                {
-                    BMCWEB_LOG_ERROR << "Failed to parse URI "
-                                     << uriEc.message();
-                    return;
-                }
-                req->url = std::string_view(req->urlView.encoded_path().data(),
-                                            req->urlView.encoded_path().size());
```
sunharis, line 596: I think there is a bug here - we need to also remove the req->urlParams related code below at line 594 till 603 because we dont have the req->urlView populated after removing this code chunk

sunharis (changes requested):

http/http_connection.hpp
gtmills: I still see 
```
                    req->urlParams = req->urlView.query_params();
#ifdef BMCWEB_ENABLE_DEBUG
                    std::string paramList = "";
                    for (const auto param : req->urlParams)
                    {
                        paramList += param->key() + " " + param->value() + " ";
                    }
                    BMCWEB_LOG_DEBUG << "QueryParams: " << paramList;
#endif
```
upstream at https://github.com/openbmc/bmcweb/blob/master/http/http_connection.hpp#L588

We could remove this, should also remove: 
https://github.com/openbmc/bmcweb/blob/597d2b142362bafa90f24fc8c30750afab91f78f/http/http_request.hpp#L26

gtmills:
> The file include/ut/multipart_test.cpp needs a change to update the crow::Request constructor while pulling this commit

Done.

http/http_connection.hpp
sunharis: > Tested: GUI works including requests like powering on

Ok. Since this worked fine during your tests

sunharis (approved):

http/http_connection.hpp
gtmills: I removed this, just an FYI

PR #191, opened 2021-12-09
Fix Environmental Metrics internal Error

lxwinspur:
Environmental Metrics has an internal error.

resolved by: https://github.com/ibm-openbmc/dev/issues/3401
upstream: https://gerrit.openbmc-project.xyz/c/openbmc/bmcweb/+/43170

Signed-off-by: Albert Zhang <zhanghaodi@inspur.com>
Change-Id: I6a75dfbc94f39d90a1368770076b874608394691

### DIFF
--- a/redfish-core/lib/environment_metrics.hpp
+++ b/redfish-core/lib/environment_metrics.hpp
@@ -47,24 +47,22 @@ inline void
                 }
                 const std::string& validPath = objectPath;
                 const std::string& connectionName = serviceName[0].first;
-                std::vector<std::string> split;
-                // Reserve space for
-                // /xyz/openbmc_project/sensors/<name>/<subname>
-                split.reserve(6);
-                boost::algorithm::split(split, validPath,
-                                        boost::is_any_of("/"));
-                if (split.size() < 6)
+                std::string sensorType;
+                std::string sensorName;
+
+                // 4 below comes from
+                // /xyz/openbmc_project/sensors/sensorType/sensorName
+                //   0      1             2        3        4
+                if (!dbus::utility::getNthStringFromPath(validPath, 3,
+                                                         sensorType) ||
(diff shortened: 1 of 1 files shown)


lxwinspur:
@gtmills 
Please review this patch, thanks

gtmills:
Looks reasonable. Thanks!!!!

redfish-core/lib/environment_metrics.hpp
```diff
@@ -117,58 +115,100 @@ inline void
...
+                    judgment = true;
+                    break;
+                }
+            }
+            if (judgment)
+            {
+                const std::array<const char*, 1> totalPowerInterfaces = {
+                    "xyz.openbmc_project.Sensor.Value"};
+                const std::string& totalPowerPath =
+                    "/xyz/openbmc_project/sensors/power/total_power";
```
aahmed-2, line 146: We should get "totalPowerPath" from the above call instead of hardcoding it here.

aahmed-2: Something like:
insert at line 130:   const std::string& totalPowerPath = "";
...
insert at line 137: totalPowerPath = path;

redfish-core/lib/environment_metrics.hpp
```diff
@@ -117,58 +115,100 @@ inline void
 inline void getPowerWatts(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
```
aahmed-2, line 115: I think we need to relook at the dbus calls here. I think we might be able to eliminate the outer dbus call and replace it with an error not found for total_power.

aahmed-2 (approved):
Did some basic curl get testing. Basic fields seemed to fill out correctly without internalError

aahmed-2 (approved):

gtmills (approved):
Defaulting some to Ali here. Seems reasonable though.

PR #192, opened 2021-12-10
GHE: Set create dump task timer to 20 minutes

raviteja-b:
Currently create dump task timer set to 6 minutes
but in latest drivers MPIPL dump creation taking more time
so setting timer to 20 minute to make it for both rainier/everest

Defect is SW540793

### DIFF
--- a/redfish-core/lib/log_services.hpp
+++ b/redfish-core/lib/log_services.hpp
@@ -935,7 +935,7 @@ inline void createDumpTaskCallback(
             createdObjPath.str +
             "',arg0='xyz.openbmc_project.Common.Progress'");
 
-    task->startTimer(std::chrono::minutes(30));
+    task->startTimer(std::chrono::minutes(20));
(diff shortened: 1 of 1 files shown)


sunharis (approved):

gtmills:
30 mins is a long time.. How did we come up with this? 
An MPIPL dump can take how long?

raviteja-b:
actually issue is earlier builds it used to take around 3-4 mins for MPIPL dump but in latest drivers it started taking 11 mins. so we decided to keep 30 mins for now.

there are too  many defects opened around this for HMC-BMC dump offload

asmithakarun (approved):

raviteja-b:
As per FVT observations, MPIPL dump create takes around 10 min 30 secs.
so keeping max 30 mins timeout

gtmills (approved):

gtmills:
Changed to 20 mins

PR #194, opened 2021-12-13
Handle empty enum values in PCIeDevice and PCIeFunctions

SunnySrivastava1984:
The commit implement changes to handle empty "DeviceType"
in PCIeDevice schema.
The property expects an enum value  and since  Dbus object
path for PCIeDevices are primed the value fetched from D-Bus
can be empty for the field.

Validator has been executed an no new error has been found.

### DIFF
--- a/redfish-core/lib/pcie.hpp
+++ b/redfish-core/lib/pcie.hpp
@@ -167,7 +167,10 @@ inline void requestRoutesSystemPCIeDevice(App& app)
                             &pcieDevProperties["DeviceType"]);
                         property)
                     {
-                        asyncResp->res.jsonValue["DeviceType"] = *property;
+                        if (!property->empty())
+                        {
+                            asyncResp->res.jsonValue["DeviceType"] = *property;
+                        }
                     }
 
                     if (std::string* property = std::get_if<std::string>(
@@ -381,7 +384,10 @@ inline void requestRoutesSystemPCIeFunction(App& app)
                                            "FunctionType"]);
                     property)
                 {
-                    asyncResp->res.jsonValue["FunctionType"] = *property;
+                    if (!property->empty())
+                    {
+                        asyncResp->res.jsonValue["FunctionType"] = *property;
+                    }
                 }
 
                 if (std::string* property = std::get_if<std::string>(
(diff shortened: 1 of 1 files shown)


redfish-core/lib/pcie.hpp
```diff
@@ -167,7 +167,11 @@ inline void requestRoutesSystemPCIeDevice(App& app)
                             &pcieDevProperties["DeviceType"]);
                         property)
                     {
-                        asyncResp->res.jsonValue["DeviceType"] = *property;
+                        std::string deviceType = analysisGeneration(*property);
```
bradbishop, line 170: I think you need a new analysis function here?  analysisDeviceType() maybe?

SunnySrivastava1984: It appears to be a mistake. We do not need any analysis for "DeviceType".
Thanks for pointing this out. I will fix it.

gtmills (approved):

redfish-core/lib/pcie.hpp
```diff
@@ -167,7 +167,10 @@ inline void requestRoutesSystemPCIeDevice(App& app)
                             &pcieDevProperties["DeviceType"]);
                         property)
                     {
-                        asyncResp->res.jsonValue["DeviceType"] = *property;
+                        if (!property->empty())
```
santoshpuranik, line 170: From the validator report George shared, we need this check for DeviceClass property in the PcieFunction too.

santoshpuranik (changes requested):

redfish-core/lib/pcie.hpp
santoshpuranik: Also the FunctionType property in PcieFunction.

bradbishop: when you do this don't forget to update the commit message from:

"Handle empty Device type in PCIeDevice"

to

Handle empty Device type in PCIeDevice and PCIeFunction

since we are broadening the scope of the patch a little bit...

SunnySrivastava1984: @santoshpuranik  Fixed.  
I left them unfixed as I planned of fixing them once we get data related fix for other fields under PCIeFunctions under separate commit but I guess it makes more sense to handle all the empty enum fields under single commit.

SunnySrivastava1984: @bradbishop  Updated.

PR #196, opened 2021-12-13
Remove link from Chassis to PCIeDevice

SunnySrivastava1984:
The commit removes current support that assumes 1:1
system:Chassis for Chassis/PCIeDevices.

Current implementation populates the same collection of
PCIeDevices w.r.t chassis as w.r.t. system.
For systems with multiple chassis the current assumption
does not hold true.

Validator has been executed with no new errors.

### DIFF
--- a/redfish-core/lib/chassis.hpp
+++ b/redfish-core/lib/chassis.hpp
@@ -313,9 +313,6 @@ inline void requestRoutesChassis(App& app)
                              {"@Redfish.ActionInfo", "/redfish/v1/Chassis/" +
                                                          chassisId +
                                                          "/ResetActionInfo"}};
-                        asyncResp->res.jsonValue["PCIeDevices"] = {
-                            {"@odata.id",
-                             "/redfish/v1/Systems/system/PCIeDevices"}};
 
                         const std::string& connectionName =
                             connectionNames[0].first;


redfish-core/lib/chassis.hpp
```diff
@@ -313,9 +313,6 @@ inline void requestRoutesChassis(App& app)
                              {"@Redfish.ActionInfo", "/redfish/v1/Chassis/" +
                                                          chassisId +
                                                          "/ResetActionInfo"}};
-                        asyncResp->res.jsonValue["PCIeDevices"] = {
-                            {"@odata.id",
-                             "/redfish/v1/Systems/system/PCIeDevices"}};
```
gtmills, line 318: Can we add this link to system ?
I don't see it here.. 
https://github.com/ibm-openbmc/bmcweb/blob/1020/redfish-core/lib/systems.hpp

There should be a link since this is a child of /redfish/v1/Systems/system

SunnySrivastava1984: We are already populating collection of PCIeDevices from system.
https://github.com/ibm-openbmc/bmcweb/blob/master/redfish-core/lib/systems.hpp#L3121

bradbishop: I'm confused.  Is the problem @gtmills that the link still remains in system but we've removed it here?

gtmills: Ignore my comment. I missed that support / didn't fully understand that in the system it is not a link to a PCIe Device collection but instead a link to the PCIe Devices themselves. 

Long term I would like to see us have the PCIe Device collection under chassis because I feel that fits a lot more naturally and that is what I see in the mockups. 

It also fits better with the new way of modeling PCIe Slots. 
 
 We would still have the system to pcie devices links but those devices would be in collections under their respective chassis. 
 
 E.g. 
 https://github.com/DMTF/Redfish/blob/8ae716c9b236b3d2188a9fe83c65786b12187f1c/mockups/public-tower/Chassis/Tower/index.json

Today we have a link from system to the pcie devices (e.g. /redfish/v1/Systems/system/PCIeDevices/pcie_card1) but we have no  link to the PCIeDevices collection (/redfish/v1/Systems/system/PCIeDevices).. Which is strange. 

My future vision for this:
system:
```
{
"@odata.id": "/redfish/v1/Systems/system",
"@odata.type": "#ComputerSystem.v1_16_0.ComputerSystem",
...
"PCIeDevices": [
{
"@odata.id": "/redfish/v1/Chassis/chassis/PCIeDevices/pcie_card0"
},
{
"@odata.id": "/redfish/v1/Chassis/mex/PCIeDevices/pcie_card10"
}
]
...    

}
```

main chassis:
```
{
"@odata.id": "/redfish/v1/Chassis/chassis",
"@odata.type": "#Chassis.v1_15_0.Chassis",
...
"PCIeDevices": {
    "@odata.id": "/redfish/v1/Chassis/chassis/PCIeDevices"
 },
...    


}
```

mex chassis:
```
{
"@odata.id": "/redfish/v1/Chassis/mex",
"@odata.type": "#Chassis.v1_15_0.Chassis",
...
"PCIeDevices": {
    "@odata.id": "/redfish/v1/Chassis/mex/PCIeDevices"
 },
...   

}
```

mex chassis PCIeDevices collection (main chassis is basically the same)
```
{
"@odata.id": "/redfish/v1/Chassis/mex/PCIeDevices",
"@odata.type": 
...
"Members": [
{
"@odata.id": "/redfish/v1/Chassis/mex/PCIeDevices/pcie_card10"
}
]
```

gtmills (approved):

PR #200, opened 2021-12-14
Link PCIeSlots to Processor schema

SunnySrivastava1984:
The commit implement channges to populate link(s) to
processor associated witha given PCIeSlot.

The association is established by fetching mapper endpoints
"upstream_processor" w.r.t given PCIeSlot.

Validator is executed and no new error has been found.

Sample Output:
{
      "HotPluggable": false,
      "Lanes": 0,
      "Links": {
        "Processors": [
          {
            "@odata.id": "/redfish/v1/Systems/system/Processors/dcm0-cpu0"
          }
        ],
        "Processors@odata.count": 1
      },
      "Location": {
        "PartLocation": {
          "ServiceLabel": "U78DA.ND0.WZS0042-P0-C0"
        }
      },
      "LocationIndicatorActive": false,
      "SlotType": "FullLength"
    }
}

### DIFF
--- a/redfish-core/lib/pcie_slots.hpp
+++ b/redfish-core/lib/pcie_slots.hpp
@@ -139,13 +139,75 @@ inline void
             "xyz.openbmc_project.Inventory.Item.PCIeDevice"});
 }
 
+inline void
+    linkAssociatedProcessor(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
+                            const std::string& pcieSlotPath, size_t index)
+{
+    crow::connections::systemBus->async_method_call(
+        [asyncResp, pcieSlotPath,
+         index](const boost::system::error_code ec,
+                const std::variant<std::vector<std::string>>& endpoints) {
+            if (ec)
+            {
+                if (ec.value() == EBADR)
+                {
+                    // This PCIeSlot have no processor association.
+                    BMCWEB_LOG_DEBUG << "No processor association found";
+                    return;
+                }
+                BMCWEB_LOG_ERROR << "DBUS response error" << ec.message();
+                messages::internalError(asyncResp->res);
+                return;
+            }
+
+            const std::vector<std::string>* value =
+                std::get_if<std::vector<std::string>>(&(endpoints));
+
+            if (value == nullptr)
+            {
+                BMCWEB_LOG_DEBUG << "Error getting processor association!";
+                messages::internalError(asyncResp->res);
+                return;
+            }
+
+            if ((*value).size() < 1)
+            {
+                BMCWEB_LOG_DEBUG << "No association found for processor";
+                messages::internalError(asyncResp->res);
+                return;
+            }
+
+            const std::vector<std::string>& processorPath = *value;
+            std::string cpuName =
+                sdbusplus::message::object_path(processorPath[0]).filename();
+            sdbusplus::message::object_path parentPath =
+                sdbusplus::message::object_path(processorPath[0]).parent_path();
+            std::string dcmName = parentPath.filename();
+            std::string processorName = dcmName + '-' + cpuName;
+
+            nlohmann::json& processorArray =
+                asyncResp->res.jsonValue["Slots"][index]["Links"]["Processors"];
+            processorArray = nlohmann::json::array();
+
+            processorArray.push_back(
+                {{"@odata.id",
+                  "/redfish/v1/Systems/system/Processors/" + processorName}});
+            asyncResp->res
+                .jsonValue["Slots"][index]["Links"]["Processors@odata.count"] =
+                processorArray.size();
+        },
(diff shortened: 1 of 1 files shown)


redfish-core/lib/pcie_slots.hpp
```diff
@@ -139,13 +139,68 @@ inline void
...
+    crow::connections::systemBus->async_method_call(
+        [asyncResp, pcieSlotPath,
+         index](const boost::system::error_code ec,
+                const std::variant<std::vector<std::string>>& endpoints) {
+            if (ec)
+            {
+                if (ec.value() == EBADR)
+                {
+                    // This PCIeSlot have no processor association.
+                    BMCWEB_LOG_DEBUG << "No processor association found";
```
bradbishop, line 155: is this really necessary?  It seems like this can be deduced by the fact that there is no link.

redfish-core/lib/pcie_slots.hpp
```diff
@@ -139,13 +139,68 @@ inline void
...
+                std::get_if<std::vector<std::string>>(&(endpoints));
+
+            if (value == nullptr)
+            {
+                BMCWEB_LOG_DEBUG << "Error getting processor association!";
+                messages::internalError(asyncResp->res);
+                return;
+            }
+
+            const std::vector<std::string> processorPath = *value;
```
bradbishop, line 173: can you avoid a copy here by making processsorPath a reference?

redfish-core/lib/pcie_slots.hpp
```diff
@@ -139,13 +139,68 @@ inline void
...
+                }
+                BMCWEB_LOG_ERROR << "DBUS response error" << ec.message();
+                messages::internalError(asyncResp->res);
+                return;
+            }
+
+            const std::vector<std::string>* value =
+                std::get_if<std::vector<std::string>>(&(endpoints));
+
+            if (value == nullptr)
```
bradbishop, line 166: do you also need to check that *value has at least one element?

redfish-core/lib/pcie_slots.hpp
```diff
@@ -139,13 +139,68 @@ inline void
...
+            {
+                BMCWEB_LOG_DEBUG << "Error getting processor association!";
+                messages::internalError(asyncResp->res);
+                return;
+            }
+
+            const std::vector<std::string> processorPath = *value;
+            const std::string& cpuName =
```
bradbishop, line 174: I would have expected:

`std::string cpuName = sdbusplus::message::object_path(processorPath[0]).filename();
`

this is a little strange, and I didn't see this pattern in upstream bmcweb.  Why are we binding these temporaries to const references?

SunnySrivastava1984: Hmm, agree.

redfish-core/lib/pcie_slots.hpp
bradbishop: I was only referring to the debug trace.  I wasn't trying to suggest you remove any branches.

bradbishop: FWIW I think I find what you originally had to be better from a readability point of view.  But the logic is the same...so up to you if you want to change it back or not.

redfish-core/lib/pcie_slots.hpp
```diff
@@ -139,13 +139,74 @@ inline void
...
+            if (value == nullptr)
+            {
+                BMCWEB_LOG_DEBUG << "Error getting processor association!";
+                messages::internalError(asyncResp->res);
+                return;
+            }
+
+            if ((*value).size() < 1)
+            {
+                BMCWEB_LOG_DEBUG << "No association found for processor";
```
bradbishop, line 175: why does a nullptr generate an internalError but this doesn't?

SunnySrivastava1984: nullptr denotes that we did not get the type that we expect as endpoint hence error but having processor associated with a slot is not always true so in case there is none just a log and return without populating that link.

This was the thought process I had in mind. Please correct me if I got that wrong.

bradbishop: I was thinking if there is no processor, the association query would fail.  If the association query does not fail, yet you get here and there are no processors - doesn't that indicate some kind of logic bug?

redfish-core/lib/pcie_slots.hpp
SunnySrivastava1984: I am not able to recollect what I had earlier. :)

redfish-core/lib/pcie_slots.hpp
SunnySrivastava1984: Agree, fixed.

bradbishop (commented):
looks good to me.

bradbishop (approved):

PR #27, opened 2019-08-13
Fix "MemorySummary" status: Disabled on IBM systems

Alpana07:
"MemorySummary" Status is always Disabled.Fixed it.

Doing GET on Interface xyz.openbmc_project.State.Decorator.OperationalStatus
instead of xyz.openbmc_project.Inventory.Item.Dimm to get the State of Dimm

Tested:
-- Ran GET request on the ComputerSystem node when Host is Up
 curl -k -H "X-Auth-Token: $bmc_token" -X GET https://${bmc}/redfish/v1/Systems/system
 "MemorySummary": {
    "Status": {
      "State": "Enabled"
    },
    "TotalSystemMemoryGiB": 0
  },

Redfish schema validator.
ComputerSystem.v1_0_0.MemorySummary:Status
    value: OrderedDict([('State', 'Enabled')]) <class 'collections.OrderedDict'>
    has Type: Resource.Status complex
    is Optional
    ***going into Complex
Resource.Status:State
    value: Enabled <class 'str'>
    has Type: Resource.State enum
    is Optional
    permission OData.Permission/Read
    Success

MemorySummary.TotalSystemMemoryGiB                          PASS
MemorySummary.Status                                 complex
MemorySummary.Status.State                              PASS

Change-Id: I272a2c6c53f39be3dafd62759be4cc65aa0dd74f
Signed-off-by: Alpana Kumari <alpankum@in.ibm.com>

Conflicts:
	redfish-core/lib/systems.hpp

PR #206, opened 2021-12-17
Implement Cable schema

sbteeks:
This commit implements Cable and Cable collection schema
on bmcweb. 

Testing:
$ curl -k -X GET https://root:0penBmc@rain100bmc:18080/redfish/v1
{
  "@odata.id": "/redfish/v1",
  "@odata.type": "#ServiceRoot.v1_5_0.ServiceRoot",
  "AccountService": {
    "@odata.id": "/redfish/v1/AccountService"
  },
  "Cables": {
    "@odata.id": "/redfish/v1/Cables"
  },
...
}

$ curl -k -X GET https://root:0penBmc@rain100bmc:18080/redfish/v1/Cables
{
  "@odata.id": "/redfish/v1/Cables",
  "@odata.type": "#CableCollection.CableCollection",
  "Description": "Collection of Cable Entries",
  "Members": [
    {
      "@odata.id": "/redfish/v1/Cables/dp0_cable0"
    },
    {
      "@odata.id": "/redfish/v1/Cables/dp0_cable1"
    },
    {
      "@odata.id": "/redfish/v1/Cables/dp0_cable2"
    },
    {
      "@odata.id": "/redfish/v1/Cables/dp0_cable3"
    }
  ],
  "Members@odata.count": 4,
  "Name": "Cable Collection"
}

$ curl -k -X GET https://root:0penBmc@rain100bmc:18080/redfish/v1/Cables/dp0_cable0
{
  "@odata.id": "/redfish/v1/Cables/dp0_cable0",
  "@odata.type": "#Cable.v1_0_0.Cable",
  "CableType": "",
  "Id": "dp0_cable0",
  "LengthMeters": NaN,
  "Name": "Cable Id"
}

$ curl -k -X GET https://root:0penBmc@rain100bmc:18080/redfish/v1/Cables/badCableId
{
  "error": {
    "@Message.ExtendedInfo": [
      {
        "Message": "The requested resource of type Cables named badCableId was not found.",
        "MessageArgs": [
          "Cables",
          "badCableId"
        ],
        "MessageId": "Base.1.8.1.ResourceNotFound",
        "Resolution": "Provide a valid resource identifier and resubmit the request."
      }
    ],
    "code": "Base.1.8.1.ResourceNotFound",
    "message": "The requested resource of type Cables named badCableId was not found."
  }
}

Validator test passed:
{
    "@odata.id": "/redfish/v1/Cables",
    "@odata.type": "#CableCollection.CableCollection",
    "Description": "Collection of Cable Entries",
    "Members": [],
    "Members@odata.count": 0,
    "Name": "Cable Collection"
}

### DIFF
--- a/redfish-core/lib/cable.hpp
+++ b/redfish-core/lib/cable.hpp
@@ -1,0 +1,184 @@
+#pragma once
+#include <boost/container/flat_map.hpp>
+#include <utils/json_utils.hpp>
+
+namespace redfish
+{
+/**
+ * @brief Fill cable specific properties.
+ * @param[in,out]   resp        HTTP response.
+ * @param[in]       ec          Error code corresponding to Async method call.
+ * @param[in]       properties  List of Cable Properties key/value pairs.
+ */
+inline void
+    fillCableProperties(crow::Response& resp,
+                        const boost::system::error_code ec,
+                        const dbus::utility::DBusPropertiesMap& properties)
+{
+    if (ec)
+    {
+        BMCWEB_LOG_DEBUG << "DBUS response error " << ec;
+        messages::internalError(resp);
+        return;
+    }
+
+    for (const auto& [propKey, propVariant] : properties)
+    {
+        if (propKey == "CableTypeDescription")
+        {
+            const std::string* cableTypeDescription =
+                std::get_if<std::string>(&propVariant);
+            if (cableTypeDescription == nullptr)
+            {
+                messages::internalError(resp);
+                return;
+            }
+            resp.jsonValue["CableType"] = *cableTypeDescription;
+        }
+        else if (propKey == "Length")
+        {
+            const double* cableLength = std::get_if<double>(&propVariant);
+            if (cableLength == nullptr)
+            {
+                messages::internalError(resp);
+                return;
+            }
+
+            if (!std::isfinite(*cableLength))
+            {
+                if (std::isnan(*cableLength))
+                {
+                    continue;
+                }
+                messages::internalError(resp);
+                return;
+            }
+
+            resp.jsonValue["LengthMeters"] = *cableLength;
+        }
+    }
+}
+
+/**
+ * @brief Api to get Cable properties.
+ * @param[in,out]   asyncResp       Async HTTP response.
+ * @param[in]       cableObjectPath Object path of the Cable.
+ * @param[in]       serviceMap      A map to hold Service and corresponding
+ * interface list for the given cable id.
+ */
+inline void
+    getCableProperties(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
+                       const std::string& cableObjectPath,
+                       const dbus::utility::MapperServiceMap& serviceMap)
+{
+    BMCWEB_LOG_DEBUG << "Get Properties for cable " << cableObjectPath;
+
+    for (const auto& [service, interfaces] : serviceMap)
+    {
+        for (const auto& interface : interfaces)
+        {
+            if (interface != "xyz.openbmc_project.Inventory.Item.Cable")
+            {
+                continue;
+            }
+
+            crow::connections::systemBus->async_method_call(
+                [asyncResp](
+                    const boost::system::error_code ec,
+                    const dbus::utility::DBusPropertiesMap& properties) {
+                    fillCableProperties(asyncResp->res, ec, properties);
+                },
+                service, cableObjectPath, "org.freedesktop.DBus.Properties",
+                "GetAll", interface);
+        }
+    }
+}
+
+/**
+ * The Cable schema
+ */
+inline void requestRoutesCable(App& app)
+{
+    BMCWEB_ROUTE(app, "/redfish/v1/Cables/<str>/")
+        .privileges(redfish::privileges::getCable)
+        .methods(boost::beast::http::verb::get)(
+            [](const crow::Request&,
+               const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
+               const std::string& cableId) {
+                BMCWEB_LOG_DEBUG << "Cable Id: " << cableId;
+                auto respHandler =
+                    [asyncResp,
+                     cableId](const boost::system::error_code ec,
+                              const dbus::utility::MapperGetSubTreeResponse&
+                                  subtree) {
+                        if (ec.value() == EBADR)
+                        {
+                            messages::resourceNotFound(
+                                asyncResp->res, "#Cable.v1_0_0.Cable", cableId);
+                            return;
+                        }
+
+                        if (ec)
+                        {
+                            BMCWEB_LOG_ERROR << "DBUS response error " << ec;
+                            messages::internalError(asyncResp->res);
+                            return;
+                        }
+
+                        for (const auto& [objectPath, serviceMap] : subtree)
+                        {
+                            sdbusplus::message::object_path path(objectPath);
+                            if (path.filename() != cableId)
+                            {
+                                continue;
+                            }
+
+                            asyncResp->res.jsonValue["@odata.type"] =
+                                "#Cable.v1_0_0.Cable";
+                            asyncResp->res.jsonValue["@odata.id"] =
+                                "/redfish/v1/Cables/" + cableId;
+                            asyncResp->res.jsonValue["Id"] = cableId;
+                            asyncResp->res.jsonValue["Name"] = "Cable";
+
+                            getCableProperties(asyncResp, objectPath,
+                                               serviceMap);
+                            return;
+                        }
+                        messages::resourceNotFound(asyncResp->res, "Cable",
+                                                   cableId);
+                    };
+
+                crow::connections::systemBus->async_method_call(
+                    respHandler, "xyz.openbmc_project.ObjectMapper",
+                    "/xyz/openbmc_project/object_mapper",
+                    "xyz.openbmc_project.ObjectMapper", "GetSubTree",
+                    "/xyz/openbmc_project/inventory", 0,
+                    std::array<const char*, 1>{
+                        "xyz.openbmc_project.Inventory.Item.Cable"});
+            });
+}
+
+/**
+ * Collection of Cable resource instances
+ */
+inline void requestRoutesCableCollection(App& app)
+{
+    BMCWEB_ROUTE(app, "/redfish/v1/Cables/")
+        .privileges(redfish::privileges::getCableCollection)
+        .methods(boost::beast::http::verb::get)(
+            [](const crow::Request&,
+               const std::shared_ptr<bmcweb::AsyncResp>& asyncResp) {
+                asyncResp->res.jsonValue["@odata.type"] =
+                    "#CableCollection.CableCollection";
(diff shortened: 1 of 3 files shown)


redfish-core/lib/cable.hpp
```diff
@@ -0,0 +1,198 @@
...
+    asyncResp->res.jsonValue["CableType"] = *cableTypeDescription;
+}
+
+/**
+ * The Cable schema
+ */
+inline void requestRoutesCable(App& app)
+{
+    BMCWEB_ROUTE(app, "/redfish/v1/Cables/<str>/")
+        .privileges({{"Login"}})
```
SunnySrivastava1984, line 57: Privileges needs to be filled from "redfish::privileges::"

redfish-core/lib/cable.hpp
```diff
@@ -0,0 +1,198 @@
...
+        }
+    }
+
+    if (cableTypeDescription == nullptr || cableLength == nullptr)
+    {
+        messages::internalError(asyncResp->res);
+        return;
+    }
+
+    std::string dataId = "/redfish/v1/Cables/" + cableId;
```
SunnySrivastava1984, line 42: We can get rid of this variable by substituting this in-place @ line 44?

redfish-core/lib/cable.hpp
```diff
@@ -0,0 +1,198 @@
...
+    asyncResp->res.jsonValue["@odata.type"] = "#Cable.v1_0_0.Cable";
+    asyncResp->res.jsonValue["@odata.id"] = dataId;
+    asyncResp->res.jsonValue["Id"] = cableId;
+    asyncResp->res.jsonValue["Name"] = "Cable Id";
+    asyncResp->res.jsonValue["LengthMeters"] = *cableLength;
```
SunnySrivastava1984, line 47: If these are not mandatory properties to be filled in schema then we can fill them directly at line 27 and 32. This way we can get rid of check at line 36 and variable at line 20 and 21.

redfish-core/lib/cable.hpp
```diff
@@ -0,0 +1,198 @@
...
+                };
+
+                crow::connections::systemBus->async_method_call(
+                    respHandler, "xyz.openbmc_project.Inventory.Manager",
+                    "/xyz/openbmc_project/inventory/cables/" + cableId,
+                    "org.freedesktop.DBus.Properties", "GetAll", "");
+            });
+
+    BMCWEB_ROUTE(app, "/redfish/v1/Cables/<str>/")
+        .privileges({{"Login"}})
```
SunnySrivastava1984, line 83: Same as line 57.

SunnySrivastava1984 (changes requested):

redfish-core/lib/cable.hpp
sbteeks: I'm not sure what exactly your ask is here, but assuming to make the changes in privilege_registry.hpp. We are not supposed to be modifying privilege_registry.hpp file directly as it gets overwritten by parse_registries.py script. 
@gtmills any thoughts?

redfish-core/lib/cable.hpp
sbteeks: I'm not sure whether they are mandatory or not. If they are mandatory then we need to have that check at line#36. Also, we need to have those variables at line#20 and 21 to get the corresponding properties if the properties are mandatory. Otherwise we need to make the change to line #27 and 32 to to make those as variables and then check if either one or both a nullptr. 

IMO, it is good to leave it as is since it is more readable and are at one place.

@gtmills, @bradbishop : Are the description and cable length are mandatory? The default description is empty string and cable length is NaN.

redfish-core/lib/cable.hpp
sbteeks: Let us see what @gtmills says...

redfish-core/lib/cable.hpp
sbteeks: '"readonly" : true' in the schema means mandatory - correct?

redfish-core/lib/cable.hpp
sbteeks: Here is one of the privilege entries in chassis.hpp:
.privileges(redfish::privileges::postChassis)

Here is where 'postChassis' is defined (privilege_registry.hpp):
// Chassis
const static auto& getChassis = privilegeSetLogin;
const static auto& headChassis = privilegeSetLogin;
const static auto& patchChassis = privilegeSetConfigureComponents;
const static auto& putChassis = privilegeSetConfigureComponents;
const static auto& deleteChassis = privilegeSetConfigureComponents;
const static auto& postChassis = privilegeSetConfigureComponents;

For Cables if you are expecting to define new entries similar to the above?
If so, it is a 'no no' to add new entries in the privilege_registry.hpp - this is an auto generated file and we are not supposed to modify it. 
Here is the comment in the privilege_registry.hpp file:
#pragma once
// privilege_registry.hpp is generated.  Do not edit directly
...

While it is possible that I'm missing something to follow based on your comment, please be more specific so that I can address it.

redfish-core/lib/cable.hpp
sbteeks: made the change taking into account whether these properties are mandatory or not...

redfish-core/lib/cable.hpp
bradbishop: I suspect patching a cable requires at least ConfigureComponents.  But I don't think we require patch support for cables anyway, so maybe we can skip patch support for now?

redfish-core/lib/cable.hpp
bradbishop: when you regenerate privilege_registry.hpp with the cables schema included, are there any new lines with any cable specific permissions?

bradbishop (changes requested):
This is looking pretty good.  I don't see any reason why this can't get a review from the upstream community.

We don't need patch support, so that can(should) be dropped.

redfish-core/lib/cable.hpp
sbteeks: @gtmills what is it we have to do for the privileges?

gtmills: There is a new Redfish_1.2.0_PrivilegeRegistry.json 
https://redfish.dmtf.org/registries/

Need to pick up by changing this line and rerunning the script
https://github.com/openbmc/bmcweb/blob/83f984b53eef3277d9d3da292152aec1250f5f4d/scripts/parse_registries.py#L145

It does include cables. 
If you don't mind please push this commit upstream. :)

redfish-core/lib/cable.hpp
```diff
@@ -0,0 +1,136 @@
...
+    getCableProperties(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
+                       ManagedPropertyType& resp, const std::string& cableId)
+{
+    for (auto& propertyMap : resp)
+    {
+        if (propertyMap.first == "CableTypeDescription")
+        {
+            std::string* cableTypeDescription =
+                std::get_if<std::string>(&propertyMap.second);
+            if (cableTypeDescription != nullptr)
```
gtmills, line 26: Typically we do the opposite check here, 
             if (cableTypeDescription == nullptr)
             {
                internalErorr;
                return
             }

redfish-core/lib/cable.hpp
```diff
@@ -0,0 +1,136 @@
...
+            {
+                asyncResp->res.jsonValue["CableType"] = *cableTypeDescription;
+            }
+        }
+        else if (propertyMap.first == "Length")
+        {
+            double* cableLength = std::get_if<double>(&propertyMap.second);
+            if (cableLength != nullptr)
```
gtmills, line 34: again typically it is  
if (cableLength == nullptr) 
{
internalError; 
return

redfish-core/lib/cable.hpp
```diff
@@ -0,0 +1,136 @@
...
+    asyncResp->res.jsonValue["Id"] = cableId;
+    asyncResp->res.jsonValue["Name"] = "Cable Id";
+}
+
+/**
+ * The Cable schema
+ */
+inline void requestRoutesCable(App& app)
+{
+    BMCWEB_ROUTE(app, "/redfish/v1/Cables/<str>/")
```
gtmills, line 52: How does this handle 
"/redfish/v1/Cables/badCableId/" ?

gtmills: See https://github.com/openbmc/bmcweb/blob/master/COMMON_ERRORS.md#11-not-responding-to-404

gtmills: Or what if you don't have any xyz.openbmc_project.Inventory.Item.Cable and you try to go to this route ?

redfish-core/lib/cable.hpp
sbteeks: If the properties are not mandatory, should we still return with an internal error?

redfish-core/lib/cable.hpp
sbteeks: Done

redfish-core/lib/cable.hpp
sbteeks: This is taken care off now...

redfish-core/lib/cable.hpp
sbteeks: Done

redfish-core/lib/cable.hpp
```diff
@@ -0,0 +1,147 @@
...
+                return;
+            }
+            asyncResp->res.jsonValue["LengthMeters"] = *cableLength;
+        }
+    }
+
+    asyncResp->res.jsonValue["@odata.type"] = "#Cable.v1_0_0.Cable";
+    asyncResp->res.jsonValue["@odata.id"] = "/redfish/v1/Cables/" + cableId;
+    asyncResp->res.jsonValue["Id"] = cableId;
+    asyncResp->res.jsonValue["Name"] = "Cable Id";
```
gtmills, line 48: Kind of an interesting name? 
We could look at prettyName here..
Or maybe call it the same as the Id, cableId? 
Or maybe call it just call it "Cable"

redfish-core/lib/cable.hpp
```diff
@@ -0,0 +1,147 @@
...
+                        BMCWEB_LOG_DEBUG << "DBUS response error " << ec;
+                        messages::internalError(asyncResp->res);
+                        return;
+                    }
+
+                    getCableProperties(asyncResp, resp, cableId);
+                };
+
+                crow::connections::systemBus->async_method_call(
+                    respHandler, "xyz.openbmc_project.Inventory.Manager",
```
gtmills, line 84: Upstream we will not be able to call Inventory Manager and will instead call the mapper. 
For downstream, I guess I am fine here.

gtmills (approved):

gtmills:
Upstream review is https://gerrit.openbmc-project.xyz/c/openbmc/bmcweb/+/50265. Has comments upstream that end addressed here as well.

gtmills:
This has merged upstream. 
I am good here

gtmills (approved):

PR #207, opened 2021-12-17
WIP: Use boost timer with 100 connection limit

gtmills:
WIP: Need to test.

This PR
Reverts https://github.com/ibm-openbmc/bmcweb/commit/6dc5b8923fcecb3d3062e2b9a127262e4bfd26ed
Cherry picks the following 3 merged commits from upstream:
https://gerrit.openbmc-project.xyz/c/openbmc/bmcweb/+/49318
https://gerrit.openbmc-project.xyz/c/openbmc/bmcweb/+/49380/
https://gerrit.openbmc-project.xyz/c/openbmc/bmcweb/+/49380/

I find this solution best because
    These have merged upstream and is the direction upstream is headed.  
    It simplifies the logic over the current timer queue 
    I find bumping the current timer queue from 100 to 300 a workaround, a bit of a hack
    We should have a connection limit


http/http_connection.hpp
```diff
@@ -300,10 +293,6 @@ class Connection :
                                         const boost::system::error_code& ec) {
                                         if (ec)
                                         {
-                                            BMCWEB_LOG_ERROR
```
geissonator, line 303: Hmm, was this a useful error that we should keep?

http/timer_queue.hpp
```diff
@@ -43,7 +43,6 @@ class TimerQueue
     {
         if (dq.size() == maxSize)
         {
-            BMCWEB_LOG_ERROR << "timer add: dq.size has hit maxSize";
```
geissonator, line 46: Still want this message?

http/http_connection.hpp
```diff
@@ -278,6 +274,12 @@ class Connection :
 
     void start()
     {
+        if (connectionCount >= 100)
+        {
+            BMCWEB_LOG_CRITICAL << this << "Max connection count exceeded.";
```
geissonator, line 279: Ahh, we get the count limit message here so we're good on that one.

http/timer_queue.hpp
gtmills: if (connectionCount >= 100)
        {
            BMCWEB_LOG_CRITICAL << this << "Max connection count exceeded.";
            
            
should cover us. 
This use case isn't really valid anymore.

http/http_connection.hpp
gtmills: Yeah.

http/http_connection.hpp
gtmills: It didn't hurt so added back. Done.

geissonator (approved):

PR #210, opened 2021-12-29
Redfish Validator: Fix Bios/Settings response

sunharis:
The Attributes listed under Bios/Settings was displayed as null
causing the validator failure

This commit initializes the Attributes property as an object. When
the GET is done on the Bios/Settings, its returned as {}

Tested by:
  1. GET on the /redfish/v1/Systems/system/Bios/Settings
  2. Ran Redfish validator

Change-Id: Ib076b49d9b84d8c71f7d9bc61567ca2d0eff4410


sunharis:
Downstream only code.

gtmills:
Can we comment upstream on the bios attributes commit so we don't forget about this later?

gtmills (approved):

sunharis:
> Can we comment upstream on the bios attributes commit so we don't forget about this later?

@gtmills I am not a reviewer there i think. Please comment there/add me as reviewer

raviteja-b (approved):

PR #211, opened 2021-12-29
Disallow no-access user login

sunharis:
An authenticated user with the NoAccess role can login but they cannot
logout.

This commit is to block the login for a user with priv-noaccess
Details are at https://github.com/openbmc/bmcweb/issues/227

Testedby:
 1. Create an LDAP user with priv-noaccess. Verify the login attempt fails
    with accessDenied error
 2. Verified the other role users can login

Change-Id: I76c0a01c152d0019ce0713da013c565ed420d0ad

### DIFF
--- a/redfish-core/lib/redfish_sessions.hpp
+++ b/redfish-core/lib/redfish_sessions.hpp
@@ -140,84 +140,131 @@ inline void requestRoutesSession(App& app)
         // endpoint responsible for giving the login privilege, and it is itself
         // its own route, it needs to not require Login
         .privileges({})
-        .methods(boost::beast::http::verb::post)(
-            [](const crow::Request& req,
-               const std::shared_ptr<bmcweb::AsyncResp>& asyncResp) -> void {
-                std::string username;
-                std::string password;
-                std::optional<nlohmann::json> oemObject;
-                std::string clientId;
-                if (!json_util::readJson(req, asyncResp->res, "UserName",
-                                         username, "Password", password, "Oem",
-                                         oemObject))
+        .methods(boost::beast::http::verb::post)([](const crow::Request& req,
+                                                    const std::shared_ptr<
+                                                        bmcweb::AsyncResp>&
+                                                        asyncResp) -> void {
+            std::string username;
+            std::string password;
+            std::optional<nlohmann::json> oemObject;
+            std::string clientId;
+            if (!json_util::readJson(req, asyncResp->res, "UserName", username,
+                                     "Password", password, "Oem", oemObject))
+            {
+                return;
+            }
+
+            if (password.empty() || username.empty() ||
+                asyncResp->res.result() != boost::beast::http::status::ok)
+            {
+                if (username.empty())
                 {
-                    return;
+                    messages::propertyMissing(asyncResp->res, "UserName");
                 }
 
-                if (password.empty() || username.empty() ||
-                    asyncResp->res.result() != boost::beast::http::status::ok)
+                if (password.empty())
                 {
-                    if (username.empty())
-                    {
-                        messages::propertyMissing(asyncResp->res, "UserName");
-                    }
-
-                    if (password.empty())
-                    {
-                        messages::propertyMissing(asyncResp->res, "Password");
-                    }
-
-                    return;
+                    messages::propertyMissing(asyncResp->res, "Password");
                 }
 
-                int pamrc = pamAuthenticateUser(username, password);
-                bool isConfigureSelfOnly = pamrc == PAM_NEW_AUTHTOK_REQD;
-                if ((pamrc != PAM_SUCCESS) && !isConfigureSelfOnly)
-                {
-                    messages::resourceAtUriUnauthorized(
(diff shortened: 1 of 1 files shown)


sunharis:
Upstream commit : https://gerrit.openbmc-project.xyz/c/openbmc/bmcweb/+/49295

gtmills:
Why can't we follow what we will do upstream and do this change in phosphor-user-manager ?

joseph-reynolds:
I reviewed the upstream discussion in https://gerrit.openbmc-project.xyz/c/openbmc/bmcweb/+/49295 
It seems to me that:
- The upstream discussion is not yet settled, and will take some time to work out.
- We must move forward with fw1020 to make our development dates.
- Removing the NoAccess role from the fw1020 release would be technically hard to do.
- The code in this PR appears correct.  It implements the solution I advocated.  Specifically, when authentication succeeds we check if the user attempting to login has the "Login" privilege, and send HTTP status 403 forbidden (access denied) if not.

So, I think it is okay to move forward with this PR.

redfish-core/lib/redfish_sessions.hpp
```diff
@@ -140,84 +140,134 @@ inline void requestRoutesSession(App& app)
...
+                        userRolePtr =
+                            std::get_if<std::string>(&userInfoIter->second);
+                    }
+                    std::string userRole{};
+                    if (userRolePtr != nullptr)
+                    {
+                        userRole = *userRolePtr;
+                        BMCWEB_LOG_DEBUG << "userName = " << username
+                                         << " userRole = " << *userRolePtr;
+                        ::redfish::Privileges userPrivileges =
```
gtmills, line 229: I have never seen it like this, ::redfish, can we have these match, e.g.  redfish::privileges::getSessionService

sunharis: Done

raviteja-b (approved):

redfish-core/lib/redfish_sessions.hpp
```diff
@@ -140,84 +140,134 @@ inline void requestRoutesSession(App& app)
...
+                    const boost::system::error_code ec,
+                    std::map<std::string,
+                             std::variant<bool, std::string,
+                                          std::vector<std::string>>>
+                        userInfo) {
+                    if (ec)
                     {
+                        BMCWEB_LOG_ERROR << "GetUserInfo failed...";
+                        asyncResp->res.result(
+                            boost::beast::http::status::internal_server_error);
```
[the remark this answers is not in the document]
sunharis: done

gtmills:
Don't we also need to do something here for the login route ?
https://github.com/ibm-openbmc/bmcweb/blob/938f25685f4b93d632fe5169cf5e258164538114/include/login_routes.hpp#L21

sunharis:
> Don't we also need to do something here for the login route ?
> 
> https://github.com/ibm-openbmc/bmcweb/blob/938f25685f4b93d632fe5169cf5e258164538114/include/login_routes.hpp#L21

Probably yes. But we can keep that separate.

joseph-reynolds:
Yes, the /login URI is similarly affected.  I agree this can be a separate PR.
What happens if we don't fix the /login route before GA?  Technically I could argue that it may contribute toward a denial of service attack (DoS), but no more than any network agent who can authenticate to the BMC.

The technical fix is to perform the same authority check in the /login method after pamAuthenticateUser() succeeded, and before we create the session.  Around here: https://github.com/ibm-openbmc/bmcweb/blob/938f25685f4b93d632fe5169cf5e258164538114/include/login_routes.hpp#L142

gtmills:
> Yes, the /login URI is similarly affected. I agree this can be a separate PR.
What happens if we don't fix the /login route before GA?

I think the /login route will be used more than this route because the GUI uses /login. 
I am not sure what the HMC uses? 
I think we should fix both or none for GA

sunharis:
> I think the /login route will be used more than this route because the GUI uses /login. I am not sure what the HMC uses? I think we should fix both or none for GA

HMC uses the redfish session which is addresses in this PR

sunharis:
> > Yes, the /login URI is similarly affected. I agree this can be a separate PR.
> > What happens if we don't fix the /login route before GA?
> 
> I think the /login route will be used more than this route because the GUI uses /login. I am not sure what the HMC uses? I think we should fix both or none for GA

Created https://github.com/ibm-openbmc/bmcweb/pull/248 to fix the GUI path

sunharis:
We will pull the upstream changes after that is concluded. Changes in this PR is not needed.

prkatti1:
NoAccess is very important feature for admins & for time based privilege.
Here Admin has an option to remove access/change privilege based on time & based on person who is going to use.
If admins wants he can remove access temporarily or can give access only while specific people are using or on time based access. So here admin thinks that users won't be able to access or read when NoAccess privilege is applied, and that is not working as expected so basic function of NoAccess privilege is broken as NoAccess user is getting un-authorized access & elevated access.

PR #218, opened 2022-01-06
EnvironmentMetrics: Fix Redfish Validator Error

lxwinspur:
Resolved: ibm-openbmc#3425

### DIFF
--- a/redfish-core/lib/environment_metrics.hpp
+++ b/redfish-core/lib/environment_metrics.hpp
@@ -243,11 +243,7 @@ inline void
 
                 asyncResp->res.jsonValue["PowerLimitWatts"]["SetPoint"] = 0;
                 asyncResp->res.jsonValue["PowerLimitWatts"]["ControlMode"] =
-                    "automatic";
-                asyncResp->res
-                    .jsonValue["PowerLimitWatts"]
-                              ["ControlMode@Redfish.AllowableValues"] = {
-                    "automatic", "disabled"};
+                    "Automatic";
 
                 crow::connections::systemBus->async_method_call(
                     [asyncResp](const boost::system::error_code ec2,
@@ -296,7 +292,7 @@ inline void
                                 {
                                     asyncResp->res.jsonValue["PowerLimitWatts"]
                                                             ["ControlMode"] =
-                                        "disabled";
+                                        "Disabled";
                                 }
                             }
                         }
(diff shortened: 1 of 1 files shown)


aahmed-2 (approved):

aahmed-2:
jenkins run tests please

redfish-core/lib/environment_metrics.hpp
```diff
@@ -243,11 +243,7 @@ inline void
 
                 asyncResp->res.jsonValue["PowerLimitWatts"]["SetPoint"] = 0;
                 asyncResp->res.jsonValue["PowerLimitWatts"]["ControlMode"] =
-                    "automatic";
-                asyncResp->res
-                    .jsonValue["PowerLimitWatts"]
-                              ["ControlMode@Redfish.AllowableValues"] = {
```
gtmills, line 249: Don't we still want AllowableValues? 
How does the client know what enum values we accept?

aahmed-2: I saw this too, but I thought AllowableValues might not be a valid property in the EnvironmentMetrics schema PowerWatts. https://redfish.dmtf.org/schemas/v1/EnvironmentMetrics.v1_1_0.json

lxwinspur: > I saw this too, but I thought AllowableValues might not be a valid property in the EnvironmentMetrics schema PowerWatts. https://redfish.dmtf.org/schemas/v1/EnvironmentMetrics.v1_1_0.json

Correct, so I removed this property.

gtmills: screenshot
From the Redfish spec. It should be allowed on any property

lxwinspur: > screenshot From the Redfish spec. It should be allowed on any property

I checked the validator code:
https://github.com/DMTF/Redfish-Service-Validator/blob/master/common/catalog.py#L382-L391
Since the excerpt of the ControlMode attribute of Control_v1.xml does not contain `String`, the obtained excerptType = NEUTRAL

http://redfish.dmtf.org/schemas/v1/Control_v1.xml

```
<Property Name="ControlMode" Type="Control.v1_0_0.ControlMode">
<Annotation Term="OData.Permissions" EnumMember="OData.Permission/ReadWrite"/>
<Annotation Term="OData.Description" String="The current operating mode of the control."/>
<Annotation Term="OData.LongDescription" String="This property shall contain the operating mode of the control."/>
<Annotation Term="Redfish.Excerpt"/>
</Property>
```

So the validator will execute here:
https://github.com/DMTF/Redfish-Service-Validator/blob/master/validateRedfish.py#L34

@gtmills Please correct me if I am wrong, thanks!

gtmills (approved):
Still think we need something to tell the clients about the allowable values but I am good getting this in

PR #228, opened 2022-01-17
Push-style event: Delete subscriber after retry failures

sunharis:
Push style eventing implements sendEvent retries based on the
maximum retry limits set at the event service. When the max retry
is hit, it means that the subscriber is gone and bmc need not
attempt sending the event to that suscriber when next event occurs.
BMC should delete the subcription in this case.

Current implementation does not delete the subscription on hitting
maximum number of sendEvent retries. This overloads the http_client
at bmc to attempt a sendEvent to a non-existing or stale subscription.

This commit implements
1. Delete subscription after max retires
2. Store subscription id as a member of subscriber
3. Some trace enhancements at http client

Tested by:
 1. Subscribe to events and close the subscriber abruptly to see
    bmc retries and after all retry attempt failures, stops sending
    the events to that subscriber
 2. A bmc restart should cleanup these stale subscribers
 3. Tested sanity of the existing push-style eventing feature

Change-Id: Idae9b4b4a8a4cc25239ea4637b7bde9f2e380b4f

### DIFF
--- a/http/http_client.hpp
+++ b/http/http_client.hpp
@@ -355,7 +355,8 @@ class HttpClient : public std::enable_shared_from_this<HttpClient>
     {
         if (retryCount >= maxRetryAttempts)
         {
-            BMCWEB_LOG_ERROR << "Maximum number of retries reached.";
+            BMCWEB_LOG_ERROR
+                << "Maximum number of retries reached for Subscriber:" << subId;
 
             // Clear queue.
             while (!requestDataQueue.empty())
@@ -366,12 +367,20 @@ class HttpClient : public std::enable_shared_from_this<HttpClient>
             BMCWEB_LOG_DEBUG << "Retry policy: " << retryPolicyAction;
             if (retryPolicyAction == "TerminateAfterRetries")
             {
-                // TODO: delete subscription
                 state = ConnState::terminated;
+                // Remove the subscription
+                BMCWEB_LOG_DEBUG << "TerminateAfterRetries is set. retryCount: "
+                                 << retryCount << " .Subscriber: " << subId
+                                 << " deleted";
+                persistent_data::EventServiceStore::getInstance()
+                    .removeSubscription(subId);
             }
             if (retryPolicyAction == "SuspendRetries")
             {
                 state = ConnState::suspended;
+                BMCWEB_LOG_ERROR
+                    << "SuspendRetries is set. retryCount: " << retryCount
+                    << " .Subscriber: " << this->subId << "suspended";
             }
             // Reset the retrycount to zero so that client can try connecting
             // again if needed
@@ -440,9 +449,13 @@ class HttpClient : public std::enable_shared_from_this<HttpClient>
                 break;
             }
             case ConnState::suspended:
-            case ConnState::terminated:
             {
                 doClose();
+                break;
+            }
+            case ConnState::terminated:
+            {
+                BMCWEB_LOG_ERROR << "Subscriber connection terminated. Stop";
                 break;
             }
             case ConnState::resolveFailed:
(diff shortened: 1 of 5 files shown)


sunharis:
The upstream bmcweb does not have the https client enabled with ssl. So the code is considerably different. I will create an upstream commit separately with this change. Same code can not be pushed upstream

include/event_service_store.hpp
```diff
@@ -258,6 +258,32 @@ class EventServiceStore
...
+    }
+
+    void persistSubscription()
+    {
+        needWrite = true;
+        return;
+    }
+    void removeSubscription(const std::string& id)
+    {
+        BMCWEB_LOG_ERROR << "removeSubscription: " << id;
```
raviteja-b, line 274: Info log?

raviteja-b (approved):

redfish-core/include/event_service_manager.hpp
```diff
@@ -559,8 +563,15 @@ class Subscription : public persistent_data::UserSubscription
         return eventSeqNum;
     }
 
+    void setSubId(const std::string& id)
+    {
+        subId = id;
+        BMCWEB_LOG_ERROR << "subId saved as: " << subId;
```
gtmills, line 569: Why Error here?

gtmills (commented):
Seems like we added a lot of BMCWEB_LOG_ERROR, do we really need these all?

redfish-core/include/event_service_manager.hpp
sunharis: removed this

include/event_service_store.hpp
sunharis: removed this

sunharis:
> Seems like we added a lot of BMCWEB_LOG_ERROR, do we really need these all?

I added some at the retry path and while removing the subscription, so that they are logged at the journal traces

include/event_service_store.hpp
```diff
@@ -258,6 +258,31 @@ class EventServiceStore
...
+        auto obj = subscriptionsConfigMap.find(id);
+        if (obj != subscriptionsConfigMap.end())
+        {
+            BMCWEB_LOG_ERROR << "Deleting " << id
+                             << " from subscriptionsConfigMap";
+            subscriptionsConfigMap.erase(obj);
+            needWrite = true;
+            return;
+        }
+        BMCWEB_LOG_ERROR << "Subscriber " << id
```
gtmills, line 283: Is this really needed as an LOG ERROR?

include/event_service_store.hpp
```diff
@@ -258,6 +258,31 @@ class EventServiceStore
...
+    {
+        needWrite = true;
+        return;
+    }
+    void removeSubscription(const std::string& id)
+    {
+        auto obj = subscriptionsConfigMap.find(id);
+        if (obj != subscriptionsConfigMap.end())
+        {
+            BMCWEB_LOG_ERROR << "Deleting " << id
```
gtmills, line 277: Is this really needed as an LOG ERROR since you have 

                BMCWEB_LOG_ERROR << "TerminateAfterRetries is set. retryCount: "
                                 << retryCount << " .Subscriber: " << subId
                                 << " deleted";
                                 
                                 above?

http/http_client.hpp
```diff
@@ -366,12 +367,20 @@ class HttpClient : public std::enable_shared_from_this<HttpClient>
             BMCWEB_LOG_DEBUG << "Retry policy: " << retryPolicyAction;
             if (retryPolicyAction == "TerminateAfterRetries")
             {
-                // TODO: delete subscription
                 state = ConnState::terminated;
+                // Remove the subscription
+                BMCWEB_LOG_ERROR << "TerminateAfterRetries is set. retryCount: "
```
gtmills, line 372: Can we have this one or the one in event_service_store.hpp on 277?

include/event_service_store.hpp
sunharis: This is hit if there is an error actually. ie the subscriber is asked to be removed, but its not found at the subscription config map

PR #230, opened 2022-01-20
Add logging to internal error

geissonator:
The goal here is to just bring in the upstream commit to add logging when a InternalError is returned to the user. The features in that commit though required c++20. Bringing in c++20 brought in other upstream commit requirements. Nothing seems all that risky here but there are quite a few commits. The only one I didn't pull in from upstream is the last one. I'm pretty sure this was also fixed upstream but in a more complex manner. I went with the simpler solution in our downstream code (utilize auto).

Upstream Commits:
https://gerrit.openbmc-project.xyz/c/openbmc/bmcweb/+/49313
https://gerrit.openbmc-project.xyz/c/openbmc/bmcweb/+/48547
https://gerrit.openbmc-project.xyz/c/openbmc/bmcweb/+/47574
https://gerrit.openbmc-project.xyz/c/openbmc/bmcweb/+/49650
https://gerrit.openbmc-project.xyz/c/openbmc/bmcweb/+/49450
https://gerrit.openbmc-project.xyz/c/openbmc/bmcweb/+/50093

### DIFF
--- a/redfish-core/src/error_messages.cpp
+++ b/redfish-core/src/error_messages.cpp
@@ -228,8 +228,11 @@ nlohmann::json internalError(void)
                        "consider resetting the service."}};
 }
 
-void internalError(crow::Response& res)
+void internalError(crow::Response& res, const bmcweb::source_location location)
 {
+    BMCWEB_LOG_CRITICAL << "Internal Error " << location.file_name() << "("
+                        << location.line() << ":" << location.column() << ") `"
+                        << location.function_name() << "`: ";
     res.result(boost::beast::http::status::internal_server_error);
     addMessageToErrorJson(res.jsonValue, internalError());
 }
--- a/include/openbmc_dbus_rest.hpp
+++ b/include/openbmc_dbus_rest.hpp
@@ -1395,8 +1395,7 @@ inline void findActionOnInterface(
                                     argumentNode->NextSiblingElement("arg");
                             }
 
-                            nlohmann::json::const_iterator argIt =
-                                transaction->arguments.begin();
+                            auto argIt = transaction->arguments.begin();
 
                             argumentNode = methodNode->FirstChildElement("arg");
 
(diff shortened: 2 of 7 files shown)


gtmills:
Why did we need to pull in C++20 ?

geissonator:
> Why did we need to pull in C++20 ? Note: I am good with pulling this change in, I think this will be really useful (we have some internal errors that are bad requests, etc that we might need to fix).

The change from Ed to figure out source info on the internal errors required c++ 20 so I had to pull it in. Seems like we still have a CI issue though that needs looking into. I was hoping this was just gonna be a real easy one to pull in....but it's never easy :(

redfish-core/src/error_messages.cpp
```diff
@@ -228,8 +228,11 @@ nlohmann::json internalError(void)
                        "consider resetting the service."}};
 }
 
-void internalError(crow::Response& res)
+void internalError(crow::Response& res, const bmcweb::source_location location)
 {
+    BMCWEB_LOG_CRITICAL << "Internal Error " << location.file_name() << "("
```
sunharis, line 233: Does BMCWEB_LOG_CRITICAL gets an entry into the journal traces by default ?

sunharis (commented):
Seems like CI fails for some downstream-only coding style due to C++20

redfish-core/src/error_messages.cpp
geissonator: Yeah, the critical level is higher then the error level so we will get both.

gtmills:
I am good with this, can you include the list of upstream commits you pulled in for this?

geissonator:
> I am good with this, can you include the list of upstream commits you pulled in for this?

Sure, I added up to the top message

gtmills (approved):
These are all merged upstream

PR #231, opened 2022-01-21
Fix bmcweb::redfish bug Fan Health does not reflect "Functional"

lxwinspur:
1. Missing fan "Health" state not updated to reflect "Functional"
   dbus property for fan
   https://github.com/ibm-openbmc/dev/issues/3449

2. Repeatedly adding [ERROR "fan.hpp":587] The fan obtained at this
   time does not belong to chassis to journal
   https://github.com/ibm-openbmc/dev/issues/3448

### DIFF
--- a/redfish-core/lib/fan.hpp
+++ b/redfish-core/lib/fan.hpp
@@ -64,7 +64,7 @@ inline void getFanHealth(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
             }
             if (*value == false)
             {
-                asyncResp->res.jsonValue["Status"]["State"] = "Absent";
+                asyncResp->res.jsonValue["Status"]["Health"] = "Critical";
             }
         },
         connectionName, path, "org.freedesktop.DBus.Properties", "Get",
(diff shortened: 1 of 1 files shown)


msbarth:
I believe you have limited systems to be able to test on, so could I get an idea on how this was tested as I'm not overly familiar with bmcweb/redfish code and am reserved over these changes possibly not being the full fixes necessary and/or cause other issues?

Also, I think there was an upstream review going with changes to this area of bmcweb/redfish code...were these changes pushed up there as well?

redfish-core/lib/fan.hpp
```diff
@@ -562,8 +562,6 @@ inline void getValidFan(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
                     }
                     if (chassisName != chassisId)
                     {
-                        BMCWEB_LOG_ERROR << "The fan obtained at this time "
```
gtmills, line 565: I am good removing these because we have multiple chassis and these would log all the time.

gtmills (approved):
Please answer Matt's questions but I am good here.

lxwinspur:
> I believe you have limited systems to be able to test on, so could I get an idea on how this was tested as I'm not overly familiar with bmcweb/redfish code and am reserved over these changes possibly not being the full fixes necessary and/or cause other issues?

Yes, we can only test on Inspur's Dubai system.
Like @gtmills said, we don't need to log these errors because we have multiple chassis and these would log all the time.

> 
> Also, I think there was an upstream review going with changes to this area of bmcweb/redfish code...were these changes pushed up there as well?

Yeah, we will update upstream sync if downstream approves.
Refer: https://gerrit.openbmc-project.xyz/c/openbmc/bmcweb/+/41086

msbarth (approved):

PR #233, opened 2022-01-24
PostCode event performance improvement

sunharis:
When the system IPLs, numbers of boot progress indicators
are generated, and each code generates an event. HMC consumes
this event and displays the boot progress at HMC GUI

While sending the events, bmcweb does a costly dbus call to get
the current postcode - "GetPostCodesWithTimeStamp"

This commit removes this dbus call. This changed logic to
calculate the current postcode url while sending the event
improves the IPL time and reduces the dbus overhead

Tested by:
  1. Manual IPL tests
  2. IPL automation
  3. GET on the postcode urls for sanity check
  4. Verified that HMC displays the progress codes correctly
     during each test

Change-Id: I21e26889bf32545d43394c6011bf561324927460

### DIFF
--- a/include/event_dbus_monitor.hpp
+++ b/include/event_dbus_monitor.hpp
@@ -18,6 +18,8 @@ static std::shared_ptr<sdbusplus::bus::match::match> matchBootProgressChange;
 static std::shared_ptr<sdbusplus::bus::match::match> matchEventLogCreated;
 static std::shared_ptr<sdbusplus::bus::match::match> matchPostCodeChange;
 
+static uint64_t postCodeCounter = 0;
+
 void registerHostStateChangeSignal();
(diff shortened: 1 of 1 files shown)


raviteja-b (approved):
changes looks good 
does this change works as expected  if  bmcweb restart/bmc reboot happens during boot?

geissonator (approved):
So as Sunitha explained, the content of the event data doesn't really matter. The HMC is always going to read the postcode interface to get the actual lattest postcode. So this seems fine to me. I still think an optimization to only send a progress code event every 1 second would also be nice (so we don't have the HMC banging on our API all the time) but that could be a separate change if needed (and also one the HMC could make possibly).

gtmills (approved):

PR #236, opened 2022-01-25
Pull ntp related upstream commits

asmithakarun:
This PR pulls two upstream commits related to ntp servers (that were merged long back).

1. Remove NTPServers duplicate values and null values
When saving the set NTPServers values from webUI, NTPServer may contain
duplicate values and null values and update them to D-Bus.
Now, need to parse and verify the value of the ntpServers attribute,and
remove duplicate values and null values.

2. Fix NTPServers are hard-coded for eth0
Since bmcweb is getting and patching NTPServers only from
'/xyz/openbmc_project/network/eth0' object, and this is hard-coded, if
we use eth1, it will broken the NTP configuration and fail to route to
the correct NTPServer.
All NTPServers of xyz.openbmc_project.Network.EthernetInterface
interface should be updated.

https://gerrit.openbmc-project.xyz/c/openbmc/bmcweb/+/47344
https://gerrit.openbmc-project.xyz/c/openbmc/bmcweb/+/47342

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -369,6 +369,7 @@ srcfiles_unittest = [
   'redfish-core/ut/lock_test.cpp',
   'redfish-core/ut/configfile_test.cpp',
   'redfish-core/ut/time_utils_test.cpp',
+  'redfish-core/ut/stl_utils_test.cpp',
   'http/ut/utility_test.cpp'
 ]
 
--- a/redfish-core/include/utils/stl_utils.hpp
+++ b/redfish-core/include/utils/stl_utils.hpp
@@ -1,0 +1,39 @@
+#pragma once
+
+#include <algorithm>
+#include <string>
+
+namespace redfish
+{
+
+namespace stl_utils
+{
+
+template <typename ForwardIterator>
+ForwardIterator firstDuplicate(ForwardIterator first, ForwardIterator last)
+{
+    auto newLast = first;
+
+    for (auto current = first; current != last; ++current)
+    {
+        if (std::find(first, newLast, *current) == newLast)
+        {
+            if (newLast != current)
+            {
+                *newLast = *current;
+            }
+            ++newLast;
+        }
+    }
+
+    return newLast;
+}
+
+template <typename T>
+void removeDuplicate(T& t)
+{
+    t.erase(firstDuplicate(t.begin(), t.end()), t.end());
+}
+
+} // namespace stl_utils
+} // namespace redfish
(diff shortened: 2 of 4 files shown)


asmithakarun:
> Can you add the links to the commits? Is this tied to a defect or anything? Or why do we need these now?

I have added in the PR description as well.
https://gerrit.openbmc-project.xyz/c/openbmc/bmcweb/+/47344
https://gerrit.openbmc-project.xyz/c/openbmc/bmcweb/+/47342

So I was investigating a defect related to NTP servers (SW540119) for which I have a pr already created (https://github.com/ibm-openbmc/phosphor-networkd/pull/21/files) and a Gerrit commit (https://gerrit.openbmc-project.xyz/c/openbmc/phosphor-networkd/+/50156).
The review is still going on, and while investigating the review comments, I found these upstream commits related to this area and pulled them downstream. I think this is needed because eth0 is hardcoded in the bmcweb code and whenever a user patches an NTP server list, it will be written only to the eth0 config file and not eth1.

gtmills:
> I think this is needed because eth0 is hardcoded in the bmcweb code and whenever a user patches an NTP server list, it will be written only to the eth0 config file and not eth1.

Like it is a guess these are needed or do you know for sure?

asmithakarun:
> > I think this is needed because eth0 is hardcoded in the bmcweb code and whenever a user patches an NTP server list, it will be written only to the eth0 config file and not eth1.
> 
> Like it is a guess these are needed or do you know for sure? i.e. I am fine pulling these in if we need.. but if we don't need I would rather this far into the cycle just leave things alone

It is needed. When user does a GET on the NetworkProtocol endpoint, they will get the list of all the ntp servers from
'/xyz/openbmc_project/network/eth0' object only. Rather it should be from both eth0 & eth1.
There might be defects coming up in the near future for this! so it is better to pull this change now.

sunharis (approved):

sunharis:
> > I think this is needed because eth0 is hardcoded in the bmcweb code and whenever a user patches an NTP server list, it will be written only to the eth0 config file and not eth1.
> 
> Like it is a guess these are needed or do you know for sure? i.e. I am fine pulling these in if we need.. but if we don't need I would rather this far into the cycle just leave things alone

This is needed to apply the NTP server configuration to both eth0 and eth1 interfaces of the bmc. Current code has hardcoded eth0 - so this makes eth1 interface non-usable with NTP server settings

sunharis:
Please push commit "Remove null values in ntp patch req" upstream

asmithakarun:
> Please push commit "Remove null values in ntp patch req" upstream

Done
https://gerrit.openbmc-project.xyz/c/openbmc/bmcweb/+/50886

gtmills (approved):

PR #243, opened 2022-01-28
Unauthorised ACF certificate upload

SunnySrivastava1984:
This commit implements the change where, when OpPanel
function 74 is active, an unauthorized agent is allowed
to PATCH in the ACF.

The following command without creating a session has been
to test the functionality.
curl  -k -X PATCH -d '{"Oem":{"IBM":{"ACF":{"ACFFile":"'${service_acf_base64}'"}}}}'
https://${BMC}/redfish/v1/AccountService/Accounts/service

In case function 74 is active this should allow to patch ACF
crtificate otherwise gives following insufficient privilege message

{
  "error": {
    "@Message.ExtendedInfo": [
      {
        "MessageArgs": [],
        "MessageId": "Base.1.8.1.InsufficientPrivilege",
      }
    ],
    "code": "Base.1.8.1.InsufficientPrivilege",
  }
}


redfish-core/lib/account_service.hpp
```diff
@@ -1377,6 +1377,141 @@ inline void updateUserProperties(std::shared_ptr<bmcweb::AsyncResp> asyncResp,
...
+                    {
+                        BMCWEB_LOG_ERROR
+                            << "Failed to read ACFWindowActive property";
+                        messages::internalError(asyncResp->res);
+                        return;
+                    }
+
+                    const bool* isACFWindowActive = std::get_if<bool>(&retVal);
+
+                    if (isACFWindowActive == nullptr)
```
joseph-reynolds, line 1488: Question: Is com.ibm.PanelApp  guaranteed to be running?  Or is it running only when an OpPanel is installed?
Would it be better to treat this condition "(isACFWindowActive == nullptr)" as if "*isACFWindowActive == false"?

redfish-core/lib/account_service.hpp
```diff
@@ -1997,23 +2132,49 @@ inline void requestAccountServiceRoutes(App& app)
             std::optional<bool> locked;
             std::optional<nlohmann::json> oem;
 
+            if (!json_util::readJson(req, asyncResp->res, "UserName",
```
joseph-reynolds, line 2135: I am giving extra review attention the privilege checking required to perform this operation (above: privileges({{"ConfigureUsers"}, {"ConfigureSelf"}})).
There are two cases:
1. When this method is invoked with an unauthenticated user (because this operation isOnWhitelist()), then "req.session == nullptr" which causes code in http/routing.hpp crow::handle() to call here without performing any privilege checks.  The check for "req.session == nullptr" below handles this case.
2. When this method is invoked with an authenticated user, then req.session is set to a user, and code in the http/routing.hpp crow::handle() checks the required privileges before calling here.
So I think this covers all the the cases correctly.

joseph-reynolds (commented):
Looks good to me, pending my question about the value of ACFWindowActive always being available.

redfish-core/lib/account_service.hpp
```diff
@@ -1377,6 +1377,141 @@ inline void updateUserProperties(std::shared_ptr<bmcweb::AsyncResp> asyncResp,
         });
 }
 
+inline void uploadACF(const std::shared_ptr<bmcweb::AsyncResp> asyncResp,
```
joseph-reynolds, line 1380: clang-tidy says to make this a const reference  parameter, e.g:
const std::shared_ptr<bmcweb::AsyncResp> &asyncResp

redfish-core/lib/account_service.hpp
```diff
@@ -1377,6 +1377,141 @@ inline void updateUserProperties(std::shared_ptr<bmcweb::AsyncResp> asyncResp,
...
+                        messages::internalError(asyncResp->res);
+                        return;
+                    }
+
+                    if (*isACFWindowActive == true)
+                    {
+                        uploadACF(asyncResp, decodedAcf);
+                        return;
+                    }
+                    else
```
joseph-reynolds, line 1500: clang-tidy wants to collapse the else section (because of the "return" in the  then clause above.  That is, remove the "else {" and closing "}", and unnest the code.

redfish-core/lib/account_service.hpp
```diff
@@ -1377,6 +1377,141 @@ inline void updateUserProperties(std::shared_ptr<bmcweb::AsyncResp> asyncResp,
         });
 }
 
+inline void uploadACF(const std::shared_ptr<bmcweb::AsyncResp> asyncResp,
+                      std::vector<uint8_t> decodedAcf)
```
joseph-reynolds, line 1381: clang-tidy says to make this a const reference parameter, e.g:
const std::vector<uint8_t> &decodedAcf

joseph-reynolds: I've added comments to address all three clang-tidy errors.

redfish-core/lib/account_service.hpp
SunnySrivastava1984: No, com.ibm.PanelApp is only available when panel app is running. Hence it is not guaranteed that we will get the value always.

redfish-core/lib/account_service.hpp
SunnySrivastava1984: Done

redfish-core/lib/account_service.hpp
SunnySrivastava1984: okay

redfish-core/lib/account_service.hpp
SunnySrivastava1984: Thanks,  I have addressed all the comments.

include/authorization.hpp
```diff
@@ -240,6 +240,16 @@ static bool isOnWhitelist(std::string_view url, boost::beast::http::verb method)
         }
     }
 
+    // This patch is allowed for service user, without authorization to upload
+    // unauthenticated ACF.
+    if (boost::beast::http::verb::patch == method)
+    {
+        if (url == "/redfish/v1/AccountService/Accounts/service")
```
gtmills, line 247: I don't think we want this.. 
This will allow unauthorized user to PATCH service, including enabling/disabling service

gtmills: Why can't we handle the PATCH here via normal auth?

redfish-core/lib/account_service.hpp
```diff
@@ -1997,23 +2130,49 @@ inline void requestAccountServiceRoutes(App& app)
...
                 effectiveUserPrivileges.isSupersetOf(configureUsers);
-            if (userHasConfigureUsers)
+            if (!userHasConfigureUsers)
             {
-                if (!json_util::readJson(req, asyncResp->res, "UserName",
-                                         newUserName, "Password", password,
-                                         "RoleId", roleId, "Enabled", enabled,
-                                         "Locked", locked, "Oem", oem))
+                // Irrespective of role can patch ACF if function
+                // 74 is active from panel.
```
gtmills, line 2167: Where do we actually check that 74 is active?

redfish-core/lib/account_service.hpp
```diff
@@ -1997,23 +2130,49 @@ inline void requestAccountServiceRoutes(App& app)
...
+                if (oem && (username == "service"))
```
gtmills, line 2168: The only one without a ConfigureUsers would be an operator  ?
Why do we need this? Why can't we just handle this normally if the user has authenticated ?

redfish-core/lib/account_service.hpp
```diff
@@ -1377,6 +1377,139 @@ inline void updateUserProperties(std::shared_ptr<bmcweb::AsyncResp> asyncResp,
...
+                std::string sDecodedAcf;
+                if (!crow::utility::base64Decode(*acfFile, sDecodedAcf))
+                {
+                    BMCWEB_LOG_ERROR << "base64 decode failure ";
+                    messages::internalError(asyncResp->res);
+                    return;
+                }
+                try
+                {
+                    std::copy(sDecodedAcf.begin(), sDecodedAcf.end(),
```
gtmills, line 1464: Would like to avoid something that throws here but I think both of my suggestions still do. 


Why a std::copy here instead of just an assign? 
Like 
https://github.com/openbmc/bmcweb/blob/26f6976f3905d7b20690bbf794fba746c1929ed3/redfish-core/lib/event_service.hpp#L406 
or just the vector constructor 

std::vector<uint8_t> vec(str.begin(), str.end());

redfish-core/lib/account_service.hpp
SunnySrivastava1984: I am not sure if username other than service should be allowed to update acf certificates. As we see in the existing code (line 2073) even an authenticated user is allowed to upload acf certificate only if username is "Service".

https://github.com/ibm-openbmc/bmcweb/pull/243/files#diff-a6441dc86bbef5cbe2192c66717e8b938f9ee3a02670ff54c702338095a0d660L2073

redfish-core/lib/account_service.hpp
SunnySrivastava1984: Just re-used the already implemented and working code so did not change that.
https://github.com/ibm-openbmc/bmcweb/pull/243/files#diff-a6441dc86bbef5cbe2192c66717e8b938f9ee3a02670ff54c702338095a0d660L2103

Let me know if you still want me to change that.

gtmills: Ack.

redfish-core/lib/account_service.hpp
gtmills: Correction, this would be operatior, readonly, and noaccess user. I am still not sure why we need this but I'll default to joseph here

gtmills:
I don't understand why an operator, readonly, and no access needs to be able to PATCH here?
I understand why we want an unauthorized user... (Admin locked out)

I guess when we discussed how to do this (PATCH of the service account vs like a POST), we probably should have thought about this having to be an unauthorized client and probably something more like a POST/PATCH of its own route would make more sense because as-is we are tearing up our auth of /redfish/v1/AccountService/Accounts/<str>/ and that really worries me (seems error-prone), (although as-is the logic looks correct). 

guess I am good here. I am looking for @joseph-reynolds to approve though.

redfish-core/lib/account_service.hpp
joseph-reynolds: If I understand correctly, in the following scenario:
- The LCD screen in not installed, so the Panel App is not running.
- Someone attempts an unauthenticcated ACF upload, and gets to this function with isACFWindowActive == nullptr
- That user gets internalError.  <-- That HTTP status seems wrong.

I think in this scenario we must handle (isACFWindowActive == nullptr) the same as (*isACFWindowActive == false).  This results in the user getting insufficientPrivilege

SunnySrivastava1984: I am sorry, I mentioned wrongly before, Panel app will be running irrespective of panel being connected to the system.

joseph-reynolds: In that case, we should always get the value of isACFWindowActive, and will only get isACFWindowActive=nullptr if the BMC is not working correctly.  The code looks good as it is.

redfish-core/lib/account_service.hpp
joseph-reynolds: I met with Sunny to discuss this.  We followed the code flows for the 3 conditions:
> (((user is not authenticated) && (isACFWindowActive==true)) ||
> ((!userHasConfigureUsers) && (isACFWindowActive==true)) ||
> userHasConfigureUsers)

The first two conditions were easy to validate.  For the third condition, we walked through the logic in 2164-2188 and through several PATCH requests:
- Line 2168: PATCHing {"Oem":{"IBM":{"ACF": ... }}}.
These two clauses are unchanged:
- Line 2177: PATCHing someone else's account (username != req.session->username).
- Line 2185: PATCHing just the RoleId, or PATCHing both "Password" and "RoleId" in the same request.

Gunnar, to address your question directly: I'm not sure what you mean by "normally".  The new code follows the logic from old line 2005 else clause (lines 2015-2029).  The old code allowed a user who does not have ConfigureUsers to change their own account password.  The new code allows that, plus it allows uploading an ACF when isACFWindowActive==true.  That looks right to me.

joseph-reynolds:
> I don't understand why an operator, readonly, and no access needs to be able to PATCH here?

Users need to be able to PATCH their own Password (per the Redfish Password property override).
I agree the authority checking performed in this code is complicated...so making changes here is error prone.

Sunny and I performed a low-level design walkthrough and a code walkthrough and cam to the conclusion that the code is correct.  I don't think we're going to be able to do better here for FW1020 GA.

gtmills (approved):
Defaulting to Joseph here

joseph-reynolds:
On way to simulate OpPanel function 74 is: `busctl set-property "com.ibm.PanelApp" "/com/ibm/panel_app" "com.ibm.panel" "ACFWindowActive" b true`

PR #248, opened 2022-02-02
Disallow NoAccess user to login to GUI

sunharis:
An authenticated user with the NoAccess role can login GUI

This commit is to block the login for a user with priv-noaccess

Testedby:

1. Create a user with priv-noaccess. Verify the login attempt fails
with accessDenied error
2. Verify NoAccess user can not login to GUI
3. Verified the other role users can login


joseph-reynolds (approved):
Looks correct to me.

sunharis:
We will pull the upstream changes after that is concluded. Changes in this PR is not needed.

PR #249, opened 2022-02-02
Move socket files to /tmp directory

raviteja-b:
Currently we store dump offload unix socket files at /var/lib.

There is no need to persist these file and persistency causing 
issues in few cases, throwing socket already in use errors after reset.

so create socket file at /tmp/DumpOffloadSockets

increase socket connect timeout and retry as well

Tested By:
BMC dump offload
System dump offload

### DIFF
--- a/include/dump_offload.hpp
+++ b/include/dump_offload.hpp
@@ -9,13 +9,14 @@
 #include <boost/beast/core/flat_static_buffer.hpp>
 #include <boost/beast/http.hpp>
 #include <http_stream.hpp>
+#include <ibm/utils.hpp>
 
 namespace crow
 {
 namespace obmc_dump
 {
 
-std::string unixSocketPathDir = "/var/lib/phosphor-debug-collector/";
+std::string unixSocketPathDir = "/tmp/DumpOffloadSockets/";
 
 inline void handleDumpOffloadUrl(const crow::Request& req, crow::Response& res,
                                  const std::string& entryId,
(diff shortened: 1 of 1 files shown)


gtmills:
Why increase socket connect timeout and retry as well ?

raviteja-b:
> Why increase socket connect timeout and retry as well ?

@gtmills I am working changing dump manager code to use async thread for dump offload, which is taking few more milli seconds to setup.
so need to increase timeout and in sometimes D-bus response are also slow.

include/dump_offload.hpp
```diff
@@ -283,6 +283,12 @@ inline void requestRoutes(App& app)
             std::string dumpType = "bmc";
             boost::asio::io_context* ioCon = conn.getIoContext();
 
+            std::filesystem::path dir(unixSocketPathDir);
+            if (!std::filesystem::exists(dir))
+            {
+                std::filesystem::create_directories(unixSocketPathDir);
```
sunharis, line 289: Please see if you can use this util method to create directories and set permissions to it https://github.com/ibm-openbmc/bmcweb/blob/1020/include/ibm/utils.hpp#L13

sunharis (approved):

include/dump_offload.hpp
```diff
@@ -289,6 +290,14 @@ inline void requestRoutes(App& app)
             handlers[&conn] = std::make_shared<Handler>(
                 *ioCon, dumpId, dumpType, unixSocketPath);
             handlers[&conn]->connection = &conn;
+
+            if (!crow::ibm_utils::createDirectory(unixSocketPathDir))
+            {
+                handlers[&conn]->connection->sendStreamErrorStatus(
+                    boost::beast::http::status::not_found);
```
gtmills, line 297: If we aren't able to create the directories, we should a not found?

raviteja-b: yeah. otherwise opening socket file will fail @gtmills

gtmills (approved):

PR #250, opened 2022-02-02
http_connection: Bump connection timeout to 3min

anoo1:
With the current 1min timeout, there are cases when the image upload
times out due to slow network connections.

Signed-off-by: Adriana Kobylak <anoo@us.ibm.com>

### DIFF
--- a/http/http_connection.hpp
+++ b/http/http_connection.hpp
@@ -47,10 +47,10 @@ constexpr uint64_t loggedOutPostBodyLimit = 4096;
 
 constexpr uint32_t httpHeaderLimit = 8192;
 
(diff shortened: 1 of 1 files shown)


anoo1:
This file has been refactored upstream with a new timer pool: https://gerrit.openbmc-project.xyz/c/openbmc/bmcweb/+/49318

The slow timeout is already selected based on whether or not the user is logged in.
If we still timeouts after bringing that change back, per Ed we might think about defining the slow timer as "time since we saw the last chunk" instead of "time to upload the whole payload".

gtmills (approved):

PR #251, opened 2022-02-03
Fix download log entry details and says bad request

gtmills:
Upstream is https://gerrit.openbmc-project.xyz/c/openbmc/bmcweb/+/50879

Defect is https://w3.rchland.ibm.com/projects/bestquest/?defect=SW542274 

Remove q-factor weighting on Accept Header

bmcweb does not do anything with the q-factor weighting (;q=) so just
remove it from the encoding.

This is needed because routes like
"/redfish/v1/Systems/system/LogServices/EventLog/Entries/<str>/attachment"
have a check for isOctetAccepted. Even though `*/* `is in the Accept
Header isOctetAccepted still fails due to the q-factor weighting.

On the system I tested, on firefox, Accept looks like:
`Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,*/*;q=0.8`

On chrome it looks like: 
`Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3;q=0.9`

The GUI reported being unable to download a AdditionalDataURI (e.g.
...attachment/)

See: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept
Tested:
/redfish/v1/Systems/system/LogServices/PostCodes/Entries/<str>/attachment/
and .../EventLog/Entries/<str>/attachment now return correctly.

Change-Id: I969f5f2c32c4acccd4d80615f17c44d0c8fabd0d

### DIFF
--- a/include/http_utility.hpp
+++ b/include/http_utility.hpp
@@ -34,8 +34,14 @@ inline bool requestPrefersHtml(std::string_view header)
 
 inline bool isOctetAccepted(std::string_view header)
 {
-    for (const std::string& encoding : parseAccept(header))
+    for (std::string& encoding : parseAccept(header))
     {
+        // ignore any q-factor weighting (;q=)
+        std::size_t separator = encoding.find(";q=");
+        if (separator != std::string::npos)
+        {
+            encoding = encoding.substr(0, separator);
+        }
         if (encoding == "*/*" || encoding == "application/octet-stream")
         {
             return true;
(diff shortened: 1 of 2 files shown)


sunharis (approved):
Seems good. Since this code ignores the incoming q value, this will not impact HMC as well.  
We can also do like below, but i am ok either way. 
if (encoding starts with "*/*") 
   then allow 
   
We are making it mandatory for the clients to send */* always with this code. The method name isOctetAccepted does not look suitable to me. But this is upstream code - so we can discuss this separately from this PR

PR #254, opened 2022-02-08
Consume availability interface for cores

lxwinspur:
In the current state, the logic is bmcweb for displaying cores is:
It checks the Present property on core dbus object to update
the [Status][State] as Enabled/Absent.

In addition to Present property, we should also use the Available
property , if the Available is false, then set the [Status][State]
should be set as "UnavailableOffline".

Refer: https://github.com/ibm-openbmc/dev/issues/3481

### DIFF
--- a/redfish-core/lib/processor.hpp
+++ b/redfish-core/lib/processor.hpp
@@ -984,6 +984,7 @@ inline void
 
                 bool present = true;
                 bool functional = true;
+                bool available = true;
 
                 for (const auto& [interface, properties] : interfaces)
                 {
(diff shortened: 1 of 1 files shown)


redfish-core/lib/processor.hpp
```diff
@@ -1052,15 +1071,24 @@ inline void
...
                 }
                 else
                 {
-                    if (!functional)
+                    if (present == false)
                     {
-                        aResp->res.jsonValue["Status"]["Health"] = "Critical";
+                        aResp->res.jsonValue["Status"]["State"] = "Absent";
+                    }
+                    else
```
gtmills, line 1085: Why not  
if
else if 
else if

gtmills:
Looks fine, @lxwinspur can you push upstream and use 
if
else if
else if

instead?

redfish-core/lib/processor.hpp
lxwinspur: Ack

redfish-core/lib/processor.hpp
```diff
@@ -1052,15 +1071,24 @@ inline void
...
+                    {
+                        if (!functional)
```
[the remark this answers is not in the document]
lxwinspur: Done

lxwinspur:
> Looks fine, @lxwinspur can you push upstream and use if else if else if
> 
> instead?

I have updated this issue.

BTW, Considering that this patch has not been reviewed, do we still need upstream?
https://gerrit.openbmc-project.xyz/c/openbmc/bmcweb/+/38570

gtmills (approved):

gtmills:
jenkins run tests please

PR #255, opened 2022-02-08
Send events for all the VMI IP property changes

asmithakarun:
Earlier, only one event will be sent to the client
indicating vmi ip has changed and this is based on the "Enabled"
property of that hypervisor ethernet ip object (if it is true, an
event will be sent, saying vmi ip has been configured). Once the event
is received by the managemnet console, in this case, it immediately does
a GET operation on the event's origin of condition (hypervisor
ethernet endpoint).

Now the problem is:
When dhcp is enabled on an ethernet interface, pldm sets the enabled property
of that ip object to true and then, there is a small
time window before the ip properties are reflected on the hypervisor
ethernet ip dbus object. Within this time lag, the management console
receives an event and does a GET on the interface, and it finds all the
ip properties to be zero (ip address, gateway and subnet mask: 0.0.0.0)
and the same will be consumed by the hmc.

To overcome this issue:
Events will be sent on any property change (ip address/gateway/prefix length/origin)
of the hypervisor ethernet ip dbus object (** only when the enabled property on that
interface is true **), thus the management console will be notified of the change and
when it does a GET on the ethernet endpoint, the correct data will be available in the
redfish response, avoiding the above scenario (tested)

Tested By:

* Set dhcp to true:
PATCH -d '{"DHCPv4": {"DHCPEnabled" :true}}' https://${bmc}/redfish/v1/Systems/hypervisor/EthernetInterfaces/eth1

* 3 events were sent to the management console indicating the change in the ip properties
(ip address, gateway, prefix length)

* Checked if the hmc has consumed the correct data from the redfish response
lssysconn -r vmi -m 9105-22B*139F2B0
resource_type=vmi,type_model_serial_num=9105-22B*139F2B0,interface=eth0,ipv4addr=0.0.0.0,gateway=0.0.0.0,net_mask=0.0.0.0,ipv4addr_orig=dhcp
resource_type=vmi,type_model_serial_num=9105-22B*139F2B0,interface=eth1,ipv4addr=192.168.1.30,gateway=192.168.1.1,net_mask=255.255.255.0,ipv4addr_orig=dhcp

* Set dhcp to false:
PATCH -d '{"DHCPv4": {"DHCPEnabled" :false}}' https://${bmc}/redfish/v1/Systems/hypervisor/EthernetInterfaces/eth1

* Events sent to the management console

* Checked if the hmc has consumed the correct data from the redfish response
lssysconn -r vmi -m 9105-22B*139F2B0
resource_type=vmi,type_model_serial_num=9105-22B*139F2B0,interface=eth0,ipv4addr=0.0.0.0,gateway=0.0.0.0,net_mask=0.0.0.0,ipv4addr_orig=dhcp
resource_type=vmi,type_model_serial_num=9105-22B*139F2B0,interface=eth1,ipv4addr=0.0.0.0,gateway=,net_mask=,ipv4addr_orig=

Change-Id: Iee42879b23310b872f26845aae9801e311705d11

### DIFF
--- a/include/event_dbus_monitor.hpp
+++ b/include/event_dbus_monitor.hpp
@@ -10,7 +10,9 @@ namespace dbus_monitor
 
 static std::shared_ptr<sdbusplus::bus::match::match> matchHostStateChange;
 static std::shared_ptr<sdbusplus::bus::match::match> matchBMCStateChange;
-static std::shared_ptr<sdbusplus::bus::match::match> matchVMIIPStateChange;
+static std::shared_ptr<sdbusplus::bus::match::match>
+    matchVMIIPEnabledPropChange;
+static std::shared_ptr<sdbusplus::bus::match::match> matchVMIIPChange;
 static std::shared_ptr<sdbusplus::bus::match::match> matchDumpCreatedSignal;
 static std::shared_ptr<sdbusplus::bus::match::match> matchDumpDeletedSignal;
 static std::shared_ptr<sdbusplus::bus::match::match> matchBIOSAttrUpdate;
@@ -20,8 +22,12 @@ static std::shared_ptr<sdbusplus::bus::match::match> matchPostCodeChange;
 
 static uint64_t postCodeCounter = 0;
 
+bool isIPEnabledOnIntfEth0;
+bool isIPEnabledOnIntfEth1;
+
 void registerHostStateChangeSignal();
 void registerBMCStateChangeSignal();
+void registerVMIIPEnabledPropChangeSignal();
 void registerVMIIPChangeSignal();
 void registerDumpCreatedSignal();
 void registerDumpDeletedSignal();
@@ -30,13 +36,47 @@ void registerBootProgressChangeSignal();
 void registerEventLogCreatedSignal();
 void registerPostCodeChangeSignal();
 
-inline void VMIIPChange(sdbusplus::message::message& msg)
+inline void setVMIIPEnabledValue(std::string intf)
 {
-    BMCWEB_LOG_DEBUG << "BMC Hypervisor IP change match fired";
+    std::string objPath =
+        "/xyz/openbmc_project/network/hypervisor/" + intf + "/ipv4/addr0";
+    crow::connections::systemBus->async_method_call(
+        [intf](const boost::system::error_code ec,
+               const std::variant<bool>& enabled) {
+            if (ec)
+            {
+                BMCWEB_LOG_ERROR << "DBUS response error " << ec;
+                return;
+            }
 
+            const bool* isEnabled = std::get_if<bool>(&enabled);
+            BMCWEB_LOG_DEBUG << "isEnabled: " << *isEnabled;
+            if (intf == "eth0")
+            {
+                isIPEnabledOnIntfEth0 = *isEnabled;
+            }
+            else if (intf == "eth1")
+            {
+                isIPEnabledOnIntfEth1 = *isEnabled;
+            }
+        },
+        "xyz.openbmc_project.Network.Hypervisor", objPath,
+        "org.freedesktop.DBus.Properties", "Get",
+        "xyz.openbmc_project.Object.Enable", "Enabled");
+}
+
+inline void setVMIIPEnabledValue()
+{
+    setVMIIPEnabledValue("eth0");
+    setVMIIPEnabledValue("eth1");
+}
+
+inline void VMIIPEnabledPropChange(sdbusplus::message::message& msg)
+{
     if (msg.is_method_error())
     {
-        BMCWEB_LOG_ERROR << "BMC Hypervisor IP property changed Signal error";
+        BMCWEB_LOG_ERROR
+            << "BMC Hypervisor IP Enabled property changed Signal error";
         return;
     }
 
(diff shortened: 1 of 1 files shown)


include/event_dbus_monitor.hpp
```diff
@@ -64,32 +103,129 @@ inline void VMIIPChange(sdbusplus::message::message& msg)
...
+    if (find != values.end())
+    {
+        const int64_t propValue =
+            static_cast<int64_t>(std::get<uint8_t>(find->second));
+        if (enabledValue)
+        {
+            BMCWEB_LOG_DEBUG
+                << "Pushing the VMI IP property change event for PrefixLength: "
+                << propValue << " with origin : " << origin;
+            sendEventOnEthIntf(origin);
```
sunharis, line 204: We need not send 3 events for 3 values. Send single event if any of the values are changed.

include/event_dbus_monitor.hpp
```diff
@@ -333,6 +479,11 @@ void registerStateChangeSignal()
 {
     registerHostStateChangeSignal();
     registerBMCStateChangeSignal();
+
+    // Get vmi ip enabled property and set it to the local variable
+    setVMIIPEnabledValue();
+
```
sunharis, line 485: nit: remove blank lines

sunharis (changes requested):

include/event_dbus_monitor.hpp
asmithakarun: Done

include/event_dbus_monitor.hpp
asmithakarun: So we listen on the property change signal on the IP interface, on each property change, we hit this method and send an event. This is only in case of dhcp configuration.

sunharis (approved):
Looks good to me.

gtmills:
> there is a small time window before the ip properties are reflected on the hypervisor ethernet ip dbus object.

Why do we have this window? 
So the enable event is sent but the dbus object still has (ip address, gateway and subnet mask: 0.0.0.0)? How?

asmithakarun:
> > there is a small time window before the ip properties are reflected on the hypervisor ethernet ip dbus object.
> 
> Why do we have this window? So the enable event is sent but the dbus object still has (ip address, gateway and subnet mask: 0.0.0.0)? How?

no, what happens is, when dhcp is enabled to true, pldm sets the enabled property to true, but all the ip properties will be 0 only at this point in time. Then vmi sends down the dhcp provided ip details, which will be updated in the bios table on bmc and then, the same will be reflected on our dbus object.
So there is a small time window between pldm setting the enabled property and dhcp ip details being sent down from vmi.

include/event_dbus_monitor.hpp
```diff
@@ -20,23 +23,87 @@ static std::shared_ptr<sdbusplus::bus::match::match> matchPostCodeChange;
...
+        [intf](const boost::system::error_code ec,
+               const std::variant<std::string>& methodOnIntf) {
+            if (ec)
+            {
+                BMCWEB_LOG_ERROR << "DBUS response error " << ec;
+                return;
+            }
+
+            const std::string* method = std::get_if<std::string>(&methodOnIntf);
+            if (intf == "eth0")
```
gtmills, line 59: Don't you need to check for null ?

include/event_dbus_monitor.hpp
```diff
@@ -64,29 +130,221 @@ inline void VMIIPChange(sdbusplus::message::message& msg)
         }
 
         const bool* propValue = std::get_if<bool>(&(find->second));
-        if (propValue != nullptr)
```
gtmills, line 67: Check for nullprt? 
if (propValue == nullptr)
{
error and return 

}

include/event_dbus_monitor.hpp
```diff
@@ -64,29 +130,221 @@ inline void VMIIPChange(sdbusplus::message::message& msg)
...
+        // Change if enabled property by pldm will
+        // take care of sending the event for static
+        // configuration
+        return;
+    }
+
+    auto find = values.find("Address");
+    if (find != values.end())
+    {
+        const std::string propValue = std::get<std::string>(find->second);
```
gtmills, line 226: This throws doesn't it ? 
std::get_if and then check null ?

include/event_dbus_monitor.hpp
```diff
@@ -64,29 +130,221 @@ inline void VMIIPChange(sdbusplus::message::message& msg)
...
+                << propValue << " with origin : " << origin;
+            sendEventOnEthIntf(origin);
+            return;
+        }
+    }
+
+    find = values.find("Gateway");
+    if (find != values.end())
+    {
+        const std::string propValue = std::get<std::string>(find->second);
```
gtmills, line 240: Same here?

gtmills:
We have been discussing here: https://ibm-systems-power.slack.com/archives/G3WAVQYTH/p1645124367926439?thread_ts=1645120872.264649&cid=G3WAVQYTH
This seems really complicated for just sending events on VMI changes

include/event_dbus_monitor.hpp
asmithakarun: done

include/event_dbus_monitor.hpp
asmithakarun: done

include/event_dbus_monitor.hpp
asmithakarun: done

sunharis:
I would also want to get some more optimization on this code. Lets hold until then

asmithakarun:
> I would also want to get some more optimization on this code. Lets hold until then

Done

sunharis (approved):
Changes looks good now.

raviteja-b (approved):

PR #256, opened 2022-02-08
Update Assembly Schema To Populate Status::State

SunnySrivastava1984:
Status::State reflects the presence status of FRU being
populated in the schema.
"Enabled" state that FRU is present.
"Absent" implies that device is not present.

Validator has been executed and no new error found.

### DIFF
--- a/redfish-core/lib/assembly.hpp
+++ b/redfish-core/lib/assembly.hpp
@@ -49,7 +49,6 @@ inline void
 
         // Set the default Status
         tempyArray.at(assemblyIndex)["Status"]["Health"] = "OK";
-        tempyArray.at(assemblyIndex)["Status"]["State"] = "Enabled";
 
         crow::connections::systemBus->async_method_call(
             [aResp, assemblyIndex, assembly](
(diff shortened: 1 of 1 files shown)


redfish-core/lib/assembly.hpp
```diff
@@ -49,7 +49,6 @@ inline void
 
         // Set the default Status
         tempyArray.at(assemblyIndex)["Status"]["Health"] = "OK";
-        tempyArray.at(assemblyIndex)["Status"]["State"] = "Enabled";
```
gtmills, line 52: You could still have this default state then you could get rid of the else on lines 276 -280. I feel that fits more with the existing Status Health and what we do upstream. I am okay here though

gtmills (approved):

gtmills:
@SunnySrivastava1984 Can this go upstream?

redfish-core/lib/assembly.hpp
SunnySrivastava1984: I will revert this.

SunnySrivastava1984:
> @SunnySrivastava1984 Can this go upstream?

Upstream the base commit for assembly is still under review and there appears to be some deviation between code upstream and downstream.

PR #258, opened 2022-02-15
Not return InternalError when there is no usbcodeupdate service

ChicagoDuan:
If usb code update service is not available, to not return the data
associated with it, instead of returning an InternalError response for
it.

### DIFF
--- a/redfish-core/lib/oem/ibm/usb_code_update.hpp
+++ b/redfish-core/lib/oem/ibm/usb_code_update.hpp
@@ -63,8 +63,7 @@ inline void getServiceName(const std::shared_ptr<bmcweb::AsyncResp>& aResp,
                 return;
             }
 
-            BMCWEB_LOG_ERROR << "Can't find usb code update service!";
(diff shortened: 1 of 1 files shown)


redfish-core/lib/oem/ibm/usb_code_update.hpp
```diff
@@ -63,8 +63,6 @@ inline void getServiceName(const std::shared_ptr<bmcweb::AsyncResp>& aResp,
                 return;
             }
 
-            BMCWEB_LOG_ERROR << "Can't find usb code update service!";
```
gtmills, line 66: Do we still want the debug?

gtmills (approved):

redfish-core/lib/oem/ibm/usb_code_update.hpp
ChicagoDuan: Done. Yes, I think we want the debug. I have added it.

PR #262, opened 2022-02-18
Check PowerSupply availability prop for State/Hlth

spinler:
This commit makes use of the Available property on the power supply
D-Bus object when determining the power supply state and health in the
PowerSupply response.  The power supply monitor code from phosphor-power
will set the property to false if it can determine that the power supply
isn't able to provide power due to certain faults.

If the PS isn't available, then the Health property will be set to
Critical, and the State will be set to UnavailableOffline (assuming it
shouldn't be Absent instead).

This is necessary because on IBM systems the Functional property
determines the fault LED state as well as the health value, and in
certain cases the fault LED needs to stay off even though the PS health
is not OK.  A specific example of this is when the PS cord is unplugged:
no fault LED is desired but it is desired for the Redfish output to show
that there is an issue with the PS.

If the Available property isn't present on D-Bus, it acts as if it had a
value of true and behaves the same as it does today.

Tested:
Available = false:
    "Health": "Critical",
    "State": "UnavailableOffline"

    "Health": "Critical",
    "State": "Enabled"

PS missing:
    "Health": "Critical",
    "State": "Absent"

Everything OK:
    "Health": "OK",
    "State": "Enabled"

    "Health": "OK",
    "State": "Enabled"

Upstream: https://gerrit.openbmc-project.xyz/c/openbmc/bmcweb/+/51336
Defect: SW542338

### DIFF
--- a/redfish-core/lib/power_supply.hpp
+++ b/redfish-core/lib/power_supply.hpp
@@ -127,15 +127,15 @@ inline void
 inline void
     getPowerSupplyState(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
                         const std::string& connectionName,
-                        const std::string& path)
+                        const std::string& path, bool available)
 {
 
     // Set the default state to Absent
     asyncResp->res.jsonValue["Status"]["State"] = "Enabled";
 
     crow::connections::systemBus->async_method_call(
-        [asyncResp](const boost::system::error_code ec,
-                    const std::variant<bool> state) {
+        [asyncResp, available](const boost::system::error_code ec,
+                               const std::variant<bool> state) {
             if (ec)
             {
                 if (ec.value() == EBADR)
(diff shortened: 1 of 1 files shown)


redfish-core/lib/power_supply.hpp
```diff
@@ -199,6 +204,43 @@ inline void
...
+            bool available = true;
+            if (ec)
+            {
+                if (ec.value() != EBADR)
+                {
+                    BMCWEB_LOG_ERROR << "Can't get PowerSupply availability!";
+                    messages::internalError(asyncResp->res);
+                    return;
+                }
+            }
```
gtmills, line 224: I guess you keep going if it isn't found.. I think that makes sense

gtmills (approved):
Seems reasonable to me

PR #263, opened 2022-02-21
Dump Offload Error message improvement

sunharis:
When Dump manager hits a org.freedesktop.DBus.Error.NoReply error
from the dbus, bmcweb returns Internal Server Error response to
the clients for the create dump request

This commit implements translating this dbus error to a meaningful
ServiceTemporarilyUnavailable error with 60 sec retry timeout

Tested by:
  Create dump call is tested for sanity

Change-Id: I2d585b685150089874a893e8d85d6c4119ca807e


raviteja-b (approved):

raviteja-b:
please check CI failure

geissonator (approved):
My only concern here is that if dump manager is hung, we will just respond like this forever. But still seems better then an InternalError.

PR #264, opened 2022-02-22
DumpOffload:Use rand() for unique unix socket

raviteja-b:
This commit uses rand() function to create unix socket file path.

Added socket cleanup code to clear socket file in error cases


include/dump_offload.hpp
```diff
@@ -351,8 +374,8 @@ inline void requestRoutes(App& app)
```
sunharis, line 374: When you hit "Unable to extract the dump id" at line 350, please return not_found error

include/dump_offload.hpp
```diff
@@ -351,8 +374,8 @@ inline void requestRoutes(App& app)
 
             boost::asio::io_context* ioCon = conn.getIoContext();
 
-            std::string unixSocketPath =
-                unixSocketPathDir + dumpType + "_dump_" + dumpId;
+            std::string unixSocketPath = unixSocketPathDir + dumpType +
```
sunharis, line 377: Please add a comment to say that you are using rand() to create different unique sockets for each dump offload request

include/dump_offload.hpp
```diff
@@ -152,11 +154,29 @@ class Handler : public std::enable_shared_from_this<Handler>
...
         });
     }
 
+    void cleanupSocketFiles()
+    {
+        std::error_code ec;
+        bool fileExists = std::filesystem::exists(unixSocketPath, ec);
+        if (ec)
+        {
+            this->connection->sendStreamErrorStatus(
```
sunharis, line 169: Add a error trace saying that the cleanup of this socket failed. Also print the unixSocketPath here

include/dump_offload.hpp
raviteja-b: done

include/dump_offload.hpp
raviteja-b: done

include/dump_offload.hpp
raviteja-b: done

raviteja-b:
jenkins run tests please

sunharis (approved):

sunharis (approved):

PR #267, opened 2022-02-28
DumpOffload: Return appropriate http error codes 

raviteja-b:
Return Service_Unavailable http code when dump manager/pldmd is not ready to offload dump

Currently bmcweb shares unix socket file path to dump manager/pldmd to connect
so that data can be transferred through socket.
In error case, pldmd/Dump manager may not setup socket and bmcweb tries to connect 3 times 
and returns 500 error code.

This commit modifies this error code to  Service_Unavailable

Tested By:
Ran offload dump in failure case where dump manager failure to start new process for dump offload.

### DIFF
--- a/http/http_stream.hpp
+++ b/http/http_stream.hpp
@@ -17,7 +17,7 @@ namespace streaming_response
 struct Connection : std::enable_shared_from_this<Connection>
 {
   public:
-    explicit Connection(const crow::Request& reqIn) : req(reqIn.req)
+    explicit Connection(const crow::Request& reqIn) : req(reqIn)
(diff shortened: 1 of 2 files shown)


sunharis (approved):

http/http_stream.hpp
```diff
@@ -99,6 +109,7 @@ class ConnectionImpl : public Connection
                 }
             });
     }
+
```
gtmills, line 112: Please try to avoid white space changes for next time

raviteja-b: sure

PR #269, opened 2022-03-02
Push-Style Eventing: FFDC improvement

sunharis:
This commit fixes the event id restoration and also adds
FFDC improvements to print the subscriberId, destination details
and eventId on failure paths

Tested by:
 1. Simulating bad subscription and triggering save-area events
 2. Connect HMC and see all good path eventing is working fine
 3. Tested event persistency when the subscriber is removed after
    send/recv step failures

Change-Id: Ifc401072a65320327d82c313fadf20b46737b850

### DIFF
--- a/http/http_client.hpp
+++ b/http/http_client.hpp
@@ -182,14 +182,28 @@ class HttpClient : public std::enable_shared_from_this<HttpClient>
                                const std::size_t& bytesTransferred) {
             if (ec)
             {
-                std::string_view eventStr = self->req.body();
-                nlohmann::json event;
-                auto eventData = event.parse(eventStr);
-
+                auto eventData =
+                    nlohmann::json::parse(self->req.body(), nullptr, false);
+                uint64_t eventId;
+                if (!eventData.is_discarded())
+                {
+                    for (const auto& item : eventData.items())
+                    {
+                        if (item.key() == "Id")
+                        {
+                            const uint64_t* id =
+                                item.value().get_ptr<const uint64_t*>();
+                            if (id != nullptr)
+                            {
+                                eventId = *id;
+                            }
+                        }
+                    }
+                }
                 BMCWEB_LOG_ERROR << "sendMessage() failed: " << ec.message()
                                  << " Destination: " << self->host << ":"
                                  << self->port << " to subId: " << self->subId
-                                 << " Event: " << eventData["Id"];
+                                 << " Event: " << eventId;
                 self->state = ConnState::sendFailed;
                 self->handleConnState();
                 return;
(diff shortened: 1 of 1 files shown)


raviteja-b (approved):

http/http_client.hpp
```diff
@@ -206,75 +217,87 @@ class HttpClient : public std::enable_shared_from_this<HttpClient>
...
-                                 << respCode << " from subId: " << self->subId;
-                self->state = ConnState::recvFailed;
-                self->handleConnState();
-                return;
-            }
+                // 2XX response is considered to be successful
+                if ((respCode < 200) || (respCode >= 300))
+                {
+                    // The listener failed to receive the Sent-Event
+                    BMCWEB_LOG_ERROR
```
gtmills, line 264: I guess this was here before, but why log this?

http/http_client.hpp
```diff
@@ -363,37 +387,56 @@ class HttpClient : public std::enable_shared_from_this<HttpClient>
             BMCWEB_LOG_ERROR
                 << "Maximum number of retries reached for Subscriber:" << subId;
 
-            // Clear queue.
-            while (!requestDataQueue.empty())
```
gtmills, line 367: Why duplicate this in 3 places now?

http/http_client.hpp
```diff
@@ -206,75 +217,87 @@ class HttpClient : public std::enable_shared_from_this<HttpClient>
...
-                self->handleConnState();
-                return;
-            }
+        auto respHandler =
+            [self(shared_from_this())](const boost::beast::error_code ec,
+                                       const std::size_t& bytesTransferred) {
+                std::string_view eventStr = self->req.body();
+                nlohmann::json event;
+                auto eventData = event.parse(eventStr);
+                if (ec && ec != boost::asio::ssl::error::stream_truncated)
```
gtmills, line 226: We just keep going when it is this ec, boost::asio::ssl::error::stream_truncated ?

http/http_client.hpp
```diff
@@ -172,23 +177,29 @@ class HttpClient : public std::enable_shared_from_this<HttpClient>
...
-                    return;
-                }
+        auto respHandler = [self(shared_from_this())](
+                               const boost::beast::error_code ec,
+                               const std::size_t& bytesTransferred) {
+            if (ec)
+            {
+                std::string_view eventStr = self->req.body();
+                nlohmann::json event;
+                auto eventData = event.parse(eventStr);
```
gtmills, line 187: If you are going to do eventData"Id" below shouldn't you check for .is_discarded() https://github.com/ibm-openbmc/bmcweb/blob/8044381620b228f99b707ea2b444c7675e714f26/include/login_routes.hpp#L42 ?

sunharis: Its not a user input. This is a json which is built within the BMC. A event record data. So it can not go bad.

http/http_client.hpp
sunharis: Yes. This was there earlier as well. I just enhanced the error trace

http/http_client.hpp
sunharis: This is to track if we receive any 4xx or 5xx errors in the response header

http/http_client.hpp
sunharis: We are clearing entire queue in case of Terminate/Suspend. For RetryForever, we are not cleaning the complete queue with a while loop.

http/http_client.hpp
```diff
@@ -363,37 +387,56 @@ class HttpClient : public std::enable_shared_from_this<HttpClient>
...
+                                "reset retrycount for subId: "
                              << subId;
             // Reset the retrycount to zero so that client can try connecting
             // again if needed
             retryCount = 0;
+
+            // All retries failed, Lets remove data from queue
+            if (!requestDataQueue.empty())
+            {
+                requestDataQueue.pop_front();
```
sunharis, line 433: We are just cleaning the current event which failed with all retries. We will try sending the next event in the queue to the subscriber

http/http_client.hpp
```diff
@@ -206,75 +217,87 @@ class HttpClient : public std::enable_shared_from_this<HttpClient>
...
+                std::string_view eventStr = self->req.body();
+                nlohmann::json event;
+                auto eventData = event.parse(eventStr);
+                if (ec && ec != boost::asio::ssl::error::stream_truncated)
+                {
+                    BMCWEB_LOG_ERROR << "recvMessage() failed: " << ec.message()
+                                     << " from subId: " << self->subId
+                                     << " Destination: " << self->host << ":"
+                                     << self->port
+                                     << " for Event: " << eventData["Id"];
```
gtmills, line 232: What if parse failed above? and you try to get eventData["Id"] ?

sunharis: Done. Addressed this at all the places

http/http_client.hpp
```diff
@@ -529,10 +597,30 @@ class HttpClient : public std::enable_shared_from_this<HttpClient>
     }
     void sendData(const std::string& data)
     {
+        auto eventData = nlohmann::json::parse(data);
```
gtmills, line 600: I think you need to turn off the exceptions 
parse(data, nullptr, false)

sunharis: Done

gtmills (approved):

PR #272, opened 2022-03-03
DumpOffload: Clean up Unix Socket on abrupt http connection close

raviteja-b:
Generally system/LPA dumps may take few mins to offload dump.
In slow network scenarios it may take few hours and client can choose to stop 
dump offload abruptly from client side.

This commit does unix socket cleanup when client abruptly closes
http connection while offloading dump.

This is downstream only since the dump offload is downstream only, 
we plan to upstream dump offload one day <or insert the real plan>
There isn't a defect for this but issue was seen without this change 

Tested by:
Interrupt curl command while offloading system dump 
Interrupt curl command while offloading LPA dump

### DIFF
--- a/include/dump_offload.hpp
+++ b/include/dump_offload.hpp
@@ -177,6 +177,7 @@ class Handler : public std::enable_shared_from_this<Handler>
         }
         if (fileExists)
         {
+            unixSocket.close();
             std::remove(unixSocketPath.c_str());
         }
         return;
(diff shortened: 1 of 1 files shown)


include/dump_offload.hpp
```diff
@@ -165,6 +165,7 @@ class Handler : public std::enable_shared_from_this<Handler>
 
     void cleanupSocketFiles()
     {
+        unixSocket.close();
```
sunharis, line 168: This should be done at line 181 ? when you successfully remove the unixSocketPath

raviteja-b: need to close socket first then remove socket file

raviteja-b: ok got it, moved unixSocket.close() inside file exists checks

PR #277, opened 2022-03-07
Enhance DUMP debug traces

raviteja-b:
These traces helps while debugging DUMP issues

1. In dump offload case, we have multiple defects where two HMCs connected
so we wanted to have trace which include client origin ip. so that we can confirm which HMC offloaded dump.
2. Printing Dump id to identify Dump in error case

### DIFF
--- a/include/dump_offload.hpp
+++ b/include/dump_offload.hpp
@@ -331,8 +331,8 @@ inline void requestRoutes(App& app)
                     boost::beast::http::status::not_found);
                 return;
             }
-            BMCWEB_LOG_INFO << "Dump offload initiated by: "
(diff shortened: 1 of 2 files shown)


gtmills:
@raviteja-b What issue are you debugging? Can you add more explaining what issues these are helping dump and why we need this PR ?

gtmills:
@raviteja-b Do we plan to remove these before GA / after we fix the above defects ?

gtmills:
I still have a concern about filling the journal

raviteja-b:
For sure These traces won't fill journal.

we are helpless in case of priority 1 defects without having these traces

gtmills (approved):

PR #281, opened 2022-03-11
Enhance logs for static ip config

asmithakarun:
This commit enhances the logs with the values of
all the ip properties when a user tries to configure
a static ip along with the ip address of the user
who initiated the request.

Tested By:

Checked journal logs for the ip properties

### DIFF
--- a/redfish-core/lib/hypervisor_system.hpp
+++ b/redfish-core/lib/hypervisor_system.hpp
@@ -510,7 +510,8 @@ inline void
 }
 
 inline void handleHypervisorIPv4StaticPatch(
-    const std::string& ifaceId, const nlohmann::json& input,
+    const crow::Request& req, const std::string& ifaceId,
+    const nlohmann::json& input,
     const std::shared_ptr<bmcweb::AsyncResp>& asyncResp)
 {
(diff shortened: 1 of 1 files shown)


redfish-core/lib/hypervisor_system.hpp
```diff
@@ -599,6 +600,12 @@ inline void handleHypervisorIPv4StaticPatch(
             return;
         }
 
+        BMCWEB_LOG_ERROR
```
raviteja-b, line 603: Can we use BMCWEB_LOG_CRITICAL << INFO:

raviteja-b (approved):

redfish-core/lib/hypervisor_system.hpp
gtmills: +1

sunharis:
> @asmithakarun Why do we need this? Can you explain what defect this will help solve?

This is to track who asks to set which network parameters for VMI ip. We felt this is an important log to capture while analyzing various VMI network issues seen from STC.

sunharis (approved):
I think its ok to have ERROR level with INFO pretag.

PR #286, opened 2022-03-14
Health and State are independent

gtmills:
If State is UnavailableOffline or Absent, still allow Health to be set
to Critical.

This should fix part of SW545991.

This matches with what we do in like other places in processor.hpp, Assemblies, Fans, Memory etc: 
https://github.com/ibm-openbmc/bmcweb/blob/9d3146f5b2710c387f353846e23fc22b8fe42a99/redfish-core/lib/processor.hpp#L481

### DIFF
--- a/redfish-core/lib/processor.hpp
+++ b/redfish-core/lib/processor.hpp
@@ -1080,7 +1080,8 @@ inline void
                 {
                     aResp->res.jsonValue["Status"]["State"] = "Absent";
                 }
-                else if (functional == false)
+
+                if (functional == false)
                 {
                     aResp->res.jsonValue["Status"]["Health"] = "Critical";
                 }


gtmills:
https://w3.rchland.ibm.com/projects/bestquest/?defect=SW546084 also looks like this might fix it

redfish-core/lib/processor.hpp
```diff
@@ -1080,7 +1080,8 @@ inline void
                 {
                     aResp->res.jsonValue["Status"]["State"] = "Absent";
                 }
-                else if (functional == false)
+
+                if (functional == false)
```
RameshIyyar, line 1084: I am not sure this will fix the SW545991 defect because the defect says the `Health` property value showed as expected but, the`State` property showed as `Enabled` when it is deallocated. Looks we are updating only Health property in this case?

RameshIyyar: Still, I am not sure this will fix SW545991 but, it will fix SW546084 so, I will approve.

RameshIyyar (approved):

PR #292, opened 2022-03-18
HW-Isolation: Return an appropriate error if the request is failed.

RameshIyyar:
It fixes [SW546086](https://w3.rchland.ibm.com/projects/bestquest/?defect=SW546086)

Gerrit patch for the first commit: https://gerrit.openbmc-project.xyz/c/openbmc/bmcweb/+/52159

### DIFF
--- a/redfish-core/include/error_messages.hpp
+++ b/redfish-core/include/error_messages.hpp
@@ -418,6 +418,25 @@ void propertyValueIncorrect(crow::Response& res, const std::string& arg1,
                             const std::string& arg2);
 
 /**
+ * @brief Formats PropertyValueResourceConflict message into JSON
+ * Message body: "The property '%1' with the requested value of '%2' could
+ * not be written because the value conflicts with the state or configuration
+ * of the resource at '%3'."
+ *
+ * @param[in] arg1 Parameter of message that will replace %1 in its body.
+ * @param[in] arg2 Parameter of message that will replace %2 in its body.
+ * @param[in] arg3 Parameter of message that will replace %3 in its body.
+ *
+ * @returns Message PropertyValueResourceConflict to JSON */
+nlohmann::json propertyValueResourceConflict(const std::string& arg1,
+                                             const std::string& arg2,
+                                             const std::string& arg3);
+
+void propertyValueResourceConflict(crow::Response& res, const std::string& arg1,
+                                   const std::string& arg2,
+                                   const std::string& arg3);
+
+/**
  * @brief Formats ResourceCreationConflict message into JSON
  * Message body: "The resource could not be created.  The service has a resource
  * at URI '<arg1>' that conflicts with the creation request."
--- a/redfish-core/include/utils/hw_isolation.hpp
+++ b/redfish-core/include/utils/hw_isolation.hpp
@@ -11,6 +11,74 @@ namespace hw_isolation_utils
 {
 
 /**
+ * @brief API used to return ChassisPowerStateOffRequiredError with
+ *        the Chassis id that will get by using the given resource
+ *        object to perform some PATCH operation on the resource object.
+ *
+ * @param[in] aResp - The redfish response to return to the caller.
+ * @param[in] resourceObjPath - The redfish resource dbus object path.
+ *
+ * @return The redfish response in given response buffer.
+ */
+inline void retChassisPowerStateOffRequiredError(
+    const std::shared_ptr<bmcweb::AsyncResp>& aResp,
+    const sdbusplus::message::object_path& resourceObjPath)
+{
+    crow::connections::systemBus->async_method_call(
+        [aResp, resourceObjPath](
+            const boost::system::error_code ec,
+            const boost::container::flat_map<
+                std::string, boost::container::flat_map<
+                                 std::string, std::vector<std::string>>>&
+                ancestors) {
+            if (ec)
+            {
+                BMCWEB_LOG_ERROR << "DBUS response error [" << ec.value()
+                                 << " : " << ec.message()
+                                 << "] when tried to get parent Chassis id for "
+                                    "the given resource object ["
+                                 << resourceObjPath.str << "]";
+                messages::internalError(aResp->res);
+                return;
+            }
+
+            if (ancestors.empty())
+            {
+                BMCWEB_LOG_ERROR
+                    << "The given resource object [" << resourceObjPath.str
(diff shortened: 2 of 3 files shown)


redfish-core/include/error_messages.hpp
```diff
@@ -417,6 +417,25 @@ nlohmann::json propertyValueIncorrect(const std::string& arg1,
 void propertyValueIncorrect(crow::Response& res, const std::string& arg1,
                             const std::string& arg2);
 
+/**
```
RameshIyyar, line 420: Please let me know whether I need to push this commit upstream even without using it in the bmcweb?

gtmills: Yes, should be able too

redfish-core/src/error_messages.cpp
```diff
@@ -928,6 +928,40 @@ void propertyValueIncorrect(crow::Response& res, const std::string& arg1,
...
 }
 
+/**
+ * @internal
+ * @brief Formats PropertyValueResourceConflict message into JSON
+ *
+ * See header file for more information
+ * @endinternal
+ */
+nlohmann::json propertyValueResourceConflict(const std::string& arg1,
```
gtmills, line 938: Isn't this generated from some tool? 

https://github.com/openbmc/bmcweb/commit/81856681e1feabe267c0d1998182c9226106aa6d

RameshIyyar: Looks like the error_messages.cpp|hpp are not auto-generated based on the [parse_registries.py](https://github.com/ibm-openbmc/bmcweb/blob/1020/scripts/parse_registries.py#L57).

I think we need to manually add the support in that files to use the `messages` abstraction layer that used to return as a redfish response. - Please correct me if I am wrong.

Even, I am not seeing any obvious changes in that commit that you pointed out for the auto generation.

But, I think I need to change the `MessageId` major and minor version from `1.12.1` to `1.10.0` since the `propertyValueResourceConflict` message registry was added in that particular version.

redfish-core/include/error_messages.hpp
RameshIyyar: OK, Gerrit patch for the same commit: https://gerrit.openbmc-project.xyz/c/openbmc/bmcweb/+/52159

redfish-core/src/error_messages.cpp
```diff
@@ -928,6 +928,40 @@ void propertyValueIncorrect(crow::Response& res, const std::string& arg1,
...
+                                             const std::string& arg2,
+                                             const std::string& arg3)
+{
+    return nlohmann::json{
+        {"@odata.type", "#Message.v1_1_1.Message"},
+        {"MessageId", "Base.1.10.0.PropertyValueResourceConflict"},
```
gtmills, line 944: Is this the only one that uses  1.10?
I see       {"MessageId", "Base.1.8.1.ResourceCreationConflict"}, ?

redfish-core/src/error_messages.cpp
```diff
@@ -928,6 +928,40 @@ void propertyValueIncorrect(crow::Response& res, const std::string& arg1,
...
+/**
+ * @internal
+ * @brief Formats PropertyValueResourceConflict message into JSON
+ *
+ * See header file for more information
+ * @endinternal
+ */
+nlohmann::json propertyValueResourceConflict(const std::string& arg1,
+                                             const std::string& arg2,
+                                             const std::string& arg3)
```
gtmills, line 940: This is the old format, which I guess is what we want for downstream

gtmills:
I am good. Just 2 questions

gtmills (approved):

redfish-core/include/utils/hw_isolation.hpp
```diff
@@ -57,29 +125,33 @@ inline void
...
+            // The Enabled property value will be "false" to isolate.
+            constexpr bool enabledPropVal = false;
             if (strcmp(dbusError->name, "xyz.openbmc_project.Common.Error."
                                         "InvalidArgument") == 0)
             {
-                constexpr bool isolate = false;
-                messages::propertyValueIncorrect(aResp->res, "@odata.id",
-                                                 std::to_string(isolate));
+                messages::propertyValueIncorrect(
+                    aResp->res, "@odata.id", std::to_string(enabledPropVal));
```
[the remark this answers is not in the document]
RameshIyyar: The backend will expect the resource inventory path to isolate and that's might be wrong based on the backend end validation and return an `InvalidArgument` error.

Here, the bmcweb returning `The property 'Enabled' with the requested value of 'false' could not be written because the value does not meet the constraints of the implementation` means the user might get confused, right? I will agree, In general, if the user assumed that passed property value is wrong if we returned `propertyValueIncorrect` error. 

So, do we need to use `ResourceNotFound` in this case since the backend D-Bus API expects inventory path alone for the manual operation (since nothing is wrong in the "Enabled" property value because it is like toggle (true/false)? 

In the future, the backed expecting more arguments then, we might need to adjust that here too i.e communicate to the user like which argument is wrong.

redfish-core/src/error_messages.cpp
RameshIyyar: There are few other properties that have used the respective major and minor versions of the message registry like this.
```
-bash-4.2$ git grep -in "MessageId" -- redfish-core/src/error_messages.cpp | grep -v 1.8.1
redfish-core/src/error_messages.cpp:587:        {"MessageId", "Base.1.11.0.ServiceDisabled"},
redfish-core/src/error_messages.cpp:944:        {"MessageId", "Base.1.10.0.PropertyValueResourceConflict"},
redfish-core/src/error_messages.cpp:1392:        {"MessageId", "Base.1.12.1.StrictAccountTypes"},
redfish-core/src/error_messages.cpp:2204:        {"MessageId", "OpenBMC.0.2.0.InvalidUpload"},
redfish-core/src/error_messages.cpp:2223:        {"MessageId", "Base.1.5.0.MutualExclusiveProperties"},
redfish-core/src/error_messages.cpp:2251:        {"MessageId", "Base.1.9.0.RestrictedRole"},
```

I think this will go away once we migrated to the upstream since its added based on the base header id - Please refer [here](https://github.com/openbmc/bmcweb/blob/master/redfish-core/src/error_messages.cpp#L129)

redfish-core/src/error_messages.cpp
RameshIyyar: yes right

PR #300, opened 2022-03-24
Display vmiIp irrespective of interfaceEnabled prop

asmithakarun:
Earlier, GET on the hypervisor ethernet interface hardcodes
InterfaceEnabled property of the redfish response as true and
vmi ip config data will be displayed based on the dbus object's
"Enabled" property - display ip when the property is true, else
return null.

With this change, the vmi ip config data will be displayed irrespective
of the "Enabled" property's value and InterfaceEnabled property of
the redfish response is mapped to the "Enabled" property of the dbus object.
This allows the client to know the configuration data, whether it is consumed
by host/not.

Tested By:

dbus introspect command for vmi eth interface:

busctl introspect xyz.openbmc_project.Network.Hypervisor /xyz/openbmc_project/network/hypervisor/eth1/ipv4/addr0

GET https://${bmc}/redfish/v1/Systems/hypervisor/EthernetInterfaces/eth1
{
  "@odata.id": "/redfish/v1/Systems/hypervisor/EthernetInterfaces/eth1",
  "@odata.type": "#EthernetInterface.v1_5_1.EthernetInterface",
  "DHCPv4": {
    "DHCPEnabled": false
  },
  "Description": "Hypervisor's Virtual Management Ethernet Interface",
  "HostName": "",
  "IPv4Addresses": [
    {
      "Address": "9.3.29.190",
      "AddressOrigin": "Static",
      "Gateway": "9.3.29.1",
      "SubnetMask": "255.255.252.0"
    }
  ],
  "IPv4StaticAddresses": [
    {
      "Address": "9.3.29.190",
      "AddressOrigin": "Static",
      "Gateway": "9.3.29.1",
      "SubnetMask": "255.255.252.0"
    }
  ],
  "Id": "eth1",
  "InterfaceEnabled": false,
  "Name": "Hypervisor Ethernet Interface"
}

Change-Id: I2dd84bd870822531a891404fc7e59de6a349471c

### DIFF
--- a/redfish-core/lib/hypervisor_system.hpp
+++ b/redfish-core/lib/hypervisor_system.hpp
@@ -432,7 +432,6 @@ inline void parseInterfaceData(
     jsonResponse["Id"] = ifaceId;
     jsonResponse["@odata.id"] =
         "/redfish/v1/Systems/hypervisor/EthernetInterfaces/" + ifaceId;
-    jsonResponse["InterfaceEnabled"] = true;
     jsonResponse["HostName"] = ethData.hostname;
     jsonResponse["DHCPv4"]["DHCPEnabled"] =
         translateDHCPEnabledToBool(ethData.DHCPEnabled, true);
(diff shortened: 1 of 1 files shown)


asmithakarun:
These changes are verified with HMC
HMC can now have the VMI interfaces shown even in Power Off state with this patch.

raviteja-b (approved):

redfish-core/lib/hypervisor_system.hpp
```diff
@@ -443,20 +442,17 @@ inline void parseInterfaceData(
     ipv4StaticArray = nlohmann::json::array();
     for (auto& ipv4Config : ipv4Data)
     {
-        if (ipv4Config.isActive)
+        jsonResponse["InterfaceEnabled"] = ipv4Config.isActive;
```
gtmills, line 445: Isn't this just the last item in the for loop?

sunharis: Only one IP address is supported on VMI interface. So there will be always only one IP in this list

gtmills: Where do we check to ensure we only support 1 VMI interface?

asmithakarun: This logic lies in the networkd repo. So when a user configures ip on an interface, the existing dbus object will be deleted and a new dbus object will be created with user given inputs.

Ref: https://github.com/ibm-openbmc/phosphor-networkd/blob/1020/src/ibm/hypervisor-network-mgr-src/hyp_ethernet_interface.cpp#L570

gtmills: > This logic lies in the networkd repo. 

Yeah it seems like we should not just rely on the networkd repo, can we add a check here?

asmithakarun: Since, this PR is intended for another purpose, I will take it up in a different PR.
I will create a defect for myself for this and work on it once all my priority 1's are done.

sunharis: SW547380 is created

gtmills (approved):

PR #303, opened 2022-03-25
Push-sytle eventing: Revert "TerminateAfterRetries

sunharis:
We had a bmcweb crash when the subscriber is deleted using
redfish command after its marked for delete via the event
retry policy "TerminateAfterRetries"

This commit reverts the "TerminateAfterRetries" support

Tested by:
  Send DELETE subscriber on a subscription which was set to
  TerminateAfterRetries. Verified there is no bmcweb crash

### DIFF
--- a/http/http_client.hpp
+++ b/http/http_client.hpp
@@ -421,12 +421,9 @@ class HttpClient : public std::enable_shared_from_this<HttpClient>
                 }
 
                 state = ConnState::terminated;
-                // Remove the subscription
+                // TODO :Remove the subscription
                 BMCWEB_LOG_ERROR << "TerminateAfterRetries is set. retryCount: "
-                                 << retryCount << " .Subscriber: " << subId
-                                 << " deleted";
-                persistent_data::EventServiceStore::getInstance()
-                    .removeSubscription(subId);
+                                 << retryCount << " .Subscriber: " << subId;
                 return;
             }
             if (retryPolicyAction == "SuspendRetries")
--- a/include/event_service_store.hpp
+++ b/include/event_service_store.hpp
@@ -247,7 +247,7 @@ class EventServiceStore
     boost::container::flat_map<std::string, std::shared_ptr<UserSubscription>>
         subscriptionsConfigMap;
     EventServiceConfig eventServiceConfig;
-    bool needWrite{false};
+
     static EventServiceStore& getInstance()
     {
         static EventServiceStore eventServiceStore;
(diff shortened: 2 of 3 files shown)


rfrandse:
fixes [SW545127](https://w3.rchland.ibm.com/projects/bestquest/?defect=SW545127)

gtmills:
Why can't we revert this whole commit? 
https://github.com/ibm-openbmc/bmcweb/commit/502574ec7ea049d51e78ae5cf2123ed4a7ce8692

Why can't we pull out the whole functions: removeSubscription, needsWrite, etc.

gtmills:
> Lets keep those code - i will work on the real fix for this issue by April. We can get that to upstream as well , and to the GA1. This change i have tested for the scenario you mentioned in the defect - and the crash is not seen any more

Would prefer if we just reverted this PR. Couple reasons:
1. Don't want us to leave some function laying around 
2. Would like a rereview of the whole functionality when we pull back in 
3. Would prefer we reuse that existing delete functionality when we reable this

sunharis:
> > Lets keep those code - i will work on the real fix for this issue by April. We can get that to upstream as well , and to the GA1. This change i have tested for the scenario you mentioned in the defect - and the crash is not seen any more
> 
> Would prefer if we just reverted this PR. [502574e](https://github.com/ibm-openbmc/bmcweb/commit/502574ec7ea049d51e78ae5cf2123ed4a7ce8692) Couple reasons:
> 
> 1. Don't want us to leave some function laying around
> 2. Would like a rereview of the whole functionality when we pull back in
> 3. Would prefer we reuse that existing delete functionality if possible when we reenable this

Fine - Reverting the complete commit will result in sending the events to the subscribers who are TerminateAfterTetries, and each event will do 5 times retry as well. This may eat up the bmcweb execution cycle.

sunharis:
> 1. Don't want us to leave some function laying around

At https://github.com/ibm-openbmc/bmcweb/commit/502574ec7ea049d51e78ae5cf2123ed4a7ce8692 

1. There are some debug traces - Needed
2. Setting the subscriber Id at the subscription object - Needed for the fixes which were pushed later
3. Not sending the events further if the connection state is set to Terminated - Needed, which will save the bmcweb cycles by avoiding sending to a subscriber which is not-valid

I agree there are some code which will not be used form https://github.com/ibm-openbmc/bmcweb/commit/502574ec7ea049d51e78ae5cf2123ed4a7ce8692 at include/event_service_store.hpp file. But they will not be executed since i removed the "delete" call in this PR

gtmills:
> I agree there are some code which will not be used form https://github.com/ibm-openbmc/bmcweb/commit/502574ec7ea049d51e78ae5cf2123ed4a7ce8692 at include/event_service_store.hpp file.

Can we trim more of this code? 

Can we just remove removeSubscription, persistSubscription, needsWrite ?

sunharis:
> > I agree there are some code which will not be used form [502574e](https://github.com/ibm-openbmc/bmcweb/commit/502574ec7ea049d51e78ae5cf2123ed4a7ce8692) at include/event_service_store.hpp file.
> 
> Can we trim more of this code?
> 
> Can we just remove removeSubscription, persistSubscription, needsWrite ?

Sure. I can do that.

gtmills:
Thanks, I think that is a good middle ground. The traces, etc can stay

sunharis:
> Thanks, I think that is a good middle ground. The traces, etc can stay

Done @gtmills

gtmills (approved):

PR #307, opened 2022-03-30
Dump Offload:Reset Dump OffloadUri as part of clean up

raviteja-b:
Recently its decided that Dump manager gets know if offload is in
progress based on OffloadUri

If OffloadUri value is set, dump manager rejects other dump offload requests
This PR resets OffloadUri property to handle error scenarios.

### DIFF
--- a/include/dump_offload.hpp
+++ b/include/dump_offload.hpp
@@ -163,6 +163,38 @@ class Handler : public std::enable_shared_from_this<Handler>
         });
     }
 
+    void resetOffloadURI()
+    {
+        std::string value{""};
+        crow::connections::systemBus->async_method_call(
+            [this,
+             self(shared_from_this())](const boost::system::error_code ec) {
+                if (ec)
+                {
+                    BMCWEB_LOG_ERROR << "DBUS response error: Unable to set "
+                                        "the dump OffloadUri "
+                                     << ec;
+
+                    if (ec.value() == EBADR)
+                    {
+                        this->connection->sendStreamErrorStatus(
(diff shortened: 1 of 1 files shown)


sunharis (approved):

raviteja-b:
> Please add Tested by section - and explain this corner case.

done

include/dump_offload.hpp
```diff
@@ -180,6 +212,7 @@ class Handler : public std::enable_shared_from_this<Handler>
             unixSocket.close();
             std::remove(unixSocketPath.c_str());
         }
+        resetOffloadURI();
```
gtmills, line 215: Don't need to do this more often? What if you hit line 202?

        if (ec)
        {
            BMCWEB_LOG_ERROR << "Unix socket file clean up failed "
                             << unixSocketPath;
            this->connection->sendStreamErrorStatus(
                boost::beast::http::status::internal_server_error);
            return;
        }
        
        
        
I guess shouldn't happen

gtmills, line 215: Don't need to do this more often? What if you hit line 202?

        if (ec)
        {
            BMCWEB_LOG_ERROR << "Unix socket file clean up failed "
                             << unixSocketPath;
            this->connection->sendStreamErrorStatus(
                boost::beast::http::status::internal_server_error);
            return;
        }
        
        
        
I guess shouldn't happen

raviteja-b: yes it should not happen unless /tmp file system screwed up
in that case system reboot needed to recover

gtmills (approved):

PR #301, opened 2022-03-24
DELETE Dump: Handle Unavailable D-bus error

raviteja-b:
Dump manager return Unavailable error for delete operation
when then same dump offload is getting offloaded.

This PR maps Unavailable D-bus error to redfish
ServiceTemporaryUnavailable error.

PR #309, opened 2022-03-31
LogService: HW-Isolation: Return an appropriate error

RameshIyyar:
It fixes [SW547295](https://w3.rchland.ibm.com/projects/bestquest/?defect=SW547295) - LogService: HW-Isolation: Return an appropriate error

- Hardware isolation D-Bus service errors:

  - Returns "NotAllowed" error when the manual hardware isolation request
    is not allowed during host power on or running.

    - Redfish server returns "ChassisPowerStateOffRequired" error.

  - Returns "InsufficientPermission" error when the user is not allowed
    to delete the hardware isolation entry that's created by the system due to
    some error.

    - Redfish server returns "ResourceCannotBeDeleted" error for the DELETE
      method request.

  - Returns "InvalidArgument" if the given resource is unable to locate in the
    backend system model (aka, device tree) to process the request.

    - Redfish server returns "PropertyValueExternalConflict" error.

### DIFF
--- a/redfish-core/include/utils/hw_isolation.hpp
+++ b/redfish-core/include/utils/hw_isolation.hpp
@@ -130,8 +130,8 @@ inline void
             if (strcmp(dbusError->name, "xyz.openbmc_project.Common.Error."
                                         "InvalidArgument") == 0)
             {
-                messages::propertyValueIncorrect(
-                    aResp->res, "@odata.id", std::to_string(enabledPropVal));
+                messages::propertyValueExternalConflict(
+                    aResp->res, "Enabled", std::to_string(enabledPropVal));
             }
             else if (strcmp(dbusError->name, "xyz.openbmc_project.Common.Error."
                                              "NotAllowed") == 0)
--- a/redfish-core/include/utils/hw_isolation.hpp
+++ b/redfish-core/include/utils/hw_isolation.hpp
@@ -138,13 +138,6 @@ inline void
             {
                 retChassisPowerStateOffRequiredError(aResp, resourceObjPath);
             }
-            else if (strcmp(dbusError->name, "xyz.openbmc_project.Common.Error."
-                                             "Unavailable") == 0)
-            {
-                messages::propertyValueResourceConflict(
-                    aResp->res, "Enabled", std::to_string(enabledPropVal),
-                    "HardwareIsolation");
-            }
             else if (strcmp(dbusError->name, "xyz.openbmc_project."
                                              "HardwareIsolation.Error."
                                              "IsolatedAlready") == 0)
@@ -271,18 +264,6 @@ inline void
                     {
                         retChassisPowerStateOffRequiredError(aResp,
                                                              resourceObjPath);
-                    }
-                    else if (strcmp(dbusError->name,
-                                    "xyz.openbmc_project.Common.Error."
-                                    "Unavailable") == 0)
-                    {
-                        // The Enabled property value will be "true" to
-                        // de-isolate.
-                        constexpr bool enabledPropVal = true;
-                        messages::propertyValueResourceConflict(
-                            aResp->res, "Enabled",
-                            std::to_string(enabledPropVal),
-                            "HardwareIsolation");
                     }
                     else if (strcmp(dbusError->name,
                                     "xyz.openbmc_project.Common.Error."
--- a/redfish-core/lib/log_services.hpp
+++ b/redfish-core/lib/log_services.hpp
@@ -5265,18 +5265,51 @@ inline void deleteSystemHardwareIsolationLogEntryById(
 
             // Delete the respective dbus entry object
             crow::connections::systemBus->async_method_call(
-                [asyncResp, entryObjPath](const boost::system::error_code ec) {
-                    if (ec)
+                [asyncResp,
+                 entryObjPath](const boost::system::error_code ec,
+                               const sdbusplus::message::message& msg) {
+                    if (!ec)
                     {
-                        BMCWEB_LOG_ERROR << "DBUS response error ["
-                                         << ec.value() << " : " << ec.message()
-                                         << "] when tried to delete the given "
-                                            "object path: "
-                                         << entryObjPath.str;
+                        messages::success(asyncResp->res);
+                        return;
+                    }
+
+                    BMCWEB_LOG_ERROR << "DBUS response error [" << ec.value()
+                                     << " : " << ec.message()
+                                     << "] when tried to delete the given "
+                                     << "entry: " << entryObjPath.str;
+
+                    const sd_bus_error* dbusError = msg.get_error();
+
+                    if (dbusError == nullptr)
+                    {
                         messages::internalError(asyncResp->res);
                         return;
                     }
-                    messages::success(asyncResp->res);
+
+                    BMCWEB_LOG_ERROR << "DBus ErrorName: " << dbusError->name
+                                     << " ErrorMsg: " << dbusError->message;
+
+                    if (strcmp(dbusError->name,
+                               "xyz.openbmc_project.Common.Error."
+                               "NotAllowed") == 0)
+                    {
+                        messages::chassisPowerStateOffRequired(asyncResp->res,
+                                                               "chassis");
+                    }
+                    else if (strcmp(dbusError->name,
+                                    "xyz.openbmc_project.Common.Error."
+                                    "InsufficientPermission") == 0)
+                    {
+                        messages::resourceCannotBeDeleted(asyncResp->res);
+                    }
+                    else
+                    {
+                        BMCWEB_LOG_ERROR << "DBus Error is unsupported so "
+                                            "returning as Internal Error";
+                        messages::internalError(asyncResp->res);
+                    }
+                    return;
                 },
                 objType[0].first, entryObjPath.str,
                 "xyz.openbmc_project.Object.Delete", "Delete");
@@ -5331,18 +5364,45 @@ inline void postSystemHardwareIsolationLogServiceClearLog(
 
             // Delete all HardwareIsolation entries
             crow::connections::systemBus->async_method_call(
-                [asyncResp](const boost::system::error_code ec) {
-                    if (ec)
+                [asyncResp](const boost::system::error_code ec,
+                            const sdbusplus::message::message& msg) {
+                    if (!ec)
                     {
-                        BMCWEB_LOG_ERROR
-                            << "DBUS response error [" << ec.value() << " : "
-                            << ec.message()
-                            << "] when tried to delete all HardwareIsolation "
-                               "entries";
+                        messages::success(asyncResp->res);
+                        return;
+                    }
+
+                    BMCWEB_LOG_ERROR
+                        << "DBUS response error [" << ec.value() << " : "
+                        << ec.message()
+                        << "] when tried to delete all HardwareIsolation "
+                           "entries";
+
+                    const sd_bus_error* dbusError = msg.get_error();
+
+                    if (dbusError == nullptr)
+                    {
                         messages::internalError(asyncResp->res);
                         return;
                     }
-                    messages::success(asyncResp->res);
+
+                    BMCWEB_LOG_ERROR << "DBus ErrorName: " << dbusError->name
+                                     << " ErrorMsg: " << dbusError->message;
+
+                    if (strcmp(dbusError->name,
+                               "xyz.openbmc_project.Common.Error."
+                               "NotAllowed") == 0)
+                    {
+                        messages::chassisPowerStateOffRequired(asyncResp->res,
+                                                               "chassis");
+                    }
+                    else
+                    {
+                        BMCWEB_LOG_ERROR << "DBus Error is unsupported so "
+                                            "returning as Internal Error";
+                        messages::internalError(asyncResp->res);
+                    }
+                    return;
                 },
                 objType[0].first, "/xyz/openbmc_project/hardware_isolation",
                 "xyz.openbmc_project.Collection.DeleteAll", "DeleteAll");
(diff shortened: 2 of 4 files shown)


devenrao (approved):

devenrao:
looks good to me

redfish-core/include/utils/hw_isolation.hpp
```diff
@@ -141,9 +141,9 @@ inline void
             else if (strcmp(dbusError->name, "xyz.openbmc_project.Common.Error."
                                              "Unavailable") == 0)
             {
-                messages::propertyValueResourceConflict(
-                    aResp->res, "Enabled", std::to_string(enabledPropVal),
-                    "HardwareIsolation");
+                messages::serviceDisabled(
```
gtmills, line 144: @raviteja-b just mapped this to Service Unavailable 
https://github.com/ibm-openbmc/bmcweb/pull/301/files

And if I read 
https://github.com/openbmc/phosphor-dbus-interfaces/blob/master/yaml/xyz/openbmc_project/Common.errors.yaml#L17
That mapping makes sense, do we need a Disabled Error?

redfish-core/include/utils/hw_isolation.hpp
```diff
@@ -275,20 +275,26 @@ inline void
                     else if (strcmp(dbusError->name,
                                     "xyz.openbmc_project.Common.Error."
                                     "Unavailable") == 0)
+                    {
+                        messages::serviceDisabled(
+                            aResp->res, "/redfish/v1/Systems/system/"
```
gtmills, line 280: Do we need a disabled Error?

redfish-core/include/utils/hw_isolation.hpp
```diff
@@ -275,20 +275,26 @@ inline void
...
+                                        "LogServices/HardwareIsolation");
+                    }
+                    else if (strcmp(dbusError->name,
+                                    "xyz.openbmc_project.Common.Error."
+                                    "InsufficientPermission") == 0)
                     {
                         // The Enabled property value will be "true" to
                         // de-isolate.
                         constexpr bool enabledPropVal = true;
                         messages::propertyValueResourceConflict(
```
gtmills, line 290: Why is InsufficientPermission mapped to propertyValueResourceConflict ?

redfish-core/lib/log_services.hpp
```diff
@@ -5249,18 +5249,59 @@ inline void deleteSystemHardwareIsolationLogEntryById(
...
+                    {
+                        messages::serviceDisabled(
+                            asyncResp->res, "/redfish/v1/Systems/system/"
+                                            "LogServices/HardwareIsolation");
+                    }
+                    else if (strcmp(dbusError->name,
+                                    "xyz.openbmc_project.Common.Error."
+                                    "InsufficientPermission") == 0)
+                    {
+                        messages::resourceCannotBeDeleted(asyncResp->res);
```
gtmills, line 5296: This maps to a 403 forbidden here, makes sense 
https://github.com/openbmc/bmcweb/blob/24861a286eb148fcec22afecffb73b596752700f/redfish-core/src/error_messages.cpp#L321

gtmills:
I am struggling with some of these mappings: 
e.g. InsufficientPermission  to propertyValueResourceConflict and Unavailable to ServiceDisabled. 
Do we need some more errors here?

redfish-core/include/utils/hw_isolation.hpp
RameshIyyar: The hardware isolation application will reject the request with an `InsufficientPermission` error If the user tried to configure/deconfigure the hardware that's isolated the system due to some failure.

Here, using the _PropertyValueResourceConflict_ error for the above case to say the user like `"The property 'Enabled' with the requested value of 'true' could not be written because the value conflicts with the state or configuration of the resource at '/redfish/v1/Systems/system/LogServices/HardwareIsolation/Entries/1'."` so he knows why it is not allowed by looking that at the pointed URI i.e he can know the severity of the deconfigured resource.

redfish-core/lib/log_services.hpp
```diff
@@ -5249,18 +5249,59 @@ inline void deleteSystemHardwareIsolationLogEntryById(
...
-                    messages::success(asyncResp->res);
+
+                    BMCWEB_LOG_ERROR << "DBus ErrorName: " << dbusError->name
+                                     << " ErrorMsg: " << dbusError->message;
+
+                    if (strcmp(dbusError->name,
+                               "xyz.openbmc_project.Common.Error."
+                               "NotAllowed") == 0)
+                    {
+                        messages::chassisPowerStateOffRequired(asyncResp->res,
```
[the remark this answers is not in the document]
RameshIyyar: The hardware isolation application won't allow configuring/deconfiguring _something_ when the host is IPLing or running so if the user attempted to configure/deconfigure some resource, that's request will be rejected with the _NotAllowed_ error and we can convey to the user like You need to poweroff the chassis to perform the requested operation.

redfish-core/lib/log_services.hpp
RameshIyyar: Why used different errors ([PropertyValueResourceConflict](https://github.com/ibm-openbmc/bmcweb/pull/309/commits/edb2243cf318414bf0f80a58a12475e95ade25ed#r839847003) and ResourceCannotBeDeleted) for the same backend returned error `InsufficientPermission`?

Because the same request comes from a different URI (one from Core/DIMM to configure and another one coming from DELETE method by using the HardwareIsolation logservices), returning the different error based on the request to give a more appropriate error.

RameshIyyar:
> I am struggling with some of these mappings: e.g. InsufficientPermission to propertyValueResourceConflict and Unavailable to ServiceDisabled. Do we need some more errors here?

I don't think we need more errors because existing errors fit our cases as mentioned below.

- InsufficientPermission -> [PropertyValueResourceConflict](https://github.com/ibm-openbmc/bmcweb/pull/309#discussion_r840261040) and  [ResourceCannotBeDeleted](https://github.com/ibm-openbmc/bmcweb/pull/309#discussion_r840267551)
- Unavailable -> [ServiceDisabled](https://github.com/ibm-openbmc/bmcweb/pull/309#discussion_r840206572)

redfish-core/include/utils/hw_isolation.hpp
gtmills: Can we add a xyz.openbmc_project.Common.Error.Disabled ?

redfish-core/lib/log_services.hpp
gtmills: > Why used different errors ([PropertyValueResourceConflict](https://github.com/ibm-openbmc/bmcweb/pull/309/commits/edb2243cf318414bf0f80a58a12475e95ade25ed#r839847003) and ResourceCannotBeDeleted) for the same backend returned error InsufficientPermission?

Yes, we should not do this. 
I think this mapping makes sense InsufficientPermission -> resourceCannotBeDeleted; the other one does not

redfish-core/include/utils/hw_isolation.hpp
RameshIyyar: IMO, an Unavailable error in the D-Bus level is fine for the backend application for this kind of use case.

But, we need to differentiate only on the Redfish side based on the context so I think we can go with it.

Still, you feel we need an error like 1-to-1 mapping i.e Redfish to D-Bus, I think we need to talk more from the design point of view i.e supporting Redfish message registry in the D-Bus or similar design since we can add arguments for the message registry but, (AFAIK) we can not do that in the current D-Bus error design (YAML file).

redfish-core/include/utils/hw_isolation.hpp
RameshIyyar: FYI, I posted the below response for this [comment](https://github.com/ibm-openbmc/bmcweb/pull/309/commits/edb2243cf318414bf0f80a58a12475e95ade25ed#r840688817) but, I could not see it in the PR so adding it here too.

IMO, an Unavailable error in the D-Bus level is fine for the backend application for this kind of use case.

But, we need to differentiate only on the Redfish side based on the context so I think we can go with it.

Still, you feel we need an error like 1-to-1 mapping i.e Redfish to D-Bus, I think we need to talk more from the design point of view i.e supporting Redfish message registry in the D-Bus or similar design since we can add arguments for the message registry but, (AFAIK) we can not do that in the current D-Bus error design (YAML file).

redfish-core/include/utils/hw_isolation.hpp
RameshIyyar: I think the context is different from the Redfish URI point of view, right?  i.e one place **the User using the property** (i.e The Enabled property as per Processor schema for the Core and Memory schema for the DIMM) and **the User using the HTTP DELETE request** (i.e DELETE request in the HardwareIsolation LogEntry) to deconfigure it.

Please share your view on this context or You are referring to preferring the 1-to-1 error mapping i.e DBus to Redfish even if the user context (aka URI) is different? IMO, we should give a detailed error message for a better understanding rather than sticking with 1-to-1 mapping.

If you have a different opinion, please share, I like to listen to your thoughts.

redfish-core/lib/log_services.hpp
RameshIyyar: I think the context is different from the Redfish URI point of view, right?  i.e one place **the User using the property** (i.e The Enabled property as per Processor schema for the Core and Memory schema for the DIMM) and **the User using the HTTP DELETE request** (i.e DELETE request in the HardwareIsolation LogEntry) to deconfigure it.

Please share your view on this context or You are referring to preferring the 1-to-1 error mapping i.e DBus to Redfish even if the user context (aka URI) is different? IMO, we should give a detailed error message for a better understanding rather than sticking with 1-to-1 mapping.

If you have a different opinion, please share, I like to listen to your thoughts.

gtmills:
I am not good with these mappings, these mappings don't make sense and we just had different mappings at https://github.com/ibm-openbmc/bmcweb/pull/301/files

gtmills:
I am not good with these mappings, these mappings don't make sense and we just had different mappings at https://github.com/ibm-openbmc/bmcweb/pull/301/files

redfish-core/include/utils/hw_isolation.hpp
raviteja-b: @gtmills  In dump use case, I have mapped  to serviceTemporarilyUnavailable wherever bmcweb has retry time interval for client to retry.
As Ramesh explained seems ServiceDisabled makes more sense as service disabled and there no retry time interval associated in this use-case.
ServiceDisabled error also underlying maps  to boost::beast::http::status::service_unavailable status code.

redfish-core/include/utils/hw_isolation.hpp
raviteja-b: @gtmills I think we don't need same naming convention followed for mapping D-bus error to redfish errors right?
As Ramesh explained seems ServiceDisabled makes more sense as service disabled and there no retry time interval associated in this use-case.
ServiceDisabled error also underlying maps to boost::beast::http::status::service_unavailable status code.
I don't think we need any changes here

redfish-core/lib/log_services.hpp
raviteja-b: @gtmills I think PropertyValueResourceConflict error suits to this particular case with respect to use-case.
 it can provide conflicting resource URI to client so that client can understands the error details and take further actions accordingly.

raviteja-b (approved):

raviteja-b:
> I am not good with these mappings, these mappings don't make sense and we just had different mappings at https://github.com/ibm-openbmc/bmcweb/pull/301/files

@gtmills I think these mappings can be done based on use-case as redfish error functions are just wrappers, we may not be able to follow same D-bus to redfish error name mappings for all use-cases.
I thinks as long as http status code and args making sense we are good

redfish-core/include/utils/hw_isolation.hpp
RameshIyyar: Hi @gtmills / @raviteja-b 

I have removed the code to handle the `Unavailable` error, we don't have support to enable the hardware isolation service through the Redfish for the user. Refer [here](https://github.com/ibm-openbmc/bmcweb/pull/309/commits/6d57fd3a3a494b92e14811bb616bbd0562914a89) to get more details.

redfish-core/include/utils/hw_isolation.hpp
RameshIyyar: I have removed the code to handle the `Unavailable` error, we don't have support to enable the hardware isolation service through the Redfish for the user. Refer [here](https://github.com/ibm-openbmc/bmcweb/pull/309/commits/6d57fd3a3a494b92e14811bb616bbd0562914a89) to get more details.

redfish-core/include/utils/hw_isolation.hpp
RameshIyyar: Done, changed to use `ResourceCannotBeDeleted` as pre request.

redfish-core/lib/log_services.hpp
RameshIyyar: Already using the `ResourceCannotBeDeleted` as per request.

gtmills (approved):
IMO NotAllowed doesn't mean chassisPowerStateOffRequired.

We are mapping NotAllowed to propertyNotWritable here:

[bmcweb/redfish-core/lib/processor.hpp](https://github.com/ibm-openbmc/bmcweb/blob/a023c353c11bf0ae0304eefca900dc3a0b6dd1b5/redfish-core/lib/processor.hpp#L1382)
I prefer an "xyz.openbmc_project.Common.Error.ChassisPowerStateOffRequired" and consistent mappings, making the code more maintainable.

That said, Ramesh and I have gone back and forth for a while here: https://ibm-systems-power.slack.com/archives/G3WAVQYTH/p1651041334942209?thread_ts=1648694265.564979&cid=G3WAVQYTH

I don't see any problems in this code that will cause issues in production; my concerns are maintainability, so going to approve. I still don't agree with this mapping, so would still be good to talk about in a design call.

PR #315, opened 2022-04-06
shell support via pty to bmcweb

patelabhishek9893:
This commit creates a terminal that allows users to log in to the
system using their session name (username), where a password is not
required. because we are passing '-f', which Does not perform
authentication; the user is pre-authenticated

This commit removes the setting user ID (Setuid), allowing users to
access the terminal.

Tested: 
tested using obmc_shell (bmcweb) and web-UI on the Rainer system. 

- Currently, any user who has access to the system can access
  WebSocket.
  Ex: service user can access as service (root user because it gets
  converted into root user)user, Admin user can access as an admin
  user, etc.
  In the future, we will add prvilege check for WebSockets

### DIFF
--- a/http/obmc_shell.hpp
+++ b/http/obmc_shell.hpp
@@ -50,7 +50,7 @@ class Handler : public std::enable_shared_from_this<Handler>
             // ERROR
             if (session)
             {
-                session->close("Error creating ssh child process");
+                session->close("Error creating child process for login shell.");
             }
             return;
         }
@@ -58,45 +58,17 @@ class Handler : public std::enable_shared_from_this<Handler>
         {
             // CHILD
 
-            // sets the effective user ID
-            // as service user
-            bool isUidSet = false;
             if (auto userName = session->getUserName(); !userName.empty())
             {
-                if (struct passwd* pw = getpwnam(userName.c_str());
-                    pw != nullptr)
-                {
-                    int uidr = setuid(pw->pw_uid);
-                    if (!uidr)
-                    {
-                        isUidSet = true;
-                    }
-                    else
-                    {
-                        BMCWEB_LOG_ERROR << "sets service user failed Error: "
-                                         << uidr;
-                    }
-                }
-                else
-                {
-                    BMCWEB_LOG_ERROR << "getpwnam return null passwd";
-                }
+                execl("/bin/login", "/bin/login", "-f", userName.c_str(), NULL);
+                // execl only returns on fail
+                BMCWEB_LOG_ERROR << "execl() for /bin/login failed: " << errno;
+                session->close("Internal Error Login failed");
             }
             else
             {
-                BMCWEB_LOG_ERROR << "userName is empty";
+                session->close("Error session user name not found");
             }
-
-            // close connection unable to set
-            // setuid()
-            if (!isUidSet)
-            {
-                session->close("sets service user failed");
-                return;
-            }
-
-            // create /bin/sh chiled process
-            execl("/bin/sh", "/bin/sh", nullptr);
             return;
         }
         if (pid > 0)
@@ -202,7 +174,7 @@ static std::map<crow::websocket::Connection*, std::shared_ptr<Handler>>
 inline void requestRoutes(App& app)
 {
     BMCWEB_ROUTE(app, "/bmc-console")
-        .privileges({{"ConfigureManager"}})
+        .privileges({{"OemIBMPerformService"}})
         .websocket()
         .onopen([](crow::websocket::Connection& conn,
                    const std::shared_ptr<bmcweb::AsyncResp>&) {


include/dbus_monitor.hpp
```diff
@@ -105,8 +105,8 @@ inline int onPropertyUpdate(sd_bus_message* m, void* userdata,
 inline void requestRoutes(App& app)
 {
     BMCWEB_ROUTE(app, "/subscribe")
-        .privileges({{"Login"}})
         .websocket()
+        .privileges({{"Login"}})
```
gtmills, line 109: Why change this?

patelabhishek9893: if I don't change then WebSocket authentication will not happen for "/subscribe". as you know previously there was no check for the socket.
so this has to change for authentication

joseph-reynolds (commented):
I had seen the 3 websocket privilege changes earlier.  I reviewed them now and then look okay, but I am not an expert in that area.

Why did you remove the code which changes the process (setuid) to the service user?  I feel this code provides important defense against unknown bugs in the code.
Consider if a non-service user is able to get to this code, it would give them root access.  Whereas leaving the code as-is would give them shell access, but not root access.

The other changes look good.

amboar:
Setting aside the setuid() stuff, there are two significant concerns on one of the upstream patches:

1. https://gerrit.openbmc-project.xyz/c/openbmc/bmcweb/+/46991/1/http/routing.hpp#1210
2. https://gerrit.openbmc-project.xyz/c/openbmc/bmcweb/+/46991/1/http/routing.hpp#1346

amboar:
> Why did you remove the code which changes the process (setuid) to the service user? I feel this code provides important defense against unknown bugs in the code. Consider if a non-service user is able to get to this code, it would give them root access. Whereas leaving the code as-is would give them shell access, but not root access.

Yes, I don't think the change is right. The user should have the capability to escalate their own privileges if necessary.

patelabhishek9893:
WebSocket privilege required more work so there will be a separate commit for that.
let's merge this in first.

http/obmc_shell.hpp
```diff
@@ -60,43 +60,15 @@ class Handler : public std::enable_shared_from_this<Handler>
...
-                    {
-                        BMCWEB_LOG_ERROR << "sets service user failed Error: "
-                                         << uidr;
-                    }
-                }
-                else
-                {
-                    BMCWEB_LOG_ERROR << "getpwnam return null passwd";
-                }
+                // create /bin/sh chiled process
```
joseph-reynolds, line 65: The child process was created by forkpty above.  The comment 4 lines above and this comment and not correct.  The execl() call logs in the service user.

joseph-reynolds (approved):
The code looks good, modulo the comments.

http/obmc_shell.hpp
patelabhishek9893: done

http/obmc_shell.hpp
```diff
@@ -58,45 +58,15 @@ class Handler : public std::enable_shared_from_this<Handler>
...
-                    {
-                        BMCWEB_LOG_ERROR << "sets service user failed Error: "
-                                         << uidr;
-                    }
-                }
-                else
-                {
-                    BMCWEB_LOG_ERROR << "getpwnam return null passwd";
-                }
+                // replaces the current process with /bin/login chiled process
```
amboar, line 63: Do we need this comment? The code tells us this.

http/obmc_shell.hpp
```diff
@@ -58,45 +58,15 @@ class Handler : public std::enable_shared_from_this<Handler>
...
+                execl("/bin/login", "/bin/login", "-f", userName.c_str(), NULL);
```
amboar, line 64: execl() can fail and you must handle the failure.

amboar (changes requested):

amboar:
The title of the PR isn't really accurate. An xterm is a specific X-windows application that is a terminal emulator. Here we're just adding shell support via pty to bmcweb, nothing really to do with xterm.

http/obmc_shell.hpp
```diff
@@ -58,45 +58,19 @@ class Handler : public std::enable_shared_from_this<Handler>
...
-                    {
-                        BMCWEB_LOG_ERROR << "sets service user failed Error: "
-                                         << uidr;
-                    }
-                }
-                else
-                {
-                    BMCWEB_LOG_ERROR << "getpwnam return null passwd";
-                }
+                execl("/bin/login", "/bin/login", "-f", userName.c_str(), NULL);
```
aahmed-2, line 63: do we want to cast the last argument to (char*) NULL? 

From the execl man page:

The list of arguments must be terminated by a NULL pointer, and, since these are variadic functions, this pointer must be cast (char *) NULL.

aahmed-2 (approved):

http/obmc_shell.hpp
patelabhishek9893: I don't this we need to do that

patelabhishek9893: FYI: https://linuxhint.com/exec_linux_system_call_c/

http/obmc_shell.hpp
patelabhishek9893: done

http/obmc_shell.hpp
patelabhishek9893: done

http/obmc_shell.hpp
```diff
@@ -58,45 +58,19 @@ class Handler : public std::enable_shared_from_this<Handler>
...
+                // execl only returns on fail
+                BMCWEB_LOG_ERROR << "unable to replaces current process, execl "
+                                    "fail, Erro number:"
```
joseph-reynolds, line 66: Nit: Typo: Erro -> Error (missing "r").
If you fix this, please also  change the comment 13 lines above "Error creating ssh child process" because it is not accurate; try: "Error creating child process for login shell."

patelabhishek9893: Done

http/obmc_shell.hpp
```diff
@@ -50,53 +50,27 @@ class Handler : public std::enable_shared_from_this<Handler>
...
-                else
-                {
-                    BMCWEB_LOG_ERROR << "getpwnam return null passwd";
-                }
+                execl("/bin/login", "/bin/login", "-f", userName.c_str(), NULL);
+                // execl only returns on fail
+                BMCWEB_LOG_ERROR << "unable to replaces current process, execl "
+                                    "fail, Error number:"
+                                 << errno;
+                session->close("Error unable to replaces current process");
```
amboar, line 68: This error message seems a bit uninformative. If you found this in a log file and you weren't familiar with what generated it, how would you respond?

http/obmc_shell.hpp
```diff
@@ -50,53 +50,27 @@ class Handler : public std::enable_shared_from_this<Handler>
...
             }
             else
             {
-                BMCWEB_LOG_ERROR << "userName is empty";
+                session->close("Error session userName name not found");
```
amboar, line 72: `... userName name ...`? Also can we put the specific username in the log message, or is this doing some magic variable interpolation?

http/obmc_shell.hpp
```diff
@@ -50,53 +50,27 @@ class Handler : public std::enable_shared_from_this<Handler>
...
-                                         << uidr;
-                    }
-                }
-                else
-                {
-                    BMCWEB_LOG_ERROR << "getpwnam return null passwd";
-                }
+                execl("/bin/login", "/bin/login", "-f", userName.c_str(), NULL);
+                // execl only returns on fail
+                BMCWEB_LOG_ERROR << "unable to replaces current process, execl "
```
amboar, line 65: What about: `"execl() for /bin/login failed: " << errno`

amboar (commented):
Minor nits about the log messages

http/obmc_shell.hpp
patelabhishek9893: done

http/obmc_shell.hpp
```diff
@@ -50,53 +50,25 @@ class Handler : public std::enable_shared_from_this<Handler>
...
-                }
+                execl("/bin/login", "/bin/login", "-f", userName.c_str(), NULL);
+                // execl only returns on fail
+                BMCWEB_LOG_ERROR << "execl() for /bin/login failed: " << errno;
+                session->close("Internal Error Login failed");
             }
             else
             {
-                BMCWEB_LOG_ERROR << "userName is empty";
+                session->close("Error session user name not found");
```
gtmills, line 70: Can't be specific about the user name as Andrew suggested?

http/obmc_shell.hpp
```diff
@@ -202,7 +174,7 @@ static std::map<crow::websocket::Connection*, std::shared_ptr<Handler>>
 inline void requestRoutes(App& app)
 {
     BMCWEB_ROUTE(app, "/bmc-console")
-        .privileges({{"ConfigureManager"}})
+        .privileges({{"OemIBMPerformService"}})
```
gtmills, line 177: As stated in the PR message, this isn't doing anything

gtmills:
Please be clear with your PR message, please use a spell checker / grammar checker. 
E.g. " if the future we will add privilege chack for WebSocket"
"In the future, we will add prvilege check for WebSockets"

http/obmc_shell.hpp
patelabhishek9893: username is not found that why I am login this error.

http/obmc_shell.hpp
patelabhishek9893: yes

patelabhishek9893: we like to change this right right now in this commit

gtmills:
I think all comments have been addressed. 
Seems reasonable to me.

gtmills (approved):

PR #319, opened 2022-04-12
Restrict the Operator role

joseph-reynolds:

Tested:
2. POST /redfish/v1/AccountService/Accounts/new with Role:Operator
Ensure it failed with message Base.1.9.0.RestrictedRole.
Posting a new Administrator user is successful.

### DIFF
--- a/redfish-core/lib/roles.hpp
+++ b/redfish-core/lib/roles.hpp
@@ -87,7 +87,7 @@ inline bool getOemPrivFromRole(std::string_view role, nlohmann::json& privArray)
 
 inline bool isRestrictedRole(const std::string& role)
 {
-    if (role == "OemIBMServiceAgent")
+    if ((role == "Operator") || (role == "OemIBMServiceAgent"))
     {
         return true;
     }


joseph-reynolds:
Note the underlying Restricted Role support was added as part of https://github.com/ibm-openbmc/bmcweb/commit/b6e0ca14f79a05cdbddff059e2def2ae8ed7ef5c
Because of that support, this change was far easier than I had anticipated.

gtmills:
This is an interesting idea. My only concern is it is still here: 
```
{
"@odata.id": ["/redfish/v1/AccountService/Roles"](https://rain111bmc.aus.stglabs.ibm.com/redfish/v1/AccountService/Roles),
"@odata.type": "#RoleCollection.RoleCollection",
"Description": "BMC User Roles",
"Members": [
{
"@odata.id": ["/redfish/v1/AccountService/Roles/Administrator"](https://rain111bmc.aus.stglabs.ibm.com/redfish/v1/AccountService/Roles/Administrator)
},
{
"@odata.id": ["/redfish/v1/AccountService/Roles/Operator"](https://rain111bmc.aus.stglabs.ibm.com/redfish/v1/AccountService/Roles/Operator)
},
{
"@odata.id": ["/redfish/v1/AccountService/Roles/ReadOnly"](https://rain111bmc.aus.stglabs.ibm.com/redfish/v1/AccountService/Roles/ReadOnly)
},
{
"@odata.id": ["/redfish/v1/AccountService/Roles/OemIBMServiceAgent"](https://rain111bmc.aus.stglabs.ibm.com/redfish/v1/AccountService/Roles/OemIBMServiceAgent)
}
],
"Members@odata.count": 4,
"Name": "Roles Collection"
}
```

gtmills:
I guess they then go to the role and see Restricted and restricted means "Can not be created or modified to"? Because the definition doesn't really say that 
```
                "Restricted": {
                    "description": "An indication of whether use of the role is restricted.",
                    "longDescription": "This property shall indicate whether use of the role is restricted by a service as defined by the 'Restricted roles and restricted privileges' clause of the Redfish Specification.  If this property is not present, the value shall be assumed to be `false`.",```

joseph-reynolds:
> I guess they then go to the role and see Restricted and restricted means "Can not be created or modified to"? Because the definition doesn't really say that

Exactly that.  I asked about this at the Redfish meeting: What does a Redfish user do when an operation fails with a RestrictedRole error message?  The general consensus was at this point the users would consult the docs (IBM product docs).

joseph-reynolds:
NOTE: This is a temporary change.  It is needed because we have not completed a review and testing of what BMC operations an Operator user should be allowed to access.  The intention is to enable the Operator role in a future fixpack.

Note: The remaining allowed Redfish Roles are: Administrator, ReadOnly.  Users who want to use the Operator role but cannot because it is restricted may select Administrator or ReadOnly, then change to Operator when it becomes available.

joseph-reynolds:
No changes are **required** for the Web GUI.  The underlying Redfish APIs will report RestrictedRole when appropriate.

However, I believe the Web GUI is making changes to remove the Operator role from in the Add user and Edit user interfaces.

gtmills (approved):

PR #320, opened 2022-04-14
Add humidity hierarchy

gtmills:
As stated in PDI humidity is a valid hierarchy
https://github.com/openbmc/phosphor-dbus-interfaces/blob/master/yaml/xyz/openbmc_project/Sensor/Value.interface.yaml

We were wondering why the humidity sensor wasn't showing at /redfish/v1/Chassis/chassis/Sensors, this is why.

Also Remove dup from node::sensors

/xyz/openbmc_project/sensors/voltage was in this list twice.
Looks like a bad merge conflict.
Match upstream which has this under BMCWEB_NEW_POWERSUBSYSTEM_THERMALSUBSYSTEM.
Upstream makes sense only show these when using the all sense behavior.
https://github.com/openbmc/bmcweb/blob/master/redfish-core/lib/sensors.hpp

Did not test, straightforward change.


gtmills:
I'll push [As stated in PDI humidity is a valid hierarchy](https://github.com/ibm-openbmc/bmcweb/pull/320/commits/6c0b091de1afef349b2bc7f6f8f6f579dc5ddfb1) upstream

derekhoward55 (approved):

PR #321, opened 2022-04-18
Send event for VMI method change on ethernet intf

asmithakarun:
This change sends an event to the client when
there is a method change on the vmi interface from
static to DHCP and vice versa.

This is needed because, say, when the client changes method
from static to dhcp and if the dhcp is not connected, the
client wont receive any event from bmcweb, thus the cient thinks
that the method is static on that interface.

Fixes: https://w3.rchland.ibm.com/projects/bestquest/?verb=view&id=SW548700&table=defect

Tested By:
Connected HMC and verified whether the events are
being sent.

### DIFF
--- a/include/event_dbus_monitor.hpp
+++ b/include/event_dbus_monitor.hpp
@@ -13,6 +13,7 @@ static std::shared_ptr<sdbusplus::bus::match::match> matchBMCStateChange;
 static std::shared_ptr<sdbusplus::bus::match::match>
     matchVMIIPEnabledPropChange;
 static std::shared_ptr<sdbusplus::bus::match::match> matchVMIIPChange;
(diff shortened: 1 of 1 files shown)


include/event_dbus_monitor.hpp
```diff
@@ -243,6 +245,16 @@ void registerVMIIPChangeSignal()
         VMINwPropertyChange);
 }
 
+void registerVMIEthIntfPropChangeSignal()
```
raviteja-b, line 248: This D-bus signal hits on IP address, gateway or any other property change on this interface as well right?

so with this change, it sends redfish events on any property change

sunharis (approved):

raviteja-b (approved):

rfrandse:
fixes SW548700

gtmills (approved):

PR #322, opened 2022-04-18
SW549923: Add asyncResp support during handleUpgrade

patelabhishek9893:
This commit mainly implies support for asyncResp to handle upgrades for
 WebSocket connection.
The current implementation uses the earlier method of using the response
object and calling response. end() to initiate completion handler.
This commit modifies the implementation to use asyncResp, where the
completion handler gets called asynchronously as the response object
goes out of scope.

This change is mainly for handling privilege into WebSocket connections.

test:
tested using bmc_console on the Rainer system. Able to test success cases.
Hard to reproduce failed test case.

### DIFF
--- a/http/http_connection.hpp
+++ b/http/http_connection.hpp
@@ -375,15 +375,36 @@ class Connection :
                               [self] { self->completeRequest(); });
         });
 
+        auto asyncResp = std::make_shared<bmcweb::AsyncResp>(res);
         if (thisReq.isUpgrade() &&
             boost::iequals(
                 thisReq.getHeaderValue(boost::beast::http::field::upgrade),
                 "websocket"))
         {
-            handler->handleUpgrade(thisReq, res, std::move(adaptor));
-            // delete lambda with self shared_ptr
-            // to enable connection destruction
-            res.setCompleteRequestHandler(nullptr);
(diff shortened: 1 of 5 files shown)


gtmills:
@patelabhishek9893 What is the upstream commit?

aahmed-2 (approved):

joseph-reynolds (approved):
Seems okay to me, but I am not an expert in this area.

http/http_connection.hpp
```diff
@@ -375,15 +375,36 @@ class Connection :
...
+                [self(shared_from_this())]() {
+                    if (self->res.resultInt() != 101)
+                    {
+                        // When any error occurs during handle upgradation,
+                        // the result in response will be set to respective
+                        // error. during success Result will be Switching
+                        // Protocols (101), which implies successful handle
+                        // upgrade. Response needs to be sent over this
+                        // connection only on failure.
+                        boost::asio::post(self->adaptor.get_executor(),
```
gtmills, line 394: This is still mostly unaddressed from upstream:
https://gerrit.openbmc-project.xyz/c/openbmc/bmcweb/+/46989/1..6/http/http_connection.hpp#b374

PR #323, opened 2022-04-19
VMI Cert: Reduce timeout to 10 seconds

raviteja-b:
Currently VMI cert timeout increased to 60 sec as a workaround due to VMI issue earlier.

VMI cert request is generally completed with in a second.
This commit sets max timeout back to 10 seconds.

Tested By:
Using SignCSR curl command

### DIFF
--- a/include/ibm/management_console_rest.hpp
+++ b/include/ibm/management_console_rest.hpp
@@ -985,7 +985,7 @@ void handleCsrRequest(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
         std::make_shared<boost::asio::steady_timer>(
             crow::connections::systemBus->get_io_context());
 
-    timeout->expires_after(std::chrono::seconds(60));
+    timeout->expires_after(std::chrono::seconds(10));
     crow::connections::systemBus->async_method_call(
         [asyncResp, timeout](const boost::system::error_code ec,
                              sdbusplus::message::message& m) {


sunharis (commented):
Change looks good. Please talk to phyp/vmi and get their change-Id which has improved the cert signing time. That can be mentioned here in this PR and then its good to go

gtmills:
Can we just revert https://github.com/ibm-openbmc/bmcweb/commit/22f81d0fbda4ba83e7052e1d1e1875dae5c5cc1a  ?

I like reverting because it cleans up some error logs and then in the history you see a revert and it is easy to know we no longer need 22f81d0f

raviteja-b:
> Can we just revert [22f81d0](https://github.com/ibm-openbmc/bmcweb/commit/22f81d0fbda4ba83e7052e1d1e1875dae5c5cc1a) ?
> 
> I like reverting because it cleans up some error logs and then in the history you see a revert and it is easy to know we no longer need [22f81d0](https://github.com/ibm-openbmc/bmcweb/commit/22f81d0fbda4ba83e7052e1d1e1875dae5c5cc1a)

Timeout increased as part of this PR as workaround for delay in signing VMI cert 
https://github.com/ibm-openbmc/bmcweb/pull/118
I think we can use this PR to revert change

gtmills:
@raviteja-b Is there more of https://github.com/ibm-openbmc/bmcweb/pull/118/files we should revert ?

raviteja-b:
> Change looks good. Please talk to phyp/vmi and get their change-Id which has improved the cert signing time. That can be mentioned here in this PR and then its good to go

I could not get exact change, but as we increased delay as workaround, its good to reduce now.
otherwise we forget

sunharis (approved):
Changes made at the original PR https://github.com/ibm-openbmc/bmcweb/pull/118/ are not available now. 
1. Traces which were added as INFO are changed to CRTITICAL level now - due to some FFDC improvement. 
2. Code part to return service_unavailable is removed already

So reverting the old PR will not help here.  The current PR looks good

gtmills (approved):

PR #329, opened 2022-04-21
SW549923: Add Support for WebSocket privilege check

patelabhishek9893:
This commit enables a privilege check for the user(s) in case of upgraded
connections.
Previously there was no chack for WebSocket Privileges. Users with no
privileges will also be able to access WebSockets connections. For
bmc_console and other WebSocket privileges are defined, but they are
useless without check.

Here there are mainly two commits retrieved from upstream.

https://gerrit.openbmc-project.xyz/c/openbmc/bmcweb/+/46990/1
https://gerrit.openbmc-project.xyz/c/openbmc/bmcweb/+/46991/1

downstream PR is also there for WebSocket connection
establishment, mainly 101 checks (when WebSocket establishes connection).
https://github.com/ibm-openbmc/bmcweb/pull/322

Tested:
Tested using the rainier system, where bmc_console only creates
a WebSocket connection for the service user. Admin, and any other
user gets discarded.

### DIFF
--- a/http/routing.hpp
+++ b/http/routing.hpp
@@ -49,7 +49,7 @@ class BaseRule
     virtual void handle(const Request&,
                         const std::shared_ptr<bmcweb::AsyncResp>&,
                         const RoutingParams&) = 0;
-    virtual void handleUpgrade(const Request& /*req*/,
+    virtual void handleUpgrade(Request& /*req*/,
                                std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
                                boost::asio::ip::tcp::socket&& /*adaptor*/)
     {
@@ -57,7 +57,7 @@ class BaseRule
     }
 #ifdef BMCWEB_ENABLE_SSL
     virtual void handleUpgrade(
-        const Request& /*req*/, std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
+        Request& /*req*/, std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
         boost::beast::ssl_stream<boost::asio::ip::tcp::socket>&& /*adaptor*/)
     {
         asyncResp->res.result(boost::beast::http::status::not_found);
@@ -362,7 +362,7 @@ class WebSocketRule : public BaseRule
         asyncResp->res.result(boost::beast::http::status::not_found);
     }
 
-    void handleUpgrade(const Request& req,
+    void handleUpgrade(Request& req,
                        std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
                        boost::asio::ip::tcp::socket&& adaptor) override
     {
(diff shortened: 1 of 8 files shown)


patelabhishek9893:
@amboar 
I try to resolve this comment downstream look like either I have to create asyn_req or create a pointer out of reference. still, life time is not a problem here I think.
https://gerrit.openbmc-project.xyz/c/openbmc/bmcweb/+/46991/1/http/routing.hpp#b1209

include/obmc_hypervisor.hpp
```diff
@@ -117,8 +117,8 @@ inline void connectHandler(const boost::system::error_code& ec)
 inline void requestRoutes(App& app)
 {
     BMCWEB_ROUTE(app, "/console1")
-        .privileges({{"ConfigureComponents", "ConfigureManager"}})
         .websocket()
+        .privileges({{"ConfigureComponents", "ConfigureManager"}})
```
joseph-reynolds, line 121: Change to OemIBMServiceAgent.  Only the service agent should be allowed to access the PHYP console.

joseph-reynolds, line 121: Change to OemIBMServiceAgent.  Only the service agent should be allowed to access the PHYP console.

joseph-reynolds (commented):
Please check the permission required for PHYP console /console1 and change if needed.
I reviewed the changes pulled from upstream, but I am not an expert.
Otherwise the changes look okay.

include/obmc_console.hpp
```diff
@@ -117,8 +117,8 @@ inline void connectHandler(const boost::system::error_code& ec)
 inline void requestRoutes(App& app)
 {
     BMCWEB_ROUTE(app, "/console0")
-        .privileges({{"ConfigureComponents", "ConfigureManager"}})
         .websocket()
+        .privileges({{"ConfigureComponents", "ConfigureManager"}})
```
gtmills, line 121: This would be Admin and above? @joseph-reynolds You good with this? Or should readonly have access too?

patelabhishek9893: @joseph-reynolds please acknowledge so I make change based on that

joseph-reynolds: This set of privileges {ConfigureComponents,ConfigureManager} is only given to Administrator (and OemBMServiceAgent) users, and is not given to Operator or ReadOnly users.  Only Administrators should have acces to /console0 (host console), so this looks correct.

joseph-reynolds (approved):
The privileges required look correct to me.

http/routing.hpp
```diff
@@ -1299,28 +1311,139 @@ class Router
...
+                {
+                    BMCWEB_LOG_ERROR << "An uncaught exception occurred. The "
+                                        "type was unknown "
+                                        "so no information was available.";
+                    asyncResp->res.result(
+                        boost::beast::http::status::internal_server_error);
+                    return;
+                }
+            },
+            "xyz.openbmc_project.User.Manager", "/xyz/openbmc_project/user",
```
gtmills, line 1444: Still comments upstream about this being a blocking call and such.. 
https://gerrit.openbmc-project.xyz/c/openbmc/bmcweb/+/46991/

PR #333, opened 2022-04-27
http_connection: Bump connection timeout to 12 min

anoo1:
Bump the connection timeout to 12min to allow firmware image uploads at
low network speeds.

Change-Id: Ia62860050c76eb61e8f65e7c5057b1ff94ebbb8e
Signed-off-by: Adriana Kobylak <anoo@us.ibm.com>

### DIFF
--- a/http/http_connection.hpp
+++ b/http/http_connection.hpp
@@ -47,10 +47,10 @@ constexpr uint64_t loggedOutPostBodyLimit = 4096;
 
 constexpr uint32_t httpHeaderLimit = 8192;
 
-// drop all connections after 3 minutes, this time limit was chosen
+// drop all connections after 12 minutes, this time limit was chosen
 // arbitrarily and can be adjusted later if needed
 static constexpr const size_t loggedInAttempts =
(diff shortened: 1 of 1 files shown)


joseph-reynolds:
Can you share the network speeds and file size you were thinking of?

joseph-reynolds:
This changes the timeout for *all* authenticated connections, not just image uploads.  It would be better to isolate this new longer timeout so it applied only to the firmware upload operation.  Can we do that?  Reset the timer for that operation?  (And I don't know the answer.)

anoo1:
> It would be better to isolate this new longer timeout so it applied only to the firmware upload operation. Can we do that?

That'd be the ideal change. We've added a TODO item for P11 since it currently there's only one timeout and it's not known how much work would it be to create a specific upload timeout.

anoo1:
> Can you share the network speeds and file size you were thinking of?

The firmware image tarball is about 150MB. Empirical testing between outside the lab to the lab show it takes about 10min for the upload to complete.

gtmills (approved):

PR #337, opened 2022-05-02
Retrieve Chassis Assembly Interface

patelabhishek9893:
williwakas is an assembly and this is how we associate slots to Flett(cable and pcie_slots).
now to describe association we need to provide a route path which we have in chassis assembly. 
chassis assembly has code that defines the route path for the chassis runtime. the code uses a member ID number in the route path which assigns runtime. Moreover, if someone removes that part from the chassis then its member ID also gets updated.

to fix this problem we have to have a similar code that generates a member ID number similar manner and provides proper indexing.

this code provides the same methodology to retrieve the member ID endpoint list with the same index.

So anyone can retrieve a list and find an index for that endpoint and use that index to build a route path.

### DIFF
--- a/redfish-core/include/utils/chassis_utils.hpp
+++ b/redfish-core/include/utils/chassis_utils.hpp
@@ -61,5 +61,238 @@ void getValidChassisID(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
     BMCWEB_LOG_DEBUG << "checkChassisId exit";
 }
 
+template <typename Callback>
+inline void
+    checkAssemblyInterface(const std::shared_ptr<bmcweb::AsyncResp>& aResp,
+                           const std::string& chassisPath,
+                           std::vector<std::string>& assemblies,
+                           Callback&& callback)
+{
+    crow::connections::systemBus->async_method_call(
+        [aResp, chassisPath, assemblies, callback{std::move(callback)}](
+            const boost::system::error_code ec,
+            const std::vector<std::pair<
+                std::string,
+                std::vector<std::pair<std::string, std::vector<std::string>>>>>&
+                subtree) {
+            if (ec)
+            {
+                BMCWEB_LOG_DEBUG << "D-Bus response error on GetSubTree " << ec;
+                messages::internalError(aResp->res);
+                return;
+            }
+
+            if (subtree.size() == 0)
+            {
+                BMCWEB_LOG_DEBUG << "No object paths found";
+                return;
+            }
+            std::vector<std::string> updatedAssemblyList;
+            for (const auto& [objectPath, serviceName] : subtree)
+            {
+                auto it =
+                    std::find(assemblies.begin(), assemblies.end(), objectPath);
+                if (it != assemblies.end())
+                {
+                    updatedAssemblyList.emplace(updatedAssemblyList.end(), *it);
+                }
+            }
+
+            if (updatedAssemblyList.size() != 0)
+            {
+                // sorting is required to facilitate patch as the array does not
+                // store and data which can be mapped back to Dbus path of
+                // assembly.
+                std::sort(updatedAssemblyList.begin(),
+                          updatedAssemblyList.end());
+                callback(updatedAssemblyList);
+            }
+        },
+        "xyz.openbmc_project.ObjectMapper",
+        "/xyz/openbmc_project/object_mapper",
+        "xyz.openbmc_project.ObjectMapper", "GetSubTree",
+        "/xyz/openbmc_project/inventory", int32_t(0), chassisAssemblyIfaces);
+}
+
+template <typename Callback>
+inline void
+    getAssemblyEndpoints(const std::shared_ptr<bmcweb::AsyncResp>& aResp,
+                         const std::string& chassisPath, Callback&& callback)
+{
+
+    BMCWEB_LOG_DEBUG << "Get assembly endpoints";
+
+    sdbusplus::message::object_path assemblyPath(chassisPath);
+    assemblyPath /= "assembly";
+
+    // if there is assembly association, look
+    // for endpoints
+    crow::connections::systemBus->async_method_call(
+        [aResp, chassisPath, callback{std::move(callback)}](
+            const boost::system::error_code ec,
+            const std::variant<std::vector<std::string>>& endpoints) {
+            if (ec)
+            {
+                BMCWEB_LOG_DEBUG << "DBUS response "
+                                    "error";
+                messages::internalError(aResp->res);
+                return;
+            }
+
+            const std::vector<std::string>* assemblyList =
+                std::get_if<std::vector<std::string>>(&(endpoints));
+
+            if (assemblyList == nullptr)
+            {
+                BMCWEB_LOG_DEBUG << "No assembly found";
+                return;
+            }
+
+            std::vector<std::string> sortedAssemblyList = *assemblyList;
+            std::sort(sortedAssemblyList.begin(), sortedAssemblyList.end());
+            checkAssemblyInterface(aResp, chassisPath, sortedAssemblyList,
+                                   std::move(callback));
+            return;
+        },
+        "xyz.openbmc_project.ObjectMapper", assemblyPath,
+        "org.freedesktop.DBus.Properties", "Get",
+        "xyz.openbmc_project.Association", "endpoints");
+}
+
+template <typename Callback>
+inline void checkForAssemblyAssociations(
+    const std::shared_ptr<bmcweb::AsyncResp>& aResp,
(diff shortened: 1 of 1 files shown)


gtmills:
This is all copied from assembly.hpp? What is our long term plan here? Move assembly.hpp to use this? 
This is more code than I expected, all is needed?

redfish-core/include/utils/chassis_utils.hpp
```diff
@@ -61,5 +61,258 @@ void getValidChassisID(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
...
+                return;
+            }
+
+            // check if the chassis path belongs to the chassis ID passed
+            for (const auto& path : chassisPaths)
+            {
+                BMCWEB_LOG_DEBUG << "Chassis Paths from Mapper " << path;
+                std::string chassis =
+                    sdbusplus::message::object_path(path).filename();
+                if (chassis != "chassis")
```
sbteeks, line 296: Chassis could be "chassis" or "chassis15363". I believe the later one is MEX/IO module or something. The assembly numbers/IDs are generated for each of chassis type w/ starting number 0. We may be good for now as we are only interested in wiliwaks/disk backplane assembly IDs. But if there is a need to get the assembly numbers for MEX/IO module, it is going to fail. I'm not sure whether we need to think about MEX assemblies or not though. 

@gtmills it's your call whether we need make it generic enough to handle all supported chassis or just stick with "chassis"

redfish-core/include/utils/chassis_utils.hpp
```diff
@@ -61,5 +61,258 @@ void getValidChassisID(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
...
+    BMCWEB_LOG_DEBUG << "Check chassis for association";
+
+    std::string chassis =
+        sdbusplus::message::object_path(chassisPath).filename();
+    if (chassis.empty())
+    {
+        BMCWEB_LOG_ERROR << "Failed to find / in Chassis path";
+        messages::internalError(aResp->res);
+        return;
+    }
```
sbteeks, line 235: I don't see the need for the above check since we already know that chassis is not empty, and is "chassis" at this point. If it is not then this method wouldn't get called.

redfish-core/include/utils/chassis_utils.hpp
```diff
@@ -61,5 +61,258 @@ void getValidChassisID(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
...
+                        "xyz.openbmc_project.Association.Definitions")
+                    {
+                        checkForAssemblyAssociations(aResp, chassisPath,
+                                                     serviceName,
+                                                     std::move(callback));
+                        return;
+                    }
+                }
+            }
+            return;
```
sbteeks, line 264: Don't need this return statement

redfish-core/include/utils/chassis_utils.hpp
```diff
@@ -61,5 +61,258 @@ void getValidChassisID(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
...
+            const associationList* value =
+                std::get_if<associationList>(&associations);
+            if (value == nullptr)
+            {
+                BMCWEB_LOG_DEBUG << "DBUS response error";
+                messages::internalError(aResp->res);
+                return;
+            }
+
+            bool isAssmeblyAssociation = false;
```
sbteeks, line 198: Don't need this variable...

redfish-core/include/utils/chassis_utils.hpp
```diff
@@ -61,5 +61,258 @@ void getValidChassisID(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
...
+            for (const auto& listOfAssociations : *value)
+            {
+                if (std::get<0>(listOfAssociations) != "assembly")
+                {
+                    // implies this is not an assembly
+                    // association
+                    continue;
+                }
+
+                isAssmeblyAssociation = true;
```
sbteeks, line 208: Call getAssemblyEndpoints(...) here

redfish-core/include/utils/chassis_utils.hpp
```diff
@@ -61,5 +61,258 @@ void getValidChassisID(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
...
+                break;
+            }
+
+            if (isAssmeblyAssociation)
+            {
+                getAssemblyEndpoints(aResp, chassisPath, std::move(callback));
+            }
```
sbteeks, line 215: Remove this - assuming you are getting rid off isAssemblyAssociation variable.

redfish-core/include/utils/chassis_utils.hpp
```diff
@@ -61,5 +61,258 @@ void getValidChassisID(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
...
+            {
+                BMCWEB_LOG_DEBUG << "No object paths found";
+                return;
+            }
+            std::vector<std::string> updatedAssemblyList;
+            for (const auto& [objectPath, serviceName] : subtree)
+            {
+                // This list will store common paths between assemblies fetched
+                // from association json and assemblies which are actually
+                // implemeted. This is to handle the case in which there is
```
sbteeks, line 95: "implemented" - not a bug :)

redfish-core/include/utils/chassis_utils.hpp
```diff
@@ -61,5 +61,258 @@ void getValidChassisID(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
...
+                {
+                    updatedAssemblyList.emplace(updatedAssemblyList.end(), *it);
+                }
+            }
+
+            if (updatedAssemblyList.size() != 0)
+            {
+                // sorting is required to facilitate patch as the array does not
+                // store and data which can be mapped back to Dbus path of
+                // assembly.
```
sbteeks, line 110: This comment could be irrelevant in our case since we are not doing a patch; but we do need to sort to find the position of the requested assembly from this list...

redfish-core/include/utils/chassis_utils.hpp
```diff
@@ -61,5 +61,258 @@ void getValidChassisID(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
...
+            return;
+        },
+        "xyz.openbmc_project.ObjectMapper",
+        "/xyz/openbmc_project/object_mapper",
+        "xyz.openbmc_project.ObjectMapper", "GetObject", chassisPath,
+        std::array<const char*, 0>{});
+}
+
+template <typename Callback>
+inline void getChassisAssempbly(const std::shared_ptr<bmcweb::AsyncResp>& aResp,
```
sbteeks, line 273: Typo in the method name...

redfish-core/include/utils/chassis_utils.hpp
patelabhishek9893: @sbteeks make it generic, not needed now, but let me know what you needed

redfish-core/include/utils/chassis_utils.hpp
patelabhishek9893: we are checking this in case object_path unable to retrieve chassis data

redfish-core/include/utils/chassis_utils.hpp
patelabhishek9893: why?

redfish-core/include/utils/chassis_utils.hpp
sbteeks: It is for @gtmills to decide whether there will be a need in the future to get assembly ID from other chassis - like chassis15363 as an example. For our current needs just "chassis" type is good.

redfish-core/include/utils/chassis_utils.hpp
sbteeks: Isn't that already checked at line number 283 for the same path?

sbteeks: Isn't that already checked at line number 283 for the same path?

patelabhishek9893:
> This is all copied from assembly.hpp? What is our long term plan here? Move assembly.hpp to use this? This is more code than I expected, all is needed?

yes, this is copied past from assembly. right now assembly code is more complicated they are doing get and patch both in the same code, where my and teek's requirements are different.
yes, all this code is needed to retrieve index information.

sbteeks (approved):
I believe there should be an issue opened to address @gtmills comment on long term plan to use this code in assembly.hpp

SunnySrivastava1984:
> I believe there should be an issue opened to address @gtmills comment on long term plan to use this code in assembly.hpp

Is there any issue opened to address/track the problem that Gunnar pointed here?

PR #350, opened 2022-05-10
SW549923: Only service user can access bmc-console/hypervisor

patelabhishek9893:
https://w3.rchland.ibm.com/projects/bestquest/?defect=SW549923

bmcweb has limitations for WebSocket privilege, where none of the
current WebSocket checks for privilege before allowing users to access
WebSocket.

bmc-console need to check for a ser (privilege) before allowing a user
to access the console.

hypervisor console needs to check for a ser (privilege) before allowing
a user to access the hypervisor  console

based on the current bmcweb implementation, it is easy to check the
username for an upcoming request. So this commits checking for a
username before allowing a user to access the console. if a user's name
is "service," they can only access bmc-console.

this change is tested using the Rainier system.

### DIFF
--- a/http/obmc_shell.hpp
+++ b/http/obmc_shell.hpp
@@ -179,6 +179,15 @@ inline void requestRoutes(App& app)
         .onopen([](crow::websocket::Connection& conn,
                    const std::shared_ptr<bmcweb::AsyncResp>&) {
             BMCWEB_LOG_DEBUG << "Connection " << &conn << " opened";
+
+            // TODO: this user check must be removed when WebSocket privileges
+            // work.
+            if (conn.getUserName() != "service")
+            {
+                BMCWEB_LOG_DEBUG
+                    << "only service user have access to obmc_shell";
+                conn.close("only service user have access to bmc console");
+            }
             if (auto it = mapHandler.find(&conn); it == mapHandler.end())
             {
                 auto insertData =
--- a/include/obmc_hypervisor.hpp
+++ b/include/obmc_hypervisor.hpp
@@ -122,7 +122,14 @@ inline void requestRoutes(App& app)
         .onopen([](crow::websocket::Connection& conn,
                    const std::shared_ptr<bmcweb::AsyncResp>&) {
             BMCWEB_LOG_DEBUG << "Connection " << &conn << " opened";
-
+            // TODO: this user check must be removed when WebSocket privileges
+            // work.
+            if (conn.getUserName() != "service")
+            {
+                BMCWEB_LOG_DEBUG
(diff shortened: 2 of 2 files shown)


http/obmc_shell.hpp
```diff
@@ -179,6 +179,15 @@ inline void requestRoutes(App& app)
...
                    const std::shared_ptr<bmcweb::AsyncResp>&) {
             BMCWEB_LOG_DEBUG << "Connection " << &conn << " opened";
+
+            // TODO: this user check must be removed when WebSocket privileges
+            // work.
+            if (conn.getUserName() != "service")
+            {
+                BMCWEB_LOG_DEBUG
+                    << "only service user have access to obmc_shell";
+                conn.close("only service user have access to bmc console");
```
gtmills, line 189: Do we need a return code here ? A 403 ?

gtmills:
How did you test?

http/obmc_shell.hpp
patelabhishek9893: we can, but that makes this more complicated.
because code reaches here because we accept the HTTP requests and upgrade (handle upgrade (convert HTTP to WSS)) also completed. that mean user already gets 101 (upgrade ok) for it. after that, sending 403 might be strange.

patelabhishek9893:
> How did you test?

so I logged in to a system using Web-UI as a service, and as an admin, where the service user was able to get access to the console whereas for the admin user there was no console, and in the log I was able to find message why it not allowed

patelabhishek9893:
> > How did you test?
> 
> so I logged in to a system using Web-UI as a service, and as an admin, where the service user was able to get access to the console whereas for the admin user, there was no console, and in the log, I was able to find message why it not allowed

hypervisor  also tested same way

joseph-reynolds (approved):
Seems okay to me, pending proper support for WebSocket connections.

http/obmc_shell.hpp
sbteeks, line 189: If the console session is established by the 'service' user, why there is a need to check or allow 'service' user to close?
I'm not sure whether it is possible for one user to setup console and another user close, so it is just for clarification.

patelabhishek9893: the service user has access to obmc_shell, which means if we get any other uses, we will restrict that user.
WebSocket privilege is core dumping that why this is a hack for restricting users.

gtmills: Ack.

gtmills (approved):

gtmills:
I am good with this for now. Long term we need to drive upstream.

sbteeks (approved):

PR #351, opened 2022-05-11
MEX not Support PowerSupply FirmwareVersion

patelabhishek9893:
Mex-supported chassis does not support the power supply firmware version.
where current bmcweb code tries to retrieve that data even that interface is not supported via dbus.
adding an error check code will determine why it is failing.
if it failed because of `EBADR` then no return error in redfish API 

Tested: using the mex rainier system 

Defect: https://w3.rchland.ibm.com/projects/bestquest/?defect=SW549511

### DIFF
--- a/redfish-core/lib/power_supply.hpp
+++ b/redfish-core/lib/power_supply.hpp
@@ -62,6 +62,12 @@ inline void getPowerSupplyFirmwareVersion(
                     const std::variant<std::string> state) {
             if (ec)
             {
+                if (ec.value() == EBADR)
+                {
+                    BMCWEB_LOG_DEBUG
+                        << "PowerSupply Firmware not found, skipping";
+                    return;
+                }
                 BMCWEB_LOG_ERROR << "Can't get PowerSupply firmware version!";
                 messages::internalError(asyncResp->res);
                 return;


redfish-core/lib/power_supply.hpp
```diff
@@ -62,6 +62,10 @@ inline void getPowerSupplyFirmwareVersion(
                     const std::variant<std::string> state) {
             if (ec)
             {
+                if (ec.value() == EBADR)
+                {
```
sbteeks, line 66: Maybe add INFO or DEBUG journal log entry?

mikecgithub (approved):

redfish-core/lib/power_supply.hpp
patelabhishek9893: @gtmills what you think

gtmills: Sure. 
<< "PowerSupply Firmware not found, skipping"

gtmills:
@patelabhishek9893 Please add the defect in the PR comment

spinler (approved):

mikecgithub (approved):

PR #354, opened 2022-05-15
DumpOffload: Avoid reset OffloadUri on dump offload success

raviteja-b:
As per design, pldm needs to reset OffloadURI while handling FileAck
after dump offload completed for resource and system dumps
for other dumps dump manager handles.

bmcweb still needs to reset OffloadUri on dump offload failure cases.
This is applies to all dump including BMC stored host dump and BMC dump 

This PR fixes SW548188 which is already MUST FIX approved

Tested by:
1.Few iterations of Offload system dump
2.Abruptly close connection while dump offloading
Note:

### DIFF
--- a/include/dump_offload.hpp
+++ b/include/dump_offload.hpp
@@ -187,6 +187,8 @@ class Handler : public std::enable_shared_from_this<Handler>
                     }
                     return;
                 }
+                BMCWEB_LOG_CRITICAL << "INFO: Reset OffloadUri of " << dumpType
+                                    << " dump id " << entryID;
             },
             "xyz.openbmc_project.Dump.Manager",
             "/xyz/openbmc_project/dump/" + dumpType + "/entry/" + entryID,
@@ -212,7 +214,6 @@ class Handler : public std::enable_shared_from_this<Handler>
             unixSocket.close();
             std::remove(unixSocketPath.c_str());
         }
-        resetOffloadURI();
         return;
     }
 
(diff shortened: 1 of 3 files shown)


include/dump_offload.hpp
```diff
@@ -408,14 +409,15 @@ inline void requestRoutes(App& app)
...
-        .onclose([](crow::streaming_response::Connection& conn) {
+        .onclose([](crow::streaming_response::Connection& conn, bool&) {
             auto handler = bmcHandlers.find(&conn);
             if (handler == bmcHandlers.end())
             {
                 BMCWEB_LOG_DEBUG << "No handler to cleanup";
                 return;
             }
             handler->second->cleanupSocketFiles();
+            handler->second->resetOffloadURI();
```
sunharis, line 420: no need to check for the completion status here before resetOffloadURI ?

raviteja-b: For BMC dumps, pldmd is not involved so bmcweb has to resetOffloadURI for both success/failure

dhruvibm: Success case can be  added dump manager

raviteja-b: Dhruv/sunitha

for bmc dumps, I have  change to reset offloadURI only in error cases,
in success case, dump manager reset offloadURI for bmc dump and host stored dumps
error cases, bmcweb resets

sunharis (approved):

asmithakarun (approved):

gtmills (approved):

PR #363, opened 2022-05-19
Fix PCIe Slot Count Check

santoshpuranik:
SW543568: When we have more than one chassis in the system, the code
would wrongly compare the total number of slots across all the chassis
to the number of slots in the PATCH request's payload (which is for a
specific chassis). This would cause the PATCH to fail without any
error.

This commit fixes the count check by properly counting the slots that
don't belong in the chassis of interest.

Tested:
Tested this on a system with Mex chassis and a CEC chassis.
Verified that the PATCH on LocationIndicatorActive now works fine
even when we have two chassis trees on D-Bus.

Signed-off-by: Santosh Puranik <santosh.puranik@in.ibm.com>

### DIFF
--- a/redfish-core/lib/pcie_slots.hpp
+++ b/redfish-core/lib/pcie_slots.hpp
@@ -559,6 +559,10 @@ void checkPCIeSlotsCount(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
                         "associated to chassis = "
                      << chassisID;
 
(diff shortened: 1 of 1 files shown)


santoshpuranik:
This fixes SW543568

ChicagoDuan (approved):

gtmills (approved):

PR #364, opened 2022-05-20
Reset LicenseString and Status property after request completed

raviteja-b:
Currently pldmd listens for LicenseString property change signal
and forwards license string to phyp.

This commit resets LicenseString and LicenseActivationStatus after request completed.
so that for further requests licenseString property change signal generated.

Increased  timeout to 20 sec to wait for response from pldmd
There are cases where phyp response arrives between 10-20 secs

This PR Fixes https://w3.rchland.ibm.com/projects/bestquest/?defect=SW544557
MUST FIX approved

Tested:
1. Upload new license key and check if its installed.
2. Upload same license key again and check if user got success
   instead of error

### DIFF
--- a/redfish-core/lib/license_service.hpp
+++ b/redfish-core/lib/license_service.hpp
@@ -93,6 +93,49 @@ inline void requestRoutesLicenseService(App& app)
             });
 }
 
+void resetLicenseActivationStatus(
+    const std::shared_ptr<bmcweb::AsyncResp>& asyncResp)
+{
+    std::string value{"com.ibm.License.LicenseManager.Status.Pending"};
+    crow::connections::systemBus->async_method_call(
+        [asyncResp](const boost::system::error_code ec) {
+            if (ec)
+            {
+                BMCWEB_LOG_ERROR << "DBUS response error: Unable to set "
+                                    "the LicenseString property "
+                                 << ec;
+                messages::internalError(asyncResp->res);
+                return;
+            }
+            licenseActivationStatusMatch = nullptr;
+        },
+        "com.ibm.License.Manager", "/com/ibm/license",
+        "org.freedesktop.DBus.Properties", "Set",
+        "com.ibm.License.LicenseManager", "LicenseActivationStatus",
+        std::variant<std::string>(value));
+}
+
+void resetLicenseString(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp)
+{
+    std::string value{""};
+    crow::connections::systemBus->async_method_call(
+        [asyncResp](const boost::system::error_code ec) {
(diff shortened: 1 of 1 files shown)


redfish-core/lib/license_service.hpp
```diff
@@ -243,6 +286,11 @@ inline void requestRoutesLicenseEntryCollection(App& app)
                                     messages::internalError(asyncResp->res);
                                     return;
                                 }
+                                if (*status == "com.ibm.License.LicenseManager."
+                                               "Status.Pending")
+                                {
+                                    return;
```
sunharis, line 292: So with this we are sending the empty response to the client, if the same license ? Which response code client is getting in this case

redfish-core/lib/license_service.hpp
```diff
@@ -181,7 +224,7 @@ inline void requestRoutesLicenseEntryCollection(App& app)
             std::shared_ptr<boost::asio::steady_timer> timeout =
                 std::make_shared<boost::asio::steady_timer>(
                     crow::connections::systemBus->get_io_context());
-            timeout->expires_after(std::chrono::seconds(10));
+            timeout->expires_after(std::chrono::seconds(20));
             crow::connections::systemBus->async_method_call(
```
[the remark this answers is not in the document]
raviteja-b: There are cases where phyp response arrives between 10-20 secs
Updated commit message

redfish-core/lib/license_service.hpp
raviteja-b: here we don't any response to client, this piece of code hit when bmcweb find a property change signal on LicenseActivationStatus.
this change is ignore D-bus signal if status value is Pending

gtmills:
Must fix approved this but the actual fix I see is different:

    fix-complexity :
        level :
            Low

        description :

            Minor change to check if same license string on D-bus property already

raviteja-b:
> Must fix approved this but the actual fix I see is different:
> 
> ```
> fix-complexity :
>     level :
>         Low
> 
>     description :
> 
>         Minor change to check if same license string on D-bus property already
> ```

yeah we discussed and decides to make these changes after MFB approval.

redfish-core/lib/license_service.hpp
gtmills: I think we need a comment here.. This also sounds pretty hacky

gtmills: I think we need a comment here.. This also sounds pretty hacky

gtmills: Can the comment say “why” we are doing this

gtmills (approved):

PR #369, opened 2022-05-25
SW551687: Restrict OCC passthrough to service

gtmills:
https://w3.rchland.ibm.com/projects/bestquest/?defect=SW551687

As decided, OCC passthrough should be service only:
https://app.slack.com/client/T0JA1U9GV/C0Q6TQP5Z/thread/C0Q6TQP5Z-1614612567.256600/1615218172.434700

For security concerns lock this interface down. 

This is downstream only. This OCC passthrough and this Service Agent are downstream only. 

Tested: For admin got Forbidden back. Service user still works. 


gtmills:
jenkins run tests please

PR #372, opened 2022-05-26
Make redfish adapter objects unique

patelabhishek9893:


redfish-core/lib/pcie.hpp
```diff
@@ -67,8 +67,36 @@ static inline void
...
+                 * @brief Workaround to handle duplicate PCI device list
+                 *
+                 * retrieve PCI device endpoint information and if path is
+                 * ~/chassisN/io_moduleN/slotN/adapterN then, replace redfish
+                 * PCI device as "chassisN_io_moduleN_slotN_adapterN"
+                 *
+                 */
+
+                if (parentofParentPath.parent_path().filename().find(
+                        "chassis") != std::string::npos)
```
aahmed-2, line 86: Should we use "chassis" here? I think we will have a Mex Drawerer chassis and maybe it will also have pcie devices idk?

gtmills: I don't think we should hardcode any chassis id here.. Why can't we just do this for all pcie devices? 
Can we just add the parent and the grandparent to all pcie devices? (or whatever the ancestors are we are adding?)

patelabhishek9893: Do you mean all parent info for all PCI devices?

patelabhishek9893: remove check is that chassis or not

redfish-core/lib/pcie.hpp
```diff
@@ -67,8 +67,36 @@ static inline void
...
+                {
+                    devName = parentofParentPath.parent_path().filename() + "_";
+                }
+                if (parentofParentPath.filename().find("io_module") !=
```
aahmed-2, line 90: might we replace some of this 'parentofParentPath' things with dbus::utility::getNthStringFromPath?
[Example here](https://github.com/ibm-openbmc/bmcweb/blob/e2f2ca0519409c45b7e28564fa3187952eadb8b5/redfish-core/lib/managers.hpp#L757)

redfish-core/lib/cable.hpp
```diff
@@ -181,14 +181,41 @@ inline void
             for (const auto& fullPath : value)
             {
                 sdbusplus::message::object_path path(fullPath);
-                std::string leaf = path.filename();
-                if (leaf.empty())
+                sdbusplus::message::object_path parentPath = path.parent_path();
```
gtmills, line 184: Anyway we can have a function here and not duplicate this function?

gtmills (approved):
I am going to approve but would like @patelabhishek9893 to look at using a util function

PR #373, opened 2022-05-26
sensors: Add Accuracy property to PS input power

smccarney:
[SW551358](https://w3.rchland.ibm.com/projects/bestquest/?defect=SW551358): Add the Accuracy property to power supply input power sensors.

The current implementation is a temporary measure to meet the near-term
deadlines.  A long term implementation that obtains the accuracy
information from D-Bus will be delivered at a later date.

Tested:
* Ran validator
* Tested with PowerSubsystem URI
  * Verified no changes in output
* Tested with PowerSubsystem/PowerSupplies URI
  * Verified no changes in output
* Tested with PowerSupply URI
  * Verified no changes in output
* Tested with ThermalSubsystem URI
  * Verified no changes in output
* Tested with ThermalSubsystem/ThermalMetrics URI
  * Verified no changes in output
* Tested with Sensors URI
  * Verified no changes in output
* Tested with Sensors/sensor_name URI
  * Tested where sensor is a power supply input power sensor
    * Verified Accuracy property added with value of 1
    * Verified other properties are the same
  * Tested where sensor is a power supply input voltage sensor
    * Verified Accuracy property not added
    * Verified other properties are the same
  * Tested where sensor is not related to power supplies
    * Verified Accuracy property not added
    * Verified other properties are the same

Signed-off-by: Shawn McCarney <shawnmm@us.ibm.com>
Change-Id: I42b99582735a9273a6b80a66ce344c6332006747

### DIFF
--- a/redfish-core/lib/sensors.hpp
+++ b/redfish-core/lib/sensors.hpp
@@ -25,6 +25,7 @@
 #include <utils/json_utils.hpp>
 
 #include <cmath>
+#include <regex>
 #include <utility>
 #include <variant>
 
@@ -51,6 +52,9 @@ static constexpr std::string_view power = "Power";
 static constexpr std::string_view sensors = "Sensors";
 static constexpr std::string_view thermal = "Thermal";
 } // namespace node
+
+// Regular expression that matches power supply input power sensor names
+static const std::regex powerSupplyInputPowerPattern{"ps[0-9]+_input_power"};
 
 namespace dbus
 {
@@ -975,6 +979,13 @@ inline void objectInterfacesToJson(
         else
         {
             sensorJson["ReadingUnits"] = readingUnits;
+        }
+
+        // If this is a power supply input power sensor, set the Accuracy to 1%.
+        // All power supplies currently used on IBM systems have this accuracy.
+        if (std::regex_match(sensorName, sensors::powerSupplyInputPowerPattern))
+        {
+            sensorJson["Accuracy"] = 1;
         }
     }
     else if (sensorType == "temperature")


derekhoward55 (approved):

redfish-core/lib/sensors.hpp
```diff
@@ -52,6 +53,9 @@ static constexpr std::string_view sensors = "Sensors";
 static constexpr std::string_view thermal = "Thermal";
 } // namespace node
 
+// Regular expression that matches power supply input power sensor names
+static const std::regex powerSupplyInputPowerPattern{"ps[0-9]+_input_power"};
```
gtmills, line 57: There has been some patches upstream in bmcweb to move away from regexes. e.g. https://gerrit.openbmc.org/c/openbmc/bmcweb/+/51094
Since this is a temporary patch.. Don't think I will enforce but just an FYI

gtmills:
We have discussed a more correct solution where we look at an Accuracy D-Bus interface for this value. Is there any ETA on that solution? Did you want this in now? Or is this the fallback to, if the correct solution doesn't make it in the next ~week?

redfish-core/lib/sensors.hpp
smccarney: OK thanks for the information.

In the referenced review, they replaced a very complex regex for URLs with a built-in library, which makes a ton of sense.  

Are they removing regex's in general though?  That seems strange.  In a case like the sensor name pattern above, a regex provides a very simple and powerful solution.  The equivalent std::string parsing solution would be longer and more error prone IMO.  But I'll abide by it for upstream code if that is what the community wants.

smccarney:
> We have discussed a more correct solution where we look at an Accuracy D-Bus interface for this value. Is there any ETA on that solution? Did you want this in now? Or is this the fallback to, if the correct solution doesn't make it in the next ~week?

I estimate that the more correct solution will take approximately 2 weeks.  Given our product schedule, I'm not sure I can complete that in time for the current release.  My hope was to merge this temporary solution for the current release and then work on the more correct solution for a future release/fix-pack.  I was hoping to implement that solution 100% upstream, which will take more time as I need the D-Bus interface merged upstream.

redfish-core/lib/sensors.hpp
gtmills: I am fine with this for now, since this whole regex is going away for a proper D-Bus <-> Redfish

gtmills:
@smccarney Can we add the defect here?

gtmills (approved):
I am approving with the assumption we will move to the correct solution as soon as possible

PR #374, opened 2022-05-27
Bump up the ConfigFile directory max limit

sunharis:
SW551763: When there is a 1000 partition configuration, HMC
is not able to install OS on all of them due to the
configFile directory size limit applied by BMC

The current limit is 10MB. HMC requires this to be
bumped up to 25MB.

This commit increases this upper limit

Tested by:
  Checked the save area upload works with this change

Change-Id: Idcde692de9758e8eca3e9444beb2a4a9ff2bc36d


sunharis:
Fixes SW551763

gtmills (approved):

PR #382, opened 2022-06-06
Refresh of PCIe Topology

patelabhishek9893:
https://w3.rchland.ibm.com/projects/bestquest/?defect=SW541518: 
bmcweb, when gets the request to refresh PCIe topology. It has to hold that request until PCIe topology doesn't get refreshed. 
This commit adds functionality, where bmcweb creates a watchdog (using timer) where timer waits for one second and then checks if PCIe Topology gets refreshed or not. If not, then wait for one more second and repeat this process eight times, and then the watchdog expires in between if it gets refreshed then unblock the request.

Tested code: it blocked requests until it not get refreshed.

### DIFF
--- a/redfish-core/lib/oem/ibm/pcie_topology_refresh.hpp
+++ b/redfish-core/lib/oem/ibm/pcie_topology_refresh.hpp
@@ -8,7 +8,9 @@
 
 namespace redfish
 {
-
+// Timer for PCIe Topology Refresh
+static std::unique_ptr<boost::asio::steady_timer> pcieTopologyRefreshTimer;
+static uint count = 0;
 /**
  * @brief Retrieves PCIe Topology Refresh properties over DBUS
  *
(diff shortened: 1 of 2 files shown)


redfish-core/lib/oem/ibm/pcie_topology_refresh.hpp
```diff
@@ -46,6 +50,82 @@ inline void
...
+{
+    if (ec)
+    {
+        BMCWEB_LOG_DEBUG << "steady_timer error " << ec;
+        messages::internalError(aResp->res);
+        pcieTopologyRefreshTimer = nullptr;
+        (*countPtr) = 0;
+        return;
+    }
+    //  This method can block incoming requests max for 10 seconds. So, each
```
gtmills, line 74: Update these comments?

redfish-core/lib/oem/ibm/pcie_topology_refresh.hpp
```diff
@@ -46,6 +50,82 @@ inline void
         "PCIeTopologyRefresh");
 }
 
+/**
+ * @brief PCIe Topology Refresh monitor. which block incoming request
+ *
+ * @param[in] t       pointer to steady timer which block call
```
gtmills, line 56: We try to avoid single char variable names

redfish-core/lib/oem/ibm/pcie_topology_refresh.hpp
```diff
@@ -46,6 +50,82 @@ inline void
...
+                t->expires_at(t->expiry() + boost::asio::chrono::seconds(1));
+                t->async_wait(
+                    [t, aResp, countPtr](const boost::system::error_code ec) {
+                        pcIeTopologyRefreshWatchdog(ec, t, aResp, countPtr);
+                    });
+            }
+            else
+            {
+                BMCWEB_LOG_DEBUG << "pcieRefreshValuePtr value refreshed";
+                // get set to false time expired.
```
gtmills, line 118: get set to false time expired? huh? Can you clarify?

redfish-core/lib/oem/ibm/pcie_topology_refresh.hpp
```diff
@@ -46,6 +48,85 @@ inline void
         "PCIeTopologyRefresh");
 }
 
+/**
+ * @brief PCIe Topology Refresh monitor. which block incoming request
+ *
+ * @param[in] timer       pointer to steady timer which block call
```
aahmed-2, line 54: Nit: which blocks further calls to refreshProperty? or blocks all further calls?

redfish-core/lib/oem/ibm/pcie_topology_refresh.hpp
```diff
@@ -46,6 +48,85 @@ inline void
...
+ * @param[in] aResp   Shared pointer for generating response message.
+ * @param[in] count   how many time this method get called.
+ *
+ * @return None.
+ */
+void pcIeTopologyRefreshWatchdog(
```
aahmed-2, line 60: nit: can we fix capitalization in pcIeTopologyRefreshWatchdog.

patelabhishek9893: ack

redfish-core/lib/oem/ibm/pcie_topology_refresh.hpp
```diff
@@ -55,18 +136,28 @@ inline void
...
                 return;
             }
+            count = 0;
+            pcieTopologyRefreshTimer =
+                std::make_unique<boost::asio::steady_timer>(*req.ioService);
+            pcieTopologyRefreshTimer->expires_after(std::chrono::seconds(1));
+            pcieTopologyRefreshTimer->async_wait(
+                [timer = pcieTopologyRefreshTimer.get(),
+                 aResp](const boost::system::error_code ec) {
+                    pcieTopologyRefreshWatchdog(ec, timer, aResp, &count);
```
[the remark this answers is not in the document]
patelabhishek9893: A0: yes. Because when the user sets that value true from there, it hast to start 8 second counter.
A1: if we get a second request then the first request gone get deleted. because it is unique_ptr.
A2: yes, second call set setPCIETopologyRefresh that is possible. but then the first user request has gone get deleted.

so this code is perfectly based on requirements, because if I am not wrong only bmcweb UI will make this refresh request. and then want to hold that request until refresh not happened.
for multiple users wants to use this feature then this code is not perfect. and I think we do not need to make this work for multiple users @gtmills please conform to that. 
if we like to make this available for multiple users then either we need more robust code or we have to move for the task(which do two replies).

aahmed-2 (approved):
mostly lgtm

gtmills (approved):
We should revisit this in a future release

PR #388, opened 2022-06-08
FabricAdapters: Add Status State and Health

gtmills:
Partially fix SW551681. This adds additional, pretty standard Redfish properties:
...
"Status": {
"Health": "OK",
"State": "Enabled"
}

These properties were already out on D-Bus some of the time. (@manojkiraneda could explain why we didn't have Status all the time.)
Did not do a default Status Health/State.

FabricAdapters isn't upstream today but we should get it in upstream and
this could follow.

Still need to figure out how to do LocationIndicatorActive (not in https://redfish.dmtf.org/schemas/FabricAdapter.v1_3_0.json so would have to be OEM)

Tested: Saw these properties and ran the validator.

### DIFF
--- a/redfish-core/lib/FabricAdapters.hpp
+++ b/redfish-core/lib/FabricAdapters.hpp
@@ -1,5 +1,7 @@
 #pragma once
 
+#include "led.hpp"
+
 #include <utils/collection.hpp>
 #include <utils/json_utils.hpp>
 #include <utils/name_utils.hpp>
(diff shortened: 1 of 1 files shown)


gtmills:
```
            else if (interface == "xyz.openbmc_project.Association.Definitions")
            {
                getLocationIndicatorActive(aResp, objPath);
            }
```
doesn't work because 
https://redfish.dmtf.org/schemas/FabricAdapter.v1_3_0.json doesn't have LocationIndicatorActive. We will have to implement as Oem. 
```
ERROR - LocationIndicatorActive not defined in schema FabricAdapter.v1_2_0 (check version, spelling and casing)
```

aahmed-2 (approved):
lgtm

PR #389, opened 2022-06-09
Link from PCIe Device to PCIe Slot

patelabhishek9893:
This commit allows us to associate the PCIe device with PCIe Slot.
PCIe Slot doesn't have an individual route path, which provides a unique
association with the PCIe device. As a solution, bmcweb includes a stander
route path with all PCIe Slot information, and based on Device naming
information users are able to determine which slot it is. 

Tested:
curl -k -H "X-Auth-Token: $bmc_token" -X GET https://${BMC_IP}/redfish/v1/Systems/system/PCIeDevices/motherboard_disk_backplane0_nvme2_dp0_drive2
{
"@odata.id": ["/redfish/v1/Systems/system/PCIeDevices/motherboard_disk_backplane0_nvme2_dp0_drive2"],
"Id": "motherboard_disk_backplane0_nvme2_dp0_drive2",
"Links": { "Oem": { "IBM": { "PCIeSlot": {
"@odata.id": ["/redfish/v1/Chassis/chassis/PCIeSlots"]
} } } }
...
...
}


redfish-core/lib/pcie.hpp
```diff
@@ -474,6 +474,9 @@ inline void requestRoutesSystemPCIeDevice(App& app)
                                  device},
                             {"Name", "PCIe Device"},
                             {"Id", device}};
+                        asyncResp->res.jsonValue["Links"]["Oem"]["IBM"]
+                                                ["PCIeSlot"]["@odata.id"] =
+                            ("/redfish/v1/Chassis/chassis/PCIeSlots");
```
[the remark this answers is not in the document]
aahmed-2: +1

redfish-core/lib/pcie.hpp
```diff
@@ -423,188 +423,193 @@ inline void requestRoutesSystemPCIeDeviceCollection(App& app)
...
-                                    *property;
-                            }
-                        }
+            asyncResp->res.jsonValue = {
+                {"@odata.type", "#PCIeDevice.v1_9_0.PCIeDevice"},
+                {"@odata.id",
+                 "/redfish/v1/Systems/system/PCIeDevices/" + device},
+                {"Name", "PCIe Device"},
+                {"Id", device}};
+            std::vector<std::string> splitText;
```
patelabhishek9893, line 468: @gtmills @aahmed-2  main code is 468 to 488 Line

this diff shows a long change because I need `std::string& path = std::get<1>(it->second);` this path on line 437 line. 
And that is why I created a new method because changing line gives more indent. that makes it hard to read

redfish-core/lib/pcie.hpp
```diff
@@ -423,188 +423,193 @@ inline void requestRoutesSystemPCIeDeviceCollection(App& app)
...
+            boost::split(splitText, path, [](char c) { return c == '/'; });
+            auto it = std::find_if(splitText.begin(), splitText.end(),
+                                   [](const std::string& data) {
+                                       return data.find("chassis") !=
```
aahmed-2, line 472: Would it be better to match on "Chassis" and then take the child? 
I know the mex drawer can have PCIeSlots as well and what happens if the name for the mex drawer doesn't have "chassis" substring in it?

gtmills (approved):
Should revisit later

PR #391, opened 2022-06-10
Map PCIeSlot path with PCIeDevice

patelabhishek9893:
This commit allows us to retrieve the PCIeSlot path from the PCIeDevice.
The old code was trying to retrieve the PCIe device name from the dbus path.
That functionality gets disabled because one of the patches updates the PCIe Device naming mechanism.

Tested:
```
curl -k -H 'X-Auth-Token: $bmc_token' -X GET https://${BMC_IP}/redfish/v1/Systems/system/PCIeDevices/motherboard_disk_backplane0_nvme2_dp0_drive2

{
"@odata.id": ["/redfish/v1/Systems/system/PCIeDevices/motherboard_disk_backplane0_nvme2_dp0_drive2"](https://rain118bmc.aus.stglabs.ibm.com:18080/redfish/v1/Systems/system/PCIeDevices/motherboard_disk_backplane0_nvme2_dp0_drive2),
"Id": "motherboard_disk_backplane0_nvme2_dp0_drive2",
"Manufacturer": "",
"Name": "NVMe Drive",
"Oem": {
"@odata.type": "#OemPCIeDevice.Oem",
"IBM": {
"@odata.type": "#OemPCIeDevice.IBM",
"LinkReset": false
}
},
...
...
}
```

### DIFF
--- a/redfish-core/lib/pcie.hpp
+++ b/redfish-core/lib/pcie.hpp
@@ -47,46 +47,37 @@ std::map<PCIeDevice, std::pair<ServiceName, PCIeDevicePath>>
     mapOfPcieDevicePaths;
 
 static inline void
-    getPCIeDeviceList(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
-                      const std::string& name)
+    retrievePCIeDeviceList(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp)
 {
-    auto getPCIeMapCallback = [asyncResp,
-                               name](const boost::system::error_code ec,
-                                     const MapperGetSubTreeResponse& subTree) {
-        if (ec)
-        {
-            BMCWEB_LOG_DEBUG << "no PCIe device paths found ec: "
-                             << ec.message();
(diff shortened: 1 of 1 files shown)


sbteeks (commented):
The change looks good. Please make sure all properties corresponding to the Device are listed. 
I also think that **mapOfPcieDevicePaths** is stale or not updated when a Device info is requested before making a Rest API call to the collection of devices. It is better if we update that map every time a request is made to display Device info.

gtmills (approved):

patelabhishek9893:
> The change looks good. Please make sure all properties corresponding to the Device are listed. I also think that **mapOfPcieDevicePaths** is stale or not updated when a Device info is requested before making a Rest API call to the collection of devices. It is better if we update that map every time a request is made to display Device info.

**mapOfPcieDevicePaths** has to get updated every time before we try to retrieve PCIe-related data or PCIe data into system data.

So this commits updates **mapOfPcieDevicePaths** every time. before calling the API information.

sbteeks (approved):

PR #401, opened 2022-06-17
Add LocationIndicatorActive for fabric adapters

lxwinspur:

### DIFF
--- a/redfish-core/lib/FabricAdapters.hpp
+++ b/redfish-core/lib/FabricAdapters.hpp
@@ -243,6 +243,74 @@ inline void
     }
 }
 
+inline void getOemLocationIndicatorActive(
+    const std::shared_ptr<bmcweb::AsyncResp>& aResp, const std::string& objPath)
+{
+    crow::connections::systemBus->async_method_call(
+        [objPath, aResp](const boost::system::error_code ec,
+                         const std::variant<std::vector<std::string>>& resp) {
+            if (ec)
+            {
+                BMCWEB_LOG_DEBUG << "DBUS response error, ec: " << ec.value();
+                return;
+            }
+
+            const std::vector<std::string>* endpoints =
+                std::get_if<std::vector<std::string>>(&resp);
+            if (endpoints == nullptr)
+            {
+                BMCWEB_LOG_DEBUG << "No endpoints, skipping get location "
+                                    "indicator active";
+                messages::internalError(aResp->res);
+                return;
+            }
+
+            for (const auto& endpoint : *endpoints)
+            {
+                bool isBlink = boost::ends_with(endpoint, "_blink");
+                auto callback = [aResp, endpoint,
+                                 isBlink](const std::string& serviceName) {
+                    crow::connections::systemBus->async_method_call(
+                        [aResp, isBlink](const boost::system::error_code ec,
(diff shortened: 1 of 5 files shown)


jinuthomas:
Was this tested using the curl commands to see if they work ?

gtmills:
@lxwinspur Does Redfish validator pass? Can you add more description?

redfish-core/lib/FabricAdapters.hpp
```diff
@@ -359,5 +360,58 @@ inline void requestRoutesFabricAdapters(App& app)
...
+
+    BMCWEB_ROUTE(app, "/redfish/v1/Systems/system/FabricAdapters/<str>/")
+        .privileges({{"Login"}})
+        .methods(boost::beast::http::verb::patch)(
+            [](const crow::Request& req,
+               const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
+               const std::string& fabricAdapter) {
+                std::optional<bool> locationIndicatorActive;
+                if (!json_util::readJson(req, asyncResp->res,
+                                         "LocationIndicatorActive",
```
gtmills, line 372: I don't see LocationIndicatorActive here https://redfish.dmtf.org/schemas/FabricAdapter.v1_3_0.json so this doesn't work and will fail the validator. 
I created a PR in Redfish to add this but won't be  out until Aug. If we need this before, this will need to be OEM 

                "Links": {
                    "$ref": "#/definitions/Links",
                    "description": "The links to other Resources that are related to this Resource.",
                    "longDescription": "The Redfish Specification-described Links Property shall contain links to Resources related to but not subordinate to this Resource."
                },
                "Location": {
                    "$ref": "http://redfish.dmtf.org/schemas/v1/Resource.json#/definitions/Location",
                    "description": "The location of the fabric adapter.",
                    "longDescription": "This property shall contain location information for the fabric adapter.",
                    "versionAdded": "v1_1_0"
                },
                "Manufacturer": {
                    "description": "The manufacturer or OEM of this fabric adapter.",
                    "longDescription": "This property shall contain a value that represents the manufacturer of the fabric adapter.",
                    "readonly": true,
                    "type": [
                        "string",
                        "null"
                    ]
                },

gtmills (changes requested):
This doesn't work since FabricAdapters doesn't have LocationIndicatorActive

jinuthomas:
jenkins run tests please

lxwinspur:
redfish validator passed

lxwinspur:
> > redfish validator passed
> 
> There is no way, because I don't this property in the fabric adapter schema

Sorry, Maybe I am not following, I added this property in the OemFabricAdapter schema

jinuthomas (approved):

gtmills:
There should a new https://redfish.dmtf.org/schemas/FabricAdapter.v1_3_0.json coming out in the coming weeks with this property, lets use that when it comes out. I don't want to add OEM since the new schema is so close to being out

gtmills (changes requested):
Let's wait for the new schema in the coming weeks

lxwinspur:
> Let's wait for the new schema in the coming weeks

Souds good, Ack

gtmills:
https://github.com/ibm-openbmc/bmcweb/pull/454 added this. This is no longer needed.

PR #407, opened 2022-06-29
Restrict host console to ConfigureManager

gtmills:
Part of [SW553027](https://w3.rchland.ibm.com/projects/bestquest/?defect=SW553027)

Currently, ReadOnly has access to the host console.
Restrict it to Admin or above.

This appears to add a half-second delay in starting the websocket (I assume this is for the dbus call). It is just noticeable but hard to measure (e.g. under a second).. Or I am going crazy. 

I followed code in https://github.com/ibm-openbmc/bmcweb/blob/8044381620b228f99b707ea2b444c7675e714f26/include/nbd_proxy.hpp#L417 

Tested: For a ReadOnly user I see 
```
websocket console0/ closed.
            code: 1000
            reason: Unathourized access
```
in the browser development tools console.

For Admin and Service the console works as normal as does the "Open in new tab". 
In the browser development tools console I see websocket console0/ opened and what I would expect flowing when booting the system. 

### DIFF
--- a/include/obmc_console.hpp
+++ b/include/obmc_console.hpp
@@ -6,6 +6,7 @@
 #include <boost/asio/local/stream_protocol.hpp>
 #include <boost/container/flat_map.hpp>
 #include <boost/container/flat_set.hpp>
+#include <privileges.hpp>
 #include <websocket.hpp>
 
 namespace crow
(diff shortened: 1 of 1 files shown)


include/obmc_console.hpp
```diff
@@ -117,23 +118,75 @@ inline void connectHandler(const boost::system::error_code& ec)
...
+
+                    const ::redfish::Privileges requiredPrivileges{
+                        "ConfigureManager"};
+
+                    if (!userPrivileges.isSupersetOf(requiredPrivileges))
+                    {
+                        BMCWEB_LOG_DEBUG
+                            << "User " << conn.getUserName()
+                            << " not authorized for host console connection";
+                        conn.close("Unathourized access");
```
joseph-reynolds, line 169: Not spelling errour: Unathourized -> Unathorized (remove the lowercase-u)

joseph-reynolds (approved):
Looks good to me modulo nit spelling error in response.

joseph-reynolds:
> This appears to add a half-second delay in starting the websocket (I assume this is for the dbus call). It is just noticeable but hard to measure (e.g. under a second).. Or I am going crazy.

You might be going crazy in addition to the noticeable delay in starting the host console.  In either case, that's okay because the latency to start a console session is not important because consoles tend to be opened infrequently and left open for long periods of time.

patelabhishek9893 (approved):

PR #408, opened 2022-06-29
ReadOnly users should not have ssh

joseph-reynolds:

Note that ReadOnly user created prior to this change will have the HostConsole AccountType privilege.  A BMC admin can PATCH a user's AccountTypes property to add or remove the HostConsole value.

Tested: Yes


gtmills:
@joseph-reynolds Can you add the defect number here?

joseph-reynolds:
This is a fix for SW553027.

joseph-reynolds:
There is a bug in line 2061: (roleId == "Administrator") should have been (roleId == "priv-admin").  SW554201 is tracking this bug.  This bug results in **all** new users not having SSH access.

PR #409, opened 2022-06-29
Disallow disabling the service user

joseph-reynolds:
This changes the PATCH /redfish/v1/AccountService/Accounts/service operation so the service user cannot be disabled.  Effectively, the way to disable service user login is to delete the ACF file.

Tested: Yes


joseph-reynolds:
Maybe this is not needed.  Meaning a BMC admin should be allowed to disable the service account login.

gtmills:
> Maybe this is not needed. Meaning a BMC admin should be allowed to disable the service account login.

I think a BMC admin should be able to disable the service account.

gtmills (changes requested):
What defect is this? I don't agree with this change here (I could be convinced) but overall this needs more discussion.

gtmills:
Can we close or is this still needed?

joseph-reynolds:
Closing.

PR #411, opened 2022-07-01
CELog requires service access

joseph-reynolds:
This changes the privilege required to access the CELog so only the service user can access (via the OemIBMPerformService privilege).

tested: compiles


spinler:
Will this break the HMC?  They need CE logs.

joseph-reynolds:
SW553847

joseph-reynolds:
> Will this break the HMC? They need CE logs.

Yes, HMC only has admin access and needs to get these logs.  Note that the entries contain no secrets, and can be given to customer/admins.

Closing this PR.

PR #417, opened 2022-07-19
Update schema to enable panel CM

SunnySrivastava1984:
The commit make required changes to enable  panel CM
by initiating the deletion of base and LCD panel FRU
and re-collection of base and LCD panel FRU as and
when initiated via GUI.

Validator was executed for this commit. No error was reported.

### DIFF
--- a/redfish-core/lib/assembly.hpp
+++ b/redfish-core/lib/assembly.hpp
@@ -299,9 +299,9 @@ inline void
                                  "xyz.openbmc_project.Inventory.Item")
                         {
                             crow::connections::systemBus->async_method_call(
-                                [aResp, assemblyIndex](
-                                    const boost::system::error_code ec,
-                                    const std::variant<bool>& property) {
+                                [aResp, assemblyIndex,
+                                 assembly](const boost::system::error_code ec,
+                                           const std::variant<bool>& property) {
                                     if (ec)
                                     {
                                         BMCWEB_LOG_DEBUG
@@ -309,6 +309,11 @@ inline void
                                         messages::internalError(aResp->res);
                                         return;
                                     }
+
+                                    std::string fru =
+                                        sdbusplus::message::object_path(
+                                            assembly)
+                                            .filename();
 
                                     nlohmann::json& assemblyArray =
                                         aResp->res.jsonValue["Assemblies"];
@@ -324,16 +329,26 @@ inline void
                                         messages::internalError(aResp->res);
                                         return;
                                     }
-                                    if (*value == false)
+
+                                    // Special handling for LCD and base
+                                    // panel CM.
+                                    if (fru == "panel0" || fru == "panel1")
                                     {
-                                        assemblyData["Status"]["State"] =
-                                            "Absent";
+                                        assemblyData["Oem"]["OpenBMC"]
+                                                    ["@odata.type"] =
+                                                        "#OemAssembly.v1_0_"
+                                                        "0.Assembly";
+
+                                        // if panel is not present,
+                                        // implies it is already
+                                        // removed or can be placed.
+                                        assemblyData["Oem"]["OpenBMC"]
+                                                    ["ReadyToRemove"] =
+                                                        !(*value);
                                     }
-                                    else
-                                    {
-                                        assemblyData["Status"]["State"] =
-                                            "Enabled";
-                                    }
+
+                                    assemblyData["Status"]["State"] =
+                                        (*value == true) ? "Enabled" : "Absent";
                                 },
                                 serviceName, assembly,
                                 "org.freedesktop.DBus.Properties", "Get",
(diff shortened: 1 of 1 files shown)


SunnySrivastava1984:
jenkins -> run tests please

geissonator:
jenkins run tests please

gtmills:
@SunnySrivastava1984 What defect here? Did you run the validator ?

gtmills:
Is this on both rainier and everest ?

SunnySrivastava1984:
> Is this on both rainier and everest ?

No it is just for Everest.

SunnySrivastava1984:
> @SunnySrivastava1984 What defect here? Did you run the validator ?

This is for 1030, I don't have a defect but I have epic defined for this work.
https://github.com/ibm-openbmc/dev/issues/3603

Yes, I ran validator. There was no error reported.

redfish-core/lib/assembly.hpp
```diff
@@ -540,6 +620,39 @@ inline void setAssemblylocationIndicators(
                         "xyz.openbmc_project.adcsensor.service", "replace");
                 }
             }
+            // Special handling for LCD and base panel CM.
```
gtmills, line 623: Can you explain more **why** we are doing this?

redfish-core/lib/assembly.hpp
```diff
@@ -540,6 +620,39 @@ inline void setAssemblylocationIndicators(
...
+            else if (sdbusplus::message::object_path(assembly).filename() ==
+                         "panel1" ||
+                     sdbusplus::message::object_path(assembly).filename() ==
+                         "panel0")
+            {
+                std::string action{};
+                if (readytoremove.value() == true)
+                {
+                    // Set action for removal of FRU.
+                    action = "deleteFRUVPD";
```
gtmills, line 633: Why does a readyToRemove mean a delete of the FRU VPD?

redfish-core/lib/assembly.hpp
```diff
@@ -106,6 +106,86 @@ inline void
                 std::array<const char*, 0>{});
         }
 
+        // Special handling for LCD and base panel CM.
+        if (sdbusplus::message::object_path(assembly).filename() == "panel1" ||
```
gtmills, line 110: And these are only found on everest? Hence on Rainier this ReadyToRemove won't be present ?

SunnySrivastava1984: The name "panel1" and "panel0" are specific to Everest systems and hence these check will fail for Rainier systems.

redfish-core/lib/assembly.hpp
SunnySrivastava1984: "readyToRemove" here act as a trigger point for removal of FRU. Once it is set to "true" in GUI implies the FRU should be prepared for removal. 
So in order to remove the FRU we need to unbind the driver attached to the FRU and set its presence to "false". So the api which collectively does steps required to remove the FRU is exposed as "deleteFRUVPD" via vpd-manager.

redfish-core/lib/assembly.hpp
SunnySrivastava1984: I assume you are asking to add more comment in the code here. Though it is pointed that it is required for CM functionality of panel. 
Is there something specific you want me to add?

redfish-core/lib/assembly.hpp
santoshpuranik, line 110: Is there a reason why this code that is setting the ReadyToRemove property isn't handled where we otherwise set the present state? line 407

redfish-core/lib/assembly.hpp
santoshpuranik: This is also the same pair of APIs that PHYP uses to indicate a CM operation for PCIe cards.

santoshpuranik (changes requested):

redfish-core/lib/assembly.hpp
SunnySrivastava1984: I put up this code here, below tod_battery CM to have similar functionalities together but overlooked that this can be moved to line 407 which will save us couple of DBus calls and also avoid redundant code as we are already getting present property there.

Updated the code accordingly.

redfish-core/lib/assembly.hpp
gtmills: Can we add a comment here explaining this, this is just unexpected

redfish-core/lib/assembly.hpp
gtmills: I asked for a comment below so as long as you do that I am good

gtmills:
There is some unexpected behavior here, please document it a little better (e.g. a commit msg more than 
`The commit make required changes to enable  panel CM by initiating the deletion of base and LCD panel FRU and re-collection of base and LCD panel FRU as and when initiated via GUI. ` and some additional comments in the code). Other than that, I am okay here.

redfish-core/lib/assembly.hpp
SunnySrivastava1984: Done

redfish-core/lib/assembly.hpp
SunnySrivastava1984: Done

redfish-core/lib/assembly.hpp
```diff
@@ -326,11 +331,42 @@ inline void
                                     }
                                     if (*value == false)
                                     {
+                                        // Special handling for LCD and base
```
santoshpuranik, line 334: Nit: These lines can simply be moved to a common place instead of repeating them in the if and else clauses

redfish-core/lib/assembly.hpp
```diff
@@ -540,6 +576,49 @@ inline void setAssemblylocationIndicators(
...
             }
+
+            // Special handling for LCD and base panel. This is required to
+            // support concurrent maintenance for base and LCD panel.
+            else if (sdbusplus::message::object_path(assembly).filename() ==
+                         "panel0" ||
+                     sdbusplus::message::object_path(assembly).filename() ==
+                         "panel1")
+            {
+                std::string action{};
```
santoshpuranik, line 587: Nit: this can be written more succinctly as:

```
std::string action = (readytoremove.value()) ? : ...;

santoshpuranik (changes requested):
Just a couple of nits

redfish-core/lib/assembly.hpp
SunnySrivastava1984: We can, but Gunnar is expecting detailed comments describing the behaviour in each case. Adding all those comments on top of one line, IMO will make it more complex to understand.
Also I am not sure if it will benefit from execution point of view.
Kept it so to have better readability. 
Let me know you say on this.

redfish-core/lib/assembly.hpp
SunnySrivastava1984: Just the comment?

santoshpuranik (approved):

gtmills (approved):

PR #424, opened 2022-07-28
Resource Dump: Fix dump input argument checks

raviteja-b:
Currently resource dump argument  parsing does not have
argument size checks so it tries to access out of bound index and crashes
when user does not pass last argument.

This commit fixes this issue by adding input argument size checks

Tested By:
Tested resource dumps with differnt resource strings
"OEMDiagnosticDataType":"Resource_vsp -globalfr"
"OEMDiagnosticDataType":"Resource_lp 2_pwd"

### DIFF
--- a/redfish-core/lib/log_services.hpp
+++ b/redfish-core/lib/log_services.hpp
@@ -1098,12 +1098,19 @@ inline void createDump(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
 
             dumpPath = "/xyz/openbmc_project/dump/resource";
 
-            createDumpParams.emplace_back(
-                std::make_pair("com.ibm.Dump.Create.CreateParameters.VSPString",
-                               resourceDumpParams[1]));
-            createDumpParams.emplace_back(
-                std::make_pair("com.ibm.Dump.Create.CreateParameters.Password",
-                               resourceDumpParams[2]));
+            if (resourceDumpParams.size() >= 2)
+            {
+                createDumpParams.emplace_back(std::make_pair(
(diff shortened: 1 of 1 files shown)


geissonator:
CI failure is valid (clang formatting related)

sunharis (approved):

gtmills:
@raviteja-b What is the defect here?

redfish-core/lib/log_services.hpp
```diff
@@ -1098,12 +1098,19 @@ inline void createDump(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
 
             dumpPath = "/xyz/openbmc_project/dump/resource";
 
-            createDumpParams.emplace_back(
-                std::make_pair("com.ibm.Dump.Create.CreateParameters.VSPString",
-                               resourceDumpParams[1]));
-            createDumpParams.emplace_back(
-                std::make_pair("com.ibm.Dump.Create.CreateParameters.Password",
-                               resourceDumpParams[2]));
+            if (resourceDumpParams.size() >= 2)
```
gtmills, line 1101: You shouldn't throw an error if the size is less than two? or is there valid resource dumps where we don't set this?

raviteja-b:
This PR fixes https://w3.rchland.ibm.com/projects/bestquest/?defect=SW554891

redfish-core/lib/log_services.hpp
raviteja-b: yes, I think there can be.

raviteja-b: there is a check added above for size >3

gtmills (approved):

PR #434, opened 2022-09-12
Reset fault LED after CM

SunnySrivastava1984:
The commit resets fault LED related to tod_battery after a successful replacement.

### DIFF
--- a/redfish-core/lib/assembly.hpp
+++ b/redfish-core/lib/assembly.hpp
@@ -367,7 +367,62 @@ inline void
 }
 
 /**
- * @brief Set location indicator for the assemblies associated to given chassis
+ * @brief API to reset fault LED.
+ * After concurrent maintenance has been performed successfully, fault LED needs
+ * to be reset.
+ * @param[in] asyncResp - Shared pointer for asynchronous calls.
+ * @param[in] assembly - Assembly object path.
+ */
+inline void
+    resetFaultLEDAfterCM(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
+                         const std::string& assembly)
+{
+    crow::connections::systemBus->async_method_call(
+        [asyncResp,
+         assembly](const boost::system::error_code ec,
+                   const std::variant<std::vector<std::string>>& resp) {
+            if (ec)
+            {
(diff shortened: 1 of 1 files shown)


redfish-core/lib/assembly.hpp
```diff
@@ -534,6 +534,26 @@ inline void setAssemblylocationIndicators(
...
+                                    if (ec2)
+                                    {
+                                        BMCWEB_LOG_ERROR
+                                            << "Failed to reset fault LED"
+                                            << ec2;
+                                        messages::internalError(asyncResp->res);
+                                        return;
+                                    }
+                                },
+                                "xyz.openbmc_project.LED.GroupManager",
```
santoshpuranik, line 551: We should use D-Bus associations to get the group object. I think you can use something similar to setLocationIndicatorActive in led.hpp and get it to look for the fault association instead.

santoshpuranik (changes requested):

redfish-core/lib/assembly.hpp
SunnySrivastava1984: Okay.

santoshpuranik (approved):

redfish-core/lib/assembly.hpp
```diff
@@ -367,7 +367,62 @@ inline void
...
+                    [asyncResp, endpoint](const boost::system::error_code ec2) {
+                        if (ec2)
+                        {
+                            BMCWEB_LOG_ERROR << "Failed to reset fault LED"
+                                             << ec2;
+                            messages::internalError(asyncResp->res);
+                            return;
+                        }
+                    },
+                    "xyz.openbmc_project.LED.GroupManager", endpoint,
```
gtmills, line 412: Doesn't this set all fault leds, what if you have a fault led unrelated to your CM?

jinuthomas (approved):

redfish-core/lib/assembly.hpp
gtmills: ack.

gtmills (approved):

SunnySrivastava1984:
Will fix - SW553263

PR #437, opened 2022-09-27
SW556104: Part 1: Bump the max password length to 64

gtmills:
This commit also needs to go in 1030. 

As mentioned in SW556104, Redfish accurately accepts passwords greater than 20 chars in almost all of our cases. 

The 20 char limit displayed in the API comes from IPMI, which is enabled upstream.

Downstream we have IPMI disabled by default, so up this password max length to 64.

Due to that, at this time, this is a downstream only change.

The GUI will then "Read the ManagerAccount object via GET /redfish/v1/AccountService/Accounts/<USER> and look at the AccountTypes property. If AccountTypes contains the "IPMI" AccountType, know that MaxPasswordLength is effectively 20 for this user.". This GUI change is part 2 of SW556104.

64 was chosen because it gives sufficient randomness for all practical purposes.

It makes sense the limit reflects the default and most users will not have IPMI enabled.
This limit is on the Account Service, not the account, so we can't change this per account. :/ 

Tested: None. Code review only.

### DIFF
--- a/redfish-core/lib/account_service.hpp
+++ b/redfish-core/lib/account_service.hpp
@@ -1789,7 +1789,7 @@ inline void requestAccountServiceRoutes(App& app)
                 {"Name", "Account Service"},
                 {"Description", "Account Service"},
                 {"ServiceEnabled", true},
-                {"MaxPasswordLength", 20},
+                {"MaxPasswordLength", 64},
                 {"Accounts",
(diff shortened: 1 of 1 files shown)


joseph-reynolds (approved):
Looks good to me.  With the provision that you have to just know that IPMI user accounts are limited to 20 characters passwords.

PR #439, opened 2022-10-03
Multipart parser fix

sunharis:
This fixes [SW557058]

Tested by:
 1. Upload multiple config files 
 2. Upload single config file

### DIFF
--- a/include/login_routes.hpp
+++ b/include/login_routes.hpp
@@ -25,6 +25,7 @@ inline void requestRoutes(App& app)
             boost::beast::http::verb::
                 post)([](const crow::Request& req,
                          const std::shared_ptr<bmcweb::AsyncResp>& asyncResp) {
+            MultipartParser parser;
             std::string_view contentType = req.getHeaderValue("content-type");
             std::string_view username;
             std::string_view password;
@@ -126,13 +127,12 @@ inline void requestRoutes(App& app)
             else if (boost::starts_with(contentType, "multipart/form-data"))
             {
                 looksLikePhosphorRest = true;
-                std::error_code ec;
-                MultipartParser parser(req, ec);
-                if (ec)
+                ParserError ec = parser.parse(req);
+                if (ec != ParserError::PARSER_SUCCESS)
                 {
                     // handle error
-                    BMCWEB_LOG_ERROR << "MIME parse failed, ec : " << ec.value()
-                                     << " , ec message : " << ec.message();
+                    BMCWEB_LOG_ERROR << "MIME parse failed, ec : "
+                                     << static_cast<int>(ec);
                     asyncResp->res.result(
                         boost::beast::http::status::bad_request);
                     return;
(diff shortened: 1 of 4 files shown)


raviteja-b:
CI failure expected?

sunharis:
> CI failure expected?

No it should not fail. Seems like some other dependency is missing. There was a lot of diff in the code upstream and what we have at 1020 branch, i will check this further. @gtmills please check if you can get some hint.

sunharis:
I think this is clang format error. I will push the fix

sunharis:
This PR pulls changes from https://gerrit.openbmc.org/c/openbmc/bmcweb/+/56796 and https://gerrit.openbmc.org/c/openbmc/bmcweb/+/56868

include/login_routes.hpp
```diff
@@ -25,6 +25,7 @@ inline void requestRoutes(App& app)
             boost::beast::http::verb::
                 post)([](const crow::Request& req,
                          const std::shared_ptr<bmcweb::AsyncResp>& asyncResp) {
+            MultipartParser parser;
```
gtmills, line 28: This is https://gerrit.openbmc.org/c/openbmc/bmcweb/+/56868/7/include/login_routes.hpp

sunharis:
> How did you test ?

Updated the tests done. All good paths are tested again for the config file upload.
I have not specifically tested for the vulnerability; as i could not recreate this memory corruption on the BMC. I have updated the functional test team to perform the related tests.

joseph-reynolds:
File login_routes.hpp should also include this fix: https://github.com/openbmc/bmcweb/commit/6071963922ae9511b822feb8908a20141bb7d821

joseph-reynolds:
This picks up the new implementation for the multi-part parser.  I think that is acceptable because that parser is an isolated function which shouldn't affect any other code.
NOTE TO TESTERS: This changes the HTTP multi-part parser,so we should regression test all functions which use that function.

include/login_routes.hpp
joseph-reynolds: Thanks Gunnar.  This is the same code I mentioned below as https://github.com/openbmc/bmcweb/commit/6071963922ae9511b822feb8908a20141bb7d821.  It is an easy fix.

sunharis, line 28: @joseph-reynolds I have the https://github.com/openbmc/bmcweb/commit/6071963922ae9511b822feb8908a20141bb7d821 change manually added here since this is a single line change - to avoid getting in any other possible merge conflict errors into our stream.

gtmills:
> The code changes look okay. I would approve this as-is.
We also need the other fix in https://github.com/openbmc/bmcweb/commit/6071963922ae9511b822feb8908a20141bb7d821 and I think we should add it to this commit; but I would be okay putting it into another commit. 

Is already in this.

gtmills (approved):

joseph-reynolds (approved):
Looks good to me.  Approved!

PR #447, opened 2022-10-17
Expand Success Case For Audit Logging

zamiseck:
This change ensures that all "success" responses are catched and audit logs the appropriate message accordingly. Prior to this change, there was a response that was missed because of the method used to determine an valid and invalid request, but this ensures it is logged as well.

### DIFF
--- a/http/http_connection.hpp
+++ b/http/http_connection.hpp
@@ -443,7 +443,8 @@ class Connection :
             (req->method() == boost::beast::http::verb::patch) ||
             (req->method() == boost::beast::http::verb::delete_))
         {
-            if (res.result() == boost::beast::http::status::ok)
+            // Look for good return codes and if so we know the operation passed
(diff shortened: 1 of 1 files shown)


http/http_connection.hpp
```diff
@@ -443,7 +443,7 @@ class Connection :
             (req->method() == boost::beast::http::verb::patch) ||
             (req->method() == boost::beast::http::verb::delete_))
         {
-            if (res.result() == boost::beast::http::status::ok)
+            if ((res.resultInt() >= 200) && (res.resultInt() < 300))
```
gtmills, line 446: A comment why here, might make this more readable. 

//  Look for good return codes and if so we know the operation passed 
            if ((res.resultInt() >= 200) && (res.resultInt() < 300))

gtmills (approved):

PR #457, opened 2022-10-27
VMI CSR: increase signCSR request timeout to 30s

isha309:
When BMC is highly loaded ,  D-bus requests took long time causing VMI client certificate requests to timeout.
This commit increases VMI signCSR timeout to 30 sec
This PR fixes SW557842

### DIFF
--- a/include/ibm/management_console_rest.hpp
+++ b/include/ibm/management_console_rest.hpp
@@ -986,7 +986,7 @@ void handleCsrRequest(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
         std::make_shared<boost::asio::steady_timer>(
             crow::connections::systemBus->get_io_context());
 
-    timeout->expires_after(std::chrono::seconds(10));
+    timeout->expires_after(std::chrono::seconds(30));
     crow::connections::systemBus->async_method_call(
         [asyncResp, timeout](const boost::system::error_code ec,
                              sdbusplus::message::message& m) {


raviteja-b (approved):

sunharis (approved):

gtmills:
jenkins run tests please

gtmills:
For next time, please use a signed-off. 
Do we plan to look at why there is so many D-bus requests?

raviteja-b:
> For next time, please use a signed-off.
> Do we plan to look at why there is so many D-bus requests?

I am not sure how to Debug these D-bus issues further, but for sure this issue is not related to vmi certificate  feature.

raviteja-b:
@gtmills @rfrandse can you please merge this PR today?? this is 1030 blocker

include/ibm/management_console_rest.hpp
```diff
@@ -986,7 +986,7 @@ void handleCsrRequest(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
         std::make_shared<boost::asio::steady_timer>(
             crow::connections::systemBus->get_io_context());
 
-    timeout->expires_after(std::chrono::seconds(10));
+    timeout->expires_after(std::chrono::seconds(30));
```
gtmills, line 989: I don't like that we are just bumping a reasonable timeout up to something high without understanding more about how this timeout hit.

raviteja-b:
> Can you continue to look at the performance here? I.e. even if we merge this, we continue to investigate and don't just be done? :)

I think we need to discuss with Andrew and ask if someone can check on D-bus performance..

include/ibm/management_console_rest.hpp
raviteja-b: this timeout basically waits for pldm response for the given CSR certificate..
As D-bus was overloaded  and it was taking more time than 10 secs, hit this 10 sec timeout.
so increasing this timeout to 30 secs which is reasonable and considering worst case.
In normal cases, these operations will be completed with in a second, only when D-bus overloaded this timeout will be used to server request.

raviteja-b:
MF approved on SW557842

sunharis:
> Can you continue to look at the performance here? I.e. even if we merge this, we continue to investigate and don't just be done? :)

This root cause investigation is not in the scope of signCSR usecase. Investigation needs to be continued as a separate performance improvement item for either bmcweb/dbus components led by the respective functional experts.

gtmills:
I am good here. 
MF approved. I ask, though, to please keep going a bit on the performance side.

PR #460, opened 2022-10-28
Remove health.hpp calls where unneeded (#237)

gtmills:
Fixes https://github.com/ibm-openbmc/dev/issues/3459

health.hpp is based on https://github.com/openbmc/docs/blob/master/designs/redfish-health-rolllup.md

It uses a xyz.openbmc_project.Inventory.Item.Global to identify resources that have a HealthRollup and then uses "critical" or "warning" associations to set that HealthRollup. It is largely unused by us downstream. The only place we are using healthRollup is chassis. This change removes health.hpp expect from chassis.hpp. Instead we tend to us Functional to set the Health and haven't been setting HealthRollup.

We have recently hit some performance problems and Andrew noted "Over a 13 second window at runtime on an Everest, there was 1.8 MB/s traffic over D-Bus. The biggest packet size (mapper←bmcweb) was 500KB."

Atleast some of these 500KB packets were theorized to be health.hpp's "busctl call xyz.openbmc_project.ObjectMapper / org.freedesktop.DBus.ObjectManager GetManagedObjects". This inefficient query should be addressed upstream, for the time being though let's just remove health.hpp except where it is needed.

Commit has been cherry-picked to 1040.

Tested: loaded on a system. 

For example, no longer see system healthrollup
```
"Status": {
"Health": "OK",
"State": "Disabled"
},
```

Other resources impacted (storage, manager, etc) look good too.

### DIFF
--- a/redfish-core/lib/ethernet.hpp
+++ b/redfish-core/lib/ethernet.hpp
@@ -1589,33 +1589,11 @@ inline void parseInterfaceData(
     const boost::container::flat_set<IPv4AddressData>& ipv4Data,
     const boost::container::flat_set<IPv6AddressData>& ipv6Data)
 {
-    constexpr const std::array<const char*, 1> inventoryForEthernet = {
-        "xyz.openbmc_project.Inventory.Item.Ethernet"};
-
     nlohmann::json& jsonResponse = asyncResp->res.jsonValue;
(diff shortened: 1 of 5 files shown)


gtmills:
jenkins run tests please

sbteeks (approved):

PR #421, opened 2022-07-27
New property to always AllowUnauthACFUpload

joseph-reynolds:
This adds a new BMC security setting as a Redfish OEM property: URI /redfish/v1/AccountService/Accounts/service property Oem.IBM.ACF.AllowUnauthACFUpload can be set by the admin.  When true, any network agent is allowed to upload an ACF file to the BMC.  The default value is false.  This capability is included in the existing URI /redfish/v1 property Oem.IBM.ACFWindowActive.  For example, AllowUnauthACFUpload=true implies ACFWindowActive=true.

Tested: via Redfish scenarios
1. Ensure AllowUnauthACFUpload=false by default.
2. Ensure a non admin user is not allowed to GET or PATCH AllowUnauthACFUpload.
3. Ensure the admin can PATCH AllowUnauthACFUpload true and false, and GET the correct values, and these values survive BMC reboot.
4. With (OpPanel fn 74 is not active) and AllowUnauthACFUpload=false, ensure ACFWindowActive=false.  And unauth ACF upload is not allowed.
5. With (OpPanel fn 74 is not active) and AllowUnauthACFUpload=true, ensure ACFWindowActive=true.  And unauth ACF upload is allowed.

PR #466, opened 2022-11-03
New property to always AllowUnauthACFUpload

joseph-reynolds:
Background: The Access Control File (ACF) service access is a 2-factor authentication scheme intended for BMC administrators to determine when authorized service and support representatives (SSRs) are allowed to access their BMC but not allow the administrator to have access. In this scheme, only the SSR can create valid ACFs,m which they then give it to the BMC admin for upload to the BMC.  Doing so authorizes the SSR to login to the BMC's service account via an unique password.

A problem with this scheme is some BMC admins lose their password.  To solve this, Operator Panel function 74 opens a 30 minute time window where anyone is allowed to upload a valid ACF to the BMC.

A problem with that solution is that in some installations there is difficulty gaining physical access to the operator panel.  To solve that, there is a new security setting to open up the ACF upload function...

This adds a new BMC security setting as a Redfish OEM property: URI /redfish/v1/AccountService/Accounts/service property Oem.IBM.ACF.AllowUnauthACFUpload can be set by the admin.  When true, any network agent is allowed to upload an ACF file to the BMC. The default value is false.  This capability is included in the existing URI /redfish/v1 property Oem.IBM.ACFWindowActive.  For example, when AllowUnauthACFUpload=true, ACFWindowActive is true.

Tested: via Redfish scenarios
1. Ensure AllowUnauthACFUpload=false by default.
2. Ensure a non admin user is not allowed to GET or PATCH AllowUnauthACFUpload.
3. Ensure the admin can PATCH AllowUnauthACFUpload true and false, and GET the correct values, and these values survive BMC reboot.
4. With (OpPanel fn 74 is not active) and AllowUnauthACFUpload=false, ensure ACFWindowActive=false.  And unauth ACF upload is not allowed.
5. With (OpPanel fn 74 is not active) and AllowUnauthACFUpload=true,

Cheatsheet:
```
busctl set-property com.ibm.PanelApp /com/ibm/panel_app com.ibm.panel ACFWindowActive b true
curl -k -H "X-Auth-Token: $TOKEN" -X PATCH -d '{"Oem":{"IBM":{"ACF":{"AllowUnauthACFUpload": true}}}}' https://${bmc}/redfish/v1/AccountService/Accounts/service
curl -k -H "X-Auth-Token: $TOKEN" -X GET https://${bmc}/redfish/v1
curl -k -X PATCH -d '{"Oem":{"IBM":{"ACF":{"ACFFile":"'${acf}'"}}}}' https://${bmc}/redfish/v1/AccountService/Accounts/service

```

### DIFF
--- a/redfish-core/lib/account_service.hpp
+++ b/redfish-core/lib/account_service.hpp
@@ -1077,6 +1077,29 @@ inline void
         std::variant<std::string>(userNameAttribute));
 }
 
+inline void setPropertyAllowUnauthACFUpload(
+    const std::shared_ptr<bmcweb::AsyncResp>& asyncResp, bool allow)
+{
+    crow::connections::systemBus->async_method_call(
+        [asyncResp, allow](const boost::system::error_code ec) {
+            if (ec)
+            {
+                BMCWEB_LOG_ERROR << "D-Bus responses error: " << ec;
+                messages::internalError(asyncResp->res);
+                return;
+            }
+            messages::success(asyncResp->res);
+            return;
+        },
+        "xyz.openbmc_project.Settings",
+        "/xyz/openbmc_project/ibmacf/allow_unauth_upload",
+        "org.freedesktop.DBus.Properties",
+        "Set",
+        "xyz.openbmc_project.Object.Enable",
+        "Enabled",
+        std::variant<bool>(allow));
+}
+
 inline void getAcfProperties(
     const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
     const std::tuple<std::vector<uint8_t>, bool, std::string>& messageFDbus)
@@ -1175,6 +1198,26 @@ inline void getAcfProperties(
     }
     asyncResp->res.jsonValue["Oem"]["IBM"]["ACF"]["ACFInstalled"] =
         acfInstalled;
+
+    crow::connections::systemBus->async_method_call(
+        [asyncResp](const boost::system::error_code ec, 
+                    const std::variant<bool>& retval) {
+            if (ec)
+            {
+                BMCWEB_LOG_ERROR << "D-Bus responses error: " << ec;
+                messages::internalError(asyncResp->res);
+                return;
+            }
+            const bool *allowed = std::get_if<bool>(&retval);
+            asyncResp->res.jsonValue["Oem"]["IBM"]["ACF"]
+                ["AllowUnauthACFUpload"] = *allowed;
+        },
+        "xyz.openbmc_project.Settings",
+        "/xyz/openbmc_project/ibmacf/allow_unauth_upload",
+        "org.freedesktop.DBus.Properties",
+        "Get",
+        "xyz.openbmc_project.Object.Enable",
+        "Enabled");
 }
 
 /**
@@ -1667,10 +1710,15 @@ inline void uploadACF(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
         "InstallACF", decodedAcf);
 }
 
+// This is called when someone either is not authenticated or is not
+// authorized to upload an ACF, and they are trying to upload an ACF;
+// in this condition, uploading an ACF is allowed when
+// AllowUnauthACFUpload is true.
 inline void triggerUnauthenticatedACFUpload(
     const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
     std::optional<nlohmann::json> oem)
 {
+    std::vector<uint8_t> decodedAcf;
     std::optional<nlohmann::json> ibm;
     if (!redfish::json_util::readJson(*oem, asyncResp->res, "IBM", ibm))
     {
@@ -1691,7 +1739,6 @@ inline void triggerUnauthenticatedACFUpload(
 
         if (acf)
         {
-            std::vector<uint8_t> decodedAcf;
             std::optional<std::string> acfFile;
             if (acf.value().contains("ACFFile") &&
                 acf.value()["ACFFile"] == nullptr)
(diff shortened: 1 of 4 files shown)


joseph-reynolds:
This is the fix for SW553048 for FW1030.
This has a pre-req on https://github.ibm.com/openbmc/openbmc/pull/3000 for the underlying new D-Bus object.
This was ported (git cherry-pick and squashed) from the FW1020 version https://github.com/ibm-openbmc/bmcweb/pull/421

gtmills:
After it passes CI, I think I am good here. Just a cherry pick from 1020 branch

joseph-reynolds:
jenkins run tests please

redfish-core/lib/account_service.hpp
```diff
@@ -1175,6 +1196,25 @@ inline void getAcfProperties(
...
+    crow::connections::systemBus->async_method_call(
+        [asyncResp](const boost::system::error_code ec,
+                    const std::variant<bool>& retval) {
+            if (ec)
+            {
+                BMCWEB_LOG_ERROR << "D-Bus responses error: " << ec;
+                messages::internalError(asyncResp->res);
+                return;
+            }
+            const bool* allowed = std::get_if<bool>(&retval);
```
gtmills, line 1209: Please check if this is nullptr like 1790

            if (allowUnauthACFUpload == nullptr)
            {
                BMCWEB_LOG_ERROR << "nullptr for allow_unauth_upload";
                return;
            }

redfish-core/lib/service_root.hpp
```diff
@@ -26,30 +26,71 @@ namespace redfish
...
-                BMCWEB_LOG_ERROR << "nullptr for ACFWindowActive";
+                BMCWEB_LOG_ERROR << "nullptr for allow_unauth_upload";
                 messages::internalError(asyncResp->res);
                 return;
             }
-            asyncResp->res.jsonValue["Oem"]["IBM"]["ACFWindowActive"] =
-                *isACFWindowActive;
+            if (*allowUnauthACFUpload == true)
+            {
+                asyncResp->res.jsonValue["Oem"]["IBM"]["ACFWindowActive"] =
```
gtmills, line 51: Shouldn't this be the new property AllowUnauthACFUpload ?

redfish-core/lib/service_root.hpp
```diff
@@ -26,30 +26,71 @@ namespace redfish
...
+                return;
+            }
+
+            // ACFWindowActive = true also when D-Bus property ACFWindowActive
+            // is true (OpPanel function 74).  Otherwise, the Window is not
+            // active.
+            crow::connections::systemBus->async_method_call(
+                [asyncResp](const boost::system::error_code ec,
+                            const std::variant<bool>& retVal) {
+                    bool isActive;
```
gtmills, line 62: I think upstream this would be wrote more like (I'm fine with this.) 

                   if (ec)
                    {
                        BMCWEB_LOG_ERROR
                            << "Failed to read ACFWindowActive property";
                        // The Panel app doesn't run on simulated systems.
                       asyncResp->res.jsonValue["Oem"]["IBM"]["ACFWindowActive"] = false;
                       return;
                    }
                    const bool* isACFWindowActive =
                            std::get_if<bool>(&retVal);
                        if (isACFWindowActive == nullptr)
                        {
                            BMCWEB_LOG_ERROR << "nullptr for ACFWindowActive";
                            return;
                        }
                         asyncResp->res.jsonValue["Oem"]["IBM"]["ACFWindowActive"] = *isACFWindowActive;

redfish-core/lib/service_root.hpp
```diff
@@ -26,30 +26,71 @@ namespace redfish
...
                 messages::internalError(asyncResp->res);
                 return;
             }
-            asyncResp->res.jsonValue["Oem"]["IBM"]["ACFWindowActive"] =
-                *isACFWindowActive;
+            if (*allowUnauthACFUpload == true)
+            {
+                asyncResp->res.jsonValue["Oem"]["IBM"]["ACFWindowActive"] =
+                    *allowUnauthACFUpload;
+                return;
```
gtmills, line 53: Why return here if allowUnauthACFUpload  is true? 
I guess I would think ACFWindowActive and AllowUnauthACFUpload  are both always shown. 
If AllowUnauthACFUpload   is false, then ACFWindowActive  is false but I would still think in all cases, you would show both

redfish-core/lib/account_service.hpp
joseph-reynolds: Good catch!  I'll add the check.

redfish-core/lib/service_root.hpp
joseph-reynolds: No.  I'll change the comments to make this more clear:
> Redfish property ACFWindowActive=true when either of these is true:
> - D-Bus property allow_unauth_upload.  (This is aka the Redfish property AllowUnauthACFUpload which the BMC admin can control.)
> - D-Bus property ACFWindowActive.  (This is aka the Redfish property ACFWindowActive under /redfish/v1/AccountService/Accounts/service property Oem.IBM.ACF.  The value is provided by the PanelApp and is true when panel function 74 is active.)
>
> Check D-Bus property allow_unauth_upload first.

redfish-core/lib/service_root.hpp
joseph-reynolds: As I mentioned in the comment above, this URL only shows the Redfish property ACFWindowActive.
This value is computed from D-Bus property allow_unauth_upload and D-Bus property ACFWindowActive.
But D-Bus property allow_unauth_upload is not shown here, only under /redfish/v1/AccountService/Accounts/service property Oem.IBM.ACF

redfish-core/lib/service_root.hpp
joseph-reynolds: Looks good.  I'll change the code.

gtmills (approved):

redfish-core/lib/account_service.hpp
```diff
@@ -2401,209 +2487,220 @@ inline void requestAccountServiceRoutes(App& app)
...
+                std::optional<std::string> newUserName;
+                std::optional<std::string> password;
+                std::optional<bool> enabled;
+                std::optional<std::string> roleId;
+                std::optional<bool> locked;
+                std::optional<nlohmann::json> oem;
+                std::optional<std::vector<std::string>> accountType;
+                bool isUserItself = false;
+
+                if (!json_util::readJson(req, asyncResp->res, "UserName",
```
sbteeks, line 2503: readJsonPatch() instaed of readJson()?
ex: https://grok.openbmc.org/xref/openbmc/bmcweb/redfish-core/lib/account_service.hpp?r=0ec8b83d#1425

joseph-reynolds: Good catch!  The code in this branch had not started using the readJsonPatch function, so we will leave it as-is for the fixpack.  We'll change to the code to use readJsonPatch when we rebase for the next release.

PR #476, opened 2022-11-30
Set Functional property after battery CM

spinler:

This does two things:

Defect PE00DX9W.  For 1030.10

### DIFF
--- a/redfish-core/lib/assembly.hpp
+++ b/redfish-core/lib/assembly.hpp
@@ -435,6 +435,95 @@ inline void
         "xyz.openbmc_project.Association", "endpoints");
 }
 
+void startOrStopADCSensor(const bool start,
+                          const std::shared_ptr<bmcweb::AsyncResp>& asyncResp)
+{
+    std::string method{"StartUnit"};
+    if (!start)
+    {
+        method = "StopUnit";
+    }
+
+    crow::connections::systemBus->async_method_call(
+        [asyncResp](const boost::system::error_code ec) {
+            if (ec)
+            {
+                BMCWEB_LOG_ERROR << "Failed to start or stop ADCSensor:" << ec;
+                messages::internalError(asyncResp->res);
+                return;
+            }
+            messages::success(asyncResp->res);
+        },
+        "org.freedesktop.systemd1", "/org/freedesktop/systemd1",
+        "org.freedesktop.systemd1.Manager", method,
+        "xyz.openbmc_project.adcsensor.service", "replace");
+}
+
+void doBatteryCM(const std::string& assembly, const bool readyToRemove,
+                 const std::shared_ptr<bmcweb::AsyncResp>& asyncResp)
+{
+    if (readyToRemove)
+    {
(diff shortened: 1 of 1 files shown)


redfish-core/lib/assembly.hpp
```diff
@@ -435,6 +435,96 @@ inline void
         "xyz.openbmc_project.Association", "endpoints");
 }
 
+void startOrStopADCSensor(bool start,
```
sbteeks, line 438: Maybe 'const bool start'?

redfish-core/lib/assembly.hpp
```diff
@@ -435,6 +435,96 @@ inline void
...
+                return;
+            }
+            messages::success(asyncResp->res);
+        },
+        "org.freedesktop.systemd1", "/org/freedesktop/systemd1",
+        "org.freedesktop.systemd1.Manager", method,
+        "xyz.openbmc_project.adcsensor.service", "replace");
+}
+
+void doBatteryCM(const std::string& assembly, bool readyToRemove,
```
sbteeks, line 462: 'const bool'?

spinler: done

gtmills:
```
clang-tidy-14 --use-color -export-fixes /tmp/tmpwiyfiv0u/tmpe4s0l3ek.yaml -p=. /home/fspcibld/openbmc-dev-jenkins/workspace/OpenBMC-Dev/build/github-public/repo-ci/public-ibm-openbmc-downstream-repo-ci-pr-bmcweb/ibm-openbmc/bmcweb/src/webserver_main.cpp
[1m../redfish-core/include/../lib/assembly.hpp:496:33: [0m[0;1;31merror: [0m[1m'serviceName' in capture list does not name a variable [clang-diagnostic-error][0m
                    [asyncResp, serviceName,
[0;1;32m                                ^
[0m197736 warnings and 1 error generated.
```
is the error

redfish-core/lib/assembly.hpp
```diff
@@ -435,6 +435,96 @@ inline void
...
+                 const std::shared_ptr<bmcweb::AsyncResp>& asyncResp)
+{
+    if (readyToRemove)
+    {
+        // Stop the adcsensor service so it doesn't monitor the battery
+        startOrStopADCSensor(false, asyncResp);
+        return;
+    }
+
+    // Find the service that has the OperationalStatus iface, set the
```
gtmills, line 472: This comment is helpful.

When readyToRemove is true, we are stopping the adcsensor, that code is above on line 468. Makes sense. 

When readyToRemove is false, this is the only time we get in here. 
readyToRemove =  false is done after the CM.. 
So when readyToRemove is false, we set functional to true and start the ADC sensor. Makes sense.

gtmills (approved):
I am good once it passes CI

spinler:
> Huh, I don't see how that's an error. Also, it works fine.

yea that is an error .fixed.

sbteeks (approved):

PR #478, opened 2022-12-08
Set Functional property after battery CM

spinler:

This does two things:

Defect PE00DX9W. For 1020.30


gtmills (approved):

spinler:
hmm, all sorts of compile errors, I guess a cherry-pick doesn't work

PR #483, opened 2022-12-12
BMC_SHELL web-base_shell

sbteeks:

Path of WebSocket:
    wss://{hostname}/bmc-console

Implementation:

    like ls, cd, etc.

Authorize user for Shell :

Test:
    Needs testing using webUI. Currently, webUI does not work with p11
    bring-up openbmc and 1050 bmcweb. The testing should be done
    using the image built with 1050-ghe. 

### DIFF
--- a/http/obmc_shell.hpp
+++ b/http/obmc_shell.hpp
@@ -1,0 +1,250 @@
+#pragma once
+
+#include <pty.h>
+#include <pwd.h>
+#include <termios.h>
+
+#include <app.hpp>
+#include <boost/asio/posix/stream_descriptor.hpp>
+#include <boost/process/io.hpp>
+#include <websocket.hpp>
+
+#include <map>
+
+namespace crow
+{
+
+namespace obmc_shell
+{
+
+class Handler : public std::enable_shared_from_this<Handler>
+{
+  public:
+    explicit Handler(crow::websocket::Connection* conn) :
+        session(conn), streamFileDescriptor(conn->getIoContext())
+    {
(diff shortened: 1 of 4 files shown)


gtmills:
jenkins run tests please

sbteeks:
jenkins run tests please

sbteeks (commented):
jenkins run tests please

http/obmc_shell.hpp
```diff
@@ -0,0 +1,250 @@
...
+    std::array<char, 4096> outputBuffer{};
+};
+
+static std::map<crow::websocket::Connection*, std::shared_ptr<Handler>>
+    mapHandler;
+
+inline void requestRoutes(App& app)
+{
+    BMCWEB_ROUTE(app, "/bmc-console")
+        .privileges({{"ConfigureManager"}})
```
gtmills, line 211: We need this to service at some point

gtmills: https://github.com/ibm-openbmc/bmcweb/blob/1030/http/obmc_shell.hpp#L177

gtmills: https://github.com/ibm-openbmc/bmcweb/blob/1030/http/obmc_shell.hpp#L185

gtmills: https://github.com/ibm-openbmc/bmcweb/commit/48b082d25e6727dbdfe8e03c08a6c160ebebc388

gtmills (approved):

PR #484, opened 2022-12-13
IBM ACF Support

anoo1:
This PR adds support for upload of ACF certificate files, in addition to a new OEM user role.

Tested: Certificate is able to be uploaded via:
```
curl -k -H "X-Auth-Token: $TOKEN" -X PATCH -d '{"Oem":{"IBM":{"ACF":{"ACFFile":"'${service_acf_base64}'"}}}}' https://${BMC}:${BMC_HTTPS_port}/redfish/v1/AccountService/Accounts/service
```
and the BMC is accessible by logging as service.

### DIFF
--- a/include/authentication.hpp
+++ b/include/authentication.hpp
@@ -239,6 +239,16 @@ static std::shared_ptr<persistent_data::UserSession>
         }
     }
 
+    // This patch is allowed for service user, without authorization to upload
+    // unauthenticated ACF.
+    if (boost::beast::http::verb::patch == method)
+    {
+        if (url == "/redfish/v1/AccountService/Accounts/service")
+        {
+            return true;
+        }
+    }
+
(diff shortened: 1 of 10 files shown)


gtmills:
@joseph-reynolds Can any of this support go upstream?

gtmills:
Can you add more to the PR message ? What does this do? ACF support for uploading/? 
Any testing?

anoo1:
> Can you add more to the PR message ? What does this do? ACF support for uploading/?
> Any testing?

I added a summary of what I think all the commits do. Also how the upload ACF file was tested. I haven't run the redfish validator.

gtmills (approved):
Newer bmcweb upstream does things slightly differently in an area or too, but I think if I want that I should address later. This is what we had before and this is needed for 1050 so going to merge. Thanks, Adriana!

PR #486, opened 2022-12-19
1050 upstream rebase new boost

geissonator:
This is a rebase with upstream bmcweb.
We needed multiple fixes for the linters (and I needed the new boost), so I went ahead and did a full rebase.

Tested:
- Built image (including new boost in it), booted in simulation, and verified bmcweb started without issue, a random assortment of Redfish queries worked, and the basic GUI functioned
- Our 1050 stream is still a bit of a work-in-progress so did not want to mess with hardware


gtmills:
Andrew tested on simics and Redfish/GUI worked. Huge PR. let's get it and we can deal with any fallout later.

gtmills (approved):

PR #487, opened 2022-12-20
Retrieve Xterminal using username

sbteeks:
This commit provides BMC console with root shell instead of
regular user shell. It also sets the privilege to oemIBMPerformService.


gtmills:
Please fill out PR msg

gtmills (approved):

PR #490, opened 2022-12-23
bmcweb: Add property value out of range error msg

sbteeks:
Current implementation displays internal error message when BIOS attribute value is not within the lower and upper bound values. Also, the same internal error message is displayed when the length of the String type value for the attribute to be updated is not within the specified limit.

This commit displays property value out of range error message for Integer type-based attributes in case of an error, and property value incorrect for String type-based attributes.

Testing:
$ curl -k -X PATCH -d '{"Attributes":{"hb_mfg_flags_current": "000000000000000000000000000000000"}}'
 https://user:password@host/redfish/v1/Systems/system/Bios/Settings
{
  "error": {
    "@Message.ExtendedInfo": [
      {
        "Message": "The property 'hb_mfg_flags_current' with the
requested value of '000000000000000000000000000000000' could not be written because the value does not meet the constraints of the implementation.",
        "MessageArgs": [
          "hb_mfg_flags_current",
          "000000000000000000000000000000000"
        ],
        "MessageId": "Base.1.8.1.PropertyValueIncorrect",
        "MessageSeverity": "Warning",
        "Resolution": "No resolution is required."
      }
    ],
    "code": "Base.1.8.1.PropertyValueIncorrect",
    "message": "The property 'hb_mfg_flags_current' with the requested
value of '000000000000000000000000000000000' could not be written because the value does not meet the constraints of the implementation."
  }
}

$ curl -k -X PATCH -d '{"Attributes":{"hb_ioadapter_enlarged_capacity": 200}}'
 https://user:password@host/redfish/v1/Systems/system/Bios/Settings
{
  "hb_ioadapter_enlarged_capacity@Message.ExtendedInfo": [
    {
      "Message": "The value 200 for the property
hb_ioadapter_enlarged_capacity is not in the supported range of acceptable values.",
      "MessageArgs": [
        "200",
        "hb_ioadapter_enlarged_capacity"
      ],
      "MessageId": "Base.1.13.0.PropertyValueOutOfRange",
      "MessageSeverity": "Warning",
      "Resolution": "Correct the value for the property in the request
body and resubmit the request if the operation failed."
    }
  ]
}

Signed-off-by: Shantappa Teekappanavar <sbteeks@yahoo.com>

### DIFF
--- a/redfish-core/lib/bios.hpp
+++ b/redfish-core/lib/bios.hpp
@@ -401,8 +401,12 @@ inline void requestRoutesBiosSettings(App& app)
                 return;
             }
 
-            boost::container::flat_map<std::string,
-                                       std::pair<bool, std::string>>
+            boost::container::flat_map<
+                std::string,
+                std::tuple<
+                    bool, std::string,
+                    boost::container::flat_map<
+                        std::string, std::variant<int64_t, std::string>>>>
                 biosAttrsType;
             const BiosBaseTableType* baseBiosTable =
                 std::get_if<BiosBaseTableType>(&retBiosTable);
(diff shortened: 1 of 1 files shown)


redfish-core/lib/bios.hpp
```diff
@@ -476,12 +525,92 @@ inline void requestRoutesBiosSettings(App& app)
...
+                    // Get Upper Bound value
+                    iter = attrBaseOptionsMap.find("UpperBound");
+                    if (iter != attrBaseOptionsMap.end())
+                    {
+                        upperBoundVal = std::get<int64_t>((*iter).second);
+                    }
+
+                    if (attrValue < lowerBoundVal || attrValue > upperBoundVal)
+                    {
+                        BMCWEB_LOG_ERROR << "Attribute value is out of range";
```
gtmills, line 553: {[m
                         BMCWEB_LOG_ERROR << "Attribute value is out of range";[m
[31m-                        std::string val =[m
[31m-                            boost::lexical_cast<std::string>(attrValue);[m
[32m+[m[32m                        std::string val = std::to_string(attrValue);[m
                         messages::propertyValueOutOfRange(asyncResp->res, val,[m
                                                           attrName);[m
                                                           
Take out that boost::lexical_cast ?

gtmills (approved):
seems reasonable

PR #507, opened 2023-01-18
Fix for some lock related defects

asmithakarun:
This commit has changes to:
1. Avoid having stale entries in the lock table in BMC
2. Fix unnecessary lock conflicts due to seg length validation
3. Remove stale lock table entries of invalid sessionid

### DIFF
--- a/include/ibm/locks.hpp
+++ b/include/ibm/locks.hpp
@@ -112,9 +112,6 @@ class Lock
 
     Lock()
     {
-        // Remove the persistent file
-        std::filesystem::remove(fileName);
-        lockTable.clear();
         transactionId = lockTable.empty() ? 0 : prev(lockTable.end())->first;
     }
 
@@ -264,7 +261,7 @@ inline RcAcquireLock Lock::acquireLock(const LockRequests& lockRequestStructure)
 
     if (status)
     {
-        BMCWEB_LOG_DEBUG << "There is a conflict within itself";
+        BMCWEB_LOG_ERROR << "There is a conflict within itself";
         return std::make_pair(true, std::make_pair(status, 1));
     }
     BMCWEB_LOG_DEBUG << "The request is not conflicting within itself";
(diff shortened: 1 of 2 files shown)


gtmills:
@asmithakarun Can you add the upstream commits here?

gtmills:
jenkins run tests please

include/ibm/locks.hpp
```diff
@@ -112,6 +112,9 @@ class Lock
 
     Lock()
     {
+        // Remove the persistent file
+        std::filesystem::remove(fileName);
+        lockTable.clear();
```
sunharis, line 117: This change is not needed. This was removed as part of other defects later.
Pls refer https://github.com/ibm-openbmc/bmcweb/commit/b2e7be91fb609fa04439ea83950a45b8d6661fd5 and https://github.com/ibm-openbmc/bmcweb/blob/1030/include/ibm/locks.hpp#L112

Can you also make the trace enhancements as in that PR https://github.com/ibm-openbmc/bmcweb/commit/b2e7be91fb609fa04439ea83950a45b8d6661fd5

sunharis (changes requested):

include/ibm/locks.hpp
asmithakarun: Done

asmithakarun:
jenkins run tests please

gtmills:
Please add the upstream commits here

include/ibm/locks.hpp
```diff
@@ -261,7 +261,7 @@ inline RcAcquireLock Lock::acquireLock(const LockRequests& lockRequestStructure)
 
     if (status)
     {
-        BMCWEB_LOG_DEBUG << "There is a conflict within itself";
+        BMCWEB_LOG_ERROR << "There is a conflict within itself";
```
gtmills, line 264: We really want an ERROR here?

sunharis: > We really want an ERROR here?

This is to get traces in the journal log if there is a lock conflict issue. There is no other way to identify this otherwise

sunharis:
> @asmithakarun Can you add the upstream commits here?

We did not upstream these fixes - as this is IBM only use case and community was not interested to review/merge.

sunharis (approved):

PR #514, opened 2023-01-25
ReadOnly users should not have ssh (#408)

baemyung:
* ReadOnly users should not have ssh

Note that ReadOnly user created prior to this change will have the HostConsole AccountType privilege.  A BMC admin can PATCH a user's AccountTypes property to add or remove the Hostconsole value.

Tested: Yes

* fixup code formatting

Tested: Compiles

### DIFF
--- a/redfish-core/lib/account_service.hpp
+++ b/redfish-core/lib/account_service.hpp
@@ -1911,13 +1911,15 @@ inline void handleAccountCollectionPost(
             return;
         }
 
-        // Create (modified) modGroupsList from allGroupsList that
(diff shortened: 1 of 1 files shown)


gtmills:
add to approvelist

joseph-reynolds (approved):
Looks good to me.  This should work okay with this related user-manager change: https://github.com/ibm-openbmc/phosphor-user-manager/commit/d6d0df0fe18221c5459f8bd7930214fdec65be5d

gtmills (approved):

PR #520, opened 2023-01-30
Populate BMC state based on state manager property

geissonator:
The "Status" of the BMC has been historically fairly simple. If the BMC is at the point where it can respond to the Redfish call then it's "OK" and it's "On". Its "Status" could be derived based on the startup systemd target, multi-user.target, as either "Starting" or "Enabled".

However, with the introduction of new BMC states it is no longer this simple. The following design introduced a new "Quiesced" state for the BMC:
https://github.com/openbmc/docs/blob/master/designs/bmc-service-failure-debug-and-recovery.md

There has was also a new BMC state introduced, "UpdateInProgress", which indicates the BMC is in the middle of a firmware update.

These new BMC states require additional logic when responding to Redfish queries for the BMC "Status".

When a BMC is in a "Quiesced" state it is an indication that a critical BMC service has entered the failed state. It is important for BMC clients to be aware of this state to take the needed steps to service the system. For example, on IBM systems, this will trigger a call home event to IBM service. The undefined aspect of this state is what makes it so critical to get serviced.

bmcweb has a precedent of abstracting systemd states behind services like x86-power-control/phosphor-state-manager. It can see it in functions like getHostState() and getChassisState(). bmcweb could contain all of the logic to determine if the BMC is being updated or has a failed critical service but that:
  a) Adds a lot more code to bmcweb and
  b) Goes away from our design of allowing other repositories to
     abstract away the complexity of these states.

Tested:
- GET /redfish/v1/Managers/bmc for different BMC states
- Redfish Validator passed

Signed-off-by: Ravi Teja <raviteja28031990@gmail.com>
Change-Id: I8ceafb7036f41678516ebda3dce5200221bba947

### DIFF
--- a/redfish-core/lib/managers.hpp
+++ b/redfish-core/lib/managers.hpp
@@ -1941,6 +1941,57 @@ inline void setDateTime(std::shared_ptr<bmcweb::AsyncResp> aResp,
     }
 }
 
+inline void getBMCState(const std::shared_ptr<bmcweb::AsyncResp>& aResp)
+{
+    aResp->res.jsonValue["PowerState"] = "On";
+    sdbusplus::asio::getProperty<std::string>(
+        *crow::connections::systemBus, "xyz.openbmc_project.State.BMC",
+        "/xyz/openbmc_project/state/bmc0", "xyz.openbmc_project.State.BMC",
+        "CurrentBMCState",
+        [aResp](const boost::system::error_code ec,
+                const std::string& bmcState) {
+        if (ec)
+        {
+            BMCWEB_LOG_DEBUG << "DBUS response error for "
+                                "Location";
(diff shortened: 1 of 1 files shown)


geissonator:
This is a cherry-pick of upstream change https://gerrit.openbmc.org/c/openbmc/bmcweb/+/50318 which we have been unsuccessful with getting into upstream. The IBM arguments for this change are:
- This area is very nuanced and we've worked through a lot of no-connect issues with the HMC team on this behavior
- Upstream code does not handle BMC quiesced and it does not handle reflecting the correct BMC state in the small window between a BMC reboot being requested and actually rebooting

I have done no additional testing of this change as it's just a cherry pick of what we had previously in prior releases.

gtmills (approved):

PR #521, opened 2023-01-30
enable error traces in bmcweb

geissonator:
In order to be able to more easily debug bmcweb related issue, lets enable error traces by default. So by default all Error and Critical level traces are now written to the journal. If the dev enables the bmcweb-logging feature then they will get all levels of trace.

There are some error traces periodically being written to the journal which do not appear to be errors (since all is working fine). The next commit will change the level of these logs.

Testing:
- Verified only the error and critical logs were displayed

Signed-off-by: Andrew Geissler <geissonator@yahoo.com>

### DIFF
--- a/http/logging.hpp
+++ b/http/logging.hpp
@@ -45,20 +45,16 @@ class Logger
            [[maybe_unused]] const size_t line, LogLevel levelIn) :
         level(levelIn)
     {
-#ifdef BMCWEB_ENABLE_LOGGING
         stringstream << "(" << timestamp() << ") [" << prefix << " "
                      << std::filesystem::path(filename).filename() << ":"
                      << line << "] ";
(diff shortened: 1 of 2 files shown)


geissonator:
@baemyung is working on getting a form of this upstreamed via https://gerrit.openbmc.org/c/openbmc/bmcweb/+/60212 but for now we need to continue to carry this patch downstream as we decided within IBM that these logs should always be written to the journal.

No additional testing of this was done as it's just a cherry pick of https://github.com/ibm-openbmc/bmcweb/commit/32c0c6b49fcbf84d677ebd2422415219fc670360 with a minor merge conflict fix to pick upstreams "NOLINTNEXTLINE" change.

gtmills (approved):

PR #523, opened 2023-01-31
WIP: Remove InternalErrors on usbCodeUpdate GET

gtmills:
Needs to go into 1020 and 1040. This code hasn't been brought into 1050 yet, so will intercept it then. 

usbCodeUpdate isn't a critical service, if it fails, just leave its properties off the response. Have a defect where on the 45th code update where we think usbCodeUpdate didn't respond (we don't have the journal so don't know more) and that caused bmcweb to return an internalError here. This internalError led to the HMC putting the BMC in No Connection. Yeah, this seems kind of fragile.. But this service isn't important so don't let it impact operations.

See https://jazz07.rchland.ibm.com:13443/jazz/web/projects/CSSD#action=com.ibm.team.workitem.viewWorkItem&id=401222 for more info 

This goes beyond https://github.com/ibm-openbmc/bmcweb/pull/258 which did this for 1 case of the usbCodeUpdate app. 

Tested: None.

### DIFF
--- a/redfish-core/lib/oem/ibm/usb_code_update.hpp
+++ b/redfish-core/lib/oem/ibm/usb_code_update.hpp
@@ -92,7 +92,6 @@ inline void
                 if (ec)
                 {
                     BMCWEB_LOG_ERROR << "DBUS response error " << ec;
-                    messages::internalError(aResp->res);
                     return;
                 }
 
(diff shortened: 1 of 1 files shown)


geissonator:
Just to be super sure, it would be great to stop the usb service and do a GET on the manager interface to make sure there is no other path we're missing here.

geissonator (approved):

ChicagoDuan (approved):

anoo1 (approved):

gtmills:
Decided we didn't need this PR at this time. Abandoning. Can reopen if we decide we want at a later date.

PR #526, opened 2023-02-01
Restrict the Operator role (#319)

joseph-reynolds:

Tested:
2. POST /redfish/v1/AccountService/Accounts/new with Role:Operator Ensure it failed with message Base.1.9.0.RestrictedRole. Posting a new Administrator user is successful.


joseph-reynolds:
This PR is downstream-only because the underlying restricted role support is downstream-only and is not yet supported by the OpenBMC community.

gtmills (approved):

PR #545, opened 2023-02-10
Update Assembly Schema To Populate Status::State

SunnySrivastava1984:
Status::State reflects the presence status of FRU being populated in the schema.
"Enabled" state that FRU is present.
"Absent" implies that device is not present.

Validator was executed and no new error was found.

### DIFF
--- a/redfish-core/lib/assembly.hpp
+++ b/redfish-core/lib/assembly.hpp
@@ -176,6 +176,45 @@ inline void
                             "LocationCode",
                             "LocationCode");
                     }
+                    else if (interface == "xyz.openbmc_project.Inventory.Item")
+                    {
+                        sdbusplus::asio::getProperty<bool>(
+                            *crow::connections::systemBus, serviceName,
+                            assembly, "xyz.openbmc_project.Inventory.Item",
+                            "Present",
+                            [aResp, assemblyIndex](
+                                const boost::system::error_code ec2,
+                                const std::variant<bool>& property) {
+                            if (ec2)
+                            {
+                                BMCWEB_LOG_DEBUG << "DBUS response error";
+                                messages::internalError(aResp->res);
+                                return;
+                            }
+
+                            nlohmann::json& assemblyArray =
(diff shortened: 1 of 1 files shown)


redfish-core/lib/assembly.hpp
```diff
@@ -176,6 +176,45 @@ inline void
...
+                            {
+                                assemblyData["Status"]["State"] = "Absent";
+                            }
+                            else
+                            {
+                                assemblyData["Status"]["State"] = "Enabled";
+                            }
+                            },
+                            serviceName, assembly,
+                            "org.freedesktop.DBus.Properties", "Get",
```
gtmills, line 215: All upstream code is using `sdbusplus::asio::getProperty<bool>(`
https://github.com/openbmc/bmcweb/blob/3ccb3adb9a14783f6bef601506de9f8bcae22d51/redfish-core/lib/led.hpp#L61

gtmills:
What is the upstream commit?

SunnySrivastava1984:
> What is the upstream commit?

There is no upstream commit for assembly as such hence this commit cannot go upstream. Once PCIeDevcie schema is merged upstream, we can take assembly as well. Till then we can keep it downstream.

redfish-core/lib/assembly.hpp
```diff
@@ -176,6 +176,45 @@ inline void
...
+
+                            const bool* value = std::get_if<bool>(&property);
+
+                            if (value == nullptr)
+                            {
+                                // illegal value
+                                messages::internalError(aResp->res);
+                                return;
+                            }
+                            if (!*value)
```
gtmills, line 208: The styling of upstream bmcweb is more like 

assemblyData["Status"]["State"] = "Enabled";
if (!*value)
{
       assemblyData["Status"]["State"] = "Absent";

}

oh well. A lot of time bmcweb also has a default in case this interface isn't there

gtmills (approved):

PR #546, opened 2023-02-10
Publish properties under PCIeDevice schema

SunnySrivastava1984:
The commit publishes following properties under
PCIeDevice schema.
- Location code
- Operational status
- Pretty Name


gtmills:
Is there an upstream commit here?

SunnySrivastava1984:
> Is there an upstream commit here?

There is an updated upstream commit for PCIeDevice schema which, as discussed, Lakshmi is supposed to take forward. 

https://gerrit.openbmc.org/c/openbmc/bmcweb/+/40051/5

Once that is there we can push these commits on top of that. As we are not aware how much that base commit is going to get modified in due course, it does not make sense to stack commits on top of that. 
Hence I have not pushed this commit upstream as of now. We can track this via our spreadsheet and once the base one is merged we can go ahead and push these as well.

baemyung (approved):

PR #551, opened 2023-02-13
Remove link from Chassis to PCIeDevice

SunnySrivastava1984:
The commit removes current support that assumes 1:1 system:Chassis for Chassis/PCIeDevices.

Current implementation populates the same collection of PCIeDevices w.r.t chassis as w.r.t. system.
For systems with multiple chassis the current assumption does not hold true.

Validator has been executed with no new errors.

### DIFF
--- a/redfish-core/lib/chassis.hpp
+++ b/redfish-core/lib/chassis.hpp
@@ -276,8 +276,6 @@ inline void
             asyncResp->res
                 .jsonValue["Actions"]["#Chassis.Reset"]["@Redfish.ActionInfo"] =
                 "/redfish/v1/Chassis/" + chassisId + "/ResetActionInfo";
-            asyncResp->res.jsonValue["PCIeDevices"]["@odata.id"] =
-                "/redfish/v1/Systems/system/PCIeDevices";
             asyncResp->res.jsonValue["PCIeSlots"]["@odata.id"] =
                 "/redfish/v1/Chassis/" + chassisId + "/PCIeSlots";
 


gtmills:
@SunnySrivastava1984 Any upstream here ?

baemyung (approved):
The logic of removal seems good.

BTW, 
>Current implementation populates the same collection of PCIeDevices w.r.t chassis as w.r.t. system.
>For systems with multiple chassis the current assumption does not hold true.

Does this mean that the upstream also needs to remove the link?
Or is it only for downstream?

gtmills:
@SunnySrivastava1984 merge conflict now. A lot going into bmcweb 1050 at the moment

SunnySrivastava1984:
> @SunnySrivastava1984 Any upstream here ?

https://gerrit.openbmc.org/c/openbmc/bmcweb/+/56996

SunnySrivastava1984:
> @SunnySrivastava1984 merge conflict now. A lot going into bmcweb 1050 at the moment

Done

SunnySrivastava1984:
> The logic of removal seems good.
> 
> BTW,
> 
> > Current implementation populates the same collection of PCIeDevices w.r.t chassis as w.r.t. system.
> > For systems with multiple chassis the current assumption does not hold true.
> 
> Does this mean that the upstream also needs to remove the link? Or is it only for downstream?

We have pushed a commit upstream but still under review.

gtmills (approved):

PR #558, opened 2023-02-15
1050: Set Functional property after battery CM (#476)

baemyung:

This does two things:


baemyung:
@spinler 
I'm not sure how to test this change?

spinler:
There is an outstanding PR in GHE to pull in a linux fix, after that the battery sensor will work and we can test. I'll keep an eye on it.

spinler (approved):
I can test the CM flow on an everest when everything is in.

PR #574, opened 2023-02-20
Correct Id parameter for two log services

gtmills:
Merged upstream at https://gerrit.openbmc.org/c/openbmc/bmcweb/+/60766 
A similar change was done for 1020, 1030, 1040 

The Journal and EventLog now fail redfish-service-validators new test. Per the specification, the "Id" parameter should match the elements odata id segment in the collection.  The validator now tests for this, and these two fail.

Tested: Redfish service validator now passes.

Change-Id: I463eea7ae7a7dfd6a9d6d562ae4f03dc887a9478


baemyung (approved):
It looks good to me

PR #580, opened 2023-02-22
1050: bmcweb hostconsole websocket access issue

ninadpalsule:
The recent changes related to using new verb class in router missed to remove the method value shifting as per the new values.

Testing:
    Tested on machine and hostconsole is accessible
    through the GUI.

Fix was reported in https://github.com/openbmc/bmcweb/issues/243 and upstream work is in progress.

### DIFF
--- a/http/routing.hpp
+++ b/http/routing.hpp
@@ -1267,7 +1267,7 @@ class Router
         }
 
         if ((rules[ruleIndex]->getMethods() &
-             (1U << static_cast<size_t>(req.method()))) == 0)
+             (static_cast<size_t>(req.method()))) == 0)
         {
             BMCWEB_LOG_DEBUG << "Rule found but method mismatch: " << req.url
                              << " with " << req.methodString() << "("


Lakshmi-y (approved):

baemyung (commented):
getMethods() is a bit-field of size_t.

```
    size_t getMethods() const
    {
        return methodsBitfield;
    }
    ...
    size_t methodsBitfield{1 << static_cast<size_t>(HttpVerb::Get)};
```

However, req.method() is an enum-value.
So, I think bitwise-& is not proper between getMethods() & req.method().

I think the issue is the size of "1U" (i.e. unsigned int) vs. "size_t".

Perhaps, should the code be written like this?

```
        if ((rules[ruleIndex]->getMethods() &
             ( ((size_t)1U) << static_cast<size_t>(req.method()))) == 0)
```

ninadpalsule:
This is the value it is failing. I think the newer implementation rules[ruleIndex]->getMethods() value and req->method() values are matching?

bmcweb[4324]: (2023-02-21 20:33:19) [DEBUG "routing.hpp":1272] Rule found but method mismatch: /console0 with GET(2) / 2

            BMCWEB_LOG_DEBUG << "Rule found but method mismatch: " << req.url
                             << " with " << req.methodString() << "("
                             << static_cast<uint32_t>(req.method()) << ") / "
                             << rules[ruleIndex]->getMethods();

gtmills:
add to approvelist

gtmills:
@ninadpalsule Have an upstream commit for this?

ninadpalsule:
I am going to backport different fix for this same issue.

ninadpalsule:
Closing this as I have pushed new commit from upstream.

PR #582, opened 2023-02-22
Add missing usage of new verb class in router

ninadpalsule:
Despite introduction of a new enum class containing method verbs, some functions were still using the one from Boost. This had caused erratic behaviour when trying to create a websocket (e.g. /nbd/<str>), because enum value of old type was compared to the one of new type. This change fixes that.

Tested:
    Verified that websockets are now created without errors and access
    hostconsole from the GUI.
    
    Upstream work:
     - Cherry-picked following commit from upstream and tested
     https://github.com/openbmc/bmcweb/commit/1c99da037b2242c6fbad3441038081c5703dc1e9

Change-Id: I52c874de9b02463618143d3b031f5c795dd42ad8

### DIFF
--- a/http/routing.hpp
+++ b/http/routing.hpp
@@ -1267,11 +1267,11 @@ class Router
         }
 
         if ((rules[ruleIndex]->getMethods() &
-             (1U << static_cast<size_t>(req.method()))) == 0)
+             (1U << static_cast<size_t>(*verb))) == 0)
         {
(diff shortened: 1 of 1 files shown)


baemyung (approved):
Since it is on upstream, it may be better to add the gerrit link in commit msg.

https://gerrit.openbmc.org/c/openbmc/bmcweb/+/59924

Lakshmi-y (approved):

gtmills (approved):

PR #589, opened 2023-02-24
Bug fix: cannot delete ACF

joseph-reynolds:
This fixes a bug with the ACF upload function which was introduced in 921ba85e0bc68779573db121790bde574332d395 (the AllowUnauthACFUpload function).  The bug prevents PATCHing the {"ACFFile":null} value from deleting the ACF.  Patching an empty string {"ACFFile":""} works and effectively deletes the ACF.

In this fix, you can PATCH the /redfish/v1/AccountService/Accounts/service property Oem.IBM.ACF.ACFFile to have either the null value, an empty string, or a correct base64-encoded ACF content

Tested: Combinations of ACFFile and AllowUnauthACFUpload property values.

### DIFF
--- a/redfish-core/lib/account_service.hpp
+++ b/redfish-core/lib/account_service.hpp
@@ -2633,7 +2633,6 @@ inline void requestAccountServiceRoutes(App& app)
                                     "AllowUnauthACFUpload",
                                     allowUnauthACFUpload);
                             }
-                            
                             if (!rc)
                             {
                                 BMCWEB_LOG_ERROR << "Illegal Property ";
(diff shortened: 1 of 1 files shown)


joseph-reynolds:
Fixes 411447 (for 1020).  Will need to cherry-pick for 1030.

redfish-core/lib/account_service.hpp
```diff
@@ -2618,10 +2618,23 @@ inline void requestAccountServiceRoutes(App& app)
                         {
                             std::optional<bool> allowUnauthACFUpload;
                             std::optional<std::string> acfFile;
-                            if (!redfish::json_util::readJson(
+                            bool rc;
```
geissonator, line 2621: nit: should initialize to a default

geissonator (approved):

redfish-core/lib/account_service.hpp
```diff
@@ -2618,10 +2618,22 @@ inline void requestAccountServiceRoutes(App& app)
...
+                            // Property ACFFile may be null or string
+                            if (acf.value().contains("ACFFile") &&
```
joseph-reynolds, line 2623: suggestion to use is_null() here

PR #590, opened 2023-02-27
1050: Make PCIe Adapter Paths Unique (#372)(#391)

baemyung:
This commit makes redfish adapter objects unique, before this commit bmcweb shows redfish adapter objects with the same path even though all the dbus objects have unique paths
{"@odata.id": "/redfish/v1/Systems/system/PCIeDevices/adapter1"}, {"@odata.id": "/redfish/v1/Systems/system/PCIeDevices/adapter1"}.

after this change bmcweb retrieve PCI device endpoint information and then, replace redfish PCI device as
"chassisN_io_moduleN_slotN_adapterN".

this change also affect below path because they also has slot and chassis value in their path.
"/xyz/openbmc_project/inventory/system/chassis/motherboard/pcieslot0/pcie_card0"

Tested result:

```
curl -k -H "X-Auth-Token: $token" -X GET  https://${bmc}/redfish/v1/Systems/system

{
...
...
"PCIeDevices": [
...
...
{"@odata.id": "/redfish/v1/Systems/system/PCIeDevices/chassis_pcieslot0_pcie_card0"},
{"@odata.id": "/redfish/v1/Systems/system/PCIeDevices/chassis_pcieslot1_pcie_card1"},
...
...
{"@odata.id": "/redfish/v1/Systems/system/PCIeDevices/chassis15363_io_module1_slot1_adapter1"},
{"@odata.id": "/redfish/v1/Systems/system/PCIeDevices/chassis15363_io_module1_slot2_adapter1"},
...
],
...
```

### DIFF
--- a/redfish-core/lib/pcie.hpp
+++ b/redfish-core/lib/pcie.hpp
@@ -26,6 +26,7 @@
 #include <sdbusplus/asio/property.hpp>
 #include <sdbusplus/unpack_properties.hpp>
 #include <utils/dbus_utils.hpp>
+#include <utils/pcie_util.hpp>
 
 #include <map>
 #include <string>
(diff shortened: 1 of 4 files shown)


redfish-core/lib/pcie.hpp
```diff
@@ -39,49 +40,93 @@ static constexpr char const* pcieDeviceInterface =
 static constexpr char const* pcieSlotInterface =
     "xyz.openbmc_project.Inventory.Item.PCIeSlot";
 
-static inline void
-    getPCIeDeviceList(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
-                      const std::string& name)
+using PCIeDevice = std::string;
```
gtmills, line 43: Why so many changes here ?

baemyung: In 1030,  PRs (e.g.  #372, #389. #391) rearranged  the codes to the most current code in this area.
One of which is to introduce - getPcieDevices(), or retrievePCIeDeviceList() etc.

Looking at them, the main restructuring can be done here via this PR so that the consecutive changes will be mostly related to the functionalities. 

For example, #389 (Link PCIeDevice to PCIeSlot) had the mix of the code restructure and functional change.
In 1050, PR #592 (i.e. #389) can be become short and simple.

Another issue which also currently exists in 1030 is to return the empty output for the first PCIe related to command after bmcweb startup.
It is due to the logic of building 'mapOfPcieDevicePaths' is done via  async_method_call as a normal function in sequence as a part of requestRoutesSystemPCIeDevice().
As the result, the very first invocation of PCIe query will only construct  mapOfPcieDevicePaths, and the the other consecutive calls are utilizing it.

This PR fixes it so that map[] will be built as a part of request...() [if map is not built] and then continue on the request function.
The other ones are capturing the code restructuring logics of 1030 PRs.

gtmills (approved):

baemyung:
Additional fix for FabricAdapter on PCIeDevice is done by https://github.com/ibm-openbmc/bmcweb/pull/605

PR #594, opened 2023-03-01
Fix redfish GET issue on Dump collection

asmithakarun:
Redfish GET on system dump entries returned null eventhough the dump entries were found in the backend. This PR fixes this issue by appending to the "Members" field of the redfish response instead of overriding it.

Tested By:

GET https://${bmc}/redfish/v1/Systems/system/LogServices/Dump/Entries

### DIFF
--- a/redfish-core/lib/log_services.hpp
+++ b/redfish-core/lib/log_services.hpp
@@ -565,7 +565,10 @@ inline void
         }
 
         nlohmann::json& entriesArray = asyncResp->res.jsonValue["Members"];
-        entriesArray = nlohmann::json::array();
+        if (entriesArray.empty())
+        {
+            entriesArray = nlohmann::json::array();
+        }
         std::string dumpEntryPath =
             "/xyz/openbmc_project/dump/" +
             std::string(boost::algorithm::to_lower_copy(dumpType)) + "/entry/";


raviteja-b:
do we have upstream commit?

raviteja-b (approved):

sunharis (approved):

sunharis:
@asmithakarun Please fix the formatting

sunharis:
> do we have upstream commit?

Not yet. They are abandoned. After the dump manager refactor, we can re-open them with some modifications by removing dump type hardcodes.

redfish-core/lib/log_services.hpp
```diff
@@ -565,7 +565,10 @@ inline void
         }
 
         nlohmann::json& entriesArray = asyncResp->res.jsonValue["Members"];
-        entriesArray = nlohmann::json::array();
+        if (entriesArray.empty())
+        {
+            entriesArray = nlohmann::json::array();
```
gtmills, line 570: why is this code needed at all if it is empty?

gtmills: but why not just delete line 568 ?

gtmills (approved):

PR #542, opened 2023-02-09
Sync HardwareIsolation 1030 commits to 1050

deepakala-k:
Sync HardwareIsolation commits from 1030 to 1050. 
Here is the upstream commits for the relevant ones. 
Implement LocationIndicatorActive for CPU resource | https://gerrit.openbmc.org/c/openbmc/bmcweb/+/37045
Implement SubProcessors for processor core | https://gerrit.openbmc.org/c/openbmc/bmcweb/+/38570
Implement LocationIndicatorActive for CPU resource | https://gerrit.openbmc.org/c/openbmc/bmcweb/+/37045

For the HWIsolation, the commits do not have respective upstream commits, as we do not have a standard guard implementation in the upstream

Test output:
After adding the patches, the Hardware Isolation records are 
-bash-4.2$ curl -k -H "X-Auth-Token: $bmc_token" -XGET https://${BMC_IP}/redfish/v1/Systems/system/LogServices/HardwareIsolation/Entries
{
  "@odata.id": "/redfish/v1/Systems/system/LogServices/HardwareIsolation/Entries",
  "@odata.type": "#LogEntryCollection.LogEntryCollection",
  "Description": "Collection of System Hardware Isolation Entries",
  "Members": [
    {
      "@odata.id": "/redfish/v1/Systems/system/LogServices/HardwareIsolation/Entries/1",
      "@odata.type": "#LogEntry.v1_9_0.LogEntry",
      "Created": "2023-02-12T15:11:24+00:00",
      "EntryType": "Event",
      "Id": "1",
      "Links": {
        "OriginOfCondition": {
          "@odata.id": "/redfish/v1/Systems/system/Memory/dimm2"
        }
      },
      "Message": "DDR Memory Port",
      "Name": "Hardware Isolation Entry",
      "Severity": "Warning"
    },

PR #597, opened 2023-03-02
Using sdbusplus::asio wrapper functions

deepakala-k:
Use sdbusplus::asio wrapper functions getProperty and getAllProperties in place of direct "Get" and "GetAll". 
This is created to address the comment https://github.com/ibm-openbmc/bmcweb/pull/542#discussion_r1107935850


redfish-core/include/utils/hw_isolation.hpp
```diff
@@ -187,10 +187,14 @@ inline void
...
-            boost::system::error_code ec,
-            const std::variant<std::vector<std::string>>& vEndpoints) {
+
+    sdbusplus::asio::getProperty<std::vector<std::string>>(
+        *crow::connections::systemBus, "xyz.openbmc_project.ObjectMapper",
+        resourceObjPath.str + "/isolated_hw_entry",
+        "xyz.openbmc_project.Association", "endpoints",
+        [aResp, resourceObjPath,
+         hwIsolationDbusName](boost::system::error_code ec,
+                              const std::vector<std::string>& vEndpoints) {
```
baemyung, line 197: I think we can use getAssociationEndPoints() like

```
    dbus::utility::getAssociationEndPoints(
        resourceObjPath.str + "/isolated_hw_entry",
        [asyncResp, pcieSlotPath,
         [aResp, resourceObjPath,
         hwIsolationDbusName](boost::system::error_code ec,
                const dbus::utility::MapperEndPoints& vEndpoints) {
```

redfish-core/include/utils/hw_isolation.hpp
```diff
@@ -216,8 +220,7 @@ inline void
         }
 
         std::string resourceIsolatedHwEntry;
-        const std::vector<std::string>* endpoints =
-            std::get_if<std::vector<std::string>>(&(vEndpoints));
+        const std::vector<std::string>* endpoints = &(vEndpoints);
```
baemyung, line 223:
```
const dbus::utility::MapperEndPoints* endpoints = &(vEndpoints);
```

redfish-core/include/utils/hw_isolation.hpp
```diff
@@ -465,10 +565,12 @@ inline void
...
-    crow::connections::systemBus->async_method_call(
-        [aResp, resourceObjPath](
-            boost::system::error_code ec,
-            const std::variant<std::vector<std::string>>& vEndpoints) {
+    sdbusplus::asio::getProperty<std::vector<std::string>>(
+        *crow::connections::systemBus, "xyz.openbmc_project.ObjectMapper",
+        resourceObjPath.str + "/event_log", "xyz.openbmc_project.Association",
+        "endpoints",
+        [aResp, resourceObjPath](boost::system::error_code ec,
+                                 const std::vector<std::string>& vEndpoints) {
```
baemyung, line 573: Similar to the above case, this can be done with `dbus::utility::getAssociationEndPoints()`

redfish-core/include/utils/hw_isolation.hpp
```diff
@@ -487,8 +589,7 @@ inline void
             return;
         }
 
-        const std::vector<std::string>* endpoints =
-            std::get_if<std::vector<std::string>>(&(vEndpoints));
+        const std::vector<std::string>* endpoints = &(vEndpoints);
```
baemyung, line 592:
```
const dbus::utility::MapperEndPoints* endpoints = &(vEndpoints);
```

redfish-core/lib/assembly.hpp
```diff
@@ -720,10 +696,12 @@ inline void
...
     // if there is assembly association, look
     // for endpoints
-    crow::connections::systemBus->async_method_call(
+    sdbusplus::asio::getProperty<std::vector<std::string>>(
+        *crow::connections::systemBus, "xyz.openbmc_project.ObjectMapper",
+        assemblyPath, "xyz.openbmc_project.Association", "endpoints",
         [aResp, chassisPath, setLocationIndicatorActiveFlag,
          req](const boost::system::error_code ec,
-              const std::variant<std::vector<std::string>>& endpoints) {
+              const std::vector<std::string>& endpoints) {
```
baemyung, line 704:
```
     const dbus::utility::MapperEndPoints& endpoints);
```

redfish-core/include/utils/hw_isolation.hpp
deepakala-k: Thank you! I addressed it.

redfish-core/lib/assembly.hpp
deepakala-k: Thank you! I addressed it.

redfish-core/include/utils/hw_isolation.hpp
```diff
@@ -452,6 +452,107 @@ inline bool setSeverity(const std::shared_ptr<bmcweb::AsyncResp>& aResp,
...
+{
+    using AssociationsValType =
+        std::vector<std::tuple<std::string, std::string, std::string>>;
+    const AssociationsValType* associations = nullptr;
+    const uint64_t* timestamp = nullptr;
+    const std::string* msgPropVal = nullptr;
+    const std::string* severity = nullptr;
+
+    const bool success = sdbusplus::unpackPropertiesNoThrow(
+        dbus_utils::UnpackErrorPrinter(), properties, "Associations",
```
RameshIyyar, line 465: > dbus_utils::UnpackErrorPrinter()

In the old code, we are adding a trace with `BMCWEB_LOG_ERROR` if the property value is not found (nullptr) with the object path that we are trying to look but in the new code, are we adding a trace with `BMCWEB_LOG_DEBUG` and with the object path?

If yes, I would like to print in the ERROR log level (and object path) since we won't enable debug log level in the field (and also I think bmcweb debug log are huge so I think we disabled it, please confirm it).

redfish-core/include/utils/hw_isolation.hpp
```diff
@@ -585,136 +682,10 @@ inline void
...
-                        {
-                            // Failed to set the severity
-                            return;
-                        }
-                    }
-                }
-                },
-                objType[0].first, hwStatusEventObj,
-                "org.freedesktop.DBus.Properties", "GetAll", "");
+                // TODO
```
RameshIyyar, line 682: why TODO? anything is spending? if yes, please include what needs todo (+ issue id)

RameshIyyar (commented):
I have reviewed hw-isolation related code and have some questions.

redfish-core/include/utils/hw_isolation.hpp
deepakala-k: Sorry. That was a note to myself when it was WIP. Now it is done. I will remove it

redfish-core/include/utils/hw_isolation.hpp
```diff
@@ -452,6 +449,107 @@ inline bool setSeverity(const std::shared_ptr<bmcweb::AsyncResp>& aResp,
...
+    const std::string* severity = nullptr;
+
+    const bool success = sdbusplus::unpackPropertiesNoThrow(
+        dbus_utils::UnpackErrorPrinter(), properties, "Associations",
+        associations, "Timestamp", timestamp, "Message", msgPropVal, "Severity",
+        severity);
+
+    if (!success)
+    {
+        messages::internalError(aResp->res);
```
gtmills, line 471: Can you add a BMCWEB_LOG_ERROR here?

redfish-core/include/utils/hw_isolation.hpp
gtmills: Just add the BMCWEB_LOG_ERROR on line 471 ?

deepakala-k: Would it be better if we update the dbus_utils::UnpackErrorPrinter() to log BMCWEB_LOG_ERROR. That way it will work everywhere and the code would be generic. @RameshIyyar Could you please share your preference?

RameshIyyar: >Would it be better if we update the dbus_utils::UnpackErrorPrinter() to log BMCWEB_LOG_ERROR. That way it will work everywhere and the code would be generic.

I would prefer that only...

but, I am worried about the hw-isolation now, and even adding a trace in the "if (!success) {...}" won't solve the issue that is, how we will know which attribute failed to read?

redfish-core/include/utils/hw_isolation.hpp
```diff
@@ -452,6 +448,107 @@ inline bool setSeverity(const std::shared_ptr<bmcweb::AsyncResp>& aResp,
...
+        return;
+    }
+
+    if (associations != nullptr)
```
gtmills, line 474: ideally this would be 

    if (associations == nullptr)
    {
    BMCWEB_LOG_ERROR
    internalError;
    return;
    }
    for (const auto& assoc : *associations)
        {
            if (std::get<0>(assoc) == "error_log")
            {
                sdbusplus::message::object_path errPath = std::get<2>(assoc);
                // we have only one condition
                nlohmann::json::json_pointer logEntryPropPath =
                    "/Status/Conditions/0/LogEntry"_json_pointer;
                error_log_utils::setErrorLogUri(aResp, errPath,
                                                logEntryPropPath, true);
            }
        }

redfish-core/include/utils/hw_isolation.hpp
```diff
@@ -452,6 +448,107 @@ inline bool setSeverity(const std::shared_ptr<bmcweb::AsyncResp>& aResp,
...
+                sdbusplus::message::object_path errPath = std::get<2>(assoc);
+                // we have only one condition
+                nlohmann::json::json_pointer logEntryPropPath =
+                    "/Status/Conditions/0/LogEntry"_json_pointer;
+                error_log_utils::setErrorLogUri(aResp, errPath,
+                                                logEntryPropPath, true);
+            }
+        }
+    }
+    else if (timestamp != nullptr)
```
gtmills, line 489: Ideally this would be 
timestamp == nullptr

redfish-core/include/utils/hw_isolation.hpp
```diff
@@ -452,6 +448,107 @@ inline bool setSeverity(const std::shared_ptr<bmcweb::AsyncResp>& aResp,
...
+    {
+        condition["Timestamp"] = redfish::time_utils::getDateTimeStdtime(
+            static_cast<std::time_t>(*timestamp));
+    }
+    else if (msgPropVal != nullptr)
```
gtmills, line 494: Why are these else if?

redfish-core/include/utils/hw_isolation.hpp
gtmills: > if (!success) {...} won't solve the issue, right

It should though. How if we add  BMCWEB_LOG_ERROR to line 471 is this different than what we had before?


Before we had some nullptr checks, which we can add. sdbusplus::unpackPropertiesNoThrow is the new cleaner way to do this and I don't see how it has any less logging 

                        if (associations == nullptr)
                        {
                            BMCWEB_LOG_ERROR
                                << "Failed to get the Associations "
                                << "from object: " << hwStatusEventObj;
                            return;
                        }

RameshIyyar: > sdbusplus::unpackPropertiesNoThrow is the new cleaner way to do this

Agree, but`UnpackErrorPrinter` was added in the debug log trace which is my concern. @gtmills cannot we use the error log trace in the `dbus_utils::UnpackErrorPrinter` function? what is the problem? any issue with it?

If we used the ERROR log trace in the `dbus_utils::UnpackErrorPrinter` then we are good since [it](https://github.com/ibm-openbmc/bmcweb/blob/1050/redfish-core/include/utils/dbus_utils.hpp#L18) prints the failed property if we are hitting failure while [reading a property](https://github.com/openbmc/sdbusplus/blob/af9556707029ca30e4c2749502f965fef28d5388/include/sdbusplus/unpack_properties.hpp#L31) value from the `properties` container and we could add the object path inside `if (!success) {...}` so no extra changes required else we need to follow the old code for each property (like you posted above).

If we are adding a check for each property and trace with an error log level then, I am fine with it.

(Just a suggestion, if we are adding a check like this for each property then we could avoid the `unpackPropertiesNoThrow` call in this function since it is not required, we could use the existing code `for` loop with `properties`).

redfish-core/include/utils/hw_isolation.hpp
RameshIyyar: Hmm, good catch. we can only honor any one property @deepakala-k.

redfish-core/include/utils/hw_isolation.hpp
deepakala-k: I added another commit to address this. 
With this Unpack will point to the error that it had issues reading. And internal error will log the source file name and line number. So with this generic solution we should be able to figure out most of issue!!

redfish-core/include/utils/chassis_utils.hpp
```diff
@@ -178,27 +167,20 @@ inline void checkForAssemblyAssociations(
     using associationList =
         std::vector<std::tuple<std::string, std::string, std::string>>;
```
baemyung, line 168: It looks like - this associationList is also used in assembly.hpp.
Perhaps, can we put this into ./include/dbus_utility.hpp like

```
    using AssociationList  =
        std::vector<std::tuple<std::string, std::string, std::string>>;
```

And implement a `dbus::utility::getAssociationList()` - using `getProperty<AssociationList>`?

deepakala-k:
Redfish Validator does not have any additional errors.
https://sys-openbmc-dev-jenkins.swg-devops.com/job/OpenBMC-Dev/job/test/job/common/job/Redfish-Service-Validator/18287/

redfish-core/include/utils/dbus_utils.hpp
```diff
@@ -14,7 +14,7 @@ struct UnpackErrorPrinter
     void operator()(const sdbusplus::UnpackErrorReason reason,
                     const std::string& property) const noexcept
     {
-        BMCWEB_LOG_DEBUG
+        BMCWEB_LOG_ERROR
```
RameshIyyar, line 17: It can go upstream, right?

include/dbus_utility.hpp
```diff
@@ -210,5 +214,19 @@ inline void
         "xyz.openbmc_project.ObjectMapper", "GetObject", path, interfaces);
 }
 
+inline void
+    getAssociationList(const std::string& service, const std::string& path,
```
RameshIyyar, line 218: It can go upstream, right?

RameshIyyar (commented):
it was missed while posting comments...

include/dbus_utility.hpp
baemyung: I think it seems good to be in upstream.

redfish-core/include/utils/hw_isolation.hpp
```diff
@@ -452,6 +452,110 @@ inline bool setSeverity(const std::shared_ptr<bmcweb::AsyncResp>& aResp,
...
+        associations, "Timestamp", timestamp, "Message", msgPropVal, "Severity",
+        severity);
+
+    if (!success)
+    {
+        messages::internalError(aResp->res);
+        return;
+    }
+
+    if (associations != nullptr)
```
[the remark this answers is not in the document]
gtmills: always best to have nullptr check so I vote leave it as is. :)

baemyung (approved):

redfish-core/lib/memory.hpp
```diff
@@ -727,35 +727,33 @@ inline void getDimmPartitionData(std::shared_ptr<bmcweb::AsyncResp> aResp,
 inline void getObjectEnable(std::shared_ptr<bmcweb::AsyncResp> aResp,
                             const std::string& service, const std::string& path)
 {
-    crow::connections::systemBus->async_method_call(
+    sdbusplus::asio::getAllProperties(
```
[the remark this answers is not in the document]
deepakala-k: I too felt that. But initially I was worried about changing the logic. My initial intention was to just convet "GetAll" to getAllProperties. I can update it now as we have modified a lot of code

RameshIyyar (approved):

include/dbus_utility.hpp
deepakala-k: But we dont have code that calls  get with "xyz.openbmc_project.Association.Definitions", "Associations" in the upstream. So would they accept to introduce this method?

deepakala-k:
Upstream commit : https://gerrit.openbmc.org/c/openbmc/bmcweb/+/61575

gtmills (approved):

PR #601, opened 2023-03-02
Read patch data earlier during an event log patch

spinler:
This fixes MalformedJSON errors when trying to patch the Resolved or ManagementSystemAck properties in an EventLog or CELog.

The first 8 bytes of the request JSON data in the Request structure was corrupt if it was read after an async D-Bus method call.  Testing showed that up until the method call was made, it was fine.

This commit changes the code to access that patch data earlier, before any D-Bus calls are made.  This makes it similar to how it was done for the Resolved property in the upstream code.

The code had been refactored when the CELog entries were added downstream to give more commonality, and that is where the change had been made to read the patch JSON data after reading the Hidden property on D-Bus.  It did work fine then though, and it also even worked when the commits were first pulled into 1050.

I'm not sure what changed to make it stop working.  Maybe it was never valid to access the Request data after code gets back to the event loop?

Change-Id: I4872558d42120321aae11f2eddb949e0a14a1886

### DIFF
--- a/redfish-core/lib/log_services.hpp
+++ b/redfish-core/lib/log_services.hpp
@@ -1896,21 +1896,11 @@ void getHiddenPropertyValue(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
         "org.open_power.Logging.PEL.Entry", "Hidden");
 }
 
-inline void updateProperty(const crow::Request& req,
+inline void updateProperty(const std::optional<bool>& resolved,
+                           const std::optional<bool>& managementSystemAck,
                            const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
                            const std::string& entryId)
 {
-    std::optional<bool> resolved;
-    std::optional<nlohmann::json> oemObject;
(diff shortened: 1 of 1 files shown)


redfish-core/lib/log_services.hpp
```diff
@@ -2675,14 +2650,40 @@ inline void requestRoutesDBusEventLogEntry(App& app)
             return;
         }
 
-        auto updatePropertyCallback =
-            [&req, asyncResp, entryId](bool hiddenPropVal) {
+        std::optional<bool> resolved;
```
spinler, line 2653: I guess I could have put this code into a function and just called it from the EventLog and CELog patch functions.

gtmills: Yeah, that would be cleaner, up to you

gtmills (approved):

redfish-core/lib/log_services.hpp
spinler: I think I'll just leave it how it is.

PR #604, opened 2023-03-05
1050: Remove Inventory.Item.Board interface from Chassis (#415)

baemyung:
IBM does not use the xyz.openbmc_project.Inventory.Item.Board interface for chassis objects. 
Remove that interfaces so only objects implementing xyz.openbmc_project.Inventory.Item.Chassis are returned for getValidChassisPath

### DIFF
--- a/redfish-core/include/utils/chassis_utils.hpp
+++ b/redfish-core/include/utils/chassis_utils.hpp
@@ -16,8 +16,7 @@ void getValidChassisPath(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
                          const std::string& chassisId, Callback&& callback)
 {
     BMCWEB_LOG_DEBUG << "checkChassisId enter";
-    const std::array<const char*, 2> interfaces = {
-        "xyz.openbmc_project.Inventory.Item.Board",
+    const std::array<const char*, 1> interfaces = {
         "xyz.openbmc_project.Inventory.Item.Chassis"};
 
     auto respHandler =


redfish-core/include/utils/chassis_utils.hpp
```diff
@@ -16,8 +16,7 @@ void getValidChassisPath(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
                          const std::string& chassisId, Callback&& callback)
 {
     BMCWEB_LOG_DEBUG << "checkChassisId enter";
-    const std::array<const char*, 2> interfaces = {
-        "xyz.openbmc_project.Inventory.Item.Board",
+    const std::array<const char*, 1> interfaces = {
         "xyz.openbmc_project.Inventory.Item.Chassis"};
```
gtmills, line 20: Do you need to do this anywhere else? Or all other places are taken care of ?

gtmills (approved):
I am good as long as this doesn't need to happen anywhere else

redfish-core/include/utils/chassis_utils.hpp
baemyung: @gtmills 
A few questions.
This PR is same as #415.
However, there are a few additional places that uses Inventory.Item.Board in 1050.

- redfish-core/lib/storage.hpp
- redfish-core/lib/thermal_metrics.hpp

Are these related to Chassis, and also need to remove? Or leave as they are?

PR #270, opened 2022-03-02
GET PATCH userGroups Information ("AccountTypes")

patelabhishek9893:
This commit enhances the redfish API to retrieve and set userGroups
information for each user account.

"Redfish" is always enabled in each user role. That's why it was
hardcoded into JSON response data in the old redfish API, where now it
gets retrieved using dbus interface xyz.openbmc_project.User.Attributes.
UserGroups.

UserGroups retrieve data are "redfish", "ssh", "web", and "ipmi", where
redfish DMTF Schema has predefined enum type as described below.

  "AccountTypes": {
    "enum":["Redfish", "SNMP", "OEM", "HostConsole", "ManagerConsole",
            "IPMI", "KVMIP", "VirtualMedia", "WebUI"]
   }

Here UserGroups ssh is mapped with two AccountTypes "HostConsole",
"ManagerConsole".

  - Redfish ManagerConsole == SSH to port 22 == Phosphor User manager
    Phosphor “ssh” privilege
  - Redfish HostConsole == SSH to port 2200 (host console), Which
    OpenBMC implements using the Phosphor User manager Phosphor “ssh”
    privilege.

This commit also enhances the redfish API to set and unset userGroups
information for each user account.

Users with ConfigureUsers level privilege can patch (Set and Unset)
AccountTypes of each user role. In addition, a user with
"ConfigureSelf" level privilege can only set or Update their password.

"Redfish" is always enabled in each user role. However,
"ConfigureUsers" can disable other user redfish services. But if
"ConfigureUsers" try to disable its redfish service, that generates an
error.

In this commit, users can enable and disable "redfish", "ssh", and
"ipmi" services from each user.

Tested: tested using the curl command. test using both privileges
"ConfigureUsers", "ConfigureSelf"

curl -k -H "X-Auth-Token: $token" -X PATCH [https://${bmc}/redfish/v1/](https://%24%7Bbmc%7D/redfish/v1/)
  AccountService/Accounts/operator_user -d '{"AccountTypes":["IPMI",
  "HostConsole","ManagerConsole","ManagerConsole"]}'
  
curl -k -H "X-Auth-Token: $token" -X GET [https://${bmc}/redfish/v1/](https://%24%7Bbmc%7D/redfish/v1/)
AccountService/Accounts/admin
{
  "@[odata.id](http://odata.id/)": "/redfish/v1/AccountService/Accounts/admin",
  "@odata.type": "#ManagerAccount.v1_4_0.ManagerAccount",
  "AccountTypes": [
[...]

PR #614, opened 2023-03-13
1050: GET PATCH userGroups Information ("AccountTypes") (#270)(#284)

baemyung:
This commit enhances the redfish API to set and unset userGroups information for each user account.

UserGroups retrieve data are "redfish", "ssh", "web", and "ipmi", where redfish DMTF Schema has predefined enum type as described below.

```
  "AccountTypes": {
    "enum":["Redfish", "SNMP", "OEM", "HostConsole", "ManagerConsole",
            "IPMI", "KVMIP", "VirtualMedia", "WebUI"]
   }
```

Users with ConfigureUsers level privilege can patch (Set and Unset) AccountTypes of each user role. In addition, a user with "ConfigureSelf" level privilege can only set or Update their password.

"Redfish" is always enabled in each user role. However, "ConfigureUsers" can disable other user redfish services. But if "ConfigureUsers" try to disable its redfish service, that generates an error.

"ssh" userGroup has to be enabled in Administrator, So if the user (admin or non-admin) tries to disable those, then it generates an error.

In this commit, users can enable and disable "redfish", "ssh", and "ipmi" services from each user.

Tested 1:
Pass Jenkins Validator

Tested 2:
```
curl -k -H "X-Auth-Token: $token" -X GET https://${bmc}/redfish/v1/AccountService/Accounts/admin
{
  "@odata.id": "/redfish/v1/AccountService/Accounts/admin",
  "@odata.type": "#ManagerAccount.v1_7_0.ManagerAccount",
  "AccountTypes": [
    "IPMI",
    "Redfish",
    "HostConsole",
    "ManagerConsole"
  ],
  "Description": "User Account",
  "Enabled": true,
  "Id": "admin",
  "Links": {
    "Role": {
      "@odata.id": "/redfish/v1/AccountService/Roles/Administrator"
    }
  },

    ...
    ...
}
```

Tested 3: tested using the curl command. test using both privilege
"ConfigureUsers", "ConfigureSelf"

```
operator_user=admin
curl -k -H "X-Auth-Token: $token" -X PATCH https://${bmc}/redfish/v1/AccountService/Accounts/${operator_user} -d '{"AccountTypes":["IPMI","HostConsole","ManagerConsole"]}'
{
  "@Message.ExtendedInfo": [
    {
      "Message": "Successfully Completed Request",
      "MessageArgs": [],
      "MessageId": "Base.1.8.1.Success",
      "MessageSeverity": "OK",
      "Resolution": "None"
    }
  ]
}
```

Change-Id: I1a0344ca45556b820bb77c3dcb459f27eb032501

### DIFF
--- a/redfish-core/lib/account_service.hpp
+++ b/redfish-core/lib/account_service.hpp
@@ -101,6 +101,7 @@ inline std::string getRoleIdFromPrivilege(std::string_view role)
     }
     return "";
 }
+
 inline std::string getPrivilegeFromRoleId(std::string_view role)
 {
     if (role == "Administrator")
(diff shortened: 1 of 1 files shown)


redfish-core/lib/account_service.hpp
```diff
@@ -126,55 +127,181 @@ inline std::string getPrivilegeFromRoleId(std::string_view role)
...
+        accountTypes.push_back("Redfish");
+    }
+    else if (userGroup == "ipmi")
+    {
+        accountTypes.push_back("IPMI");
+    }
+    else if (userGroup == "ssh")
     {
-        if (userGroup == "redfish")
+        accountTypes.push_back("HostConsole");
```
gtmills, line 146: @joseph-reynolds do we need a downstream patch to look at that downstream HostConsole group?

gtmills: This is what we had for 1030, we can address later

gtmills (approved):

PR #618, opened 2023-03-14
1050: Follow-up the upstream schema handling approach

baemyung:
This updates schema handling as to upstream

- Generate *json files from DMTF schema file
- generate redfish-core/include/*.hpp & redfish-core/include/generated/enums/*.hpp from schema

This upstream method also contains

- Aggregation: Generate collections from schema https://github.com/openbmc/bmcweb/commit/9e03140656b1d93c2f4d0f3894eee8d664eb646e

In aggregation we need to account for when a satellite BMC supports a resource collection that the aggregating BMC does not.  We need to add links in responses from upstream resources to indicate that these satellite only resources exist.  These top level collections do not always have the location format of /redfish/v1/<collection_id>.  We determine all of those from the schema rather than hardcoding a few choice exceptions that do not fit that format such as /redfish/v1/UpdateService/FirmwareInventory.

This patch is the first step in accomplishing this.  We parse all xml schema files in order to determine what are the top level collection URIs.  For URIs that are upstream from top level collections we'll be able to derive what URIs could appear in the response which are also on the same path to a top level collection.

The xml files for resources that aren't currently supported by bmcweb are ignored when creating the schema index and later deleted once the above parsing completes.

Later patches will use this information to make sure that we are properly handling all schema defined top level collections.

Tested:
Compiles with aggregation enabled

Change-Id: Icff825f4e4e9d84c96254561350f82839abdc074

- Fix OemAccountService to OpenBMCAccountService https://github.com/openbmc/bmcweb/commit/5b5574ace856c7f787528ab8f5e5352a90f2a6cb

This schema as-written broke a bunch of Redfish specification rules for schemas, naming and other things.  This commit starts to fix them.

It can be used as a starting point for fixing the existing schemas. This working through testing.

Tested: OpenBMCAccountService_v1.xml now passes the redfish odata schema validator.

Change-Id: Id991794344b8b7d36af3c44bba7214bf025deec8


gtmills:
What upstream commits does this all include?

baemyung:
> What upstream commits does this all include?

This includes the all schemas with VERSION = "DSP8010_2022.2".
The new version VERSION = "DSP8010_2022.3" may be done later.

gtmills:
>>     What upstream commits does this all include?

> This includes the all schemas with VERSION = "DSP8010_2022.2".
> The new version VERSION = "DSP8010_2022.3" may be done later.

Looking for the upstream gerrit commits here

baemyung:
> 

The schema related upstream changes  are:

https://gerrit.openbmc.org/c/openbmc/bmcweb/+/60757
https://gerrit.openbmc.org/c/openbmc/bmcweb/+/61064
https://gerrit.openbmc.org/c/openbmc/bmcweb/+/60223
https://gerrit.openbmc.org/c/openbmc/bmcweb/+/57441
https://gerrit.openbmc.org/c/openbmc/bmcweb/+/55215

gtmills (approved):

PR #623, opened 2023-03-16
1050: Add LocationIndicator support for fabric adapters

jpallenibm:
This update adds the GET and PATCH support for LocationIndicatorActive to fabric adapters. Only some fabric adapters support this option.

I manually tested this using the following commands:

curl -k -H "X-Auth-Token: $token" -X  GET https://${bmc}/redfish/v1/Systems/system/FabricAdapters/chassis-motherboard-disk_backplane0

and

curl -k -H "X-Auth-Token: $token" -X  PATCH -d '{"LocationIndicatorActive":true}' https://${bmc}/redfish/v1/Systems/system/FabricAdapters/chassis-motherboard-disk_backplane0

The first GET command returned LocationIndicatorActive as false. The PATCH command completed without error.
The subsequent GET command returned LocationIndicatorActive as true.

### DIFF
--- a/redfish-core/lib/fabric_adapters.hpp
+++ b/redfish-core/lib/fabric_adapters.hpp
@@ -238,45 +238,6 @@ inline void doAdapterGet(const std::shared_ptr<bmcweb::AsyncResp>& aResp,
     }
 }
 
-inline void doAdapterPatch(const crow::Request& req,
-			   const std::shared_ptr<bmcweb::AsyncResp>& aResp,
-			   const std::string& adapterId,
-			   const std::string& fabricAdapterPath)
-{
-    std::optional<bool> locationIndicatorActive;
-        
-    if (!json_util::readJsonPatch(req, aResp->res,
-			     "LocationIndicatorActive",
-			     locationIndicatorActive))
-    {
-      return;
-    }
-    
-    crow::connections::systemBus->async_method_call(
-       [fabricAdapterPath, adapterId,locationIndicatorActive, aResp](
-	 const boost::system::error_code ec) {
-	 if (ec)
-	 {
-	     handleAdapterError(ec, aResp->res, adapterId);
-	     return;
-	 }
-	 
-	 if (locationIndicatorActive)
-	 {
-	     setLocationIndicatorActive(
-					aResp, fabricAdapterPath,
-					*locationIndicatorActive);
-	 }
-    
-	},
-	"xyz.openbmc_project.ObjectMapper",
-	"/xyz/openbmc_project/object_mapper",
-	"xyz.openbmc_project.ObjectMapper", "GetSubTree",
-	"/xyz/openbmc_project/inventory", 0,
-	std::array<const char*, 1>{
-	  "xyz.openbmc_project.Inventory.Item.FabricAdapter"});
-}
-
 inline void getValidFabricAdapterPath(
     const std::string& adapterId, const std::string& systemName,
     const std::shared_ptr<bmcweb::AsyncResp>& aResp,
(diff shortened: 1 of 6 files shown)


gtmills:
add to approvelist

redfish-core/lib/fabric_adapters.hpp
```diff
@@ -236,6 +238,45 @@ inline void doAdapterGet(const std::shared_ptr<bmcweb::AsyncResp>& aResp,
...
+	 {
+	     setLocationIndicatorActive(
+					aResp, fabricAdapterPath,
+					*locationIndicatorActive);
+	 }
+    
+	},
+	"xyz.openbmc_project.ObjectMapper",
+	"/xyz/openbmc_project/object_mapper",
+	"xyz.openbmc_project.ObjectMapper", "GetSubTree",
```
gtmills, line 274: doAdapterPatch already did a GetSubTree? Don't you already know the fabricAdapterPath ?

redfish-core/lib/fabric_adapters.hpp
```diff
@@ -236,6 +238,45 @@ inline void doAdapterGet(const std::shared_ptr<bmcweb::AsyncResp>& aResp,
...
+{
+    std::optional<bool> locationIndicatorActive;
+        
+    if (!json_util::readJsonPatch(req, aResp->res,
+			     "LocationIndicatorActive",
+			     locationIndicatorActive))
+    {
+      return;
+    }
+
```
gtmills, line 254: Can this be easy simple like 
```
inline void doAdapterPatch(const crow::Request& req,
			   const std::shared_ptr<bmcweb::AsyncResp>& aResp,
			   const std::string& adapterId,
			   const std::string& fabricAdapterPath)
{
    std::optional<bool> locationIndicatorActive;

    if (!json_util::readJsonPatch(req, aResp->res,
			     "LocationIndicatorActive",
			     locationIndicatorActive))
    {
      return;
    }
            if (locationIndicatorActive)
            {
                setLocationIndicatorActive(aResp, fabricAdapterPath,
                                           *locationIndicatorActive);
            }
```

redfish-core/lib/fabric_adapters.hpp
```diff
@@ -225,6 +226,7 @@ inline void doAdapterGet(const std::shared_ptr<bmcweb::AsyncResp>& aResp,
     getFabricAdapterAsset(aResp, serviceName, fabricAdapterPath);
     getFabricAdapterState(aResp, serviceName, fabricAdapterPath);
     getFabricAdapterHealth(aResp, serviceName, fabricAdapterPath);
+    getLocationIndicatorActive(aResp, fabricAdapterPath);
```
gtmills, line 229: Can you debug more why we are getting this property even when it is null?

jpallenibm: Yes, I'm still investigating that issue.

redfish-core/lib/fabric_adapters.hpp
jpallenibm: Do I need the systemBus->async_method_call to send this to the backend?  Frank's changes originally did this, but this code has been refactored since then. So is that not needed>

redfish-core/lib/fabric_adapters.hpp
jpallenibm: This is resolved in the change to the issues below

redfish-core/lib/fabric_adapters.hpp
jpallenibm: I verified this simplified version does work!  Thanks

baemyung: The current code is still doing another level of GetSubTree inside doAdapterPatch().
So, this would need to be rewritten as to what @gtmills suggested.
Also, readJsonPatch() is better to be invoked at the outer function to avoid the multiple calls inside doAdapterPatch().

If do so, doAdapterPatch() becomes a very small # of codes. 
Then, doAdapterPatch() part can be merged into handleFabricAdapterPatch().

So, this will become like this.

```

inline void
handleFabricAdapterPatch(App &app, const crow::Request &req,
                         const std::shared_ptr<bmcweb::AsyncResp> &aResp,
                         const std::string &systemName,
                         const std::string &adapterId) {
  if (!redfish::setUpRedfishRoute(app, req, aResp)) {
    return;
  }

  std::optional<bool> locationIndicatorActive;

  if (!json_util::readJsonPatch(req, aResp->res, "LocationIndicatorActive",
                                locationIndicatorActive)) {
    return;
  }

  getValidFabricAdapterPath(
      adapterId, systemName, aResp,
      [req, aResp, adapterId, locationIndicatorActive](
          const std::string &fabricAdapterPath,
          [[maybe_unused]] const std::string &serviceName,
          [[maybe_unused]] const dbus::utility::InterfaceList &interfaces) {
        if (locationIndicatorActive) {
          setLocationIndicatorActive(aResp, fabricAdapterPath,
                                     *locationIndicatorActive);
        }
      });
}
```

baemyung:
Another one is to run the formatter to avoid the CI failure.

`$ clang-format-15 -i fabric_adapters.hpp`

redfish-core/lib/fabric_adapters.hpp
jpallenibm: This has been removed in latest commit.

redfish-core/lib/fabric_adapters.hpp
jpallenibm: These changes have been made in latest commit.

PR #625, opened 2023-03-16
Fix issues while creating dumps

asmithakarun:
There were a few changes missed during the rebase activity

This change will fix the following issues:
* BMC dumps on successful creation is reported as Resource_<bmc dump id> in the task header
* Resource creation fails to initiate with "Wrong parameter values passed" error

Tested-By:

* Created a BMC dump and checked the task URI associated with this dump creation for correct bmc dump id in the "Location" field
* Successful reation of a resource dump

### DIFF
--- a/redfish-core/include/resource_messages.hpp
+++ b/redfish-core/include/resource_messages.hpp
@@ -37,7 +37,7 @@ inline nlohmann::json resourceCreated()
 inline nlohmann::json resourceRemoved()
 {
     return getLogResourceEvent(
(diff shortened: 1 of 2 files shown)


sunharis (approved):

gtmills (approved):

gtmills:
Do we need to push upstream ?

asmithakarun:
> Do we need to push upstream ?

https://gerrit.openbmc.org/c/openbmc/bmcweb/+/61692 I pushed it now

PR #291, opened 2022-03-18
Enhance Name Utils Loggin

patelabhishek9893:
getPrettyName method gets called more than one time.
required Service information to debug the error messages.
Enhancing util logging message to retire service information.

Signed-off-by: Abhishek Patel <Abhishek.Patel@ibm.com>

PR #626, opened 2023-03-16
1050: Logging invalid service name for debug (#291)

baemyung:
The method getPrettyName() may be called more than one time.
This will generate journal logging messages for the required Service name
if there are unexpected service names.


gtmills:
Can we push upstream ?

baemyung:
> Can we push upstream ?

Upstream does not use getPrettyName().
So, we can push getPrettyName() and its usages to upstream.
I'll update this when ready for it.

gtmills (approved):

baemyung:
> 
Upstream: https://gerrit.openbmc.org/c/openbmc/bmcweb/+/43967
However, the community decided to abandon.

PR #627, opened 2023-03-16
Remove the default assembly Health

gtmills:
Needs to go into 1040 
Defect is https://jazz07.rchland.ibm.com:13443/jazz/web/projects/CSSD#action=com.ibm.team.workitem.viewWorkItem&id=454681

Since it is different lines, I'll push the 1050 myself. 
https://github.com/ibm-openbmc/bmcweb/blob/1050/redfish-core/lib/assembly.hpp#L105

Have a defect where assemblies for the I/O expansion chassis's assemblies don't have an operationalStatus because we don't know the health of the FRU but we still have a Redfish Health for these FRUs. Upstream we have been default Health and State, the thinking is "Hey I reported this inventory item, mark it as good by default.". Upstream maybe we should do something different.

### DIFF
--- a/redfish-core/lib/assembly.hpp
+++ b/redfish-core/lib/assembly.hpp
@@ -60,9 +60,6 @@ inline void
         tempyArray.at(assemblyIndex)["Name"] =
             sdbusplus::message::object_path(assembly).filename();
 
-        // Set the default Status
-        tempyArray.at(assemblyIndex)["Status"]["Health"] = "OK";
-
         // Handle special case for tod_battery assembly OEM ReadyToRemove
         // property NOTE: The following method for the special case of the
         // tod_battery ReadyToRemove property only works when there is only ONE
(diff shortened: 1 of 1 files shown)


gtmills:
Looks like (a assembly without a OperationStatus) 
```
{
"@odata.id": ["/redfish/v1/Chassis/chassis15873/Assembly#/Assemblies/2"](https://z1634abmc.rch.stglabs.ibm.com:18080/redfish/v1/Chassis/chassis15873/Assembly#/Assemblies/2),
"@odata.type": "#Assembly.v1_3_0.AssemblyData",
"Location": {
"PartLocation": {
"ServiceLabel": "U50EE.001.WZS000H-P3"
}
},
"LocationIndicatorActive": null,
"MemberId": "2",
"Name": "motherboard3",
"Status": {
"State": "Enabled"
}
},
```

PR #628, opened 2023-03-16
1050: Remove the default assembly Health

gtmills:
Defect is https://jazz07.rchland.ibm.com:13443/jazz/web/projects/CSSD#action=com.ibm.team.workitem.viewWorkItem&id=454681

Have a defect where assemblies for the I/O expansion chassis's assemblies don't have an operationalStatus because we don't know the health of the FRU but we still have a Redfish Health for these FRUs. Upstream we have been default Health and State, the thinking is "Hey I reported this inventory item, mark it as good by default.". Upstream maybe we should do something different.

Didn't test directly but tested 1030.

https://github.com/ibm-openbmc/bmcweb/pull/627/ is the 1030/1040 PR.

### DIFF
--- a/redfish-core/lib/assembly.hpp
+++ b/redfish-core/lib/assembly.hpp
@@ -101,9 +101,6 @@ inline void
         tempyArray.at(assemblyIndex)["Name"] =
             sdbusplus::message::object_path(assembly).filename();
 
-        // Set the default Status
-        tempyArray.at(assemblyIndex)["Status"]["Health"] = "OK";
-
         // Handle special case for tod_battery assembly OEM ReadyToRemove
         // property NOTE: The following method for the special case of the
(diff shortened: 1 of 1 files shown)


baemyung:
Isn't the value of null still valid?

`https://www.dmtf.org/sites/default/files/standards/documents/DSP0268_2022.1.pdf`
says:

> Attributes - Designates whether the property is read-only or read-write, if supported by the implementation, and
> whether the service might return a null value if the property value is temporarily unavailable.
> 
> And 

> LocationIndicatorActive - boolean - read-write (null) - This property shall contain the state of the indicator used to physically identify
> or locate this resource.

Does this mean that it is expected to have "null" if it is yet evaluated?

baemyung (approved):
It looks good

PR #629, opened 2023-03-16
Add Support for privilege check in handleUpgrade

baemyung:
This commit enables privilege check for user(s) in case of upgraded connections.
Currently users with no privileges will also be able to access Websockets connections (Ex: KVM).

The privilege check was already in place for normal connections (i.e. router->handle()). This commit lifts off the privilege check code and moves it into a common function (validatePrivilege()), which can be used both by handle() and handleUpgrade() and register required callback to be called.

Also, the const qualifier for Request in the handleUpgrade() function's signature is removed to enable setting "isConfigureSelf" field of request. The signature of handleUpgrade() is made identical to handle()

Tested:
 - websocket_test.py Passed
 - Admin and Operator users are able to access KVM on WebUI
 - Readonly User was unable to access KVM on WebUI

Change-Id: I6f743c27e7e6077f1c6c56e6958922027e4404e8

### DIFF
--- a/http/app.hpp
+++ b/http/app.hpp
@@ -58,9 +58,11 @@ class App
     App& operator=(const App&&) = delete;
 
     template <typename Adaptor>
-    void handleUpgrade(const Request& req, Response& res, Adaptor&& adaptor)
+    void handleUpgrade(Request& req,
+                       const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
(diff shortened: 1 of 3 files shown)


baemyung:
I need more testing for this.

baemyung:
CI keeps failing without good explanation.
So, I'll redo via another clean PR.

PR #635, opened 2023-03-22
Fix obmc-console handling

baemyung:
obmc .onopen is doing the unecessary connection attempt because the user role validation.

This will validate the user first and then allow the connection.

Tested:
 - Redfish validator passed.
 - websocket_test.py
 - Admin and Operator users are able to access console on WebUI
 - Readonly User was unable to access console on WebUI

Example test method of 'console' on webui:
 - Use web-browser, and open 'inspect' sub-page tab
 - Under 'admin',  access & switch between https://${bmc-ip}:18080/#/operations/service-login https://${bmc-ip}:18080/#/operations/host-console

### DIFF
--- a/include/obmc_console.hpp
+++ b/include/obmc_console.hpp
@@ -5,6 +5,7 @@
 #include <async_resp.hpp>
 #include <boost/asio/local/stream_protocol.hpp>
 #include <boost/container/flat_set.hpp>
+#include <dbus_utility.hpp>
 #include <privileges.hpp>
 #include <websocket.hpp>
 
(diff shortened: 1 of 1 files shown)


include/obmc_console.hpp
```diff
@@ -118,29 +119,12 @@ inline void requestRoutes(App& app)
     BMCWEB_ROUTE(app, "/console0")
         .privileges({{"ConfigureManager"}})
         .websocket()
-        .onopen(
-            [](crow::websocket::Connection& conn) {
-        BMCWEB_LOG_DEBUG << "Connection " << &conn << " opened";
-
-        sessions.insert(&conn);
-        if (hostSocket == nullptr)
```
gtmills, line 126: Why remove the hostSocket==nullptr check?

gtmills:
Can you add what you told me in slack and on scrum about why we want this? 
E.g. this is what we had for 1030, etc.

baemyung (commented):
The new code flow is also the same as 1030, by validating the user and allowing the connection.

baemyung:
> 
The new code flow is the same as 1030, by validating the user and then allowing the connection.

include/obmc_console.hpp
baemyung: The check is done after the user validation at the line 169.

```
            const ::redfish::Privileges requiredPrivileges{"ConfigureManager"};
            if (!userPrivileges.isSupersetOf(requiredPrivileges))
            {
...
                return;
            }
            sessions.insert(&conn);
L169>    if (hostSocket == nullptr)
            {
                const std::string consoleName("\0obmc-console", 13);
                boost::asio::local::stream_protocol::endpoint ep(consoleName);
                hostSocket = std::make_unique<
                    boost::asio::local::stream_protocol::socket>(
                    conn.getIoContext());
            }
```

gtmills (approved):

baemyung:
Sorry about the error.
It turns out - this fix is incorrect and causing bmcweb failure.
Can we remove this from merge?

PR #636, opened 2023-03-22
1050: Update schema pack to 2022.3

baemyung:
Update scripts/update_schemas.py to point at 2022.3 and run.

Schema pack 2022.3 is the latest Redfish release, released 01/23/2023. It contains several new schemas and support for Multi-factor Authentication and CXL support along other things.

Update done by automation.
Schema 'AggregationSource' (which was omitted before) is also added. 

Tested: Redfish validator passed with 1050-ghe/openbmc.

Change-Id: I55a64d7cda26572e7b75135acc324cb44d316fe6


gtmills:
Can you add why this is needed?

baemyung:
> Can you add why this is needed?

- I missed one schema 'AggregationSource' in the previous commit, and I added it.   I think this would be needed for the further Aggregation development.
- Upstream is using 2022.3, and it may be easier by using the same one.
- Some seem based on the new schema set (?)

gtmills (approved):

PR #643, opened 2023-03-27
Handle patchCpuCoreMembers request

deepakala-k:
This was missed while rebasing 1030 changes. NA for Upstream as we do not have the method to patch Cpu Core Members on upstream
It is fixed now and handles 
curl -k -H "X-Auth-Token: $bmc_token" -X PATCH
https://${BMC_IP}/redfish/v1/Systems/system/Processors/dcm0-cpu0/SubProcessors/core8 -d '{"Enabled" : <true/false> }' successfully

### DIFF
--- a/redfish-core/lib/processor.hpp
+++ b/redfish-core/lib/processor.hpp
@@ -1991,6 +1991,11 @@ inline void requestRoutesSubProcessorsCore(App& app)
 
         getSubProcessorData(asyncResp, processorId, coreId);
         });
+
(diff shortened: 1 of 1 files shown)


raviteja-b:
do we have upstream commit for this one?

deepakala-k:
> do we have upstream commit for this one?

No. As this function is not part of upstream

raviteja-b (approved):

gtmills (approved):

PR #652, opened 2023-04-04
Fix dump offload issue

asmithakarun:
The dumps were not getting offloaded and this is because the route handlers for the same were not invoked on a GET request.

This issue is resolved by modifying the dump offload code as per the upstream http connection framework and allows clients to offload dumps using redfish URI.

This PR:
1. Enables dynamic response support
2. Defines new bmcweb rule to support dump streaming.
3. Adds routes for bmc and system dump offload.

The existing http connection is implemented based on http string buffer and now the http stream connection is implemented using http dynamic buffer body that enables dump data streaming over https. This dynamic buffer sends the data size that is to be received by the client in the http header, so that the client waits till data with the given size is received before the http connection is closed. Added DynamicResponse struct to deal with dynamic buffer body response type.

Tested By:
[1] GET https://${bmc}/redfish/v1/Managers/bmc/LogServices/Dump/Entries/< dumpId >/attachment  >  < filename >

Change-Id: I08fe2840f4f6e4d9b8f53a2f72f7eb7714c6aeb0

### DIFF
--- a/http/http_connection.hpp
+++ b/http/http_connection.hpp
@@ -390,6 +390,19 @@ class Connection :
             asyncResp->res.setCompleteRequestHandler(nullptr);
             return;
         }
+
+        std::string url(req->target());
+        if (boost::contains(url, "/Dump/Entries/") &&
+            boost::ends_with(url, "/attachment"))
+        {
+            BMCWEB_LOG_DEBUG << "upgrade stream connection";
+            handler->handleUpgrade(*req, res, std::move(adaptor));
+            // delete lambda with self shared_ptr
+            // to enable connection destruction
+            res.completeRequestHandler = nullptr;
+            return;
+        }
+
         std::string_view expected =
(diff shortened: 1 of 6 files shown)


gtmills:
What is the upstream story here? Why so many changes ?

raviteja-b (approved):

sunharis (approved):
Changes looks good. Please verify other bmcweb basic use cases - GET, POST, Save-area file uploads and Events are working fine with this http files (though there is no visible impact)

asmithakarun:
> @asmithakarun CI failed - seems that needs changes as per the new bmcweb coding style.

Done

gtmills:
Can you add more to the PR msg? Explain a bit of the DynamicResponse / StreamingResponseRule. Upstream any commit with any complexity has a few paragraphs, e.g. https://gerrit.openbmc.org/c/openbmc/bmcweb/+/59940

PR #660, opened 2023-04-11
IBM GHE only: Fix IPv4 address configuration

raviteja-b:
This commit avoids IPv4 DHCP and Static address co-existence 
Disables DHCPv4 while configuring ipv4 static addresses and 
Deletes ipv4 static addresses while enabling DHCPv4.

This behaviour needs to be fixed in the backend Network D-bus design.
This is downstream only change needed untill networkd fixes.

Tested by:
Enable DHCPv4 when ipv4 static addresses configured on the interface.
Configure IPv4 static addresses when DHCPv4 enabled on the interface

### DIFF
--- a/redfish-core/lib/ethernet.hpp
+++ b/redfish-core/lib/ethernet.hpp
@@ -1103,6 +1103,70 @@ inline void
         dbus::utility::DbusVariantType(macAddress));
 }
 
+void deleteIPv4StaticAddresses(
+    const std::string& ethifaceId,
+    const std::shared_ptr<bmcweb::AsyncResp>& asyncResp)
+{
+    crow::connections::systemBus->async_method_call(
+        [asyncResp, ethifaceId](const boost::system::error_code& errorCode,
+                                dbus::utility::ManagedObjectType& resp) {
+        if (errorCode)
+        {
+            BMCWEB_LOG_ERROR << "D-Bus response error: " << errorCode;
+            messages::internalError(asyncResp->res);
+        }
+        const std::string ipPathStart =
+            "/xyz/openbmc_project/network/" + ethifaceId;
+
+        for (const auto& objpath : resp)
+        {
+            if (objpath.first.str.starts_with(ipPathStart + "/"))
+            {
+                for (const auto& interface : objpath.second)
+                {
+                    if (interface.first == "xyz.openbmc_project.Network.IP")
+                    {
+                        auto type = std::find_if(interface.second.begin(),
+                                                 interface.second.end(),
+                                                 [](const auto& property) {
+                            return property.first == "Type";
+                        });
+                        const std::string* typeStr =
+                            std::get_if<std::string>(&type->second);
+
+                        if (typeStr == nullptr ||
+                            (*typeStr !=
+                             "xyz.openbmc_project.Network.IP.Protocol.IPv4"))
+                        {
+                            continue;
+                        }
+                        auto origin = std::find_if(interface.second.begin(),
+                                                   interface.second.end(),
+                                                   [](const auto& property) {
+                            return property.first == "Origin";
+                        });
+                        const std::string* originStr =
+                            std::get_if<std::string>(&origin->second);
+
+                        if (originStr == nullptr ||
+                            (*originStr !=
+                             "xyz.openbmc_project.Network.IP.AddressOrigin.Static"))
+                        {
+                            continue;
+                        }
+
+                        std::string id =
+                            objpath.first.str.substr(ipPathStart.size());
+                        deleteIPv4(ethifaceId, id, asyncResp);
+                    }
+                }
+            }
+        }
+        },
+        "xyz.openbmc_project.Network", "/xyz/openbmc_project/network",
+        "org.freedesktop.DBus.ObjectManager", "GetManagedObjects");
+}
+
 inline void setDHCPEnabled(const std::string& ifaceId,
                            const std::string& propertyName, const bool v4Value,
                            const bool v6Value,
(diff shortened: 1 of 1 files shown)


nkotania (approved):

redfish-core/lib/ethernet.hpp
```diff
@@ -2001,6 +2072,20 @@ inline void requestEthernetInterfacesRoutes(App& app)
...
+                bool ipv4Active =
+                    translateDhcpEnabledToBool(ethData.dhcpEnabled, true);
+                bool ipv6Active =
+                    translateDhcpEnabledToBool(ethData.dhcpEnabled, false);
+                if (ipv4Active)
+                {
+                    // Disable DHCPv4 while configuring ipv4 static addresses
+                    setDHCPEnabled(ifaceId, "DHCPEnabled", false, ipv6Active,
+                                   asyncResp);
+                }
```
sunharis, line 2087: Can this be moved into createIPv4 method? So that some more error checks are passed? Just before creating the static IP object, we can disable the DHCPv4

redfish-core/lib/ethernet.hpp
```diff
@@ -1974,6 +2037,14 @@ inline void requestEthernetInterfacesRoutes(App& app)
...
             if (dhcpv4 || dhcpv6)
             {
+                // TODO this is workaround needed to avoid ipv4 static and DHCP
+                // address co-existence This behaviour should be fixed in
+                // backend networkd D-bus application
+                if (dhcpv4)
+                {
+                    // Delete IPv4 static addesses while enabling DHCPv4
+                    deleteIPv4StaticAddresses(ifaceId, asyncResp);
+                }
```
sunharis, line 2047: This can be moved into the handleDHCPPatch method before invoking setDHCPEnabled

redfish-core/lib/ethernet.hpp
raviteja-b: can be moved into createIPv4 method, but need to change createIPv4 input args.
 and createIPv4 method which was getting used in other places..

redfish-core/lib/ethernet.hpp
raviteja-b: done

redfish-core/lib/ethernet.hpp
raviteja-b: done

redfish-core/lib/ethernet.hpp
```diff
@@ -1103,20 +1103,91 @@ inline void
...
+                    {
+                        auto type = std::find_if(interface.second.begin(),
+                                                 interface.second.end(),
+                                                 [](const auto& property) {
+                            return property.first == "Type";
+                        });
+                        const std::string* typeStr =
+                            std::get_if<std::string>(&type->second);
+
+                        if (typeStr == nullptr ||
```
gtmills, line 1136: upstream more common to throw an error if nullptr (it should not be nullptr so something terrible happened if nullptr)

redfish-core/lib/ethernet.hpp
```diff
@@ -1103,20 +1103,91 @@ inline void
...
+void deleteIPv4StaticAddresses(
+    const std::string& ethifaceId,
+    const std::shared_ptr<bmcweb::AsyncResp>& asyncResp)
+{
+    crow::connections::systemBus->async_method_call(
+        [asyncResp, ethifaceId](const boost::system::error_code errorCode,
+                                dbus::utility::ManagedObjectType& resp) {
+        if (errorCode)
+        {
+            messages::internalError(asyncResp->res);
```
gtmills, line 1115: A LOG_ERROR here with the errorCode would be useful

redfish-core/lib/ethernet.hpp
```diff
@@ -1103,20 +1103,91 @@ inline void
         dbus::utility::DbusVariantType(macAddress));
 }
 
+void deleteIPv4StaticAddresses(
+    const std::string& ethifaceId,
+    const std::shared_ptr<bmcweb::AsyncResp>& asyncResp)
+{
+    crow::connections::systemBus->async_method_call(
+        [asyncResp, ethifaceId](const boost::system::error_code errorCode,
```
gtmills, line 1111: Pass error_code by reference please. See https://gerrit.openbmc.org/c/openbmc/bmcweb/+/61013

gtmills:
Why GHE only?

sunharis (approved):

redfish-core/lib/ethernet.hpp
raviteja-b: in this case its just about reading property value to check if its null, it just ignores and continue to check other objects.
same followed in the existing.. I don't anything terrible can happen with bmcweb if this property ptr is null
https://github.com/openbmc/bmcweb/blob/master/redfish-core/lib/ethernet.hpp#L456

redfish-core/lib/ethernet.hpp
raviteja-b: done

redfish-core/lib/ethernet.hpp
raviteja-b: done

raviteja-b:
> Why GHE only?

This workaround for now to turnaround our FVT defects and stabilise 1050 driver, actually networkd design should fixed to handle this configuration, so this won't be acceptable in the upstream for sure.
we have already started discussion with networkd maintainer regarding the same.

redfish-core/lib/ethernet.hpp
```diff
@@ -1103,20 +1103,92 @@ inline void
...
+        [asyncResp, ifaceId, v4Value](const boost::system::error_code& ec) {
         if (ec)
         {
             BMCWEB_LOG_ERROR << "D-Bus responses error: " << ec;
             messages::internalError(asyncResp->res);
             return;
         }
+        // TODO this is workaround needed to avoid ipv4 static and DHCP
+        // address co-existence This behaviour should be fixed in
+        // backend networkd D-bus application
```
gtmills, line 1186: Why can't we fix this in the network D-Bus app?

raviteja-b: networkd design needs to be changed to fix this, have tried to fix but seeing issues in the networkd
we have started discussions for networkd design changes with upstream maintainers.

gtmills: I would really prefer we fix this in networkd
bmcweb has enough hacks

gtmills: Can we fix in downstream networkd ? Work those upstream?

raviteja-b: this PR clears at least 5 critical sev1 network defects

gtmills: Upstream wouldn’t allow this in bmcweb, I’m struggling why we should allow this downstream in bmcweb. How much effort is the networkd change? A few days or weeks? Can we wait until we make the changes in networkd?

gtmills: If you discuss Monday at scrum of scrums overflow and decide this is the correct course, I’m good but I really think we should just fix the right way in networkd now.

raviteja-b: Gunnar 
upstream networkd design completely changed and as part 1050 rebase we pulled master code which allows both ipv4 dhcp and static IP co-existence & networkd broken completely, we need to figure out networkd fix yet..
as I mentioned we need to pull this workaround for sometime to unblock fvt

PR #668, opened 2023-04-13
1050: Redfish check snmpTrap port

ChicagoDuan:
The upstream rule is to set the default value if the port is not compliant.

IBM requires an error to be returned if the port is not compliant. 
ref: https://github.com/ibm-openbmc/dev/issues/3626

Test:
```
curl -k -H "X-Auth-Token: $token" -X POST https://${bmc}/redfish/v1/EventService/Subscriptions -d '{"Destination":"snmp://9.41.166.76:65536", "SubscriptionType":"SNMPTrap", "Protocol":"SNMPv2c"}'
{
  "Destination@Message.ExtendedInfo": [
    {
      "Message": "The value 'snmp://9.41.166.76:65536' for the property Destination is of a different format than the property can accept.",
      "MessageArgs": [
        "snmp://9.41.166.76:65536",
        "Destination"
      ],
      "MessageId": "Base.1.13.0.PropertyValueFormatError",
      "MessageSeverity": "Warning",
      "Resolution": "Correct the value for the property in the request body and resubmit the request if the operation failed."
    }
  ]
}
```

### DIFF
--- a/http/utility.hpp
+++ b/http/utility.hpp
@@ -10,6 +10,7 @@
 #include <nlohmann/json.hpp>
 
 #include <array>
+#include <charconv>
 #include <chrono>
 #include <cstddef>
 #include <cstdint>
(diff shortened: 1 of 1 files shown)


gtmills:
> The upstream rule is to set the default value if the port is not compliant.

Why?

gtmills (approved):

gtmills:
>       "MessageId": "Base.1.13.0.PropertyValueFormatError",

Is this the error we had for 1030 ?

ChicagoDuan:
> 
Please see this code:
https://github.com/openbmc/bmcweb/blob/master/http/utility.hpp#L742
It will set inappropriate ports to 0 or default ports

ChicagoDuan:
>  Is this the error we had for 1030 ?

No, they are different, but they all return HTTPS 400.
1030 ：`propertyValueTypeError`
https://github.com/ibm-openbmc/bmcweb/blob/1030/redfish-core/lib/event_service.hpp#L237

PR #658, opened 2023-04-08
1050: Fix redfish browser - html redfish logo display

baemyung:
Fix redfish browser to show html redfish logo.
It is also reported by tester - https://jazz07.rchland.ibm.com:13443/jazz/web/projects/CSSD#action=com.ibm.team.workitem.viewWorkItem&id=483135

On bmc redish web view, DMTF Redfish logo fails to display.
It is because the headers in the Request is no longer valid
after the use of move().

The following upstream commits are needed.

- Add asyncResp support to handleUpgrade https://gerrit.openbmc.org/c/openbmc/bmcweb/+/46989
  Change-Id: I1c6c91f126b734e1b5573d5ef204fe2bf6ed6c26

- Add Support for privilege check in handleUpgrade https://gerrit.openbmc.org/c/openbmc/bmcweb/+/46991?usp=search
  Change-Id: I6f743c27e7e6077f1c6c56e6958922027e4404e8

- Fix some moves
  Change-Id: Ib6d99726a64326b7c8bad15bc9d4ca774ab6256d https://gerrit.openbmc.org/c/openbmc/bmcweb/+/61677

- Fix Request use-after-move
  Change-Id: Iae68a68817146c28377bbcade04716725e4a6096 https://gerrit.openbmc.org/c/openbmc/bmcweb/+/61768

Tested:

- Redfish validator passed
- [https://${bmc}/redfish/v1](https://%24%7Bbmc%7D/redfish/v1) is showing the pretty redfish logo
- Tried some GUI pages and worked without issues.
- Tried consoles on GUI and worked without issues.

PR #669, opened 2023-04-13
Revert commits to fix bmcweb crash hang

gtmills:
This reverts this PR: 
https://github.com/ibm-openbmc/bmcweb/pull/658

This PR was the only bmcweb PR to go into 2.22 
https://github.ibm.com/openbmc/release-notes/wiki/fw1050.00-2.22

2.22 is were we first saw bmcweb hanging and core dumping when shut down. Defects are: 
487666
487798
487741

The root cause can be investigated more but for now, let's get the train back on the tracks. 

Tested: bmcweb with 2.21 works on one of the systems we saw this happening on. Built bmcweb with this commit and loaded onto 2 of the bmcs we saw this on and no longer are hanging.

### DIFF
--- a/http/app.hpp
+++ b/http/app.hpp
@@ -58,7 +58,7 @@ class App
     App& operator=(const App&&) = delete;
 
     template <typename Adaptor>
-    void handleUpgrade(Request& req,
+    void handleUpgrade(const Request& req,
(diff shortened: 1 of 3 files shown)


Lakshmi-y (approved):

ninadpalsule (commented):
Looks good to me.

PR #673, opened 2023-04-19
Fix system dump delete functionality

asmithakarun:
Currently, when a user does a DELETE operation on
any of the system dumps, "Resource Not Found" error is being sent back by bmcweb.

This is because, all the system dump entry IDs will be of the format: < dumpType >_< ID >. For example, "Resource_1"/"System_1". Currently, no logic is present to get dump id from the above string to make dbus call and the call fails. This issue is resolved in this PR.

Fixes:
https://jazz07.rchland.ibm.com:13443/jazz/web/projects/CSSD#action=com.ibm.team.workitem.viewWorkItem&id=493478

Change-Id: Ifde5a532bcc90e6c5c150a704c8e045b8befbf48

### DIFF
--- a/redfish-core/lib/log_services.hpp
+++ b/redfish-core/lib/log_services.hpp
@@ -4059,44 +4059,56 @@ inline void handleLogServicesDumpEntryDelete(
 inline void handleLogServicesDumpEntryComputerSystemDelete(
     crow::App& app, const crow::Request& req,
     const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
-    const std::string& chassisId, const std::string& dumpId)
+    const std::string& systemId, const std::string& dumpEntryId)
 {
     if (!redfish::setUpRedfishRoute(app, req, asyncResp))
     {
         return;
     }
-    if (chassisId != "system")
+    if (systemId != "system")
     {
-        messages::resourceNotFound(asyncResp->res, "ComputerSystem", chassisId);
+        messages::resourceNotFound(asyncResp->res, "ComputerSystem", systemId);
         return;
     }
 
-    if (boost::starts_with(dumpId, "System"))
+    std::size_t pos = dumpEntryId.find_first_of('_');
+    if (pos == std::string::npos || (pos + 1) >= dumpEntryId.length())
+    {
+        // Requested ID is invalid
+        messages::invalidObject(
+            asyncResp->res, crow::utility::urlFromPieces(
+                                "redfish", "v1", "Systems", systemId,
+                                "LogServices", "Dump", "Entries", dumpEntryId));
+        return;
+    }
+
+    const std::string& dumpId = dumpEntryId.substr(pos + 1);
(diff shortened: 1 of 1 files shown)


redfish-core/lib/log_services.hpp
```diff
@@ -4071,32 +4071,44 @@ inline void handleLogServicesDumpEntryComputerSystemDelete(
...
     }
     else
     {
-        messages::invalidObject(asyncResp->res,
-                                crow::utility::urlFromPieces(
-                                    "redfish", "v1", "Systems", "system",
-                                    "LogServices", "Dump", "Entries", dumpId));
+        messages::invalidObject(
+            asyncResp->res, crow::utility::urlFromPieces(
+                                "redfish", "v1", "Systems", chassisId,
```
gtmills, line 4110: chassisId here doesn't look correct

asmithakarun: The chassisId will anyways be "system" right?
Because we are checking if chassisId is system and if not, we are throwing an error.

redfish-core/lib/log_services.hpp
```diff
@@ -4071,32 +4071,44 @@ inline void handleLogServicesDumpEntryComputerSystemDelete(
...
 
-    if (boost::starts_with(dumpId, "System"))
+    std::size_t pos = dumpEntryId.find_first_of('_');
+    if (pos == std::string::npos || (pos + 1) >= dumpEntryId.length())
+    {
+        // Requested ID is invalid
+        messages::invalidObject(
+            asyncResp->res,
+            crow::utility::urlFromPieces("redfish", "v1", "Systems", chassisId,
+                                         "EthernetInterfaces", dumpEntryId));
```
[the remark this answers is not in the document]
asmithakarun: Done

raviteja-b (approved):

redfish-core/lib/log_services.hpp
```diff
@@ -4059,7 +4059,7 @@ inline void handleLogServicesDumpEntryDelete(
 inline void handleLogServicesDumpEntryComputerSystemDelete(
     crow::App& app, const crow::Request& req,
     const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
-    const std::string& chassisId, const std::string& dumpId)
+    const std::string& chassisId, const std::string& dumpEntryId)
```
gtmills, line 4062: this should be systemId

redfish-core/lib/log_services.hpp
gtmills: Can be fix this var name to be correct so systemId ?

sunharis (approved):
@asmithakarun to remove the confusions, you may rename the variable.

redfish-core/lib/log_services.hpp
asmithakarun: Done

gtmills (approved):

PR #674, opened 2023-04-19
log_services: Fix isContentTypeAllowed checks

spinler:
The commit https://gerrit.openbmc.org/c/openbmc/bmcweb/+/56694 inadvertently changed the polarity on the if checks using the newly introduced isContentTypeAllowed function which caused 'Bad Request' to be returned when the content type was allowed.

Tested:
Getting the EventLog and PostCodes attachment would return the data instead of 'Bad Request'.

Change-Id: Iafcdeaba1a0723326347bb2a832b53bbf0aab230

### DIFF
--- a/redfish-core/lib/log_services.hpp
+++ b/redfish-core/lib/log_services.hpp
@@ -3181,7 +3181,7 @@ inline void requestRoutesDBusEventLogEntryDownload(App& app)
         {
             return;
         }
-        if (http_helpers::isContentTypeAllowed(
+        if (!http_helpers::isContentTypeAllowed(
                 req.getHeaderValue("Accept"),
                 http_helpers::ContentType::OctetStream, true))
         {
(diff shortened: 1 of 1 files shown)


spinler:
Upstream review is https://gerrit.openbmc.org/c/openbmc/bmcweb/+/62700
Defect 492002

gtmills (approved):

spinler:
This is slightly different than the upstream commit since CELogs aren't in upstream so I had to fix that one just downstream.

redfish-core/lib/log_services.hpp
```diff
@@ -3226,7 +3226,7 @@ inline void requestRoutesDBusCELogEntryDownload(App& app)
         {
             return;
         }
-        if (http_helpers::isContentTypeAllowed(
+        if (!http_helpers::isContentTypeAllowed(
```
gtmills, line 3229: This is the downstream only one (CElog)

PR #676, opened 2023-04-25
Add support for fault data dump

asmithakarun:
Fault data dump contains the server fault information and is collected by the BMC and is stored on the bmc.

This dump entry will be stored as one other bmc dump entry in the dbus. This code adds support for creation of this dump. Listing and deletion of this dump will be the same as how it is done for bmc dump.

Tested By:

* Create dump command: POST https://${bmc}/redfish/v1/Managers/bmc/LogServices/Dump\ /Actions/LogService.CollectDiagnosticData -d \
'{"DiagnosticDataType":"Manager", "OEMDiagnosticDataType":"FaultData"}'

* List dumps: GET https://${bmc}/redfish/v1/Managers/bmc/LogServices/Dump/Entries/

* Get dump: GET https://${bmc}/redfish/v1/Managers/bmc/LogServices/Dump/Entries/<id>

* Delete dump: DELETE https://${bmc}/redfish/v1/Managers/bmc/LogServices/Dump/Entries/<id>

Change-Id: I5f258a4ca201fe882778fba73037f44a3d4ab44b

### DIFF
--- a/redfish-core/lib/log_services.hpp
+++ b/redfish-core/lib/log_services.hpp
@@ -1178,6 +1178,23 @@ inline void createDump(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
             messages::internalError(asyncResp->res);
             return;
         }
+        if (oemDiagnosticDataType)
+        {
+            if (*oemDiagnosticDataType == "FaultData")
+            {
+                // Fault data dump
+                createDumpParams.emplace_back(
+                    "xyz.openbmc_project.Dump.Internal.Create.Type",
+                    "xyz.openbmc_project.Dump.Internal.Create.Type.FaultData");
+            }
+            else
+            {
+                BMCWEB_LOG_ERROR
+                    << "Wrong parameter value passed for 'OEMDiagnosticDataType'";
+                messages::internalError(asyncResp->res);
+                return;
+            }
+        }
         dumpPath = "/redfish/v1/Managers/bmc/LogServices/Dump/";
     }
     else


raviteja-b:
@asmithakarun we need to support offload also right?
Seems we need to make some changes to support offload for this new dump type

raviteja-b (approved):

sunharis (approved):

asmithakarun:
> @asmithakarun we need to support offload also right? if we are treating this fault dump as new dump type then we need to make some changes to support offload

Actually this new dump type will not be stored as a new dump entry dbus object. It will be like just another bmc dump entry -- /xyz/openbmc_project/dump/bmc/entry/1

So offload support additionally is not necessary. the available implementation will work.

asmithakarun:
Currently no support for dump types. Once debug-collector code restructuring is done, we can target to get all the dump types support upstream. This is a downstream only change.

gtmills:
How does the client know this is a fault/NAG dump when querying the dumps ?

asmithakarun:
> How does the client know this is a fault/NAG dump when querying the dumps ?

Currently, there is no way to differentiate the dumps - bmc and nag; It will be displayed as BMC dumps
Dhruv had mentioned about range-based ids that can be used in the backend for different types of bmc dumps.

Right now, hmc initiates the nag dump, and then offloads it at the same time, itseems - as per discussion with Dhruv.

gtmills (approved):
Approving but this doesn't feel like a very good client experience. The client doesn't know if the dump is a Nag dump.
We don't advertise this capability to the client when the client inspects the API

nkotania (approved):

PR #681, opened 2023-05-05
Fix broadcast eventing issue

asmithakarun:
Broadcast events were not sent in the right format to the connected client due to which the client was unable to process the event.

Currently the events are sent in the following format: events:{
  "Message": "<>",
  "Name": "Broadcast Message",
  "OriginOfCondition": "/ibm/v1/HMC/BroadcastService",
  "Timestamp": "2023-05-05T03:15:50+00:00"
}

With this PR, the event json is changed to the following format which is the common format for all kinds of events: {
  "@odata.type":"#Event.v1_4_0.Event",
  "Events":{
     "EventId":"Broadcast Event",
     "EventTimestamp":"2020-07-15T12:03:30+00:00",
     "MemberId":"Broadcast Event",
     "Message":"Hi This is a message to forward to a single HMC",
     "OriginOfCondition":"/ibm/v1/HMC/BroadcastService"
  },
  "Id":1,
  "Name":"Event Log"
}

Fixes: https://jazz07.rchland.ibm.com:13443/jazz/web/projects/CSSD#action=com.ibm.team.workitem.viewWorkItem&id=516553

Tested By:
POST https://${bmc}/ibm/v1/HMC/BroadcastService -d '{"Message":"A Test message"}'

### DIFF
--- a/redfish-core/include/event_service_manager.hpp
+++ b/redfish-core/include/event_service_manager.hpp
@@ -1085,23 +1085,6 @@ class EventServiceManager
             }
         }
     }
-    void sendBroadcastMsg(const std::string& broadcastMsg)
-    {
-        for (const auto& it : this->subscriptionsMap)
-        {
-            std::shared_ptr<Subscription> entry = it.second;
-            nlohmann::json msgJson;
-            msgJson["Timestamp"] =
-                redfish::time_utils::getDateTimeOffsetNow().first;
-            msgJson["OriginOfCondition"] = "/ibm/v1/HMC/BroadcastService";
-            msgJson["Name"] = "Broadcast Message";
-            msgJson["Message"] = broadcastMsg;
-
-            std::string strMsg = msgJson.dump(
-                2, ' ', true, nlohmann::json::error_handler_t::replace);
-            entry->sendEvent(strMsg);
-        }
-    }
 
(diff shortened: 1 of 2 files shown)


redfish-core/include/event_service_manager.hpp
```diff
@@ -1085,23 +1085,6 @@ class EventServiceManager
             }
         }
     }
-    void sendBroadcastMsg(const std::string& broadcastMsg)
-    {
-        for (const auto& it : this->subscriptionsMap)
-        {
-            std::shared_ptr<Subscription> entry = it.second;
-            nlohmann::json msgJson;
-            msgJson["Timestamp"] =
```
gtmills, line 1094: What about these fields Timestamp, Origin, Name?

gtmills: Don't need them anymore?

raviteja-b (approved):

redfish-core/include/event_service_manager.hpp
asmithakarun: This method will take care of the other fields

https://github.com/openbmc/bmcweb/blob/master/redfish-core/include/event_service_manager.hpp#L1014

gtmills:
Some this looks upstream ? 
https://github.com/search?q=repo%3Aopenbmc%2Fbmcweb%20sendBroadcastMsg&type=code

Push upstream?

PR #686, opened 2023-05-12
Introducing DHCPv6 parameters to network config

cmjishnu:
At present, DHCP parameters like DNSEnabled, NTPEnabled etc. are
shared between DHCPv4 and DHCPv6 in the network configuration.

This change is to enable the possibility to have separate writable
parameters for DHCPv4 and DHCPv6 in the network configuration to be
able to configure them independently.

Tested by:

Used the redfish commands to set and get the parameter values
individually for both DHCPv4 and DHCPv6.
Verified that modifying the new DHCPv6 parameters are not disturbing
the existing DHCPv4 parameters and vice versa.
The values are shared between the ethernet interfaces. When we modify
a DHCPv6 value both interfaces are updated

Upstream PR : https://gerrit.openbmc.org/c/openbmc/bmcweb/+/63390

### DIFF
--- a/redfish-core/lib/ethernet.hpp
+++ b/redfish-core/lib/ethernet.hpp
@@ -86,8 +86,11 @@ struct EthernetInterfaceData
     size_t mtuSize;
     bool autoNeg;
     bool dnsEnabled;
+    bool dnsv6Enabled;
     bool ntpEnabled;
+    bool ntpv6Enabled;
     bool hostNameEnabled;
+    bool hostNamev6Enabled;
     bool linkUp;
     bool nicEnabled;
     bool ipv6AcceptRa;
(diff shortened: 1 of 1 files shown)


redfish-core/lib/ethernet.hpp
```diff
@@ -1313,84 +1343,54 @@ inline void handleDHCPPatch(const std::string& ifaceId,
...
-        {
-            messages::generalError(asyncResp->res);
-            return;
-        }
-        nextDNS = *v4dhcpParms.useDnsServers;
-    }
-    else if (v4dhcpParms.useDnsServers)
+    bool nextDNS{}, nextDNSv6{};
+    nextDNS = nextDNSv6 = ethData.dnsEnabled;
+    if (v4dhcpParms.useDnsServers)
```
sunharis, line 1348: These removed changes in this method could have been retained, and only the setDHCPConfig() calls would have changed to send the separate set command to the dbus.
I think this code already had checked the v4 and v6 settings separately using v4dhcpParms and v6dhcpParms

sunharis (changes requested):

redfish-core/lib/ethernet.hpp
cmjishnu: It was mandating v4dhcpParms and v6dhcpParms to have the same values by returning generalError when they're not. It was removed because different values for different network types are supported now.

PR #698, opened 2023-05-25
Processor: Add processor throttle status

cjcain:
- Update Processor Schema to 18.0
- Add processor throttle status and cause https://gerrit.openbmc.org/c/openbmc/phosphor-dbus-interfaces/+/63063 Throttled: An indication of whether the processor is throttled. ThrottledCauses: An array of reasons that the processor is throttled.
- Remove processor version from messages
- Change getPrettyName to take specific service name instead of map

Ran validator and no new errors were found.

Change-Id: Ia4a58ae0f26ffc6177f418420ba45063471323da

### DIFF
--- a/Redfish.md
+++ b/Redfish.md
@@ -905,6 +905,8 @@ other.
 - Socket
 - SparePartNumber
 - Status
+- ThrottleCauses
+- Throttled
 - TotalCores
 - TotalThreads
 - Version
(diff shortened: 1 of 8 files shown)


gtmills:
I know getPrettyName  is up to for review, can you comment on it, so this doesn't get lost?

gtmills:
And the PrettyName is still working with the other resources (CPU, processors)?

cjcain:
Yes those resources all had Names:
```
    "Name": "chassis",
    "Name": "Assembly Collection"
            "Name": "System Backplane",
            "Name": "Operator Panel Base",
    "Name": "Chassis Collection"
    "Name": "Fabric Adapter Collection"
    "Name": "NVMe Backplane",
    "Name": "Storage Collection"
    "Name": "Storage",
    "Name": "Processor Module",
```

PR #701, opened 2023-06-06
Hypervisor: Add support for default IPv6 gateway

asmithakarun:
This commit adds redfish support for patching
"IPv6DefaultGateway" property for each ethernet interfaces for a static configuration. In case of dynamic configuration, patching of this property will not be permitted in the backend networkd app and an error response will be sent back to the client.

Address the following defect:
https://jazz07.rchland.ibm.com:13443/jazz/web/projects/CSSD#action=com.ibm.team.workitem.viewWorkItem&id=543926

The phosphor networkd changes are here:
https://github.com/ibm-openbmc/phosphor-networkd/pull/88

Tested By:

[1] PATCH -d '{"IPv6DefaultGateway":"<ipv6Gateway>"}' https://${bmc}/redfish/v1/Systems/hypervisor/EthernetInterfaces/eth<0/1>

Change-Id: Iaa48431147ba510eac6be9eec2f330303d588005

### DIFF
--- a/redfish-core/lib/hypervisor_system.hpp
+++ b/redfish-core/lib/hypervisor_system.hpp
@@ -1068,6 +1068,26 @@ inline void handleHypervisorIPv6StaticPatch(
     }
 }
 
+inline void handleHypV6DefaultGatewayPatch(
+    const std::string& ifaceId, const std::string& ipv6DefaultGateway,
+    const std::shared_ptr<bmcweb::AsyncResp>& asyncResp)
+{
+    crow::connections::systemBus->async_method_call(
+        [asyncResp](const boost::system::error_code ec) {
+        if (ec)
+        {
+            BMCWEB_LOG_ERROR << "D-Bus responses error: " << ec;
+            messages::internalError(asyncResp->res);
+            return;
+        }
+        },
+        "xyz.openbmc_project.Network.Hypervisor",
+        "/xyz/openbmc_project/network/hypervisor/" + ifaceId,
+        "org.freedesktop.DBus.Properties", "Set",
+        "xyz.openbmc_project.Network.EthernetInterface", "DefaultGateway6",
+        std::variant<std::string>(ipv6DefaultGateway));
+}
+
 inline void requestRoutesHypervisorSystems(App& app)
 {
     /**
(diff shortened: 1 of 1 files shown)


redfish-core/lib/hypervisor_system.hpp
```diff
@@ -1237,13 +1258,15 @@ inline void requestRoutesHypervisorSystems(App& app)
...
+        std::optional<std::string> ipv6DefaultGateway;
 
         if (!json_util::readJsonPatch(
                 req, asyncResp->res, "HostName", hostName,
                 "IPv4StaticAddresses", ipv4StaticAddresses,
                 "IPv6StaticAddresses", ipv6StaticAddresses, "IPv4Addresses",
                 ipv4Addresses, "DHCPv4", dhcpv4, "DHCPv6", dhcpv6,
-                "StatelessAddressAutoConfig", statelessAddressAutoConfig))
+                "StatelessAddressAutoConfig", statelessAddressAutoConfig,
+                "IPv6DefaultGateway", ipv6DefaultGateway))
```
sunharis, line 1269: This is readonly property. Can not patch.

For patching you must use IPv6StaticDefaultGateways property

sunharis (changes requested):

redfish-core/lib/hypervisor_system.hpp
asmithakarun: Done

redfish-core/lib/hypervisor_system.hpp
```diff
@@ -1416,6 +1439,31 @@ inline void requestRoutesHypervisorSystems(App& app)
...
+                    {
+                        defaultGwStr += "\"" + ipv6StaticDefaultGw[i] + "\"";
+                        if (i != ipv6StaticDefaultGw.size() - 1)
+                        {
+                            defaultGwStr += ",";
+                            continue;
+                        }
+                        defaultGwStr += "]}";
+                    }
+                    messages::propertyValueFormatError(
```
raviteja-b, line 1459: say if user tries to configure multiple ipv6 default gateways, its correct as per schema,
so returning propertyValueFormatError does not Look right.
thinking if we can do other work around like take only first or last input entry or return a error description to say single entry supported

raviteja-b (approved):

redfish-core/lib/hypervisor_system.hpp
asmithakarun: >  its correct as per schema,
Done

sunharis (approved):

gtmills (approved):

PR #702, opened 2023-06-07
Hypervisor config: Send relevant error on patch

asmithakarun:
This commit will send back relevant redfish error back if the client tries to configure the hypervisor network with invalid IPv6 address.

Tested By:

[1] PATCH -d '{"IPv6StaticAddresses": [{"Address": \ "9.3.2.3:1A:1B:1c:1D:1E:1F", "PrefixLength": 64}]}' \ https://${bmc}/redfish/v1/Systems/hypervisor/EthernetInterfaces/eth0

Change-Id: I262fe3bd222e8e35ac48e4433ba52a4b6d5fb649

### DIFF
--- a/redfish-core/lib/hypervisor_system.hpp
+++ b/redfish-core/lib/hypervisor_system.hpp
@@ -569,8 +569,7 @@ inline void
                        const std::shared_ptr<bmcweb::AsyncResp>& asyncResp)
 {
     crow::connections::systemBus->async_method_call(
-        [asyncResp, gateway, ifaceId,
-         address](const boost::system::error_code ec) {
+        [asyncResp, ifaceId, address](const boost::system::error_code ec) {
         if (ec)
         {
             BMCWEB_LOG_DEBUG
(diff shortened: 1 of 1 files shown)


redfish-core/lib/hypervisor_system.hpp
```diff
@@ -585,6 +585,12 @@ inline void
                                             "hypervisor", "EthernetInterfaces",
                                             ifaceId));
             }
+            else if (ec == boost::system::errc::io_error)
+            {
+                messages::propertyValueFormatError(
+                    asyncResp->res, "Address/PrefixLength",
```
sunharis, line 591: Are you sure we can give 2 properties like this?
Prefix length is an integer value. There is no format for that as IPv6 addresses.

asmithakarun: We can give the whole json data here instead. But I checked giving an invalid prefix length and it failed with a different error itself - invalid value for prefix length. So we are good if we just give address here

sunharis (approved):

raviteja-b (approved):

gtmills (approved):

PR #709, opened 2023-06-12
Fix internal error due to system_attention_indicator while booting

lxwinspur:
We should not return an internal error if the service is not up.

Fixes https://github.com/ibm-openbmc/dev/issues/3618

### DIFF
--- a/redfish-core/lib/oem/ibm/system_attention_indicator.hpp
+++ b/redfish-core/lib/oem/ibm/system_attention_indicator.hpp
@@ -44,8 +44,7 @@ inline void getSAI(const std::shared_ptr<bmcweb::AsyncResp>& aResp,
                         const dbus::utility::MapperGetObject& object) {
         if (ec || object.empty())
         {
-            BMCWEB_LOG_ERROR << "DBUS response error " << ec.message();
-            messages::internalError(aResp->res);
+            BMCWEB_LOG_DEBUG << "Failed to get LED DBus name: " << ec.message();
             return;
         }
 


lxwinspur:
> Is this downstream-only code we're fixing here or is there something we need to send upstream?

downstream-only
Because there is no `oem` upstream.

redfish-core/lib/oem/ibm/system_attention_indicator.hpp
```diff
@@ -114,8 +113,7 @@ inline void setSAI(const std::shared_ptr<bmcweb::AsyncResp>& aResp,
                              const dbus::utility::MapperGetObject& object) {
         if (ec || object.empty())
         {
-            BMCWEB_LOG_ERROR << "DBUS response error " << ec.message();
```
geissonator, line 117: I get not returning an error if it's just a generic GET on the system interface. But if someone has specifically done a POST to set the LED, and that fails, we should return some sort of error?

lxwinspur: Yeah, I think you are correct, updated.

geissonator (commented):
Looks good to me, I'll let Gunnar review/approve.

redfish-core/lib/oem/ibm/system_attention_indicator.hpp
```diff
@@ -44,8 +44,7 @@ inline void getSAI(const std::shared_ptr<bmcweb::AsyncResp>& aResp,
                         const dbus::utility::MapperGetObject& object) {
         if (ec || object.empty())
         {
-            BMCWEB_LOG_ERROR << "DBUS response error " << ec.message();
-            messages::internalError(aResp->res);
+            BMCWEB_LOG_DEBUG << "Failed to get LED DBus name: " << ec.message();
```
gtmills, line 47: We are ignoring all errors here.. but I think that is fine, since isn't that important of a property and that is what we did for 1030 it looks like 
https://github.com/ibm-openbmc/bmcweb/blob/1030/redfish-core/lib/oem/ibm/system_attention_indicator.hpp#L38

gtmills (approved):

PR #712, opened 2023-06-20
1050: Fix PCIeSlot LocationIndicatorActive Patch Handling

baemyung:
PCIe Slot LocationIndicatorActive PATCH is not working correctly all the time.

https://github.com/ibm-openbmc/dev/issues/3631

Below is an Everest PATCH of all LEDs to true.
- You will notice not all go true.
- This has also been reported on Rainier.
- This has also been reported when moving the LocationIndicatorActive to false.
- This has been reported on the GUI.

```
Set all slots to be true.

$ curl -k  -X PATCH https://$bmc/redfish/v1/Chassis/chassis/PCIeSlots \
    -d '{"Slots":[{"LocationIndicatorActive":true},{"LocationIndicatorActive":true},  \
      ... \
        {"LocationIndicatorActive":true},{"LocationIndicatorActive":true}]}'

$ curl -s -k https://$bmc/redfish/v1/Chassis/chassis/PCIeSlots | grep LocationIndicatorActive
      "LocationIndicatorActive": true,
      "LocationIndicatorActive": true,
      ...
      "LocationIndicatorActive": true,
      "LocationIndicatorActive": false,     <--- Sometimes incorrect - esp at 21st entry
```

This can be addressed by getting the valid pcieSlots first, and then validate it before performing PATCH.

 - Get valid PCIeSlots `getValidPCIeSlotList()` - which filters out the improper slot and invoke callback at the last slots.
 - Validate it with input slots. (the number of inputs should match with the number of valid slots)
 - Execute the PATCH (i.e. `setLocationIndicatorActive()`) using the the obtained list.

Tested:
 - Redfish Validator performed
 - GET PCIeSlots and compare it with the existing output.
 - PATCH PCIeSlots and verify the correctness (e.g. number of input slots, and content)

### DIFF
--- a/redfish-core/lib/pcie_slots.hpp
+++ b/redfish-core/lib/pcie_slots.hpp
@@ -16,9 +16,13 @@
 #include <utils/json_utils.hpp>
 #include <utils/pcie_util.hpp>
 
+#include <algorithm>
+#include <array>
+#include <cstdint>
+#include <map>
 #include <memory>
+#include <optional>
 #include <string>
-#include <variant>
 #include <vector>
 
 namespace redfish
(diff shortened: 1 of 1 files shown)


lxwinspur (approved):
LGTM

redfish-core/lib/pcie_slots.hpp
```diff
@@ -605,14 +558,50 @@ inline void handlePCIeSlotCollectionGet(
...
+        asyncResp, chassisID,
+        [asyncResp, chassisID, total,
+         locationIndicatorActiveMap{std::move(locationIndicatorActiveMap)}](
+            const std::vector<std::pair<std::string, std::string>>&
+                slotPathConnNames) {
+        if (slotPathConnNames.size() != total)
+        {
+            BMCWEB_LOG_ERROR
+                << "The actual number of pcieslots is different from the number of the input slots";
+            messages::internalError(asyncResp->res);
```
gtmills, line 590: This is a PATCH.. I think we should return a 404 
For 1020 we returned a resourceNotFound 
https://github.com/ibm-openbmc/bmcweb/blob/1020/redfish-core/lib/pcie_slots.hpp#LL646C1-L648C60

gtmills: You could use 

        "ArraySizeTooLong": {
        
or 
        
                "ArraySizeTooShort": {
                
                
or 

        "InvalidIndex": {

redfish-core/lib/pcie_slots.hpp
```diff
@@ -605,14 +558,50 @@ inline void handlePCIeSlotCollectionGet(
...
+    redfish::chassis_utils::getValidChassisPath(
+        asyncResp, chassisID,
         [asyncResp,
-         chassisID](const boost::system::error_code ec,
-                    const dbus::utility::MapperGetSubTreeResponse& subtree) {
-        onMapperSubtreeDone(asyncResp, chassisID, ec, subtree);
+         chassisID](const std::optional<std::string>& validChassisPath) {
+        if (!validChassisPath)
+        {
+            BMCWEB_LOG_ERROR << "Not a valid chassis ID:" << chassisID;
```
gtmills, line 567: should be warning, oh well

redfish-core/lib/pcie_slots.hpp
```diff
@@ -489,110 +380,172 @@ inline void
...
+        {
+            sdbusplus::asio::getAllProperties(
+                *crow::connections::systemBus, connectionName, pcieSlotPath,
+                "xyz.openbmc_project.Inventory.Item.PCIeSlot",
+                [asyncResp, connectionName, pcieSlotPath](
+                    const boost::system::error_code& ec,
+                    const dbus::utility::DBusPropertiesMap& propertiesList) {
+                if (ec)
+                {
+                    BMCWEB_LOG_ERROR << "Can't get PCIeSlot properties!";
```
gtmills, line 540: include the ec

redfish-core/lib/pcie_slots.hpp
```diff
@@ -489,110 +380,172 @@ inline void
...
+                    {
+                        // This PCIeSlot has no chassis association.
+                        if (slotIndex == slotCount)
+                        {
+                            // Reach to the end of slotElement
+                            callback(*validSlotPathConnNames);
+                        }
+                        return;
+                    }
+                    BMCWEB_LOG_ERROR << "DBUS response error";
```
gtmills, line 485: the ec would help here

redfish-core/lib/pcie_slots.hpp
```diff
@@ -605,14 +558,50 @@ inline void handlePCIeSlotCollectionGet(
...
+{
+    getValidPCIeSlotList(
+        asyncResp, chassisID,
+        [asyncResp, chassisID, total,
+         locationIndicatorActiveMap{std::move(locationIndicatorActiveMap)}](
+            const std::vector<std::pair<std::string, std::string>>&
+                slotPathConnNames) {
+        if (slotPathConnNames.size() != total)
+        {
+            BMCWEB_LOG_ERROR
```
gtmills, line 588: probably warning

redfish-core/lib/pcie_slots.hpp
baemyung: I think - InvalidIndex seems better because it represents the inconsistent count of array inputs to the valid/active pcie slots.

redfish-core/lib/pcie_slots.hpp
baemyung: Changed

redfish-core/lib/pcie_slots.hpp
baemyung: Changed

redfish-core/lib/pcie_slots.hpp
baemyung: Changed to WARNING

baemyung:
In addition, I updated `getValidPCIeSlotList()` so that it should invoke `callback()` even if an individual `getAssociationEndPoints()` call may be failed.
It is because not all pcieslots may have the proper/correct association.  
In that case, the subset of pcieslots would need to be returned as valid slots, rather than resulting it as the failure of `getValidPCIeSlotList()`.

baemyung:
> In addition, I updated `getValidPCIeSlotList()` so that it should invoke `callback()` even if an individual `getAssociationEndPoints()` call may be failed. It is because not all pcieslots may have the proper/correct association. In that case, the subset of pcieslots would need to be returned as valid slots, rather than resulting it as the failure of `getValidPCIeSlotList()`.

After I think this more, I think it seems more appropriate to keep the current way of handling getAssociationEndPoints() error case.
- if getAssociationEndPoints() gives D-Bus error (other than EBADR), it returns an error.
- If the number of associated chassises is more than 1, it returns an error
- This way will provide the underline cause that needs to be addressed beyond the PATCH request.

redfish-core/lib/pcie_slots.hpp
gtmills: still don't see this

gtmills (approved):

PR #718, opened 2023-06-28
Audit Log: Factor out function to determine which events to audit

jeaaustx:
Factored out function which determines which Redfish events to audit.
Change is in preparation for pushing upstream. I anticipate this is an area which will receive upstream comments.

I will push this into 1050 only.

Tested:
- Added unit test for new function, wantAudit().
- Confirmed expected events still being audited.


baemyung (approved):
It looks good to me.

gtmills (approved):

PR #720, opened 2023-06-30
SSDP Server prototype

cmjishnu:
This PR contains the server side implementation of the SSDP


gtmills:
@cm-jishnu Is this still needed? Can we close?

PR #725, opened 2023-07-10
Audit Log: Hide details of building audit message into auditEvent()

jeaaustx:
Added utility functions to centralize checking whether detail data for the request being logged should be included in the log message. Moved the building of the audit message into the audit namespace so main bmcweb code doesn't have to understand the details. Changed building of message buffer to use std::string to simplify existing logic.

Tested:
 - CI passes with new unit tests and audit-events enabled.
 - Confirmed sampling of events have the correct detail data included.
 - Confirmed user session events do not have detail data.

### DIFF
--- a/include/audit_events.hpp
+++ b/include/audit_events.hpp
@@ -95,22 +95,22 @@ inline void auditSetState(bool enable)
 }
 
 /**
- * @brief Checks if POST request should be audited on completion
+ * @brief Checks if POST request is a user connection event
  *
  * Login and Session requests are audited when the authentication is attempted.
  * This allows failed requests to be audited with the user detail.
  *
- * @return True if request should be audited
+ * @return True if request is a user connection event
  */
-inline bool checkPostAudit(const crow::Request& req)
+inline bool checkPostUser(const crow::Request& req)
 {
     if ((req.target() == "/redfish/v1/SessionService/Sessions") ||
         (req.target() == "/redfish/v1/SessionService/Sessions/") ||
         (req.target() == "/login"))
     {
-        return false;
+        return true;
     }
-    return true;
+    return false;
 }
 
 /**
@@ -123,7 +123,7 @@ inline bool wantAudit(const crow::Request& req)
         (req.method() == boost::beast::http::verb::put) ||
         (req.method() == boost::beast::http::verb::delete_) ||
         ((req.method() == boost::beast::http::verb::post) &&
-         checkPostAudit(req)))
+         !checkPostUser(req)))
     {
         return true;
     }
@@ -131,11 +131,62 @@ inline bool wantAudit(const crow::Request& req)
     return false;
 }
 
-inline void auditEvent(const char* opPath, const std::string& userName,
-                       const std::string& ipAddress, bool success)
+/**
+ * @brief Checks if request should include additional data
+ *
+ * - Accounts requests data may contain passwords
+ * - IBM Management console events data is not useful. It can be binary data or
+ *   contents of file.
+ * - User login and session data may contain passwords
+ *
+ * @return True if request's data should not be logged
+ */
+inline bool checkSkipDetail(const crow::Request& req)
 {
-    int code = __LINE__;
+    if (req.target().starts_with("/redfish/v1/AccountService/Accounts") ||
+        req.target().starts_with("/ibm/v1") ||
+        ((req.method() == boost::beast::http::verb::post) &&
+         checkPostUser(req)))
+    {
+        return true;
+    }
+    return false;
+}
 
+/**
+ * @brief Checks if request's detail data should be logged
+ *
+ * @return True if request's detail data should be logged
+ */
+inline bool wantDetail(const crow::Request& req)
+{
+    switch (req.method())
+    {
+        case boost::beast::http::verb::patch:
+        case boost::beast::http::verb::post:
+            if (checkSkipDetail(req))
+            {
+                return false;
+            }
+            return true;
+
+        case boost::beast::http::verb::put:
+            return (!req.target().starts_with("/ibm/v1"));
+
+        case boost::beast::http::verb::delete_:
+            return true;
+
+        default:
+            // Shouldn't be here, don't log any data
+            BMCWEB_LOG_DEBUG << "Unexpected verb " << req.method();
+            return false;
+    }
+}
+
+inline void auditEvent(const crow::Request& req, const std::string& userName,
+                       bool success)
+{
+    std::string opPath;
     char cnfgBuff[256];
     size_t bufLeft = 256; // Amount left available in cnfgBuff
     char* user = NULL;
@@ -143,13 +194,18 @@ inline void auditEvent(const char* opPath, const std::string& userName,
     size_t userLen = 0;
     int rc;
     int origErrno;
+    std::string ipAddress;
+    std::string detail = "";
 
     if (auditOpen() == false)
     {
         return;
     }
 
-    opPathLen = std::strlen(opPath) + 1;
+    opPath = "op=" + std::string(req.methodString()) + ":" +
+             std::string(req.target()) + " ";
+    // Account for NULL
+    opPathLen = opPath.length() + 1;
     if (opPathLen > bufLeft)
     {
         // Truncate event message to fit into fixed sized buffer.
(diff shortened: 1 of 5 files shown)


include/audit_events.hpp
```diff
@@ -123,76 +123,157 @@ inline bool wantAudit(const crow::Request& req)
...
+        }
+        else
+        {
+            strncat(cnfgBuff, detail.c_str(), detail.length());
+            bufLeft -= detail.length();
+        }
+    }
+
     // encode user account name to ensure it is in an appropriate format
     user = audit_encode_nv_string("acct", userName.c_str(), 0);
```
baemyung, line 245: Similar,  the variables can defined here.
```
    char* user = audit_encode_nv_string("acct", userName.c_str(), 0);
    size_t userLen = 0;
```

include/audit_events.hpp
```diff
@@ -123,76 +123,157 @@ inline bool wantAudit(const crow::Request& req)
...
+            return false;
+    }
+}
 
+inline void auditEvent(const crow::Request& req, const std::string& userName,
+                       bool success)
+{
+    std::string opPath;
     char cnfgBuff[256];
     size_t bufLeft = 256; // Amount left available in cnfgBuff
```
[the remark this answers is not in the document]
jeaaustx: Good point. Corrected.
Note: I'm including the count for '\0' in the opPathLen adjustment to bufLeft rather than bufLeft.

include/audit_events.hpp
```diff
@@ -123,76 +123,157 @@ inline bool wantAudit(const crow::Request& req)
...
     char* user = NULL;
     size_t opPathLen;
     size_t userLen = 0;
     int rc;
     int origErrno;
+    std::string ipAddress;
+    std::string detail = "";
```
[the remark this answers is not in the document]
jeaaustx: Made the change.

include/audit_events.hpp
```diff
@@ -123,76 +123,157 @@ inline bool wantAudit(const crow::Request& req)
...
 
     if (auditOpen() == false)
```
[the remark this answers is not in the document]
jeaaustx: Changed. When I coded this I wasn't sure if the shorthand checking was acceptable in this repo.

include/audit_events.hpp
```diff
@@ -123,76 +123,157 @@ inline bool wantAudit(const crow::Request& req)
...
     {
         return;
     }
 
-    opPathLen = std::strlen(opPath) + 1;
+    opPath = "op=" + std::string(req.methodString()) + ":" +
+             std::string(req.target()) + " ";
+    // Account for NULL
+    opPathLen = opPath.length() + 1;
```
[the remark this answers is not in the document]
jeaaustx: Done.

include/audit_events.hpp
```diff
@@ -123,76 +123,157 @@ inline bool wantAudit(const crow::Request& req)
...
     cnfgBuff[opPathLen - 1] = '\0';
     bufLeft -= opPathLen;
 
+    // Determine any additional info for the event
+    if (wantDetail(req))
+    {
+        detail = req.body + " ";
+    }
+
+    if (detail.length() > 0)
```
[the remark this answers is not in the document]
jeaaustx: Yes, I've made the change.

include/audit_events.hpp
baemyung: Either one should be okay in this case as the only one and here is accessing it.

include/audit_events.hpp
```diff
@@ -188,24 +188,21 @@ inline void auditEvent(const crow::Request& req, const std::string& userName,
...
-    size_t bufLeft = 256; // Amount left available in cnfgBuff
-    char* user = NULL;
-    size_t opPathLen;
-    size_t userLen = 0;
+    size_t bufLeft = sizeof(cnfgBuff); // Amount left available in cnfgBuff
     int rc;
     int origErrno;
     std::string ipAddress;
-    std::string detail = "";
+    std::string detail;
```
baemyung, line 196: Itis generally better to define the variables close to where they are needed & with the proper initialization.

For example, rc & origErrno are better be defined with the initial values.
Or those variables are not needed to define if `auditOpen()' is false.

So I would move like

```
inline void auditEvent(const crow::Request& req, const std::string& userName,
                       bool success)
{
    if (!auditOpen())
    {
        return;
    }

==> Declare opPath & buff
    char cnfgBuff[256];
    size_t bufLeft = sizeof(cnfgBuff); // Amount left available in cnfgBuff
   std::string  opPath = "op=" + std::string(req.methodString()) + ":" +
             std::string(req.target()) + " ";
...

==> 'detail' seems needed here
    // Determine any additional info for the event
   std::string detail;
    if (wantDetail(req))
    {
        detail = req.body + " ";
    }

...
==> ipAddress are needed here
   std::string     ipAddress = req.ipAddress.to_string();
....
==> 'rc' is to handle audit_log_...()
   int     rc = audit_log_user_message(auditfd, AUDIT_USYS_CONFIG, cnfgBuff,
                                boost::asio::ip::host_name().c_str(),
                                ipAddress.c_str(), NULL, int(success));
    if (rc <= 0)
    {
        // Something failed with existing connection. Try to establish a new
        // connection and retry if successful.
        // Preserve original errno to report if the retry fails.
==> origErrno is a temporary variable to save the current  errno
        int origErrno = errno;
.....
        if (rc <= 0)
        {
            BMCWEB_LOG_ERROR << "Error writing audit message: " << origErrno;
        }
    }

    return;
}
```

include/audit_events.hpp
```diff
@@ -242,7 +239,17 @@ inline void auditEvent(const crow::Request& req, const std::string& userName,
...
+    std::unique_ptr<char, void (*)(char*)> userUP(user, [](char* ptr) {
+        if (ptr != nullptr)
+        {
+            ::free(ptr);
+        }
+    });
+
     if (user == NULL)
     {
         BMCWEB_LOG_ERROR << "Error appending user to audit msg : " << errno;
```
baemyung, line 255: This `errno` would need to be used right after `audit_encode_nv_string()`

So, wouldn't this code may be

```
    size_t userLen = 0;
    char* user = audit_encode_nv_string("acct", userName.c_str(), 0);
    if (user == NULL)
    {
        BMCWEB_LOG_ERROR << "Error appending user to audit msg : " << errno;
    }
   else
   {

        // setup a unique_ptr to handle freeing memory from user
        std::unique_ptr<char, void (*)(char*)> userUP(user, [](char* ptr) {
                      if (ptr != nullptr)
                      {
                                  ::free(ptr);
                      }
        });

        userLen = std::strlen(user);

        if (userLen > bufLeft)
        {
            // Username won't fit into fixed sized buffer. Leave it off.
            BMCWEB_LOG_WARNING << "Audit buffer too small for username:"
                               << " bufLeft=" << bufLeft
                               << " userLen=" << userLen;
        }
        else
        {
            strncat(cnfgBuff, user, userLen);
            bufLeft -= userLen;
        }
    }

```

baemyung, line 255: BMCWEB_LOG_ERROR  (in L255 ) seems not a fatal error case, but more of informational or warning message?

include/audit_events.hpp
jeaaustx: That makes sense, I've made these changes.

include/audit_events.hpp
jeaaustx: Good catch, I should have caught that myself!

jeaaustx: Agreed, I changed to warning.

include/audit_events.hpp
```diff
@@ -123,86 +123,168 @@ inline bool wantAudit(const crow::Request& req)
...
+inline void auditEvent(const crow::Request& req, const std::string& userName,
+                       bool success)
+{
+    if (!auditOpen())
     {
         return;
     }
 
-    opPathLen = std::strlen(opPath) + 1;
+    char cnfgBuff[256];
```
baemyung, line 194: It looks like - the code logic seems trying to combine 3 attrs (e.g. opPath, detail, user) into a limited-sized buffer, and each append-logic seems essentially same.

So, it could be refactored into a small append function and use it for each attr like

```
inline bool append_item_to_buffer(std::string& cnfgBuffer, size_t maxOutSize,
                                  const std::string& item)
{
    if (cnfgBuf.length() + item.length() > maxOutSize)
    {
        return false;
    }
    cnfgBuffer += item;
    return true;
}

inline void auditEvent(const crow::Request& req, const std::string& userName,
                       bool success)
{
    if (!auditOpen())
    {
        return;
    }

    constexpr size_t cnfgBuffMaxSize = 256;
    std::string cnfgBuffer;

    std::string opPath = "op=" + std::string(req.methodString()) + ":" +
                         std::string(req.target()) + " ";
    if (!append_item_to_buffer(cnfgBuffer, cnfgBuffMaxSize, opPath))
    {
        BMCWEB_LOG_WARNING << "Audit buffer too small, truncating:"
                           << " bufLeft=" << bufLeft
                           << " opPathLen=" << opPathLen;
    }

    if (wantDetail(req))
    {
        std::string detail = req.body + " ";

        if (!append_item_to_buffer(cnfgBuffer, cnfgBuffMaxSize, detail))
        {
            BMCWEB_LOG_WARNING << "Audit buffer too small for data:"
                               << " bufLeft=" << bufLeft
                               << " detailLen=" << detail.length();
        }
    }

    char* user = audit_encode_nv_string("acct", userName.c_str(), 0);
    if (user == NULL)
    {
        BMCWEB_LOG_WARNING << "Error encoding user for audit msg : " << errno;
    }
    else
    {
        // setup a unique_ptr to handle freeing memory from user
        std::unique_ptr<char, void (*)(char*)> userUP(user, [](char* ptr) {
            if (ptr != nullptr)
            {
                ::free(ptr);
            }
        });
        if (!append_item_to_buffer(cnfgBuffer, cnfgBuffMaxSize,
                                   std::string(user)))
        {
            // Username won't fit into fixed sized buffer. Leave it off.
            BMCWEB_LOG_WARNING << "Audit buffer too small for username:"
                               << " bufLeft=" << bufLeft
                               << " userLen=" << userLen;
        }
    }
....

```

jeaaustx: Thank you Myung, this is a good idea. I've changed to use this approach. Note: When adding opPath I do want to truncate it rather than just not include it. (So any audit message will have at least a partial Redfish path included.) So when opPath is added I'm using std::substr instead to get the truncation.

baemyung (approved):

PR #727, opened 2023-07-11
Incorrect value displayed for total system memory

PriyangaRamasamy:
GUI displays 0 as total system memory when the total size is not exactly divisible by 1024 as the code truncates the floating point value.

This commit has the fix for it.


PriyangaRamasamy:
jenkins run tests please

gtmills (approved):

SunnySrivastava1984 (approved):

PR #728, opened 2023-07-11
Use std::string_view interface type in getDbusObject

Lakshmi-y:
Change the interface type to std::string_view in the getDbusObject function to ensure consistency with other functions such as getSubTree and getSubTreePaths.

Upstream commit:
https://gerrit.openbmc.org/c/openbmc/bmcweb/+/60230

Tested: Redfish Validator passed

### DIFF
--- a/redfish-core/lib/pcie.hpp
+++ b/redfish-core/lib/pcie.hpp
@@ -35,10 +35,10 @@
 namespace redfish
 {
 
-static constexpr const char* pcieDeviceInterface =
-    "xyz.openbmc_project.Inventory.Item.PCIeDevice";
-static constexpr const char* pcieSlotInterface =
-    "xyz.openbmc_project.Inventory.Item.PCIeSlot";
+static constexpr std::array<std::string_view, 1> pcieDeviceInterface = {
+    "xyz.openbmc_project.Inventory.Item.PCIeDevice"};
+static constexpr std::array<std::string_view, 1> pcieSlotInterface = {
+    "xyz.openbmc_project.Inventory.Item.PCIeSlot"};
 
 using FindPcieSlotCbFunc = std::function<void(
     const std::string&, const dbus::utility::MapperServiceMap&)>;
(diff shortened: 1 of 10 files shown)


baemyung:
> Upstream commit:
> https://gerrit.openbmc.org/c/openbmc/bmcweb/+/60230

Would you add "Change-Id" & the previous "Signer" in both commit message to ensure the reference and credit?

redfish-core/lib/power_supply_metrics.hpp
```diff
@@ -43,8 +43,8 @@ static const char* historyMaximumInterface =
 static const char* historyAverageInterface =
     "org.open_power.Sensor.Aggregation.History.Average";
 
-static std::array<const char*, 2> historyInterfaces{historyMaximumInterface,
-                                                    historyAverageInterface};
+static std::array<std::string_view, 2> historyInterfaces{
+    historyMaximumInterface, historyAverageInterface};
```
baemyung, line 47: Since `historyMaximumInterface` & `historyAverageInterface` are used only for constructing `historyInterfaces`, I think we can simply write it as

```
static constexpr std::array<std::string_view, 2> historyInterfaces{
    "org.open_power.Sensor.Aggregation.History.Maximum",
     "org.open_power.Sensor.Aggregation.History.Average" };
```

redfish-core/lib/pcie.hpp
```diff
@@ -35,10 +35,10 @@
...
 {
 
-static constexpr const char* pcieDeviceInterface =
-    "xyz.openbmc_project.Inventory.Item.PCIeDevice";
-static constexpr const char* pcieSlotInterface =
-    "xyz.openbmc_project.Inventory.Item.PCIeSlot";
+static constexpr std::array<std::string_view, 1> pcieDeviceInterface = {
+    "xyz.openbmc_project.Inventory.Item.PCIeDevice"};
+static constexpr std::array<std::string_view, 1> pcieSlotInterface = {
+    "xyz.openbmc_project.Inventory.Item.PCIeSlot"};
```
baemyung, line 41: Since these may be (are) referenced in other places - e.g. fabric_adapters.hpp, pcie_slots.hpp etc, can they be located in `utils/pcie_util.hpp`, or in a new utils (e.g. utils/interface_utils.hpp)?

Or, we can do those refactoring later...

redfish-core/lib/power_supply_metrics.hpp
gtmills: This change is changing to std::string_view, I think outside this commit

gtmills (approved):
This is to match upstream..

baemyung (approved):

PR #739, opened 2023-07-31
1050: Reduce memory usage during Firmware update

baemyung:
bmcweb currently consumes the excessive memory during firmware update. It is due to the replications of the req.body (which is the full content of firmware).

- req.body is captured by-value instead of by-reference in a lambda when the user-validation.

- audit copies the entire req.body before truncating it to the buffer.

Tested:

- Redfish validator passes
- Monitor memory consumption of bmcweb process periodically. e.g., top -b -d 0.5 | grep /usr/bin/bmcweb
- With the fix, the max memory usage should only be increased about the size of the image.

### DIFF
--- a/http/routing.hpp
+++ b/http/routing.hpp
@@ -1460,7 +1460,7 @@ class Router
         std::string username = req.session->username;
 
         crow::connections::systemBus->async_method_call(
-            [req, asyncResp, &rule, params](
+            [&req, asyncResp, &rule, params](
                 const boost::system::error_code ec,
                 const dbus::utility::DBusPropertiesMap& userInfoMap) mutable {
             if (ec)
--- a/include/audit_events.hpp
+++ b/include/audit_events.hpp
@@ -229,7 +229,7 @@ inline void auditEvent(const crow::Request& req, const std::string& userName,
     std::string detail;
     if (wantDetail(req))
     {
-        detail = req.body + " ";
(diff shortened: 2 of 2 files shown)


baemyung:
This is to address https://jazz07.rchland.ibm.com:13443/jazz/web/projects/CSSD#action=com.ibm.team.workitem.viewWorkItem&id=552854

Upstream is not affected as it already captures (`&req`) by reference.

jeaaustx:
Long term I should change to not include the detail for these events. Doesn't make any sense for the detail to be part of the event if it is just binary data.

baemyung:
> 
It is a part of firmware update, and thus this would need to be put into the audit event records.
`curl -k -H Content-Type: application/octet-stream -X POST -T ${image-file}  https://${bmc}/redfish/v1/UpdateService/update`
Regardless of the filtering, it would still be proper to limit the string copy to the buffer size, instead of copying it first and then truncating it.

baemyung:
For the reference, here is the analysis of the memory consumption.

- Image octet size = `162m`
- (1) req.body is captured by-value for GetUserInfo causes the additional copy of the image = `+162m` => `330m`
- (2) audit event handling causes about 3 more copies temporarily

  `detail = req.body + " "`;

  During the operation, it is duplicated upto 3 more copies (i.e. around additional of `+480m`) ==> `645m`. (even with the fix of (1))

PR #740, opened 2023-07-31
RedfishEvents: Avoid setting SNI for IP addresses

raviteja-b:
ssl_handshake fails while establishing connection to IPv6 destination address. as IPv6 addressses considered as invalid value for SNI hostname due to special characters.

SNI allows valid HostName which Allows characters are only {alphabetic
characters (A-Z), numeric characters (0-9), the minus sign.

Tested By: Verified redfish events 1. Subscribing Destination with IPv6 address.  2. Subscribing Destination with IPv4 address.

Change-Id: I32d30292bbc29c753f1c1815c66fcc93e8074eaa

### DIFF
--- a/http/http_client.hpp
+++ b/http/http_client.hpp
@@ -37,6 +37,7 @@
 #include <boost/beast/version.hpp>
 #include <boost/container/devector.hpp>
 #include <boost/system/error_code.hpp>
+#include <boost/url/url_view.hpp>
 #include <logging.hpp>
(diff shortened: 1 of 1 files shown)


sunharis (approved):
Changes worked good for the HMC usecases. 
I am not sure how will be accepted at the community though. But this is needed for the downstream - for the HMC to work with the push style events.

gtmills (approved):

PR #741, opened 2023-07-31
Audit Log: Limit additional data passed to audit log

jeaaustx:
Avoid making copy of large body data by limiting to maximum that could be used.

### DIFF
--- a/http/http_connection.hpp
+++ b/http/http_connection.hpp
@@ -467,7 +467,7 @@ class Connection :
                     (req->method() == boost::beast::http::verb::put &&
                      !req->target().starts_with("/ibm/v1")))
                 {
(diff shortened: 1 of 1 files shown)


gtmills:
Can you add a bit more detail, the defect, the 1050 fix this was included in, etc.

gtmills (approved):

baemyung (approved):

baemyung:
For the reference, here is the analysis of the memory consumption.
This previous audit event handling causes about 3 more copies temporarily

  `detail = req.body + " "`;

  During the operation, it is duplicated upto 3 more copies (i.e. around additional of `+480m`) ==> `645m`.

PR #747, opened 2023-08-21
Push Style Event: Cleanup at maxRequestQueueSize

sunharis:
HMC subscribes with keep_alive and RetryForever option. At some unusual circumstances, HMC abruptly closes the socket and does not send error response, BMC keeps waiting for the response and keeps adding the events to send queue. This ultimately results in not sending the events to HMC and hits max queue size.

This change is to clear the queue at this point and close the connection and do related cleanups. The next event will be sent good by creating new socket and connections.

Tested by:
 Tested event sanity by connecting a HMC and performing 10
 iterations of PowerOn and PoweOff.
 Could not recreate the original issue with this patch

### DIFF
--- a/http/http_client.hpp
+++ b/http/http_client.hpp
@@ -451,10 +451,6 @@ class ConnectionInfo : public std::enable_shared_from_this<ConnectionInfo>
             {
                 state = ConnState::suspended;
             }
-            if (connPolicy->retryPolicyAction == "RetryForever")
-            {
-                state = ConnState::idle;
-            }
 
             // We want to return a 502 to indicate there was an error with
             // the external server
@@ -714,6 +710,18 @@ class ConnectionPool : public std::enable_shared_from_this<ConnectionPool>
             return;
         }
 
+        if (requestQueue.size() == maxRequestQueueSize)
+        {
+            // We can remove the request from the queue at this point
+            BMCWEB_LOG_ERROR << "requestQueue is full. Clearing the queue for "
+                             << destIP << ":" << std::to_string(destPort);
+            requestQueue.clear();
+            // Let's close the connection and restart from resolve for next
+            // event.
+            conn->doClose();
+            return;
+        }
+
         // No more messages to send so close the connection if necessary
         if (keepAlive)
         {
@@ -796,13 +804,6 @@ class ConnectionPool : public std::enable_shared_from_this<ConnectionPool>
             Response dummyRes;
             dummyRes.result(boost::beast::http::status::too_many_requests);
             resHandler(dummyRes);
-        }
-        if (requestQueue.size() == maxRequestQueueSize)
-        {
-            // We can remove the request from the queue at this point
-            BMCWEB_LOG_ERROR << "requestQueue is full. Clearing the queue for "
-                             << destIP << ":" << std::to_string(destPort);
-            requestQueue.pop_front();
         }
     }
 


sunharis:
Upstreaming this will be added to story https://jsw.ibm.com/browse/PFEBMC-1016

sunharis:
Fixes defect https://jazz07.rchland.ibm.com:13443/jazz/web/projects/CSSD#action=com.ibm.team.workitem.viewWorkItem&id=563569

http/http_client.hpp
```diff
@@ -714,6 +710,18 @@ class ConnectionPool : public std::enable_shared_from_this<ConnectionPool>
             return;
         }
 
+        if (requestQueue.size() == maxRequestQueueSize)
```
baemyung, line 713: Will this part code be always executed everytime `requestQueue` needs to be changed?
Or is there any case that `requestQueue` may be changed more than `1` at time?

Perhaps would this be more safer?

```
if (requestQueue.size() >= maxRequestQueueSize)
```

http/http_client.hpp
```diff
@@ -451,10 +451,6 @@ class ConnectionInfo : public std::enable_shared_from_this<ConnectionInfo>
             {
                 state = ConnState::suspended;
             }
-            if (connPolicy->retryPolicyAction == "RetryForever")
```
gtmills, line 454: Why get rid of this?

gtmills:
> HMC subscribes with keep_alive and RetryForever option. At some unusual circumstances, HMC abruptly closes the socket and does not send error response, BMC keeps waiting for the response and keeps adding the events to send queue. This ultimately results in not sending the events to HMC and hits max queue size.
 
For 1060, we plan to have the GUI use events and changes like this (specific to HMC, make that harder :/ ) 
This feels like we are moving more towards SuspendRetries / TerminateAfterRetries, which is the policy I think HMC should be using. Long term, is that something the HMC agrees with (move to SuspendRetries / TerminateAfterRetries)?


My concerns here are:
 1) This is adding TerminateAfterRetries behavior but still calling it RetryForever. 

2) We are hacking around an HMC bug "HMC abruptly closes the socket and does not send error response". 

3) This isn't something upstream is going to accept as is. Can we talk with upstream now?

4) This is a really intense intervention. We are clearing the .queue(), what if there are events from the GUI or another client in the queue?

Could we discuss this with Ed? 
Could we talk with the HMC about the long-term plan here?

baemyung:
@sunharis 

I'm wondering about the semantics of "RetryForever" option.

Are these correct assumptions?
1) a client sends a request to bmcweb and bmcweb remembers the request (i.e. connection between the client & bmcweb). 
2) bmcweb keeps trying to connect to the destination continuously.

In other words, it is not the client who keeps trying to send the requests repeatedly, right?

If so, 
1) the previously received requests should be kept rather than setting them as "ConnState::idle".  I think your proposed change seems doing it.

2) Now if MaxPool is already full, the current & your proposed logic seems trying to remove the older one.  
    Isn't this contradicting the semantics (RetryForever)?
    Shouldn't it be left as full so that any new requests are to be denied?
    The older ones should be removed by closing the older clients?

----
In summary, the proposed change seems 
-  removing the option of "RetryForever" and
- also always clearing up the queued requests if it becomes full.

baemyung:
So, if the semantics for RetryForever is to retry forever, the area to change seems at `waitAndRetry()` so that it will keep retrying regardless of the `retryCount`.

```
diff --git a/http/http_client.hpp b/http/http_client.hpp
index 52eaa209..57c1d8e9 100644
--- a/http/http_client.hpp
+++ b/http/http_client.hpp
@@ -434,7 +434,8 @@ class ConnectionInfo : public std::enable_shared_from_this<ConnectionInfo>

     void waitAndRetry()
     {
-        if ((retryCount >= connPolicy->maxRetryAttempts) ||
+        if ((connPolicy->retryPolicyAction != "RetryForever" &&
+             retryCount >= connPolicy->maxRetryAttempts) ||
             (state == ConnState::sslInitFailed))
         {
             BMCWEB_LOG_ERROR << "Maximum number of retries reached."
@@ -451,10 +452,6 @@ class ConnectionInfo : public std::enable_shared_from_this<ConnectionInfo>
             {
                 state = ConnState::suspended;
             }
-            if (connPolicy->retryPolicyAction == "RetryForever")
-            {
-            }

             // We want to return a 502 to indicate there was an error with
             // the external server
@@ -797,13 +794,6 @@ class ConnectionPool : public std::enable_shared_from_this<ConnectionPool>
             dummyRes.result(boost::beast::http::status::too_many_requests);
             resHandler(dummyRes);
         }
-        if (requestQueue.size() == maxRequestQueueSize)
-        {
-            // We can remove the request from the queue at this point
-            BMCWEB_LOG_ERROR << "requestQueue is full. Clearing the queue for "
-                             << destIP << ":" << std::to_string(destPort);
-            requestQueue.pop_front();
-        }
     }

```

sunharis:
> In summary, the proposed change seems
> 
> removing the option of "RetryForever" and
> also always clearing up the queued requests if it becomes full.

Yes, thats what i did. This change is kept on hold as of now. We need more design discussion to get this implemented properly. The current "Retryforever" implementation is not correct as you mentioned in your comments. But this is how it was done for HMC usecases. We have a story to work this upstream and design in the right way.

sunharis:
Closing this PR as we are perusing different solutions. JIRA story will contain the updates

PR #752, opened 2023-08-28
VMI:Not able to set static IPv6 address when DHCPv6 is enabled

abhilashraju:
Bug Fix for 562664.
Pre-Condition:
DHCPv4::DHCPEnabled:true
DHCPv6::OperationMode:Enabled
Try to set IPV6 static address.
IPV6 static adress is not set.

Fix:
There is an oreder mismatch in setting ip adress and disabling operation mode. Disabling operation mode should be called before setting the static Ip address. Otherwise operation mode call will reset Ip adress to zero.

Change-Id: If1337e2eccb1078e44d3cfa0902b728e0e536241

### DIFF
--- a/redfish-core/lib/pcie_slots.hpp
+++ b/redfish-core/lib/pcie_slots.hpp
@@ -229,12 +229,13 @@ inline void linkAssociatedDiskBackplane(
                                 drivePath);
             if (it != assemblyList.end())
             {
+                asyncResp->res.jsonValue["Slots"][index]["Links"]["Oem"]
+                                        ["@odata.type"] = "#OemPCIeSlots.Oem";
                 asyncResp->res.jsonValue["Slots"][index]["Links"]["Oem"]["IBM"]
-                                        ["AssociatedAssembly"] = {
-                    {{"@odata.id",
-                      "/redfish/v1/Chassis/" + chassisId +
-                          "/Assembly#/Assemblies/" +
-                          std::to_string(it - assemblyList.begin())}}};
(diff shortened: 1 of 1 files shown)


redfish-core/lib/hypervisor_system.hpp
```diff
@@ -980,17 +980,18 @@ inline void handleHypervisorIPv4StaticPatch(
         {
             return;
         }
-
+
```
gtmills, line 983: I don't think we want these spaces

gtmills:
Is there an upstream commit here?

gtmills:
For next time, please use a branch name that describes the commit. (e.g. fix-vmi-ipv6-address) Thanks!

redfish-core/lib/hypervisor_system.hpp
raviteja-b: +1

raviteja-b (approved):
Please add test cases verified

sunharis (approved):
Please update the test cases in the commit message. Code changes looks good

asmithakarun (approved):
If you could add some more description in the commit msg/description, it would be better

"DHCP should be disabled first when there is a static configuration request. Right now, static ip is configured and then dhcp mode is disabled causing the ip properties should be resetted"

gtmills (changes requested):
spaces need to be addressed, please add the upstream commit, and answer any other questions. Thanks!

gtmills:
jenkins run tests please

PR #754, opened 2023-08-30
Remove location header

gtmills:
https://jazz07.rchland.ibm.com:13443/jazz/web/projects/CSSD#action=com.ibm.team.workitem.viewWorkItem&id=548905 is the defect 

The location header is not used. This header is causing problems because it looks like the following when returned Location
	�Kõ�����/v1/AccountService/LDAP/Certificates/1
and chrome is rejecting that response with a
ERR_INVALID_HTTP_RESPONSE


gtmills:
This is a hack, https://github.com/ibm-openbmc/bmcweb/pull/757 is a real fix. Let's go with it

PR #760, opened 2023-09-07
Enable more RC traces

gtmills:
The return code coming back from D-Bus is really useful. For example on 568739, assume these were all 110 (timeout) but don't know that without an return code trace.

These are all downstream only :/
LEDs are up for review at
https://gerrit.openbmc.org/c/openbmc/bmcweb/+/57765 but Lakshmi has already added this tracing.


baemyung (approved):
The changes seem good for knowing the cause of those mysterious symptoms.

PR #763, opened 2023-09-13
1050: Fix the removal of unrelated queued http request

baemyung:
During RetryForever support implementation, the front queued item is removed if the queud becomes full, regardless whether it is related to RetryForever option or not.

This logic may cause the side-effect.
For example, theoretically, if the queued item is still being handled and left in the async-call schedulers, the removed object may be referenced when the scheduler pops.

This change is to preserve the queued items to be active as long as they are still connected.  If the newer requests cause the queue pool overflowed, they will be rejected (rather than the queue item is to removed).

### DIFF
--- a/http/http_client.hpp
+++ b/http/http_client.hpp
@@ -797,13 +797,6 @@ class ConnectionPool : public std::enable_shared_from_this<ConnectionPool>
             dummyRes.result(boost::beast::http::status::too_many_requests);
             resHandler(dummyRes);
         }
-        if (requestQueue.size() == maxRequestQueueSize)
-        {
-            // We can remove the request from the queue at this point
-            BMCWEB_LOG_ERROR << "requestQueue is full. Clearing the queue for "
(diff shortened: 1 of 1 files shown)


baemyung:
Related references:
- https://syswiki02.rchland.ibm.com/e/en/openbmc/openbmc-downstream-design-meeting-topics
- slack: https://ibm-systems-power.slack.com/archives/GBEFW3F5J/p1692674871586019
- Discussion over PR: https://github.com/ibm-openbmc/bmcweb/pull/747

Related STG defect:  https://jazz07.rchland.ibm.com:13443/jazz/web/projects/CSSD#action=com.ibm.team.workitem.viewWorkItem&id=560112

sunharis (approved):

rfrandse:
https://ibm-systems-power.slack.com/archives/C04KG79NAUA/p1694809081579159?thread_ts=1694807550.433449&cid=C04KG79NAUA

PR #700, opened 2023-06-06
Configure DHCP4 and DHCP6 independently

cmjishnu:
At present, DHCP parameters like DNSEnabled, NTPEnabled etc. are
shared between DHCPv4 and DHCPv6 in the network configuration.

This change is to enable the possibility to have separate writable
parameters for DHCPv4 and DHCPv6 in the network configuration 
to be able to configure them independently.

The DHCP configuration was in the network manager level earlier,
it has been moved to interface level with
https://github.com/ibm-openbmc/phosphor-networkd/pull/87.
This bmcweb change is to move the redfish settings to the interface level.

Tested by:

Patch request:
  
 curl -k -H "X-Auth-Token: $bmc_token" -X  PATCH -d 
 '{"DHCPv6":{"UseNTPServers":false}}' 
 https://${bmc}/redfish/v1/Managers/bmc/EthernetInterfaces/eth1
 
Patch Response:
 {
  "@Message.ExtendedInfo": [
    {
      "@odata.type": "#Message.v1_1_1.Message",
      "Message": "The request completed successfully.",
      "MessageArgs": [],
      "MessageId": "Base.1.13.0.Success",
      "MessageSeverity": "OK",
      "Resolution": "None"
    }
  ],

Get Request:
curl -k -H "X-Auth-Token: $bmc_token" -X GET 
https://${bmc}/redfish/v1/Managers/bmc/EthernetInterfaces/eth1
  
Get Response:
{
  "@odata.id": "/redfish/v1/Managers/bmc/EthernetInterfaces/eth1",
  "@odata.type": "#EthernetInterface.v1_8_0.EthernetInterface",
  "DHCPv4": {
    "DHCPEnabled": true,
    "UseDNSServers": true,
    "UseDomainName": true,
    "UseNTPServers": true
  },
  "DHCPv6": {
    "OperatingMode": "Enabled",
    "UseDNSServers": true,
    "UseDomainName": true,
    "UseNTPServers": false
  },  
  

Change-Id: I5db29b6dfc8966ff5af51041da11e5b79da7d1dd

PR #766, opened 2023-09-28
Configure DHCP4 and DHCP6 parameters independently

cmjishnu:
This is a follow-up PR for https://github.com/ibm-openbmc/bmcweb/pull/700
It includes the review comments incorporations.

Upstream : https://gerrit.openbmc.org/c/openbmc/bmcweb/+/63390
tested by:
Used the redfish commands to set and get the parameter values individually for both DHCPv4 and DHCPv6.
Verified the network configuration file is updated accordingly.
Verified unwanted logs are not printed in the journal logs

Change-Id: I5db29b6dfc8966ff5af51041da11e5b79da7d1dd

### DIFF
--- a/redfish-core/lib/ethernet.hpp
+++ b/redfish-core/lib/ethernet.hpp
@@ -1582,8 +1582,7 @@ inline void setDHCPConfig(const std::string& propertyName, const bool& value,
         path /= "dhcp6";
     }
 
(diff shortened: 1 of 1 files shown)


gtmills:
@cmjishnu Is this still needed or can we close?

PR #767, opened 2023-09-29
Bypass authentication for requests sent internally

gtmills:
PATCH request containing If-Match header with the value of etag causes the BMCWEB to crash. To compute the etag, a new GET request for the same URI is internally generated. Same req.session is assigned to the new request. The new req gets authenticated and when privileges are assigned in  populateUserInfo() BMCWEB will get crashed when req.session members are accessed.

For the new GET request which have the same req.session created for original PATCH request, bypass authentication and privilege assignment, since the same credentials are already validated and its a overhead to validate again.

Tested: Run Redfish Protocol validator and got passed.

Change-Id: Iaad2b1f0e982b480e9fab24005ce8b7c258882c3

Upstream is https://gerrit.openbmc.org/c/openbmc/bmcweb/+/66685

### DIFF
--- a/http/app.hpp
+++ b/http/app.hpp
@@ -66,9 +66,10 @@ class App
     }
 
     void handle(Request& req,
-                const std::shared_ptr<bmcweb::AsyncResp>& asyncResp)
+                const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
+                bool bypassAuth = false)
     {
-        router.handle(req, asyncResp);
+        router.handle(req, asyncResp, bypassAuth);
     }
 
     DynamicRule& routeDynamic(std::string&& rule)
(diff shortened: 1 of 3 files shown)


baemyung:
Looking at the the chains of the gerrit reviews and the related references (https://github.com/openbmc/bmcweb/issues/262) and the codes, it seems that the changes are to bypass auth when a new req is internally created.

The change seems still safer as its change scope is only related to do bypassing the auth under the same session by original PATCH request. 

So, I think we can include into 1050.

baemyung (approved):
Looking at the the chains of the gerrit reviews and the related references (https://github.com/openbmc/bmcweb/issues/262) and the codes, it seems that the changes are to bypass auth when a new req is internally created.

The change seems still safer as its change scope is only related to do bypassing the auth under the same session by original PATCH request.

So, I think we can include into 1050.

PR #775, opened 2023-10-17
BMCWEB log flooding : move ERROR trace log to DEBUG

isha309:
-   below error trace seems to flood the BMCWEB LOGS
`"Max pool size reached. Adding data to queue. destIP: "`
- will move the log level from ERROR to Debug to solve the problem 

UPSTREAM :
the log level of the mentioned trace is already DEBUG IN upstream 
so no changes need for upstream .
https://github.com/openbmc/bmcweb/blob/master/http/http_client.hpp#L757

DEFECT :
https://jazz07.rchland.ibm.com:13443/jazz/web/projects/CSSD#action=com.ibm.team.workitem.viewWorkItem&id=580659

### DIFF
--- a/http/http_client.hpp
+++ b/http/http_client.hpp
@@ -783,7 +783,7 @@ class ConnectionPool : public std::enable_shared_from_this<ConnectionPool>
         }
         else if (requestQueue.size() < maxRequestQueueSize)
         {
-            BMCWEB_LOG_ERROR << "Max pool size reached. Adding data to queue."
+            BMCWEB_LOG_DEBUG << "Max pool size reached. Adding data to queue."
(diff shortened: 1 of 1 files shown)


sunharis (commented):
Please correct the commit message according to the bmcweb guidelines.

isha309:
> Please correct the commit message according to the bmcweb guidelines.

corrected the commit message

baemyung:
In upstream, it was changed by https://github.com/openbmc/bmcweb/commit/bf2ddede1c5f68f81ece754f511aefdafcdb2815 and there are more entries.

If we choose to do cherry-picking, wouldn't it be better to include all of those?

PR #789, opened 2023-11-16
No Connection: Fix for http client stale connection

abhilashraju:
http_client is not handling broken connections well. Eg: Connection termination by server due to keep alive timeout.

At present client is not aware of connection termination from server. So whenever next event ready to be sent client will try send/receives data over broken connection. After failed operation the client will try to restart the connection after closing the current connection. During the process the ssl shutdown will hang when it is made over a broken connection.

Solution:
Restart the connection using fresh new socket object rather than reusing object.Thus the ssl shutdown over broken connection can be avoided.

Tested:
Developer test has been done using bmc and hmc located in different network and let the keep-alive timeout occured. The connection has been restablished on arrival or new event and events are published succesfully to hmc

Change-Id: Ie983f57cbe6e688176501013401009ef0e057ac2

### DIFF
--- a/http/http_client.hpp
+++ b/http/http_client.hpp
@@ -48,7 +48,6 @@
 #include <memory>
 #include <queue>
 #include <string>
-
 namespace crow
 {
 
@@ -502,7 +501,9 @@ class ConnectionInfo : public std::enable_shared_from_this<ConnectionInfo>
     }
     void restartConnection()
     {
-        BMCWEB_LOG_ERROR << "restartConnection ";
+        BMCWEB_LOG_DEBUG << host << ":" << std::to_string(port)
+                         << ", id: " << std::to_string(connId)
+                         << " restartConnection ";
         conn = makeConnection(ioc, sslConn.has_value());
         doResolve();
     }
@@ -530,6 +531,7 @@ class ConnectionInfo : public std::enable_shared_from_this<ConnectionInfo>
         {
             // Now let's try to resend the data
             state = ConnState::retry;
+            // reconnect to server using new socket
             restartConnection();
         }
         else
@@ -618,7 +620,6 @@ class ConnectionInfo : public std::enable_shared_from_this<ConnectionInfo>
     std::unique_ptr<boost::asio::ip::tcp::socket>
         makeConnection(boost::asio::io_context& iocIn, bool useSsl)
     {
-        BMCWEB_LOG_ERROR << "makeConnection ssl" << useSsl;
         auto newconn = std::make_unique<boost::asio::ip::tcp::socket>(iocIn);
         if (useSsl)
         {
@@ -782,9 +783,6 @@ class ConnectionPool : public std::enable_shared_from_this<ConnectionPool>
                 }
                 else
                 {
-                    BMCWEB_LOG_ERROR << "Reusing existing connection "
-                                     << commonMsg << "with state "
-                                     << static_cast<int>(conn->state);
                     conn->restartConnection();
                 }
                 return;
--- a/http/http_client.hpp
+++ b/http/http_client.hpp
@@ -497,7 +497,7 @@ class ConnectionInfo : public std::enable_shared_from_this<ConnectionInfo>
         }
 
         // Let's close the connection and restart from resolve.
-        doCleanup(true);
+        shutdownConn(true);
     }
     void restartConnection()
     {
@@ -538,11 +538,6 @@ class ConnectionInfo : public std::enable_shared_from_this<ConnectionInfo>
         {
             state = ConnState::closed;
         }
-    }
-    void doCleanup(bool retry)
-    {
-        BMCWEB_LOG_ERROR << "doCleanup called with retry " << retry;
-        shutdownConn(retry);
     }
 
     void doClose(bool retry = false)
--- a/http/http_client.hpp
+++ b/http/http_client.hpp
@@ -643,7 +643,6 @@ class ConnectionInfo : public std::enable_shared_from_this<ConnectionInfo>
                 boost::asio::ssl::context::tlsv13_client};
             sslConn.emplace(*newconn.get(), sslCtx);
             setCipherSuiteTLSext();
-            BMCWEB_LOG_ERROR << "created  ssl Conn" << sslConn.has_value();
         }
         return newconn;
     }
--- a/http/http_client.hpp
+++ b/http/http_client.hpp
@@ -149,9 +149,10 @@ class ConnectionInfo : public std::enable_shared_from_this<ConnectionInfo>
     // Ascync callables
     std::function<void(bool, uint32_t, Response&)> callback;
     crow::async_resolve::Resolver resolver;
-    boost::asio::ip::tcp::socket conn;
+    boost::asio::io_context& ioc;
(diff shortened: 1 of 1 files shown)


http/http_client.hpp
```diff
@@ -497,14 +498,19 @@ class ConnectionInfo : public std::enable_shared_from_this<ConnectionInfo>
         }
 
         // Let's close the connection and restart from resolve.
-        doClose(true);
+        doCleanup(true);
+    }
+    void restartConnection()
+    {
+        BMCWEB_LOG_ERROR << "restartConnection ";
```
sunharis, line 505: Please enhance this trace by adding the destination ip and port number. And edit as "INFO: restart connection to <host> : <ip>

http/http_client.hpp
```diff
@@ -604,18 +615,12 @@ class ConnectionInfo : public std::enable_shared_from_this<ConnectionInfo>
...
-        const std::string& destIPIn, uint16_t destPortIn, bool useSSL,
-        unsigned int connIdIn) :
-        subId(idIn),
-        connPolicy(connPolicyIn), host(destIPIn), port(destPortIn),
-        connId(connIdIn), conn(iocIn), timer(iocIn)
+    std::unique_ptr<boost::asio::ip::tcp::socket>
+        makeConnection(boost::asio::io_context& iocIn, bool useSsl)
     {
-        if (useSSL)
+        BMCWEB_LOG_ERROR << "makeConnection ssl" << useSsl;
```
sunharis, line 621: please remove this

http/http_client.hpp
```diff
@@ -635,10 +640,24 @@ class ConnectionInfo : public std::enable_shared_from_this<ConnectionInfo>
             } */
             boost::asio::ssl::context sslCtx{
                 boost::asio::ssl::context::tlsv13_client};
-            sslConn.emplace(conn, sslCtx);
+            sslConn.emplace(*newconn.get(), sslCtx);
             setCipherSuiteTLSext();
+            BMCWEB_LOG_ERROR << "created  ssl Conn" << sslConn.has_value();
```
sunharis, line 645: Add a space between the statement and value. Also please add the destination and port in the traces

http/http_client.hpp
```diff
@@ -763,9 +782,10 @@ class ConnectionPool : public std::enable_shared_from_this<ConnectionPool>
                 }
                 else
                 {
-                    BMCWEB_LOG_DEBUG << "Reusing existing connection "
-                                     << commonMsg;
-                    conn->doResolve();
+                    BMCWEB_LOG_ERROR << "Reusing existing connection "
```
sunharis, line 785: INFO: please. This is to identify this as information log and not an error

http/http_client.hpp
```diff
@@ -763,9 +782,10 @@ class ConnectionPool : public std::enable_shared_from_this<ConnectionPool>
...
+                                     << commonMsg << "with state "
```
sunharis, line 786: need spaces when you add the values. Else it becomes unreadable at journal traces

sunharis (changes requested):
Please apply the commit message 50/72 rule at downstream as well.

http/http_client.hpp
```diff
@@ -364,7 +365,7 @@ class ConnectionInfo : public std::enable_shared_from_this<ConnectionInfo>
         }
 
         timer.cancel();
-        if (ec && ec != boost::asio::ssl::error::stream_truncated)
```
gtmills, line 367: why remove the  boost::asio::ssl::error::stream_truncated ?

http/http_client.hpp
gtmills: How often does this log? We aren't going to be spamming the journal a crazy amount ?

http/http_client.hpp
```diff
@@ -524,13 +530,18 @@ class ConnectionInfo : public std::enable_shared_from_this<ConnectionInfo>
         {
             // Now let's try to resend the data
             state = ConnState::retry;
-            doResolve();
+            restartConnection();
```
gtmills, line 533: I think a comment here and a few other places would make this code more maintainable

http/http_client.hpp
```diff
@@ -149,9 +149,10 @@ class ConnectionInfo : public std::enable_shared_from_this<ConnectionInfo>
     // Ascync callables
     std::function<void(bool, uint32_t, Response&)> callback;
     crow::async_resolve::Resolver resolver;
-    boost::asio::ip::tcp::socket conn;
+    boost::asio::io_context& ioc;
     std::optional<boost::beast::ssl_stream<boost::asio::ip::tcp::socket&>>
         sslConn;
+    std::unique_ptr<boost::asio::ip::tcp::socket> conn;
```
gtmills, line 155: Why did you need to move this to a pointer?

gtmills (commented):
Neat stuff! 
Can you add a few comments to this code..?

http/http_client.hpp
abhilashraju: because we are not  planning to reuse the socket.  I don't know what benefit we get in reuse. But there are troubles seen after keep-alive time out happens.

http/http_client.hpp
```diff
@@ -497,14 +498,19 @@ class ConnectionInfo : public std::enable_shared_from_this<ConnectionInfo>
         }
 
         // Let's close the connection and restart from resolve.
-        doClose(true);
+        doCleanup(true);
```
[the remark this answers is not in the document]
abhilashraju: yes, we can do it. But giving explicit name similar to close looked readable. Also I felt it is easier to add new code (in case needed in future) before calling the shutdown. I am open to remove this function.

http/http_client.hpp
abhilashraju: Ack

http/http_client.hpp
abhilashraju: Ack

http/http_client.hpp
abhilashraju: Ack

http/http_client.hpp
abhilashraju: Ack

http/http_client.hpp
abhilashraju: Ack

http/http_client.hpp
sunharis: Not done?

sunharis (approved):
Upstream is being worked via https://jsw.ibm.com/browse/PFEBMC-1644

http/http_client.hpp
```diff
@@ -48,7 +48,6 @@
 #include <memory>
 #include <queue>
 #include <string>
-
```
gtmills, line 51: be careful with empty spare changes. Oh well, for next time. ;)

http/http_client.hpp
gtmills: Ack

http/http_client.hpp
gtmills: I think a few more comments, a more descriptive comment on why would help..

gtmills:
What is the upstream plan here?

gtmills (approved):
I am good with this but can this be pushed upstream?

PR #796, opened 2024-01-08
Work In Progress : Add flightRecorder support to dump redfish events

raviteja-b:
This commit adds flight recorder support to dump redfish events into a file which will be helpful for debugging eventing defects.
This event data will be collected as part BMC dump collection.

Tested by:
Created a BMC dump and check event data in file

### DIFF
--- a/http/http_server.hpp
+++ b/http/http_server.hpp
@@ -21,6 +21,7 @@
 #include <vector>
 #ifdef BMCWEB_ENABLE_IBM_MANAGEMENT_CONSOLE
 #include <persistent_data.hpp>
+#include <include/ibm/flight_recorder.hpp>
 #endif
 namespace crow
 {
@@ -161,6 +162,7 @@ class Server
                         << "INFO: Receivied USR1 signal to dump latest session "
                            "data for bmc dump";
                     persistent_data::getConfig().writeCurrentSessionData();
+                    bmcweb::flightrecorder::FlightRecorder::GetInstance().playRecorder();
                     this->startAsyncWaitForSignal();
                 }
 #endif
--- a/include/ibm/flight_recorder.hpp
+++ b/include/ibm/flight_recorder.hpp
@@ -1,0 +1,116 @@
+#pragma once
+
+#include <fstream>
+#include <iomanip>
+#include <iostream>
+#include <vector>
+
+#define FLIGHT_RECORDER_MAX_ENTRIES 300
+
+namespace bmcweb
+{
+namespace flightrecorder
+{
+using FlightRecorderData = nlohmann::json;
+using FlightRecorderTimeStamp = std::string;
+using FlightRecorderRecord =
+    std::tuple<FlightRecorderTimeStamp, FlightRecorderData>;
+using FlightRecorderCassette = std::vector<FlightRecorderRecord>;
+static constexpr auto flightRecorderDumpPath = "/tmp/redfish_events_flight_recorder";
+
+std::string getCurrentSystemTime()
+{
+    using namespace std::chrono;
+    const time_point<system_clock> tp = system_clock::now();
+    std::time_t tt = system_clock::to_time_t(tp);
+    auto ms = duration_cast<microseconds>(tp.time_since_epoch()) -
+              duration_cast<seconds>(tp.time_since_epoch());
+
+    std::stringstream ss;
+    ss << std::put_time(std::localtime(&tt), "%F %Z %T.")
+       << std::to_string(ms.count());
+    return ss.str();
+}
+
+/** @class FlightRecorder
+ *
+ *  The class for implementing the BMCWEB flight recorder logic. This class
+ *  handles the insertion of the data into the recorder and also provides
+ *  API's to dump the flight recorder into a file.
+ */
+
+class FlightRecorder
+{
+  private:
+    FlightRecorder() : index(0)
+    {
+        flightRecorderPolicy = FLIGHT_RECORDER_MAX_ENTRIES ? true : false;
+        if (flightRecorderPolicy)
+        {
+            tapeRecorder = FlightRecorderCassette(FLIGHT_RECORDER_MAX_ENTRIES);
+        }
+    }
+
+  protected:
+    unsigned int index;
+    FlightRecorderCassette tapeRecorder;
+    bool flightRecorderPolicy;
+
+  public:
+    FlightRecorder(const FlightRecorder&) = delete;
+    FlightRecorder(FlightRecorder&&) = delete;
+    FlightRecorder& operator=(const FlightRecorder&) = delete;
+    FlightRecorder& operator=(FlightRecorder&&) = delete;
+    ~FlightRecorder() = default;
+
+    static FlightRecorder& GetInstance()
+    {
+        static FlightRecorder flightRecorder;
+        return flightRecorder;
+    }
+ 
+    /** @brief Add records to the flightRecorder
+     *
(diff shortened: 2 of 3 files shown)


http/http_server.hpp
```diff
@@ -161,7 +162,8 @@ class Server
                         << "INFO: Receivied USR1 signal to dump latest session "
                            "data for bmc dump";
                     persistent_data::getConfig().writeCurrentSessionData();
-                    this->startAsyncWaitForSignal();
+		     bmcweb::flightrecorder::FlightRecorder::GetInstance().playRecorder();
```
sunharis, line 165: Did you run the clang?

include/ibm/flight_recorder.hpp
```diff
@@ -0,0 +1,116 @@
...
+    void saveRecord(const FlightRecorderData& eventJson)
+    {
+        // if the flight recorder policy is enabled, then only insert the
+        // messages into the flight recorder, if not this function will be just
+        // a no-op
+        if (flightRecorderPolicy)
+        {
+    	    unsigned int currentIndex = index++;
+            tapeRecorder[currentIndex] = std::make_tuple(
+                getCurrentSystemTime(), eventJson);
```
sunharis, line 87: Will this add another time entry in the file, other than the one inside the event record ?

include/ibm/flight_recorder.hpp
```diff
@@ -0,0 +1,116 @@
+#pragma once
+
+#include <fstream>
+#include <iomanip>
+#include <iostream>
+#include <vector>
+
+#define FLIGHT_RECORDER_MAX_ENTRIES 300
```
gtmills, line 8: Just do this like we do other limits 
https://github.com/ibm-openbmc/bmcweb/blob/ac4fd01279f5f8873cca21c3e16c16da5d7d2cd6/http/http_connection.hpp#L53

include/ibm/flight_recorder.hpp
```diff
@@ -0,0 +1,116 @@
...
+
+namespace bmcweb
+{
+namespace flightrecorder
+{
+using FlightRecorderData = nlohmann::json;
+using FlightRecorderTimeStamp = std::string;
+using FlightRecorderRecord =
+    std::tuple<FlightRecorderTimeStamp, FlightRecorderData>;
```
gtmills, line 17: #include <tuple>

include/ibm/flight_recorder.hpp
```diff
@@ -0,0 +1,116 @@
...
+#include <vector>
+
+#define FLIGHT_RECORDER_MAX_ENTRIES 300
+
+namespace bmcweb
+{
+namespace flightrecorder
+{
+using FlightRecorderData = nlohmann::json;
+using FlightRecorderTimeStamp = std::string;
```
gtmills, line 15: `#include <string>`

include/ibm/flight_recorder.hpp
```diff
@@ -0,0 +1,116 @@
...
+using FlightRecorderRecord =
+    std::tuple<FlightRecorderTimeStamp, FlightRecorderData>;
+using FlightRecorderCassette = std::vector<FlightRecorderRecord>;
+static constexpr auto flightRecorderDumpPath = "/tmp/redfish_events_flight_recorder";
+
+std::string getCurrentSystemTime()
+{
+    using namespace std::chrono;
```
gtmills, line 23: `#include <chrono>` 
Include what you use

gtmills: Upstream we would not allow this `using namespace std::chrono;`
I am fine with `using chrono std::chrono` if you want

gtmills:
What is the upstream plan here? 
Do we plan to deploy this to the field for 1060?
Will this be in the dump?

include/ibm/flight_recorder.hpp
```diff
@@ -0,0 +1,116 @@
...
+            index = (currentIndex == FLIGHT_RECORDER_MAX_ENTRIES - 1) ? 0
+                                                                      : index;
+        }
+    }
+
+    /** @brief play flight recorder
+     *
+     *  @return void
+     */
+
```
gtmills, line 97: extra space

include/ibm/flight_recorder.hpp
```diff
@@ -0,0 +1,116 @@
...
+        {
+    	    unsigned int currentIndex = index++;
+            tapeRecorder[currentIndex] = std::make_tuple(
+                getCurrentSystemTime(), eventJson);
+            index = (currentIndex == FLIGHT_RECORDER_MAX_ENTRIES - 1) ? 0
+                                                                      : index;
+        }
+    }
+
+    /** @brief play flight recorder
```
gtmills, line 93: If you do these, would be good to go beyond the function name. 
`    /** @brief write flight recorder to file`

include/ibm/flight_recorder.hpp
raviteja-b: yes  current time stamp is also printed and following this format
`timestamp : event data `
Thinking again its not needed, will remove timestamp

include/ibm/flight_recorder.hpp
baemyung, line 8: Shouldn't this `FLIGHT_RECORDER_MAX_ENTRIES (or flightRecorderMaxEntries)` be larger than max connection count at least to hold any potential pending connections?

https://github.com/ibm-openbmc/bmcweb/blob/ac4fd01279f5f8873cca21c3e16c16da5d7d2cd6/http/http_connection.hpp#L294

Perhaps should we introduce maxConnectionCount constant and use it to define flightRecorderMaxEntries?

```
http_connection.hpp:
constexpr int maxConnectionCount = 200;

L294:
        if (connectionCount >= maxConnectionCount)
```

And here

```
constexpr int flightRecorderMaxEntries = (maxConnectionCount + 100);   //maxcount with some extras
```

raviteja-b: maxConnectionCount is not all related for redfish evening.

include/ibm/flight_recorder.hpp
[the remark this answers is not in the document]
raviteja-b: I am still working on fixing format of data that needs to logged, event data already has timestamp so we don't this timestamp again

include/ibm/flight_recorder.hpp
abhilashraju, line 1: can we use a boost circular buffer instead? I mean would it be simpler to use an existing library feature? 
You can see the implementation [here](https://godbolt.org/z/P3zYee4ds) that covers all of the cases handled in this file.

redfish-core/include/event_service_manager.hpp
```diff
@@ -1071,6 +1071,9 @@ class EventServiceManager
                     2, ' ', true, nlohmann::json::error_handler_t::replace);
                 entry->sendEvent(strMsg);
                 eventId++; // increament the eventId
+#ifdef BMCWEB_ENABLE_IBM_MANAGEMENT_CONSOLE
+                bmcweb::flightrecorder::FlightRecorder::GetInstance().saveRecord(msgJson);
```
abhilashraju, line 1075: this should be moved to a handler that is called after a successful sendEvent and a callback that is called after unsuccessful retries finish.

include/ibm/flight_recorder.hpp
raviteja-b: sure, lets continue discussion in discord and finalise on the mechanism to use

raviteja-b:
we are still discussing on t better approach that needs to be used in this case, will re-work on this PR later

PR #801, opened 2024-01-17
redfish: remove ForceRestart from AllowableValues

geissonator:
IBM hypervisor firmware requires host reboots to always be graceful to ensure certain BIOS settings are correctly set.

There is no easy way to upstream this type of configuration without some fairly messy #ifdef's and it seems to be a pretty specific IBM requirement.

P10 PR: https://github.com/ibm-openbmc/bmcweb/pull/522

Tested:
```
curl -k -H "X-Auth-Token: $token" -X GET https://${bmc}/redfish/v1/Systems/system/ResetActionInfo {
  "@odata.id": "/redfish/v1/Systems/system/ResetActionInfo",
  "@odata.type": "#ActionInfo.v1_1_2.ActionInfo",
  "Id": "ResetActionInfo",
  "Name": "Reset Action Info",
  "Parameters": [
    {
      "AllowableValues": [
        "On",
        "ForceOff",
        "ForceOn",
        "GracefulRestart",
        "GracefulShutdown",
        "PowerCycle",
        "Nmi"
      ],
      "DataType": "String",
      "Name": "ResetType",
      "Required": true
    }
  ]
}
```

### DIFF
--- a/redfish-core/lib/systems.hpp
+++ b/redfish-core/lib/systems.hpp
@@ -3074,11 +3074,6 @@ inline void handleComputerSystemResetActionPost(
         command = "xyz.openbmc_project.State.Chassis.Transition.Off";
         hostCommand = false;
     }
-    else if (resetType == "ForceRestart")
-    {
-        command = "xyz.openbmc_project.State.Host.Transition.ForceWarmReboot";
-        hostCommand = true;
-    }
     else if (resetType == "GracefulShutdown")
     {
         command = "xyz.openbmc_project.State.Host.Transition.Off";
@@ -3560,7 +3555,6 @@ inline void handleSystemCollectionResetActionGet(
     allowableValues.emplace_back("On");
     allowableValues.emplace_back("ForceOff");
     allowableValues.emplace_back("ForceOn");
-    allowableValues.emplace_back("ForceRestart");
     allowableValues.emplace_back("GracefulRestart");
     allowableValues.emplace_back("GracefulShutdown");
     allowableValues.emplace_back("PowerCycle");


gtmills:
> There is no easy way to upstream this type of configuration without some fairly messy #ifdef's and it seems to be a pretty specific IBM requirement.

Anyway bmcweb could look at D-Bus to decide if it supports this? 
🤷‍♂️ easy enough to carry if you don't think it is worth supporting something in state manager

gtmills (approved):

geissonator:
> Anyway bmcweb could look at D-Bus to decide if it supports this? enough to carry if you don't think it is worth supporting something in state manager

The interfaces are defined with PDI, so there's no easy way to optionally remove one there. We could add a config flag and compile options to PSM to reject the command but that doesn't help with the Redfish API that tells the user the supported one's. So yeah, I think we just gotta carry this one.

gtmills:
> Redfish API that tells the user the supported one

We could propose a new PDI property for AllowableValues

geissonator:
Yeah, I def struggle with a small downstream patch like this vs. the PDI, PSM, bmcweb path we'll have to go through to upstream for something I haven't seen anyone else needing, but ... never hurts to try. https://gerrit.openbmc.org/c/openbmc/phosphor-dbus-interfaces/+/68933 is the first part of this.

And it's always nice to take a downstream patch off the list.

gtmills:
Whenever you are ready, merge this

geissonator:
All function in upstream via https://gerrit.openbmc.org/q/topic:%22AllowedHostTransitions%22

PR #807, opened 2024-01-31
WIP: To support Bonnell change Default Model to "BMC on System Backplane"

gtmills:
WIP!

Discussed 1/31.
If we have to do something here for 586924.. We could do this.

https://github.com/ibm-openbmc/bmcweb/blob/a4a534e865211515608cdf3b4fa777a31c176002/redfish-core/lib/managers.hpp#L2235 gets the Model when we have an Inventory.Item.Decorator.Asset (rainier, everest) 
https://github.com/openbmc/phosphor-dbus-interfaces/blob/1abc0c57692b679753955b7273c1393369ef4987/yaml/xyz/openbmc_project/Inventory/Decorator/Asset.interface.yaml#L22

Since Bonnell doesn't have an Inventory.Item.Decorator.Asset interface, it is showing "OpenBmc"


gtmills:
We decided not to do anything here so closing

PR #820, opened 2024-02-21
Audit Log: Rebase service audit log

jeaaustx:
Rebase of all bmcweb commits to support service audit log.

This introduces a new bmcweb meson option, audit-events, which is defaulted to disabled. Recipe changes in GHE openbmc/openbmc enable audit logging.

When audit-events is enabled all bmcweb PATCH, POST, PUT, and DELETE events are logged using the Linux kernel auditd subsystem. Additionally, login events coming through bmcweb are also recorded.

The body of the events is recorded as well except in the following cases:
  - /redfish/v1/AccountService/Accounts PATCH/POST events - body is not recorded as it may contain clear text password
  - /ibm/v1 PATCH/POST/PUT events - body is not recorded as it contains HMC config file binary data

Events are recorded in /var/log/audit/. User type dreport will gather these log files.

Tested:
 - Enabled auditing then initiated a variety of Redfish events using curl.
 - Confirmed the events were recorded and confirmed the data recorded was accurate.
 - Confirmed password was not included in audit data recorded.

Commits included:
```
 -  4bc3e20f Audit Log: Hide details of building audit message into auditEvent() (#725)
 -  7483bb00 Audit Log: Factor out function to determine which events to audit (#718)
 -  2a4acf86 Audit Log: Add utility function to enable/disable audit connection
 -  bda2e617 Audit Log: Handle breaking of unexpected connection with auditd
 -  fab81331 Audit Log: Add unit test for auditOpen() and auditClose()
 -  3e352fcd Audit Log: Open and save fd for auditing future events
 -  b0901dd3 Audit Log: Correct logging of success login case. (#690)
 -  3976090f Merge pull request #680 from jeaaustx/1050-enableAuditLog
 -  f5745fc6 Audit Log: Exclude body data for all IBM management console events
 -  db0ccb2e Audit Log: Add PUT events to audit log for HMC config file management
 -  9bd97269 Audit log enhancements and fixs (#500)
 -  5ced90bd Expand Success Case For Audit Logging (#447)
 -  6dc7529e Add Audit Event Points to All POSTs
 -  9241b341 Add Audit Event Through bmcweb
```

### DIFF
--- a/include/audit_events.hpp
+++ b/include/audit_events.hpp
@@ -1,0 +1,296 @@
+#pragma once
+
+#include "http_request.hpp"
+#include "logging.hpp"
+
+#include <libaudit.h>
+
+#include <boost/asio/ip/host_name.hpp>
+
+#include <cstring>
+#include <string>
+
+namespace audit
+{
+
+// NOLINTBEGIN(cppcoreguidelines-avoid-non-const-global-variables)
+static bool tryOpen = true;
+static int auditfd = -1;
+// NOLINTEND(cppcoreguidelines-avoid-non-const-global-variables)
+
+/**
+ * @brief Closes connection for recording audit events
+ */
+inline void auditClose()
+{
+    if (auditfd >= 0)
+    {
+        audit_close(auditfd);
+        auditfd = -1;
+        BMCWEB_LOG_DEBUG("Audit log closed.");
+    }
+}
+
+/**
+ * @brief Opens connection for recording audit events
+ *
+ * Reuses prior connection if available.
+ *
+ * @return If connection was successful or not
+ */
+inline bool auditOpen()
+{
+    if (auditfd < 0)
+    {
+        /* Blocking opening of audit connection */
+        if (!tryOpen)
+        {
+            BMCWEB_LOG_DEBUG("Audit connection disabled");
+            return false;
+        }
+
+        auditfd = audit_open();
+
+        if (auditfd < 0)
+        {
+            BMCWEB_LOG_ERROR("Error opening audit socket : {}", errno);
+            return false;
+        }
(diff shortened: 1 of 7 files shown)


include/audit_events.hpp
```diff
@@ -0,0 +1,295 @@
...
+inline void auditEvent(const crow::Request& req, const std::string& userName,
+                       bool success)
+{
+    if (!auditOpen())
+    {
+        return;
+    }
+
+    std::string opPath = "op=" + std::string(req.methodString()) + ":" +
+                         std::string(req.target()) + " ";
```
baemyung, line 200: Perhaps, it may be easier to read if we do 
```
std::string opPath = std::format("op={}:{} ",  req.methodString(),  req.target());
```

baemyung (approved):

include/audit_events.hpp
jeaaustx: Thank you, I agree that is lots easier to read. (In fact it is also one of the comments from upstream as well.) I will make this change.

baemyung (approved):

include/audit_events.hpp
```diff
@@ -0,0 +1,296 @@
...
+ * - Accounts requests data may contain passwords
+ * - IBM Management console events data is not useful. It can be binary data or
+ *   contents of file.
+ * - User login and session data may contain passwords
+ *
+ * @return True if request's data should not be logged
+ */
+inline bool checkSkipDetail(const crow::Request& req)
+{
+    return req.target().starts_with("/redfish/v1/AccountService/Accounts") ||
```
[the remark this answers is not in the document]
jeaaustx: Yes, start_with matches anything with the prefix specified. That is the intention here.

PR #824, opened 2024-02-22
Unauthorized ACF certificate upload (#243)

Alpana07:
This commit implements the change where, when OpPanel function 74 is active, an unauthorized agent is allowed to PATCH in the ACF.

The following command without creating a session has been to test the functionality.
```
curl  -k -X PATCH -d '{"Oem":{"IBM":{"ACF":{"ACFFile":"'${service_acf_base64}'"}}}}'
https://${BMC}/redfish/v1/AccountService/Accounts/service
```

In case function 74 is active this should allow to patch ACF crtificate otherwise gives following insufficient privilege message

```
{
  "error": {
    "@Message.ExtendedInfo": [
      {
        "MessageArgs": [],
        "MessageId": "Base.1.8.1.InsufficientPrivilege",
      }
    ],
    "code": "Base.1.8.1.InsufficientPrivilege",
  }
}
```

### DIFF
--- a/redfish-core/lib/account_service.hpp
+++ b/redfish-core/lib/account_service.hpp
@@ -1541,6 +1541,128 @@ inline void updateUserProperties(
     });
 }
 
+inline void uploadACF(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
+                      const std::vector<uint8_t>& decodedAcf)
+{
+    crow::connections::systemBus->async_method_call(
+        [asyncResp](const boost::system::error_code ec,
+                    sdbusplus::message::message& m,
+                    const std::tuple<std::vector<uint8_t>, bool, std::string>&
+                        messageFDbus) {
+        if (ec)
+        {
+            BMCWEB_LOG_ERROR("DBUS response error: {}", ec.value());
+            if (m.get_error()->name ==
+                std::string_view(
+                    "xyz.openbmc_project.Certs.Error.InvalidCertificate"))
+            {
+                redfish::messages::invalidUpload(
+                    asyncResp->res,
+                    "/redfish/v1/AccountService/Accounts/service",
+                    "Invalid Certificate");
+            }
+            else
+            {
+                messages::internalError(asyncResp->res);
+            }
+            return;
+        }
+        getAcfProperties(asyncResp, messageFDbus);
+    },
+        "xyz.openbmc_project.Certs.ACF."
+        "Manager",
+        "/xyz/openbmc_project/certs/ACF", "xyz.openbmc_project.Certs.ACF",
+        "InstallACF", decodedAcf);
+}
+
+inline void triggerUnauthenticatedACFUpload(
+    const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
+    std::optional<nlohmann::json> oem)
+{
+    std::optional<nlohmann::json> ibm;
+    if (!redfish::json_util::readJson(*oem, asyncResp->res, "IBM", ibm))
+    {
+        BMCWEB_LOG_ERROR("Illegal Property ");
+        messages::propertyMissing(asyncResp->res, "IBM");
+        return;
+    }
+
+    std::optional<nlohmann::json> acf;
+    if (ibm)
+    {
+        if (!redfish::json_util::readJson(*ibm, asyncResp->res, "ACF", acf))
+        {
+            BMCWEB_LOG_ERROR("Illegal Property ");
+            messages::propertyMissing(asyncResp->res, "ACF");
+            return;
+        }
+    }
+
+    if (acf)
+    {
+        std::vector<uint8_t> decodedAcf;
+        std::optional<std::string> acfFile{};
+        if (!redfish::json_util::readJson(*acf, asyncResp->res, "ACFFile",
+                                          acfFile))
+        {
+            BMCWEB_LOG_ERROR("Illegal Property ");
+            messages::propertyMissing(asyncResp->res, "ACFFile");
+            return;
+        }
+
+        std::string sDecodedAcf;
+        if (!crow::utility::base64Decode(*acfFile, sDecodedAcf))
+        {
+            BMCWEB_LOG_ERROR("base64 decode failure ");
+            messages::internalError(asyncResp->res);
+            return;
+        }
+        try
+        {
+            std::copy(sDecodedAcf.begin(), sDecodedAcf.end(),
+                      std::back_inserter(decodedAcf));
+        }
+        catch (const std::exception& e)
+        {
+            BMCWEB_LOG_ERROR("Exception thrown: {}", e.what());
+            messages::internalError(asyncResp->res);
+            return;
+        }
+
+        crow::connections::systemBus->async_method_call(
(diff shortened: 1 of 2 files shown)


redfish-core/lib/account_service.hpp
```diff
@@ -1541,6 +1541,138 @@ inline void updateUserProperties(
...
+                      const std::vector<uint8_t>& decodedAcf)
+{
+    crow::connections::systemBus->async_method_call(
+        [asyncResp](const boost::system::error_code ec,
+                    sdbusplus::message::message& m,
+                    const std::tuple<std::vector<uint8_t>, bool, std::string>&
+                        messageFDbus) {
+        if (ec)
+        {
+            BMCWEB_LOG_ERROR("DBUS response error: {}", ec);
```
baemyung, line 1554: In bmcweb, it is suggested to use `ec.value()` in log.

redfish-core/lib/account_service.hpp
```diff
@@ -1541,6 +1541,138 @@ inline void updateUserProperties(
...
+            if (strcmp(m.get_error()->name, "xyz.openbmc_project.Certs.Error."
+                                            "InvalidCertificate") == 0)
```
baemyung, line 1556: The new CI would suggest to combine the parts of strings into one like `xyz.openbmc_project.Certs.Error.InvalidCertificate`.

I also think the better/safer comparison method seems like
```
            if ( m.get_error()->name == std::string_view("xyz.openbmc_project.Certs.Error.InvalidCertificate"))
```

redfish-core/lib/account_service.hpp
```diff
@@ -1541,6 +1541,138 @@ inline void updateUserProperties(
...
+            {
+                redfish::messages::invalidUpload(
+                    asyncResp->res,
+                    "/redfish/v1/AccountService/Accounts/service",
+                    "Invalid Certificate");
+            }
+            else
+            {
+                messages::internalError(asyncResp->res);
```
baemyung, line 1565: Please add `BMCWEB_LOG_ERROR() with ec.value()` prior to `internalError()`

redfish-core/lib/account_service.hpp
```diff
@@ -1541,6 +1541,138 @@ inline void updateUserProperties(
...
+{
+    std::optional<nlohmann::json> ibm;
+    if (!redfish::json_util::readJson(*oem, asyncResp->res, "IBM", ibm))
+    {
+        BMCWEB_LOG_ERROR ("Illegal Property ");
+        messages::propertyMissing(asyncResp->res, "IBM");
+        return;
+    }
+
+    if (ibm)
```
baemyung, line 1589: I think this deeper nested code block can be simplified by doing

```
   std::optional<nlohmann::json> acf;
   std::vector<uint8_t> decodedAcf;
   std::optional<std::string> acfFile;

    if (ibm)
    {
        if (!redfish::json_util::readJson(*ibm, asyncResp->res, "ACF", acf))
        {
            BMCWEB_LOG_ERROR ("Illegal Property ");
            messages::propertyMissing(asyncResp->res, "ACF");
            return;
        }
   }

  if(acf)
  {
     ... acf & acfFile reading...
   }

   std::string sDecodedAcf;
   if (!crow::utility::base64Decode(*acfFile, sDecodedAcf))
   {
                    BMCWEB_LOG_ERROR ("base64 decode failure ");
                    return;
   }
  
```

SunnySrivastava1984:
> It looks like this Change-Id is not valid.
> 
> ```
> Change-Id: I6f85b3d6cbc937c6c11e91704ffdc9e5795429b5 
> ```
> 
> Could you upstream this and add its Change-Id in this commit msg?

This is OEM specific changes and can't be upstreamed.

baemyung:
> > It looks like this Change-Id is not valid.
> > ```
> > Change-Id: I6f85b3d6cbc937c6c11e91704ffdc9e5795429b5 
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Could you upstream this and add its Change-Id in this commit msg?
> 
> This is OEM specific changes and can't be upstreamed.

Then, would you remove Change-Id from this.

Alpana07:
> > > It looks like this Change-Id is not valid.
> > > ```
> > > Change-Id: I6f85b3d6cbc937c6c11e91704ffdc9e5795429b5 
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Could you upstream this and add its Change-Id in this commit msg?
> > 
> > 
> > This is OEM specific changes and can't be upstreamed.
> 
> Then, would you remove Change-Id from this.

okay

redfish-core/lib/account_service.hpp
Alpana07: We have 
BMCWEB_LOG_ERROR("DBUS response error: {}", ec);  <---1571 , is not ok?

redfish-core/lib/account_service.hpp
Alpana07: Ack.
For variable declaration I think we should declare it need basis, so it is placed just before it's use.

redfish-core/lib/account_service.hpp
Alpana07: okay

baemyung: In the past, there have been mixed uses of ec in logging.
Now, it has been suggested to use `ec.value()` in logging -- which will reduce the amount of logging output.

redfish-core/lib/account_service.hpp
baemyung: You're right. It is already there before internalError.

redfish-core/lib/account_service.hpp
baemyung: I think it is about balancing of the readability...

I meant it can be short-cut by returning it s

```
  if(!acf)
  {
      return;
 }

  std::vector<uint8_t> decodedAcf;
 std::optional<std::string> acfFile{};
   ... acf & acfFile reading...
  if(failed) {
    return;
 }
...

````

**Review by baemyung — APPROVED** [review]
Besides a minor comment to use `ec.value()` in logging, the code seems good to me.

File: redfish-core/lib/account_service.hpp [hunk 5]
```diff
@@ -1541,6 +1541,130 @@ inline void updateUserProperties(
...
+                return;
+            }
+
+            if (*isACFWindowActive)
+            {
+                uploadACF(asyncResp, decodedAcf);
+                return;
+            }
+
+            BMCWEB_LOG_ERROR("ACF window not set to "
```
baemyung, line 1657: Should this an WARNING, rather than ERROR, as it is the unexpected condition to proceed, rather than something wrong in service?

https://github.com/openbmc/bmcweb/blame/e715d14b258a764f0e41559fe4a5e9f7d2efe314/DEVELOPING.md#L223-L229C58

baemyung (approved):

gtmills:
If Myung is good, I am good

PR #825, opened 2024-02-22
Rebase fabric adapter and  chassis 

Alpana07:
Link Fabric adapter to PCIeDevice schema: 
This commit implement changes to populate link to PCIeDevice schema in
case the given Fabric adapter also implements PCIeDevice interface.

Remove link from Chassis to PCIeDevice:
The commit removes current support that assumes 1:1
system:Chassis for Chassis/PCIeDevices.

### DIFF
--- a/include/dbus_utility.hpp
+++ b/include/dbus_utility.hpp
@@ -83,6 +83,9 @@ using DBusInterfacesMap =
 using ManagedObjectType =
     std::vector<std::pair<sdbusplus::message::object_path, DBusInterfacesMap>>;
 
+// List of interfaces
+using InterfaceList = std::vector<std::string>;
+
 // Map of service name to list of interfaces
 using MapperServiceMap =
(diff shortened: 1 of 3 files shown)


include/dbus_utility.hpp
```diff
@@ -83,6 +83,9 @@ using DBusInterfacesMap =
 using ManagedObjectType =
     std::vector<std::pair<sdbusplus::message::object_path, DBusInterfacesMap>>;
 
+// List of interfaces
+using InterfaceList = std::vector<std::string>;
+
```
baemyung, line 88: Since the second of `MapperServiceMap` is a interfaceList, we can also update `MapperServiceMap`.

```
// List of interfaces
using InterfaceList = std::vector<std::string>;

// Map of service name to list of interfaces
using MapperServiceMap = std::vector<std::pair<std::string, InterfaceList>;
```

baemyung (commented):
Cherry-pick caused merge-conflict.

PR #835, opened 2024-02-28
Redfish : Look for OperationalStatus for the memory Health

lxwinspur:

Upstream link: https://gerrit.openbmc.org/c/openbmc/bmcweb/+/69859
Commit Id: Ifb7e4b5e6bfadd417122fda66ef3f6590ccc43b1

Tested: Validator passes.
```
{
  "Status": {
    "Health": "Critical",
    "State": "Enabled"
  }
}
```

### DIFF
--- a/redfish-core/lib/memory.hpp
+++ b/redfish-core/lib/memory.hpp
@@ -424,6 +424,7 @@ inline void
     const std::string* sparePartNumber = nullptr;
     const std::string* model = nullptr;
     const std::string* locationCode = nullptr;
+    const bool* functional = nullptr;
 
     const bool success = sdbusplus::unpackPropertiesNoThrow(
         dbus_utils::UnpackErrorPrinter(), properties, "MemoryDataWidth",
@@ -436,7 +437,7 @@ inline void
         memoryConfiguredSpeedInMhz, "MemoryType", memoryType, "Channel",
         channel, "MemoryController", memoryController, "Slot", slot, "Socket",
         socket, "SparePartNumber", sparePartNumber, "Model", model,
-        "LocationCode", locationCode);
+        "LocationCode", locationCode, "Functional", functional);
 
     if (!success)
     {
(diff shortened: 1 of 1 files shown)


gtmills:
Jenkins run tests please

gtmills:
@lxwinspur Where is upstream here?

lxwinspur:
> @lxwinspur Where is upstream here?

There is an updated design document here(https://gerrit.openbmc.org/c/openbmc/docs/+/43485), but it has been abandoned. The issue created by IBM is here:
https://github.com/ibm-openbmc/dev/issues/3193

I'm not sure if this patch will be accepted by Ed after upstream.

baemyung:
Even if the upstream commit is abandoned, I think we should add its Change-Id into this commit message for the future reference.
Also if ready, should it be restored as active?

redfish-core/lib/memory.hpp
```diff
@@ -599,6 +599,14 @@ inline void
                                 ["ServiceLabel"] = *locationCode;
     }
 
+    if (functional != nullptr)
```
gtmills, line 602: Why can't this go upstream?

redfish-core/lib/memory.hpp
```diff
@@ -607,13 +615,6 @@ inline void getDimmDataByService(std::shared_ptr<bmcweb::AsyncResp> asyncResp,
                                  const std::string& service,
                                  const std::string& objPath)
 {
-    if constexpr (bmcwebEnableHealthPopulate)
```
gtmills, line 610: We don't have this enabled so this is a no-op

gtmills: Maybe we will enable this for https://github.com/ibm-openbmc/bmcweb/pull/836/files ?

lxwinspur:
> @lxwinspur Where is upstream here?

https://gerrit.openbmc.org/c/openbmc/bmcweb/+/69859

redfish-core/lib/memory.hpp
gtmills: https://gerrit.openbmc.org/c/openbmc/bmcweb/+/69859 is upstream

gtmills (approved):

baemyung (approved):

PR #836, opened 2024-02-28
Redfish: Implement HealthRollup in chassis schema

lxwinspur:
The health status of the current CPUs and powersupplies will not rollup to chassis. This submission implements rollup of the health status of CPUs and powersupplies to chassis.

Upstream Link: https://gerrit.openbmc.org/c/openbmc/bmcweb/+/69861
Commit ID: Iafacbfacff081d09c38405a83b7fabb7074f61a1

Tested:
We need add the xyz.openbmc_project.Inventory.Item.Global interface to the chassis.
I create a critical association of powersupply1 using the same method as phosphor-logging:
[1] https://github.com/openbmc/phosphor-logging/blob/76198a2ea2e3deb14078e0d3dd3932b657ddd9b8/extensions/openpower-pels/service_indicators.cpp#L83

```
curl -k -H "X-Auth-Token: $token" -X GET https://${bmc}/redfish/v1/Chassis/chassis
"Status": {
    "Health": "OK",
    "HealthRollup": "Critical",
    "State": "Enabled"
  },
```

### DIFF
--- a/redfish-core/lib/chassis.hpp
+++ b/redfish-core/lib/chassis.hpp
@@ -485,6 +485,35 @@ inline void handleChassisGetSubTree(
                     return; // no sensors = no failures
                 }
                 health->inventory = resp;
+
+                constexpr std::array<std::string_view, 13> interfaces = {
+                    "xyz.openbmc_project.Inventory.Item.Dimm",
+                    "xyz.openbmc_project.Inventory.Item.Cpu",
+                    "xyz.openbmc_project.Inventory.Item.PowerSupply",
+                    "xyz.openbmc_project.Inventory.Item.Fan",
+                    "xyz.openbmc_project.Inventory.Item.PCIeSlot",
+                    "xyz.openbmc_project.Inventory.Item.Vrm",
+                    "xyz.openbmc_project.Inventory.Item.Tpm",
+                    "xyz.openbmc_project.Inventory.Item.Panel",
+                    "xyz.openbmc_project.Inventory.Item.Battery",
+                    "xyz.openbmc_project.Inventory.Item.DiskBackplane",
+                    "xyz.openbmc_project.Inventory.Item.Board",
+                    "xyz.openbmc_project.Inventory.Item.Board.Motherboard",
+                    "xyz.openbmc_project.Inventory.Item.Connector"};
+                dbus::utility::getSubTreePaths(
+                    "/", 0, interfaces,
+                    [health](const boost::system::error_code& ec3,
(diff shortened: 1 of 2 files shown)


gtmills:
Jenkins run tests please

baemyung:
Should this be upstreamed?
Please add Change-Id and gerrit url.

lxwinspur:
> Should this be upstreamed? Please add Change-Id and gerrit url.

Updated

gtmills:
Do we need to enable this? I thought health.hpp was turned off by default ?

lxwinspur:
> Do we need to enable this? I thought health.hpp was turned off by default ?

Right, But we can't see the 1110 branch of ibm-openbmc, so I can't enable redfish-health-populate in the bmcweb_%.bbappend file

gtmills:
@rfrandse Is working on pushing an 1110 openbmc/openbmc branch out to ibm-openbmc/

redfish-core/lib/chassis.hpp
```diff
@@ -485,6 +485,35 @@ inline void handleChassisGetSubTree(
...
+                    [health](const boost::system::error_code& ec3,
+                             const dbus::utility::MapperGetSubTreePathsResponse&
+                                 resp2) {
+                    if (ec3)
+                    {
+                        // no inventory
+                        return;
+                    }
+
+                    health->inventory.insert(health->inventory.end(),
```
baemyung, line 514: I made the comment on upstream - https://gerrit.openbmc.org/c/openbmc/bmcweb/+/69861/comment/5831805e_1c269f7c/

----

This logic seems adding all inventory subtree paths of interfaces (even if they may not be associated with path), if there is an association of `path + "/all_sensors"?
Is it an intention?

Or, is this logic trying to find all associated subtree paths that are having all_sensors?
If so, wouldn't these chains of getAssociationEndPoints() & getSubTreePaths() be the same as getAssociatedSubtreePaths() like

```
 dbus::utility::getAssociatedSubTreePaths(
        path + "/all_sensors", "/", 0,
        interfaces,
        [health](
            const dbus::utility::MapperGetSubTreePathsResponse& resp) {

        if (ec)
        {
             return; // no sensors = no failures
        }
        health->inventory = resp;
```

gtmills:
We had this in P10, so let's just go with this for Pnext

gtmills (approved):

rfrandse:
Merged with [1110: Implement HealthRollup in chassis schema](https://github.com/ibm-openbmc/bmcweb/pull/862#top)
#862

PR #583, opened 2023-02-23
Link Fabric adapter to PCIeDevice schema

SunnySrivastava1984:
This commit implement changes to populate link to PCIeDevice schema in case the given Fabric adapter also implements PCIeDevice interface.

Validator has been executed and no new error was found.

Sample Output:
{
  "@odata.id": "/redfish/v1/Systems/system/FabricAdapters/pcie_card8",
  "@odata.type": "#FabricAdapter.v1_4_0.FabricAdapter",
  "Id": "pcie_card8",
  "Links": {
    "PCIeDevices": [
      {
        "@odata.id": "/redfish/v1/Systems/system/PCIeDevices/pcie_card8"
      }
    ],
    "PCIeDevices@odata.count": 1
  },
  "Location": {
    "PartLocation": {
      "ServiceLabel": "U78DB.ND0.WZS000G-P0-C8"
    }
  },
  "Name": "Fabric Adapter",
  "Status": {
    "Health": "OK",
    "State": "Absent"
  }
}

PR #840, opened 2024-02-29
Link Fabric adapter to PCIeDevice schema (#583)

Alpana07:
This commit implement changes to populate link to PCIeDevice schema in
case the given Fabric adapter also implements PCIeDevice interface.

Test:
Validator has been executed and no new error was found.

Sample Output:
curl -k -H "X-Auth-Token: $bmc_token" -X GET https://${bmc}/redfish/v1/Systems/system/FabricAdapters/pcie_card11                        
{
  "@odata.id": "/redfish/v1/Systems/system/FabricAdapters/pcie_card11",
  "@odata.type": "#FabricAdapter.v1_4_0.FabricAdapter",
  "Id": "pcie_card11",
  "Links": {
    "PCIeDevices": [
      {
        "@odata.id": "/redfish/v1/Systems/system/PCIeDevices/pcie_card11"
      }
    ],
    "PCIeDevices@odata.count": 1
  },
…

### DIFF
--- a/include/dbus_utility.hpp
+++ b/include/dbus_utility.hpp
@@ -83,9 +83,11 @@ using DBusInterfacesMap =
 using ManagedObjectType =
     std::vector<std::pair<sdbusplus::message::object_path, DBusInterfacesMap>>;
 
+// List of interfaces
+using InterfaceList = std::vector<std::string>;
+
 // Map of service name to list of interfaces
-using MapperServiceMap =
-    std::vector<std::pair<std::string, std::vector<std::string>>>;
+using MapperServiceMap = std::vector<std::pair<std::string, InterfaceList>>;
 
 // Map of object paths to MapperServiceMaps
 using MapperGetSubTreeResponse =
--- a/redfish-core/lib/fabric_adapters.hpp
+++ b/redfish-core/lib/fabric_adapters.hpp
@@ -184,11 +184,37 @@ inline void afterDoCheckFabricAdapterChassis(
         messages::internalError(asyncResp->res);
         return;
     }
-
     sdbusplus::message::object_path path(chassisPaths[0]);
     std::string chassisName = path.filename();
 
     callback(chassisName, pcieSlotPaths);
+}
+
+inline void
+    linkAsPCIeDevice(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
+                     const std::string& fabricAdapterPath)
+{
+    const std::string pcieDeviceName =
+        sdbusplus::message::object_path(fabricAdapterPath).filename();
+
+    if (pcieDeviceName.empty())
+    {
+        BMCWEB_LOG_ERROR("Failed to find / in pcie device path");
+        messages::internalError(asyncResp->res);
+        return;
+    }
+
+    nlohmann::json::object_t device;
+    nlohmann::json::array_t deviceArray;
+
+    device["@odata.id"] = boost::urls::format(
+        "/redfish/v1/Systems/system/PCIeDevices/{}", pcieDeviceName);
+
+    deviceArray.emplace_back(device);
+
+    asyncResp->res.jsonValue["Links"]["PCIeDevices@odata.count"] =
+        deviceArray.size();
+    asyncResp->res.jsonValue["Links"]["PCIeDevices"] = std::move(deviceArray);
 }
 
 inline void doCheckFabricAdapterChassis(
@@ -261,7 +287,8 @@ inline void doAdapterGet(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
                          const std::string& systemName,
                          const std::string& adapterId,
                          const std::string& fabricAdapterPath,
-                         const std::string& serviceName)
+                         const std::string& serviceName,
+                         const dbus::utility::InterfaceList& interfaces)
 {
     asyncResp->res.addHeader(
         boost::beast::http::field::link,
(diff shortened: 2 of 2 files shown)


redfish-core/lib/fabric_adapters.hpp
```diff
@@ -173,11 +173,36 @@ inline void
...
+        messages::internalError(asyncResp->res);
+        return;
+    }
+
+    nlohmann::json& deviceArray = asyncResp->res.jsonValue["Links"]["PCIeDevices"];
+    deviceArray = nlohmann::json::array();
+
+    deviceArray.push_back(
+        {{"@odata.id",
+          "/redfish/v1/Systems/system/PCIeDevices/" + pcieDeviceName}});
```
baemyung, line 194: See https://github.com/openbmc/bmcweb/blob/master/COMMON_ERRORS.md#13-complete-replacement-of-the-response-object.

Or, similar pattern is https://github.com/openbmc/bmcweb/blob/41b25aa3a541a0f92dee53b89064f655ac5e11cb/redfish-core/lib/systems.hpp#L3406

It is suggested to right like

```
nlohmann::json::object_t item;
item["@odata.id"] = boost::urls::format("/redfish/v1/Systems/system/PCIeDevices/{}", pcieDeviceName);
deviceArray.emplace_back( std::move(item) );
```

redfish-core/lib/fabric_adapters.hpp
```diff
@@ -173,11 +173,36 @@ inline void
...
+
+    asyncResp->res.jsonValue["Links"]["PCIeDevices@odata.count"] =
+        deviceArray.size();
```
baemyung, line 197: I think L189-L194 seems better to restructure so that 
- construct the content in a tempArray
- Move it to asyncResp->res.

```
nlohmann::json::array_t deviceArray;
... <<< construct deviceArray.> .....
asyncResp->res.jsonValue["Links"]["PCIeDevices"] = std::move(deviceArray);
```

redfish-core/lib/fabric_adapters.hpp
Alpana07: Ack

redfish-core/lib/fabric_adapters.hpp
Alpana07: Ack

Alpana07:
> Changed-Id in the commit message is not valid
> 
> ```
> Change-Id: Id5cf30cb9cff76c92171bb9c667ceeb1d60e9667
> ```
> 
> I would suggest to push this to upstream and use the change-Id here.

removed change-id

Alpana07:
Jenkins run tests please

Alpana07:
Jenkins run tests please

redfish-core/lib/fabric_adapters.hpp
```diff
@@ -173,11 +173,37 @@ inline void
...
+        BMCWEB_LOG_ERROR("Failed to find / in pcie device path");
+        messages::internalError(asyncResp->res);
+        return;
+    }
+
+    nlohmann::json::array_t deviceArray;
+    nlohmann::json& device = deviceArray.emplace_back();
+
+    device["@odata.id"] = boost::urls::format(
+        "/redfish/v1/Systems/system/PCIeDevices/{}", pcieDeviceName);
```
baemyung, line 194: I think the current code is first add an empty item and then update it.
Instead, I think the better approach may be:

```
  nlohmann::json::object_t device;

    device["@odata.id"] = boost::urls::format(
        "/redfish/v1/Systems/system/PCIeDevices/{}", pcieDeviceName);

   deviceArray.emplace_back( device );
```

baemyung (approved):
I'm approving this commit with my latest comment of L190-L194 reordering.

include/dbus_utility.hpp
```diff
@@ -83,6 +83,9 @@ using DBusInterfacesMap =
 using ManagedObjectType =
     std::vector<std::pair<sdbusplus::message::object_path, DBusInterfacesMap>>;
 
+// List of interfaces
+using InterfaceList = std::vector<std::string>;
+
 // Map of service name to list of interfaces
 using MapperServiceMap =
```
baemyung, line 90: Since `InterfaceList` is introduced, `MapperServiceMap` is better to be updated like
```
using MapperServiceMap =
    std::vector<std::pair<std::string, InterfaceList>>;
```

redfish-core/lib/fabric_adapters.hpp
```diff
@@ -196,6 +224,15 @@ inline void doAdapterGet(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
     getFabricAdapterAsset(asyncResp, serviceName, fabricAdapterPath);
     getFabricAdapterState(asyncResp, serviceName, fabricAdapterPath);
     getFabricAdapterHealth(asyncResp, serviceName, fabricAdapterPath);
+
+    // if the adapter also implements this interface, link the adapter schema to
+    // PCIeDevice schema for this adapter.
+    if (std::find(interfaces.begin(), interfaces.end(),
```
baemyung, line 230: Similar to `std::sort()` to `std::ranges::sort()`, this can be written using `std::ranges::find()`.
```
    if (std::ranges::find(interfaces, "xyz.openbmc_project.Inventory.Item.PCIeDevice") !=
```

baemyung:
> > Changed-Id in the commit message is not valid
> > ```
> > Change-Id: Id5cf30cb9cff76c92171bb9c667ceeb1d60e9667
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I would suggest to push this to upstream and use the change-Id here.
> 
> removed change-id

This is still in this PR description above.

BTW, why is it not going to do upstream?
Should we do it?

Alpana07:
> > > Changed-Id in the commit message is not valid
> > > ```
> > > Change-Id: Id5cf30cb9cff76c92171bb9c667ceeb1d60e9667
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > I would suggest to push this to upstream and use the change-Id here.
> > 
> > 
> > removed change-id
> 
> This is still in this PR description above.
> 
> BTW, why is it not going to do upstream? Should we do it?

Yes, will do, as base commit is different in upstream, cannot cherry-pick there, will make the change manually .
And I was told to merge the PR 1st, then will push upstream changes.

baemyung (approved):

gtmills:
If Myung is good, I am good

PR #842, opened 2024-03-06
Audit Log: Handle nullptr for message args

jeaaustx:
fillAuditLogEntryJson() expects a string for each MessageArgs entry. When one of the entries was a nullptr bmcweb would core dump.

Handle a nullptr by replacing it with an empty string.

Note: Root cause here is that the entry being returned by
      phosphor-auditlog is not of the expected type. Added a warning
      message for this case for future debugging.

Tested:
  Recreated core dump. Then using the same audit log source files
  confirmed the nullptr was correctly handled and no core dump happened.

Fixes: 600216

Fix for phosphor-auditlog also coded: https://github.com/ibm-openbmc/phosphor-logging/pull/51

### DIFF
--- a/redfish-core/lib/log_services.hpp
+++ b/redfish-core/lib/log_services.hpp
@@ -6728,9 +6728,22 @@ static LogParseError
     // Fill the MessageArgs into the Message
     if (messageArgs.size() > 0)
     {
-        int i = 0;
+        if (messageArgs[0] != "USYS_CONFIG")
+        {
+            BMCWEB_LOG_WARNING << "Unexpected audit log entry type: "
+                               << messageArgs[0];
+        }
+
+        uint i = 0;
         for (auto messageArg : messageArgs)
         {
+            if (messageArg == nullptr)
+            {
+                BMCWEB_LOG_DEBUG << "Handle null messageArg";
+                messageArg = "";
+                messageArgs[i] = "";
+            }
+
             std::string argStr = "%" + std::to_string(++i);
             size_t argPos = msg.find(argStr);
             if (argPos != std::string::npos)


redfish-core/lib/log_services.hpp
```diff
@@ -6728,9 +6728,21 @@ static LogParseError
...
+        if (messageArgs[0] != "USYS_CONFIG")
+        {
+            BMCWEB_LOG_WARNING << "Unexpected audit log entry type: "
+                               << messageArgs[0];
+        }
+
         int i = 0;
         for (auto messageArg : messageArgs)
         {
+            if (messageArg == nullptr)
```
baemyung, line 6740: If this happens, how should the original `messageArgs` be used at L6764?

```
    logEntryJson["MessageArgs"] = std::move(messageArgs);
```

Should `messageArgs` be changed with "" (after reconstructing it)?

jeaaustx: Oh, good point. Yes it should probably also be adjusted to use a null string so the GUI doesn't have to worry about getting a null entry.

baemyung (approved):

redfish-core/lib/log_services.hpp
```diff
@@ -6728,9 +6728,22 @@ static LogParseError
...
+        }
+
+        uint i = 0;
         for (auto messageArg : messageArgs)
         {
+            if (messageArg == nullptr)
+            {
+                BMCWEB_LOG_DEBUG << "Handle null messageArg";
+                messageArg = "";
+                messageArgs[i] = "";
```
baemyung, line 6744: I don't know much about json-array but it may be different from the other types of array.
Can it be okay to modify an element in a loop?

jeaaustx: I have limited knowledge. But reading the JSON documentation I don't think it should be a problem to change the value part of an array element inside an iterator. The key part is used for determining the order of the iterator, so I think changing or adding a key might confuse the iterator.

https://json.nlohmann.me/features/iterators/#range-based-for-loops

Gunnar can you confirm if this is okay?

baemyung: It seems okay for doing the value part during the loop.

PR #852, opened 2024-03-13
1060: Catching up a few bug fixes

baemyung:
A few number of bugs (which could be critical) have been fixed, and it is catching up into downstream.
They are

- https://github.com/openbmc/bmcweb/commit/a6695a8441445e441009ad64ff04c732f2538426
    Clear json object

-  https://github.com/openbmc/bmcweb/commit/f3cb5df94799e94df37f9e2e6c2fcc473e0cae4a
    http_client:  fix for broken connection

- https://github.com/openbmc/bmcweb/commit/a88942019fdd3d8fc366999f7c178f3e1c18b2fe
    bmcweb: Add nullptr check for weakptr

### DIFF
--- a/http/http_client.hpp
+++ b/http/http_client.hpp
@@ -380,6 +380,16 @@ class ConnectionInfo : public std::enable_shared_from_this<ConnectionInfo>
         unsigned int respCode = parser->get().result_int();
         BMCWEB_LOG_DEBUG << "recvMessage() Header Response Code: " << respCode;
 
+        // Handle the case of stream_truncated.  Some servers close the ssl
+        // connection uncleanly, so check to see if we got a full response
+        // before we handle this as an error.
+        if (!parser->is_done())
+        {
+            state = ConnState::recvFailed;
+            waitAndRetry();
+            return;
+        }
+
         // Make sure the received response code is valid as defined by
         // the associated retry policy
         if (connPolicy->invalidResp(respCode))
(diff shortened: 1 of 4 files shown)


gtmills:
@baemyung Do we think this fixes the core dump we are seeing ?

gtmills (approved):

raviteja-b:
@baemyung have you verified basic event flow with this PR with HMC connected?

http/http_client.hpp
```diff
@@ -380,6 +380,16 @@ class ConnectionInfo : public std::enable_shared_from_this<ConnectionInfo>
         unsigned int respCode = parser->get().result_int();
         BMCWEB_LOG_DEBUG << "recvMessage() Header Response Code: " << respCode;
 
+        // Handle the case of stream_truncated.  Some servers close the ssl
```
abhilashraju, line 383: can we take the latest code from upstream? It does not uses socket pointer

http/http_response.hpp
```diff
@@ -153,7 +153,7 @@ struct Response
     {
         BMCWEB_LOG_DEBUG << this << " Clearing response containers";
         stringResponse.emplace(response_type{});
-        jsonValue.clear();
+        jsonValue = nullptr;
```
abhilashraju, line 156: jsonValue is an object not a pointer right? should't it be like jsonValue=nlohmann::json();

baemyung: I think both ways can be used - https://github.com/nlohmann/json/issues/2156
Here, we're trying to sync up with what upstream does - https://github.com/openbmc/bmcweb/blob/9970e93f794d0110de436828775e333e81330ad5/http/http_response.hpp#L194

baemyung:
> @baemyung have you verified basic event flow with this PR with HMC connected?

I have asked the tester of 595145 to try it out,and waiting for the result.

baemyung:
> @baemyung Do we think this fixes the core dump we are seeing ?

I don't know at this time, but these seem critical ones to resolve regardless.

gtmills:
Let's get this in. We can do something more later

PR #854, opened 2024-03-16
1110: Make PCIe Adapters Unique Path  (#590)

baemyung:
This commit makes redfish adapter objects unique, before this commit bmcweb shows redfish adapter objects with the same path even though all the dbus objects have unique paths
```
{"@odata.id": "/redfish/v1/Systems/system/PCIeDevices/adapter1"},
{"@odata.id": "/redfish/v1/Systems/system/PCIeDevices/adapter1"}
```

after this change bmcweb retrieve PCI device endpoint information and then, replace redfish PCI device as
"chassisN_io_moduleN_slotN_adapterN".

this change also affect below path because they also has slot and chassis value in their path.
"/xyz/openbmc_project/inventory/system/chassis/motherboard/pcieslot0/pcie_card0"

Tested result:

```
curl -k -H "X-Auth-Token: $token" -X GET  https://${bmc}/redfish/v1/Systems/system

{
...
...
"PCIeDevices": [
...
...
{"@odata.id": "/redfish/v1/Systems/system/PCIeDevices/chassis_pcieslot0_pcie_card0"},
{"@odata.id": "/redfish/v1/Systems/system/PCIeDevices/chassis_pcieslot1_pcie_card1"},
...
...
{"@odata.id": "/redfish/v1/Systems/system/PCIeDevices/chassis15363_io_module1_slot1_adapter1"},
{"@odata.id": "/redfish/v1/Systems/system/PCIeDevices/chassis15363_io_module1_slot2_adapter1"},
...
],
...
```

### DIFF
--- a/redfish-core/include/utils/pcie_util.hpp
+++ b/redfish-core/include/utils/pcie_util.hpp
@@ -15,6 +15,7 @@
 #include <array>
 #include <memory>
 #include <optional>
+#include <ranges>
 #include <string>
 #include <string_view>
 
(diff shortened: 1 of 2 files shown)


gtmills:
Can we add getting this upstream to https://jsw.ibm.com/browse/PFEBMC-2003 ?

redfish-core/include/utils/pcie_util.hpp
```diff
@@ -24,27 +24,90 @@ namespace pcie_util
...
+ * @return string: unique PCIe device name
  */
+inline std::string buildPCIeUniquePath(const std::string& fullPath)
+{
+    sdbusplus::message::object_path path(fullPath);
+    sdbusplus::message::object_path parentPath = path.parent_path();
+    sdbusplus::message::object_path parentofParentPath =
+        parentPath.parent_path();
+    std::string devName;
+
```
jeaaustx, line 44: Looks like this is making an assumption that the PCI device endpoint would be at most three levels deep. Is that a safe assumption?

redfish-core/include/utils/pcie_util.hpp
```diff
@@ -24,27 +24,90 @@ namespace pcie_util
...
+ * @param[i]   fullPath  object path of PCIe device
+ *
+ * @return string: unique PCIe device name
  */
+inline std::string buildPCIeUniquePath(const std::string& fullPath)
+{
+    sdbusplus::message::object_path path(fullPath);
+    sdbusplus::message::object_path parentPath = path.parent_path();
+    sdbusplus::message::object_path parentofParentPath =
+        parentPath.parent_path();
```
jeaaustx, line 42: I can't find the parent_path() method source or documentation, will it handle if parentPath is empty?

redfish-core/include/utils/pcie_util.hpp
```diff
@@ -24,27 +24,90 @@ namespace pcie_util
...
+    }
+    if (!parentofParentPath.filename().empty())
+    {
+        devName += parentofParentPath.filename() + "_";
+    }
+    if (!parentPath.filename().empty())
+    {
+        devName += parentPath.filename() + "_";
+    }
+    devName += path.filename();
```
jeaaustx, line 57: Is this in a performance sensitive path?

If so, could revert and nest the order of the checks to avoid the parent checks for the levels above the one which is empty. E.g.

if (!parentPath.filename().empty())
{
   if (!parentofParentPath.filename().empty())
  {
     if (!parentofParentPath.parent_path().filename().empty())
     {
        devName = parentofParentPath.parent_path().filename() + "_";
     }
     devName +=...
  }
  devName += ...
}

redfish-core/include/utils/pcie_util.hpp
```diff
@@ -24,27 +24,90 @@ namespace pcie_util
...
-        boost::urls::url("/redfish/v1/Systems/system/PCIeDevices");
 
-    collection_util::getCollectionToKey(
-        asyncResp, pcieDeviceUrl, pcieDeviceInterface,
-        "/xyz/openbmc_project/inventory", jsonKeyName);
+    dbus::utility::getSubTreePaths(
+        "/xyz/openbmc_project/inventory", 0, pcieDeviceInterface,
+        [asyncResp, jsonKeyName](
+            const boost::system::error_code& ec,
+            const dbus::utility::MapperGetSubTreePathsResponse& paths) {
```
jeaaustx, line 78: Should this > 10 line lambda be in a named function?

redfish-core/include/utils/pcie_util.hpp
```diff
@@ -24,27 +24,90 @@ namespace pcie_util
...
+        nlohmann::json::json_pointer jsonCountKeyName = jsonKeyName;
+        std::string back = jsonCountKeyName.back();
```
jeaaustx, line 80: If there was an error can we count on jsonKeyName being set and valid? Looks like back() will throw if there is no parent. (As will pop_back() below.)

Seems like these assignments should be after checking for ec.

baemyung:
Upstream commit: https://gerrit.openbmc.org/c/openbmc/bmcweb/+/70090

redfish-core/include/utils/pcie_util.hpp
baemyung: It will return "/" if path does not contain parent.
https://github.com/openbmc/sdbusplus/blob/e12a23cc2e28b49234d042ac60eafec4ce4e3a3b/src/message/native_types.cpp#L97

redfish-core/include/utils/pcie_util.hpp
baemyung: I haven't actually read the code `buildPCIeUniquePath()` which has been inherited for long.
Reading the code, I think it should be okay  even if its depth is less than 3. It that case, it will be return "".
However,  I think this is used only for PCIeDevice which will be more than 3 depths....

redfish-core/include/utils/pcie_util.hpp
baemyung: Right. I think we can make the code better.

redfish-core/include/utils/pcie_util.hpp
baemyung: I refactored it.

redfish-core/include/utils/pcie_util.hpp
baemyung: This is the logic to get  a json pointer to have "<Name>.@odataid" from "<Name>" -- copying from the upstream code - https://github.com/openbmc/bmcweb/blob/c3805d43db7b5f4e11af39582a48aa8255383716/redfish-core/include/utils/collection.hpp#L30

redfish-core/include/utils/pcie_util.hpp
baemyung: I modified it to go thru 3 depths

gtmills (approved):
Let's drive upstream a better way to do this. Seems reasonable though

baemyung:
> https://gerrit.openbmc.org/c/openbmc/bmcweb/+/70090

Upstream conversations is going - https://gerrit.openbmc.org/c/openbmc/bmcweb/+/70090/comment/eca02937_d7dcd16d/

PR #863, opened 2024-03-20
1060: Sync http client socket handling to upstream

baemyung:
The commit 2523e03[1] changes `conn` of http_client as a pointer and This makes the discrepancy from the upstream and causes the issue.

```
- boost::asio::ip::tcp::socket conn;
+ std::unique_ptr<boost::asio::ip::tcp::socket> conn;
```

This commit is to sync up with upstream in handling http client.

[1] https://github.com/ibm-openbmc/bmcweb/commit/2523e033cc84c2a0d9405a64e682969a0e8a8664

### DIFF
--- a/http/http_client.hpp
+++ b/http/http_client.hpp
@@ -151,7 +151,7 @@ class ConnectionInfo : public std::enable_shared_from_this<ConnectionInfo>
     boost::asio::io_context& ioc;
     std::optional<boost::beast::ssl_stream<boost::asio::ip::tcp::socket&>>
         sslConn;
-    std::unique_ptr<boost::asio::ip::tcp::socket> conn;
+    boost::asio::ip::tcp::socket conn;
 
     boost::asio::steady_timer timer;
 
@@ -194,7 +194,7 @@ class ConnectionInfo : public std::enable_shared_from_this<ConnectionInfo>
         timer.async_wait(std::bind_front(onTimeout, weak_from_this()));
 
         boost::asio::async_connect(
-            *conn, endpointList,
+            conn, endpointList,
             std::bind_front(&ConnectionInfo::afterConnect, this,
                             shared_from_this()));
     }
@@ -293,7 +293,7 @@ class ConnectionInfo : public std::enable_shared_from_this<ConnectionInfo>
         else
         {
             boost::beast::http::async_write(
-                *conn, req,
+                conn, req,
                 std::bind_front(&ConnectionInfo::afterWrite, this,
                                 shared_from_this()));
         }
@@ -346,7 +346,7 @@ class ConnectionInfo : public std::enable_shared_from_this<ConnectionInfo>
         else
         {
             boost::beast::http::async_read(
-                *conn, buffer, *parser,
+                conn, buffer, *parser,
                 std::bind_front(&ConnectionInfo::afterRead, this,
                                 shared_from_this()));
         }
@@ -364,7 +364,7 @@ class ConnectionInfo : public std::enable_shared_from_this<ConnectionInfo>
         }
 
         timer.cancel();
-        if (ec)
+        if (ec && ec != boost::asio::ssl::error::stream_truncated)
         {
             BMCWEB_LOG_ERROR << "recvMessage() failed: " << ec.message()
                              << " from " << host << ": "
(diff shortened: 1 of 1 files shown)


baemyung:
I also go thru the other upstream commits to find out any critical ones.
However,  most of them are synced-up.  

There are additional upstream commits (not in 1060), but I think they are not required.

- https://github.com/openbmc/bmcweb/commit/b2896149c39967dd9d1ee79357bdc53537cfabd7
- https://github.com/openbmc/bmcweb/commit/a716aa74954617e32a1d2e691d184580402b3eaf
- https://github.com/openbmc/bmcweb/commit/e01d0c36af115ed46d54b5dbbacfe3ad92226bd3
- https://github.com/openbmc/bmcweb/commit/52e31629978bbcefc0e14c96272ef77c2dad6a9c
- https://github.com/openbmc/bmcweb/commit/f8ca6d79390b7f7fac5dbb82635dd6384810c668
- https://github.com/openbmc/bmcweb/commit/e1452bea2809f7189a69ee562e5078e452f0a251

baemyung:
This is for the defect - https://jazz07.rchland.ibm.com:13443/jazz/web/projects/CSSD#action=com.ibm.team.workitem.viewWorkItem&id=595145

It is also syncing up to upstream.
- https://github.com/openbmc/bmcweb/commit/f3cb5df94799e94df37f9e2e6c2fcc473e0cae4a)
- https://gerrit.openbmc.org/c/openbmc/bmcweb/+/67961

gtmills:
@baemyung Any testing here?

http/http_client.hpp
```diff
@@ -364,7 +367,7 @@ class ConnectionInfo : public std::enable_shared_from_this<ConnectionInfo>
         }
 
         timer.cancel();
-        if (ec)
+        if (ec && ec != boost::asio::ssl::error::stream_truncated)
```
abhilashraju, line 370: There is a parser check missing in line 380

http/http_client.hpp
```diff
@@ -626,10 +628,10 @@ class ConnectionInfo : public std::enable_shared_from_this<ConnectionInfo>
 
     void initializeConnection(bool ssl)
     {
-        conn = std::make_unique<boost::asio::ip::tcp::socket>(ioc);
+        conn = boost::asio::ip::tcp::socket(ioc);
         if (ssl)
         {
-            /* std::optional<boost::asio::ssl::context> sslCtx =
+            std::optional<boost::asio::ssl::context> sslCtx =
```
abhilashraju, line 634: this will have problem in downstream.  I see that downstream was not using proper ssl context as upstream for some reason. It is better to keep it that way

sunharis:
> @baemyung Any testing here?

This needs to be tested for the concurrent code update regression as mentioned in the defect. At least 15 iterations must pass.

http/http_client.hpp
sunharis: Yes. This was commented out in downstream - to work with the HMC's SSL context.

sunharis:
> > > @baemyung Any testing here?
> > 
> > 
> > This needs to be tested for the concurrent code update regression as mentioned in the defect. At least 15 iterations must pass.
> 
> I don't know how to test it with sandbox build - esp like what the tester is doing? Could you help me out how to do it?
> 
> BTW, the problem is happening only when it was done via HMC. Also note that, the problem is not only related to code update, but happening on other cases as long as it is done via HMC.

Yeah. Abhilash is on it. Lets see how the test goes.

http/http_client.hpp
gtmills: This as is matches what we have upstream. 
https://github.com/openbmc/bmcweb/blob/1fccd0d5eafd56049a8a381543dd83d5559fbf25/http/http_client.hpp#L365

I don't understand why we want something different. @abhilashraju ?

gtmills: If we want something different, then let's do that upstream

http/http_client.hpp
gtmills: Why couldn't we have gone upstream with https://github.com/ibm-openbmc/bmcweb/commit/d104e6f83a28a291c4637f8d4bafa796c175b493  ? 

Let's just comment out for now. 
Then to me, this looks good.

gtmills (approved):

PR #864, opened 2024-03-21
Events: modified state to close from idle

abhilashraju:
~Fix for 595145 by taking upstream changes~
Plus modified state to close from idle.
we reach retry max condition when we are not able to send data to hmc due connection error. we must recreate new socket before attempting to try again. close will make sure that the next try is with new socket else it will attempt with old socket and evnentually fails. We can avoid this failure if we make state closed

### DIFF
--- a/redfish-core/lib/assembly.hpp
+++ b/redfish-core/lib/assembly.hpp
@@ -110,7 +110,7 @@ inline void
             "tod_battery")
         {
             tempyArray.at(assemblyIndex)["Oem"]["OpenBMC"]["@odata.type"] =
-                "#OemAssembly.v1_0_0.Assembly";
+                "#OemAssembly.v1_0_0.OpenBMC";
 
             crow::connections::systemBus->async_method_call(
                 [aResp, assemblyIndex](const boost::system::error_code ec) {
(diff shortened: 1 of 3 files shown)


gtmills:
What upstream changes? 
How does this compare with https://github.com/ibm-openbmc/bmcweb/pull/863 ?

baemyung:
It is same as #863 except this changes the way of handling "RetryForever" (from 1050/1060/upstream).
Does this mean that the connection will be re-attempted? 
Is it expected for HMC?

```
            if (connPolicy->retryPolicyAction == "RetryForever")
            {
+                state = ConnState::closed;
            }
```

http/http_client.hpp
```diff
@@ -468,7 +468,7 @@ class ConnectionInfo : public std::enable_shared_from_this<ConnectionInfo>
             }
             if (connPolicy->retryPolicyAction == "RetryForever")
             {
-                state = ConnState::idle;
+                state = ConnState::closed;
```
gtmills, line 471: Can you explain in the PR message, why we need this?

abhilashraju: we reach here when we are not able to send data to hmc due connection error. we must recreate new socket before attempting to try again. close will make sure that the next try is with new socket else it will attempt with old socket and evnetually fails. We can avoid this failure if we make state closed

baemyung (approved):
It looks good to me

http/http_client.hpp
gtmills: I updated the title and crossed some text out of the PR message. For next time, please make sure your PR msg, etc reflects the latest state.

PR #870, opened 2024-03-25
Populate BMC state based on state manager property (#520)

geissonator:
The "Status" of the BMC has been historically fairly simple. If the BMC is at the point where it can respond to the Redfish call then it's "OK" and it's "On". Its "Status" could be derived based on the startup systemd target, multi-user.target, as either "Starting" or "Enabled".

However, with the introduction of new BMC states it is no longer this simple. The following design introduced a new "Quiesced" state for the BMC[1]

There has was also a new BMC state introduced, "UpdateInProgress", which indicates the BMC is in the middle of a firmware update.

These new BMC states require additional logic when responding to Redfish queries for the BMC "Status".

When a BMC is in a "Quiesced" state it is an indication that a critical BMC service has entered the failed state. It is important for BMC clients to be aware of this state to take the needed steps to service the system. For example, on IBM systems, this will trigger a call home event to IBM service. The undefined aspect of this state is what makes it so critical to get serviced.

bmcweb has a precedent of abstracting systemd states behind services like x86-power-control/phosphor-state-manager. It can see it in functions like getHostState() and getChassisState(). bmcweb could contain all of the logic to determine if the BMC is being updated or has a failed critical service but that:
  a) Adds a lot more code to bmcweb and
  b) Goes away from our design of allowing other repositories to
     abstract away the complexity of these states.

Tested:
- GET /redfish/v1/Managers/bmc for different BMC states
- Redfish Validator passed

[1]: https://github.com/openbmc/docs/blob/master/designs/bmc-service-failure-debug-and-recovery.md

Change-Id: I8ceafb7036f41678516ebda3dce5200221bba947

### DIFF
--- a/redfish-core/lib/managers.hpp
+++ b/redfish-core/lib/managers.hpp
@@ -45,6 +45,7 @@
 #include <memory>
 #include <ranges>
 #include <sstream>
+#include <string>
 #include <string_view>
 #include <variant>
 
@@ -1962,6 +1963,55 @@ inline void
     });
 }
 
+inline void getBMCState(const std::shared_ptr<bmcweb::AsyncResp>& aResp)
+{
+    aResp->res.jsonValue["PowerState"] = "On";
+    sdbusplus::asio::getProperty<std::string>(
+        *crow::connections::systemBus, "xyz.openbmc_project.State.BMC",
+        "/xyz/openbmc_project/state/bmc0", "xyz.openbmc_project.State.BMC",
+        "CurrentBMCState",
+        [aResp](const boost::system::error_code ec,
+                const std::string& bmcState) {
+        if (ec)
+        {
+            BMCWEB_LOG_DEBUG("DBUS response error reading CurrentBmcState");
+            aResp->res.jsonValue["Status"]["State"] = "Enabled";
+            aResp->res.jsonValue["Status"]["Health"] = "OK";
+            return;
+        }
+
+        if (bmcState == "xyz.openbmc_project.State.BMC.BMCState.Ready")
+        {
+            aResp->res.jsonValue["Status"]["State"] = "Enabled";
+            aResp->res.jsonValue["Status"]["Health"] = "OK";
+        }
+        else if (bmcState == "xyz.openbmc_project.State.BMC.BMCState."
+                             "Quiesced")
+        {
+            aResp->res.jsonValue["Status"]["State"] = "Quiesced";
+            aResp->res.jsonValue["Status"]["Health"] = "Critical";
+        }
+        else if (bmcState == "xyz.openbmc_project.State.BMC.BMCState."
+                             "NotReady")
+        {
+            aResp->res.jsonValue["Status"]["State"] = "Starting";
+            aResp->res.jsonValue["Status"]["Health"] = "OK";
+        }
+        else if (bmcState == "xyz.openbmc_project.State.BMC.BMCState."
+                             "UpdateInProgress")
+        {
+            aResp->res.jsonValue["Status"]["State"] = "Updating";
+            aResp->res.jsonValue["Status"]["Health"] = "OK";
+        }
+        else
+        {
+            BMCWEB_LOG_DEBUG("Unsupported D-Bus CurrentBMCState: {}", bmcState);
+            aResp->res.jsonValue["Status"]["State"] = "Enabled";
+            aResp->res.jsonValue["Status"]["Health"] = "OK";
+        }
+    });
+}
+
 inline void requestRoutesManager(App& app)
 {
     std::string uuid = persistent_data::getConfig().systemUuid;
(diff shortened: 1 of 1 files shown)


geissonator:
Similar statement as those made in https://github.com/ibm-openbmc/bmcweb/pull/520

This is a cherry-pick of upstream change https://gerrit.openbmc.org/c/openbmc/bmcweb/+/50318 which we have been unsuccessful with getting into upstream. The IBM arguments for this change are:

- This area is very nuanced and we've worked through a lot of no-connect issues with the HMC team on this behavior
- Upstream code may not handle reflecting the correct BMC state in the small window between a BMC reboot being requested and actually rebooting (we have had multiple HMC defects due to this)

I verified the expected "Status" response for all possible BMC states. 
I did not run Redfish validation as this had been done previously and this does not change any responses.

redfish-core/lib/managers.hpp
```diff
@@ -1962,6 +1962,55 @@ inline void
...
+        }
+        else if (bmcState == "xyz.openbmc_project.State.BMC.BMCState."
+                             "UpdateInProgress")
+        {
+            aResp->res.jsonValue["Status"]["State"] = "Updating";
+            aResp->res.jsonValue["Status"]["Health"] = "OK";
+        }
+        else
+        {
+            BMCWEB_LOG_DEBUG("Unsupported D-Bus CurrentBMCState: {}", bmcState);
```
gtmills, line 2007: Not sure why you need this, but 🤷 
You are sweeping this error under the rug. Typically we would do an internal error here but since this is what we had before 🤷 I am okay with it

redfish-core/lib/managers.hpp
```diff
@@ -1962,6 +1962,55 @@ inline void
...
 
+inline void getBMCState(const std::shared_ptr<bmcweb::AsyncResp>& aResp)
+{
+    aResp->res.jsonValue["PowerState"] = "On";
+    sdbusplus::asio::getProperty<std::string>(
+        *crow::connections::systemBus, "xyz.openbmc_project.State.BMC",
+        "/xyz/openbmc_project/state/bmc0", "xyz.openbmc_project.State.BMC",
+        "CurrentBMCState",
+        [aResp](const boost::system::error_code ec,
+                const std::string& bmcState) {
```
gtmills, line 1973: `#include <string>`

Up to you if you want to change

gtmills (approved):
Change if you want, I am okay as is

redfish-core/lib/managers.hpp
```diff
@@ -1962,6 +1962,55 @@ inline void
...
+        if (ec)
+        {
+            BMCWEB_LOG_DEBUG("DBUS response error reading CurrentBmcState");
+            aResp->res.jsonValue["Status"]["State"] = "Enabled";
```
gtmills, line 1977: 3 of these Enabled/OK could rewrite as  

```
    aResp->res.jsonValue["PowerState"] = "On";
    aResp->res.jsonValue["Status"]["State"] = "Enabled";
    aResp->res.jsonValue["Status"]["Health"] = "OK";
   
    sdbusplus::asio::getProperty<std::string>(
        *crow::connections::systemBus, "xyz.openbmc_project.State.BMC",
        "/xyz/openbmc_project/state/bmc0", "xyz.openbmc_project.State.BMC",
        "CurrentBMCState",
        [aResp](const boost::system::error_code ec,
                const std::string& bmcState) {
       if (ec)
        {
            BMCWEB_LOG_DEBUG("DBUS response error reading CurrentBmcState");
            // Just use the default
            return;
        }

       // .BMC.BMCState.Ready is just the default
       
        if (bmcState == "xyz.openbmc_project.State.BMC.BMCState."
                             "Quiesced")
        {
            aResp->res.jsonValue["Status"]["State"] = "Quiesced";
            aResp->res.jsonValue["Status"]["Health"] = "Critical";
        }
        else if (bmcState == "xyz.openbmc_project.State.BMC.BMCState."
                             "NotReady")
        {
            aResp->res.jsonValue["Status"]["State"] = "Starting";
        }
        else if (bmcState == "xyz.openbmc_project.State.BMC.BMCState."
                             "UpdateInProgress")
        {
            aResp->res.jsonValue["Status"]["State"] = "Updating";
        }      
        else
        {
            BMCWEB_LOG_DEBUG("Unsupported D-Bus CurrentBMCState: {}", bmcState);
            // Just use the default 
        }
 ```

gtmills (approved):
Change if you want, I am okay as is

redfish-core/lib/managers.hpp
geissonator: I think it was to stick with the previous bmcweb default (hard code to this to start with). I'm pretty in-tune with any new bmc states so lets just leave as-is.

redfish-core/lib/managers.hpp
geissonator: Done

redfish-core/lib/managers.hpp
gtmills: Ack

redfish-core/lib/managers.hpp
geissonator: I like a bit more code that's a bit more obvious and readable, and also code that's been tested (1020-1060) so lets leave as-is.

gtmills (approved):

baemyung (approved):
Upstream renamses `aResp` to  'asyncResp`.

Other than those, I think I'm good with the code.

PR #871, opened 2024-03-27
1060: Use shared_ptr for request body

baemyung:

```
struct Request
{
   boost::beast::http::request<boost::beast::http::string_body> req;
...
}
```

This may exhibit a few issues.

- If it is moved to a lambda of user authentication async method call, the original `completeRequest()` may observe the invalid content of it [1]. This is causing a problem for audit recording.

```
            [req{std::move(req)}, asyncResp, &rule, params](
```

- It may be copied (e.g. captured by value) and causing the excessive memory consumption [2], in case `req` is large like the case of code update.

```
        afterIfMatchRequest, std::ref(app), asyncResp, req, ifMatch));
```

To address those issues, the content of`req` would need to be valid until all references are done, and also the copy of `req` would need to be less overhead.

So, it can be changed as `shared_ptr` and captured by value for lambda.

```
   std::shared_ptr<boost::beast::http::request<boost::beast::http::string_body>> reqPtr;
```

```
-            [req{std::move(req)}, asyncResp, &rule, params](
+            [req, asyncResp, &rule, params](

```

Tested:
- Audit runs correctly
- Debug stmt shows the content correct at `completeRequest()` on PATCH  like

```
```

- Validator passes

[1] https://github.com/ibm-openbmc/bmcweb/blob/ef23e48288b35d30baaf57b61299597ca26b1930/http/routing.hpp#L1462C1-L1463C61
[2] https://github.com/ibm-openbmc/bmcweb/blob/ef23e48288b35d30baaf57b61299597ca26b1930/redfish-core/include/query.hpp#L97C1-L98C71

### DIFF
--- a/http/http_request.hpp
+++ b/http/http_request.hpp
@@ -10,6 +10,7 @@
 #include <boost/beast/websocket.hpp>
 #include <boost/url/url_view.hpp>
 
+#include <memory>
 #include <string>
 #include <string_view>
 #include <system_error>
@@ -19,7 +20,11 @@ namespace crow
 
 struct Request
 {
-    boost::beast::http::request<boost::beast::http::string_body> req;
+    using http_request_body =
+        boost::beast::http::request<boost::beast::http::string_body>;
+    std::shared_ptr<http_request_body> reqPtr;
+    http_request_body& req;
+
     boost::beast::http::fields& fields;
     std::string_view url{};
     boost::urls::url_view urlView{};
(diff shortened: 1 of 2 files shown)


jeaaustx (approved):
Looks good to me.

jeaaustx:
Note: This fixes defect 600746

gtmills:
Upstream first here

baemyung:
Upstream: https://gerrit.openbmc.org/c/openbmc/bmcweb/+/70355

gtmills (approved):

PR #876, opened 2024-04-03
Pull Upstream commit: Create Redfish specific setProperty call

asmithakarun:
There are currently 78 sdbusplus::asio::setProperty calls in redfish-core. 

Tested:
```
curl  -k --user "root:0penBmc" -H "Content-Type: application/json" -X PATCH -d '{"HostName":"openbmc@#"}' https://192.168.7.2/redfish/v1/Managers/bmc/EthernetInterfaces/eth0
```

Returns the appropriate error in the response
Base.1.16.0.PropertyValueIncorrect

Change-Id: If033a1112ba516792c9386c997d090c8f9094f3a

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -370,6 +370,7 @@ fs = import('fs')
 
 srcfiles_bmcweb = files(
   'redfish-core/src/error_messages.cpp',
+  'redfish-core/src/utils/dbus_utils.cpp',
   'redfish-core/src/utils/json_utils.cpp',
   'src/boost_asio_ssl.cpp',
   'src/boost_asio.cpp',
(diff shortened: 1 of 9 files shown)


meson.build
```diff
@@ -13,41 +13,40 @@ project('bmcweb', 'cpp',
             'werror=true',
            ])
 
-# Project related links
+#Project related links
```
gtmills, line 16: What is going on here?

asmithakarun: i ran clang-tidy on this file and found these changes.
(clang-format-17 -i meson.build)

gtmills: This is a meson file. This doesn't match with what is upstream. 
https://github.com/openbmc/bmcweb/blob/master/meson.build#L386

gtmills (changes requested):

meson.build
gtmills: Can you revert your changes to this file and only add the one line you need?

gtmills:
What is the upstream commit here? Can you add that to the PR msg?

meson.build
asmithakarun: okay

raviteja-b (approved):
Thank you so much @asmithakarun for pulling all these changes into downstream PR.
This improves error handling and cleans up few outstanding defects

baemyung (approved):
It looks good to me

asmithakarun:
> What is the upstream commit here? Can you add that to the PR msg?

done added in the PR description

gtmills (approved):

baemyung:
I thought this is 1110 but it is for 1060.
I would suggest to add "1060:" or "1110:" to PR subject line.

I think this would also needs to be done for 1110.

PR #877, opened 2024-04-05
Pull Upstream change: Create Redfish specific setProperty call

asmithakarun:
The following PR is pulled for 1050 branch explicitly as it had merge conflicts:
https://github.com/ibm-openbmc/bmcweb/pull/876

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -338,6 +338,7 @@ fs = import('fs')
 
 srcfiles_bmcweb = files(
   'redfish-core/src/error_messages.cpp',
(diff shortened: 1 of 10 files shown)


gtmills (changes requested):
Let's see what happens with https://ibm-systems-power.slack.com/archives/C1HT3FHNK/p1712322538533119 first

gtmills:
Decided not to do this for 1020-1060. Closing. Thanks for these changes, Asmitha. We will go with this direction for 1110

PR #879, opened 2024-04-05
Pull Upstream commit: Create Redfish specific setProperty call

asmithakarun:
PR https://github.com/ibm-openbmc/bmcweb/pull/876 is pulled for 1050 branch explicitly as it had merge conflicts.


baemyung:
For 1050, MUSTFIX approval would be needed.

On the other hand, since https://github.com/ibm-openbmc/bmcweb/pull/876 is already merged for 1060, it is better to put it into 1050 (as 1050/1060 code base is pretty same or similar).
Or, should we undo 1060 #876?  Is it required for 1050/1060?

Even if we include it into 1050, I think it is still better to sync up between them -- i.e. `redfish-core/lib/hypervisor_system.hpp` is better not to be changed for 1050.

PR #890, opened 2024-04-08
1110: ReadOnly users should not have ssh (#408) (#514)

baemyung:
* ReadOnly users should not have ssh

Note that ReadOnly user created prior to this change will have the HostConsole AccountType privilege.  A BMC admin can PATCH a user's AccountTypes property to add or remove the Hostconsole value.

Tested: Yes


gtmills:
We had for 1020-1060 so seems reasonable. 
For the future, I think we need to review the AccountTypes/Permissions.

gtmills (approved):

PR #898, opened 2024-04-12
Check PowerSupply availability prop for State/Hlth

spinler:
This commit makes use of the Available property on the power supply D-Bus object when determining the power supply state and health in the PowerSupply response.  The power supply monitor code from phosphor-power will set the property to false if it can determine that the power supply isn't able to provide power due to certain faults.

If the PS isn't available, then the Health property will be set to Critical, and the State will be set to UnavailableOffline (assuming it shouldn't be Absent instead).

This is necessary because on IBM systems the Functional property determines the fault LED state as well as the health value, and in certain cases the fault LED needs to stay off even though the PS health is not OK.  A specific example of this is when the PS cord is unplugged: no fault LED is desired but it is desired for the Redfish output to show that there is an issue with the PS.

If the Available property isn't present on D-Bus, it acts as if it had a value of true and behaves the same as it does today.

Tested:
Available = false:
    "Health": "Critical",
    "State": "UnavailableOffline"

    "Health": "Critical",
    "State": "Enabled"

PS missing:
    "Health": "Critical",
    "State": "Absent"

Everything OK:
    "Health": "OK",
    "State": "Enabled"

    "Health": "OK",
    "State": "Enabled"

Change-Id: I8758e1e6f0ce7e7ab5ff83fa540abe1b20c6c0e2

### DIFF
--- a/redfish-core/lib/power_supply.hpp
+++ b/redfish-core/lib/power_supply.hpp
@@ -205,12 +205,14 @@ inline void getValidPowerSupplyPath(
 
 inline void
     getPowerSupplyState(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
-                        const std::string& service, const std::string& path)
+                        const std::string& service, const std::string& path,
+                        bool available)
 {
     sdbusplus::asio::getProperty<bool>(
         *crow::connections::systemBus, service, path,
         "xyz.openbmc_project.Inventory.Item", "Present",
-        [asyncResp](const boost::system::error_code& ec, const bool value) {
+        [asyncResp, available](const boost::system::error_code& ec,
+                               const bool value) {
         if (ec)
         {
             if (ec.value() != EBADR)
@@ -226,17 +228,23 @@ inline void
         {
             asyncResp->res.jsonValue["Status"]["State"] = "Absent";
         }
+        else if (!available)
+        {
+            asyncResp->res.jsonValue["Status"]["State"] = "UnavailableOffline";
+        }
     });
 }
 
 inline void
     getPowerSupplyHealth(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
-                         const std::string& service, const std::string& path)
+                         const std::string& service, const std::string& path,
+                         bool available)
 {
     sdbusplus::asio::getProperty<bool>(
         *crow::connections::systemBus, service, path,
         "xyz.openbmc_project.State.Decorator.OperationalStatus", "Functional",
-        [asyncResp](const boost::system::error_code& ec, const bool value) {
+        [asyncResp, available](const boost::system::error_code& ec,
+                               const bool value) {
         if (ec)
         {
             if (ec.value() != EBADR)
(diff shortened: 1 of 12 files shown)


spinler:
This is a cherry pick of https://github.com/ibm-openbmc/bmcweb/pull/528, which was a cherry pick of https://github.com/ibm-openbmc/bmcweb/pull/262.

Upstream: https://gerrit.openbmc-project.xyz/c/openbmc/bmcweb/+/51336
 - people didn't like it
Original Defect: SW542338

We would like to remove this feature from downstream and instead implement https://jsw.ibm.com/browse/PFEBMC-766, which would show the power supply input power line status on the web UI.  That would take fairly small changes to psu-monitor, bmcweb, and the webui.

It's in the backlog at the moment, but we plan on doing it for 1110 if we can get the resources on the power team.

redfish-core/lib/power_supply.hpp
```diff
@@ -205,12 +205,14 @@ inline void getValidPowerSupplyPath(
...
     getPowerSupplyState(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
-                        const std::string& service, const std::string& path)
+                        const std::string& service, const std::string& path,
+                        bool available)
 {
     sdbusplus::asio::getProperty<bool>(
         *crow::connections::systemBus, service, path,
         "xyz.openbmc_project.Inventory.Item", "Present",
-        [asyncResp](const boost::system::error_code& ec, const bool value) {
+        [asyncResp, available](const boost::system::error_code ec,
```
gtmills, line 214: We should stay with passing by reference https://github.com/openbmc/bmcweb/commit/e81de5126d8b9412bd703b8b11a1f5daa7bd83d4

redfish-core/lib/power_supply.hpp
```diff
@@ -226,17 +228,23 @@ inline void
...
     getPowerSupplyHealth(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
-                         const std::string& service, const std::string& path)
+                         const std::string& service, const std::string& path,
+                         bool available)
 {
     sdbusplus::asio::getProperty<bool>(
         *crow::connections::systemBus, service, path,
         "xyz.openbmc_project.State.Decorator.OperationalStatus", "Functional",
-        [asyncResp](const boost::system::error_code& ec, const bool value) {
+        [asyncResp, available](const boost::system::error_code ec,
```
gtmills, line 246: reference

redfish-core/lib/power_supply.hpp
```diff
@@ -248,13 +256,37 @@ inline void
...
+    sdbusplus::asio::getProperty<bool>(
+        *crow::connections::systemBus, service, path,
+        "xyz.openbmc_project.State.Decorator.Availability", "Available",
+        [asyncResp, service, path](const boost::system::error_code ec,
+                                   const bool available) {
+        if (ec)
+        {
+            if (ec.value() != EBADR)
+            {
+                messages::internalError(asyncResp->res);
```
gtmills, line 279: If you want to add logging, it might be helpful one day 
                BMCWEB_LOG_ERROR("DBUS response error {}",
                                 ec.value());

gtmills (commented):
I am good here, just a couple minor things

redfish-core/lib/power_supply.hpp
spinler: done

redfish-core/lib/power_supply.hpp
spinler: done

redfish-core/lib/power_supply.hpp
spinler: done

redfish-core/lib/power_supply.hpp
spinler, line 214: hmm, I guess not resolved yet even though I thought a pushed the fix.  will investigate.

gtmills (approved):

PR #553, opened 2023-02-13
Redfish: Implements Redfish Assembly Health

ChicagoDuan:
Look for the OperationalStatus to set the Assembly Health. 
If the OperationalStatus is false, then set the Health to "Critical" 
If the OperationalStatus is true, then the Health is "OK"

upstream:
https://gerrit.openbmc.org/c/openbmc/bmcweb/+/50317

Signed-off-by: Chicago Duan <duanzhijia01@inspur.com>

PR #900, opened 2024-04-15
1110: Redfish: Implements Redfish Assembly Health (#553)(#628)

baemyung:
Redfish: Implements Redfish Assembly Health (#553)(#628)

Look for the OperationalStatus to set the Assembly Health. 
If the OperationalStatus is false, then set the Health to "Critical"
If the OperationalStatus is true, then the Health is "OK"

```
$curl -k -X GET https://${bmc}:18080/redfish/v1/Chassis/chassis/Assembly
{
{
...
    {
      "@odata.id": "/redfish/v1/Chassis/chassis/Assembly#/Assemblies/2",
      "Status": {
        "Health": "OK",
      }
    },
    {
      "@odata.id": "/redfish/v1/Chassis/chassis/Assembly#/Assemblies/3",
      "Status": {
        "Health": "Critical",
      }
    },
...
}
```

### DIFF
--- a/redfish-core/lib/assembly.hpp
+++ b/redfish-core/lib/assembly.hpp
@@ -278,6 +278,36 @@ void getAssemblyPresence(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
     });
 }
 
+void getAssemblyHeath(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
+                      const auto& serviceName, const auto& assembly,
+                      const auto& assemblyIndex)
+{
+    sdbusplus::asio::getProperty<bool>(
+        *crow::connections::systemBus, serviceName, assembly,
+        "xyz.openbmc_project.State.Decorator.OperationalStatus", "Functional",
+        [asyncResp, assemblyIndex](const boost::system::error_code& ec,
+                                   bool functional) {
+        if (ec)
+        {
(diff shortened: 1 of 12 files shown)


redfish-core/lib/assembly.hpp
```diff
@@ -278,6 +278,36 @@ void getAssemblyPresence(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
...
+        }
+
+        nlohmann::json& assemblyArray = asyncResp->res.jsonValue["Assemblies"];
+        nlohmann::json& assemblyData = assemblyArray.at(assemblyIndex);
+
+        if (!functional)
+        {
+            assemblyData["Status"]["Health"] = "Critical";
+        }
+        else
```
gtmills, line 304: I assume we purposely don't want a default? If this is how 1020-1060 was, then yeah we want to follow this.

gtmills (approved):

redfish-core/lib/assembly.hpp
baemyung: Right, this incorporated https://github.com/ibm-openbmc/bmcweb/pull/628 which does not set a default value.

baemyung:
Upstream: https://gerrit.openbmc.org/c/openbmc/bmcweb/+/39574
Once https://gerrit.openbmc.org/c/openbmc/bmcweb/+/39574 is done, this would be upstreamed.

PR #903, opened 2024-04-16
Update Privilege Registry from 1.3.0 to 1.5.0

gtmills:
Upstream is https://gerrit.openbmc.org/c/openbmc/bmcweb/+/70854 

Change 1 line in scripts/parse_registries.py and rerun the script.

Long term OpenBMC/bmcweb need more direction here on the Privilege Registry, but for now continue with the current direction of using the Privilege Registry and taking it from the DMTF.

There is new entries in 1.4/1.5 that are needed for future development.

Tested: It builds.

Change-Id: I4337dc44e794c58f00f7307ea0508b84f14e8c8f

### DIFF
--- a/redfish-core/include/redfish.hpp
+++ b/redfish-core/include/redfish.hpp
@@ -190,7 +190,7 @@ class RedfishService
         requestRoutesMemory(app);
 
         requestRoutesSystems(app);
-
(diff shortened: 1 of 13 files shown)


baemyung (approved):
It looks good to me

PR #908, opened 2024-04-17
FFDC: Collect current in-memory sessions for BMC dump 

nkotania:
Currently it collects bmcweb_persistent_data.json as part of BMC
dump, but this file does not contain all existing redfish sessions

This commit is to collect all sessions into bmcweb_current_session_persistent_data.json
during bmc dump creation, on receiving SIGUSR1 from dump manager.

* Add event subscriptions data along with session data


nkotania:
Not Required

PR #909, opened 2024-04-17
FFDC: Collect current in-memory sessions for BMC dump

nkotania:
Currently it collects bmcweb_persistent_data.json as part of BMC dump, but this file does not contain all existing redfish sessions

This commit is to collect all sessions into bmcweb_current_session_persistent_data.json during bmc dump creation, on receiving SIGUSR1 from dump manager.

* Add event subscriptions data along with session data

### DIFF
--- a/http/http_server.hpp
+++ b/http/http_server.hpp
@@ -19,7 +19,9 @@
 #include <memory>
 #include <utility>
 #include <vector>
-
+#ifdef BMCWEB_ENABLE_IBM_MANAGEMENT_CONSOLE
+#include <persistent_data.hpp>
+#endif
 namespace crow
 {
 
@@ -36,7 +38,11 @@ class Server
         acceptor(std::move(acceptorIn)),
         signals(*ioService, SIGINT, SIGTERM, SIGHUP), handler(handlerIn),
         adaptorCtx(std::move(adaptorCtxIn))
-    {}
+    {
+#ifdef BMCWEB_ENABLE_IBM_MANAGEMENT_CONSOLE
+        signals.add(SIGUSR1);
+#endif
+    }
 
     Server(Handler* handlerIn, uint16_t port,
            const std::shared_ptr<boost::asio::ssl::context>& adaptorCtxIn,
@@ -148,6 +154,15 @@ class Server
                     }
                     startAsyncWaitForSignal();
                 }
+#ifdef BMCWEB_ENABLE_IBM_MANAGEMENT_CONSOLE
+                if (signalNo == SIGUSR1)
+                {
+                    BMCWEB_LOG_CRITICAL(
+                        "INFO: Receivied USR1 signal to dump latest session  data for bmc dump");
+                    persistent_data::getConfig().writeCurrentSessionData();
+                    this->startAsyncWaitForSignal();
+                }
+#endif
                 else
                 {
                     stop();
--- a/include/persistent_data.hpp
+++ b/include/persistent_data.hpp
@@ -23,7 +23,8 @@ class ConfigFile
   public:
     // todo(ed) should read this from a fixed location somewhere, not CWD
     static constexpr const char* filename = "bmcweb_persistent_data.json";
-
+    static constexpr const char* dumpFilename =
+        "bmcweb_current_session_snapshot.json";
     ConfigFile()
     {
         readData();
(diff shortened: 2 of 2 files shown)


nkotania:
Referenced PR: https://github.com/ibm-openbmc/bmcweb/commit/9c8f72077191ec5632f43d0f172eaa81b9df925f?diff=unified&w=0

include/persistent_data.hpp
```diff
@@ -188,6 +189,84 @@ class ConfigFile
         }
     }
 
+#ifdef BMCWEB_ENABLE_IBM_MANAGEMENT_CONSOLE
```
gtmills, line 192: Anyway we could use `void writeData()`  ?

raviteja-b: since this is specific to session data used this method name writeCurrentSessionData

raviteja-b (approved):

gtmills (approved):

include/persistent_data.hpp
gtmills: Is a lot of dupped code :/

include/persistent_data.hpp
```diff
@@ -188,6 +189,84 @@ class ConfigFile
...
+        {
+            if (p.second->persistence !=
+                persistent_data::PersistenceType::SINGLE_REQUEST)
+            {
+                sessions.push_back({
+                    {"unique_id", p.second->uniqueId},
+                    {"username", p.second->username},
+                    {"client_ip", p.second->clientIp},
+                    {"client_id", p.second->clientId},
+                });
```
baemyung, line 227: This would be better to be done like https://github.com/ibm-openbmc/bmcweb/blob/292e2f278ecdcd62c14b744b19c0b6451eefe95d/include/persistent_data.hpp#L228

```
               nlohmann::json::object_t session;
                session["unique_id"] = p.second->uniqueId;
               ...
                sessions.emplace_back(std::move(session));
```

I think the current code may cause the compiler error....but not sure why it did not hit - perhaps it is because `ibm-management-console` is not defined.

include/persistent_data.hpp
```diff
@@ -188,6 +189,84 @@ class ConfigFile
...
+        nlohmann::json& subscriptions = data["subscriptions"];
+        subscriptions = nlohmann::json::array();
+        for (const auto& it :
+             EventServiceStore::getInstance().subscriptionsConfigMap)
+        {
+            std::shared_ptr<UserSubscription> subValue = it.second;
+            if (subValue->subscriptionType == "SSE")
+            {
+                BMCWEB_LOG_DEBUG
+                    << "The subscription type is SSE, so skipping.";
```
baemyung, line 240: This should be
```
               BMCWEB_LOG_DEBUG( "The subscription type is SSE, so skipping.");
```

I think somehow this part of code path is not enabled for compilation - eg. ibm-management-console is not defined.

baemyung: Needs to include 
```
#include "logging.hpp"
```

baemyung (changes requested):
I think there must be some problem in compilation as it is reported as succeeded.
However, the current code should cause the compilation failure - because it uses the old style of logging
```
BMCWEB_LOG_DEBUG << "....";
```
The correct way is to use `BMCWEB_LOG_DEBUG("...");`

In addition, I think some of inline-json data assignment may also cause the compilation issue -- i.e. L222-L227.

raviteja-b:
> I think there must be some problem in compilation as it is reported as succeeded. However, the current code should cause the compilation failure - because it uses the old style of logging
> 
> ```
> BMCWEB_LOG_DEBUG << "....";
> ```
> 
> The correct way is to use `BMCWEB_LOG_DEBUG("...");`
> 
> In addition, I think some of inline-json data assignment may also cause the compilation issue -- i.e. L222-L227.

yeah right

http/http_server.hpp
```diff
@@ -148,6 +154,16 @@ class Server
                     }
                     startAsyncWaitForSignal();
                 }
+#ifdef BMCWEB_ENABLE_IBM_MANAGEMENT_CONSOLE
+                if (signalNo == SIGUSR1)
+                {
+                    BMCWEB_LOG_CRITICAL
```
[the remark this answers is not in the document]
raviteja-b: +1

include/persistent_data.hpp
raviteja-b: +1

gtmills:
> CI does not have BMCWEB_ENABLE_IBM_MANAGEMENT_CONSOLE enabled?
> can we enable this by default for ibm builds?

I am good with that

http/http_server.hpp
```diff
@@ -19,7 +19,9 @@
 #include <memory>
 #include <utility>
 #include <vector>
-
+#ifdef BMCWEB_ENABLE_IBM_MANAGEMENT_CONSOLE
```
gtmills, line 22: I wonder if we should get rid of the ifdef here.. upstream bmcweb has been doing that

include/persistent_data.hpp
gtmills: I wonder if we should get rid of the ifdef here.. upstream bmcweb has been doing that 

We would keep the ifdefs around 

```
#ifdef BMCWEB_ENABLE_IBM_MANAGEMENT_CONSOLE
        signals.add(SIGUSR1);
#endif
```


```
#ifdef BMCWEB_ENABLE_IBM_MANAGEMENT_CONSOLE
                if (signalNo == SIGUSR1)
                {
                    BMCWEB_LOG_CRITICAL


```

raviteja-b:
@gtmills closed this PR? any reason?

gtmills:
> @gtmills closed this PR? any reason?

Sorry, my bad

baemyung:
@nkotania 
This PR contains many of already-merged commits.
I think you may need to refresh your own 1110 branch (to ibm-openbmc/bmcweb/ 1110) forcefully and rebase this PR to the clean 1110, or rebase this PR against the latest ibm-openbmc/bmcweb 1110 branch.

baemyung (approved):
It looks good to me (besides Gunnar's comments).

PR #910, opened 2024-04-17
1110: Get rid of include_list to have all schemas (#662)

baemyung:
To do the same way as before, update_schemas.py will not require to specify 'include_list' so that all available schemas.

This will effectively make schema handling same as how it was done in the previous releases.

Upstream: https://gerrit.openbmc.org/c/openbmc/bmcweb/+/60220
          Currently it is WIP but would incorporate
          when it is finalized in the future.

This also adds another script change to pass CI to specify the explicit
schema array size in `schemas.hpp`.
Its upstream commit: https://gerrit.openbmc.org/c/openbmc/bmcweb/+/70908

Tested:  Redfish validator passed (except the known issues so far)


gtmills:
> Upstream: https://gerrit.openbmc.org/c/openbmc/bmcweb/+/60220
> Currently it is WIP but would incorporate
> when it is finalized in the future.

I opened https://jsw.ibm.com/browse/PFEBMC-2665 for this work

gtmills (approved):

PR #911, opened 2024-04-18
1110: Get rid of include_list to have all schemas (#662) (#910)

rfrandse:
#### Get rid of include_list to have all schemas (#662) (#910)
```
To do the same way as before, update_schemas.py will not require
to specify 'include_list' so that all available schemas.

This will effectively make schema handling same as how it was done
in the previous releases.

Upstream: https://gerrit.openbmc.org/c/openbmc/bmcweb/+/60220
          Currently it is WIP but would incorporate
          when it is finalized in the future.

This also adds another script change to pass CI to specify the explicit
schema array size in `schemas.hpp`.
Its upstream commit: https://gerrit.openbmc.org/c/openbmc/bmcweb/+/70908

Tested:  Redfish validator passed (except the known issues so far)

Signed-off-by: Myung Bae <myungbae@us.ibm.com>```


baemyung (approved):
It is good.

PR #241, opened 2022-01-27
OEM - Enclosure Firmware Version

patelabhishek9893:
Mex is updating via a lid in the host image, updated via PHYP.

PHYP sends an FRU record set PDR and sends the io enclosure version
data via the frurecord table and pldm parses that and hosts the
Version Interface on the mex chassis object.

This implementation retrieves that mex chassis FirmwareVersion data,
which defined via the manufacturer for the associated chassis, and
displayed on Redfish OEM schema for chassis.

Test: Tested Using Jankins validator. All Oem related checks are failing.

PR #912, opened 2024-04-18
OEM - Enclosure Firmware Version (#241)

baemyung:
* OEM - Enclosure Firmware Version

Mex is updating via a lid in the host image, updated via PHYP.

PHYP sends an FRU record set PDR and sends the io enclosure version data via the frurecord table, and pldm parses that and hosts the Version Interface on the mex chassis object.

This implementation retrieves that mex chassis FirmwareVersion data, which define via the manufacturer for the associated chassis, and displayed on Redfish OEM schema for chassis.


gtmills:
What do you think about moving this to standard Redfish since we have:

```
                "Version": {
                    "description": "The hardware version of this chassis.",
                    "longDescription": "This property shall contain the hardware version of this chassis as determined by the vendor or supplier.",
                    "readonly": true,
                    "type": [
                        "string",
                        "null"
                    ],
                    "versionAdded": "v1_21_0"
                },

```

gtmills:
>  Can we simply replace Oem/FirmwareVersion to this Version (and tell WebGui team about it)?

I feel like we could just replace it for 1110

baemyung:
@gtmills 

It looks like the standard Chassis Version seems implying "hardware version".
```
                "Version": {
                    "description": "The hardware version of this chassis.",
                    "longDescription": "This property shall contain the hardware version of this chassis as determined by the vendor or supplier.",
```

On the other hand, the current implementation seems as "FirmwareVersion" - https://github.com/ibm-openbmc/phosphor-dbus-interfaces/blob/959ed7eefc160f65e7f78b4d94511fce6212e860/yaml/xyz/openbmc_project/Software/Version.interface.yaml#L2

Can we treat it as Chassis's Version?

gtmills:
Added this to the agenda for Monday

gtmills (approved):
I am good if we end up going this direction but a candidate to drop OEM

PR #914, opened 2024-04-19
Ignore if no ACF date

gtmills:
As we have done in the past, ignore the error and move on if we can't get non-critical property, in this case, the ACF date.

This was bad we were clearing the response, remove that but go a step further and not throw an internalError.

This helps with 598833. Now the GUI logs in with an expired certificate.

An alternative change was https://github.com/ibm-openbmc/phosphor-certificate-manager/commit/66cbd8d4faea7dee870182dc4a16bf05c6aaeecb but the verify in ibm-acf was also deleting the expire date (code in this area)
https://github.com/ibm-openbmc/ibm-acf/blob/028f49460c04d94c29b67735a63e6f859aab8fac/subprojects/ce-login/celogin/src/CeLoginV2.cpp#L196

I decided this change is much safer 1. I understand this code and 2) this isn't acf validation or anything like that.. 

Tested:
```
  "Oem": {
    "IBM": {
      "@odata.type": "#OemManagerAccount.v1_0_0.IBM",
      "ACF": {
        "@odata.type": "#OemManagerAccount.v1_0_0.ACF",
        "ACFFile": "",
        "ExpirationDate": "",
        "WarningLongDatedExpiration": null
      }
    }
  },
  ```

### DIFF
--- a/redfish-core/lib/account_service.hpp
+++ b/redfish-core/lib/account_service.hpp
@@ -1122,8 +1122,6 @@ inline void getAcfProperties(
         if (expirationDateCpy.length() != 10)
         {
             BMCWEB_LOG_ERROR << "expirationDate format invalid";
-            asyncResp->res = {};
-            messages::internalError(asyncResp->res);
             return;
         }
         while ((pos = expirationDateCpy.find(delimiter)) != std::string::npos)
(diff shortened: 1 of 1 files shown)


baemyung (approved):
I think it is reasonable.

PR #915, opened 2024-04-19
1110: Use shared_ptr for request body (#871)

baemyung:

```
struct Request
{
   boost::beast::http::request<boost::beast::http::string_body> req;
...
}
```

This may exhibit a few issues.

- If it is moved to a lambda of user authentication async method call, the original `completeRequest()` may observe the invalid content of it [1]. This is causing a problem for audit recording, esp on PATCH..

```
            [req{std::move(req)}, asyncResp, &rule, params](
```

- It may be copied (e.g. captured by value) and causing the excessive memory consumption [2], in case `req` is large like the case of code update.

```
        afterIfMatchRequest, std::ref(app), asyncResp, req, ifMatch));
```

To address those issues, the content of`req` would need to be valid until all references are done, and also the copy of `req` would need to be less overhead.

So, it can be changed as `shared_ptr` and captured by value for lambda.

```
   std::shared_ptr<boost::beast::http::request<boost::beast::http::string_body>> reqPtr;
```

```
-            [req{std::move(req)}, asyncResp, &rule, params](
+            [req, asyncResp, &rule, params](

```

This is temporary solution like we had in 1050/1060. We will drive the real solution upstream and maybe pull that back in to 1110.

Tested:
- Audit runs correctly
- Debug stmt shows the content correct at `completeRequest()` on PATCH For example,
```
```
- Validator passes

[1] https://github.com/ibm-openbmc/bmcweb/blob/ef23e48288b35d30baaf57b61299597ca26b1930/http/routing.hpp#L1462C1-L1463C61
[2] https://github.com/ibm-openbmc/bmcweb/blob/ef23e48288b35d30baaf57b61299597ca26b1930/redfish-core/include/query.hpp#L97C1-L98C71

Change-Id: I40a5cc224c1d12e6ee65c99963b9bc6b46379226

### DIFF
--- a/http/http2_connection.hpp
+++ b/http/http2_connection.hpp
@@ -35,7 +35,7 @@ namespace crow
 
 struct Http2StreamData
 {
-    crow::Request req{};
+    crow::Request req;
     crow::Response res{};
     size_t sentSofar = 0;
 };
--- a/http/http_request.hpp
+++ b/http/http_request.hpp
@@ -10,6 +10,7 @@
 #include <boost/beast/websocket.hpp>
 #include <boost/url/url.hpp>
 
+#include <memory>
 #include <string>
 #include <string_view>
 #include <system_error>
@@ -19,7 +20,10 @@ namespace crow
 
 struct Request
 {
-    boost::beast::http::request<boost::beast::http::string_body> req;
+    using http_request_body =
+        boost::beast::http::request<boost::beast::http::string_body>;
+    std::shared_ptr<http_request_body> reqPtr;
+    http_request_body& req;
 
   private:
     boost::urls::url urlBase{};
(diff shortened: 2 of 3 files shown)


baemyung:
Upstream: https://gerrit.openbmc.org/c/openbmc/bmcweb/+/70355

This PR  is an intermediate approach -- until upstream may redo with the shared_ptr<Request>.

jeaaustx (approved):

gtmills:
This fixes the html pretty? And the Audit log?

baemyung:
> This fixes the html pretty? And the Audit log?

This makes Audit log to work

gtmills (approved):

gtmills:
This is temporary solution like we had in 1050/1060. We will drive the real solution upstream and maybe pull that back in to 1110.

PR #916, opened 2024-04-19
Leave off USB code update if no backend

gtmills:
This appears what we had in 1020-1060.

Tested: Enable USB code update. Load on rainier with no backend. 200 rc for manager resource.


baemyung (approved):
It looks good to me

PR #488, opened 2022-12-20
Added Hypervisor Serial Socket

sbteeks:
This commit adds `wss://${window.location.host}/console1` WebSocket into bmcweb. it retrieves 'PHYP debug shell' info and displays WebSocket data into WebUI using SOL (Serial over LAN) console.

console1 reads data from UNIX domain socket '@obmc-console.hypervisor' and display it in WebUI.

Privilege is set to {"ConfigureComponents", "ConfigureManager"}.

Tested:
  Used webUI to open Hypervisor console. Console window was displayed.
  However, we are not getting the shell because the BMC goes to
  Quiecsed state and the Host does not move to 'Running' state.
  There is still some level of missing support on 1050-ghe, and this
  needs testing once we have all the missing support is enabled.

Signed-off-by: Shantappa Teekappanavar <shantappa.teekappanavar@ibm.com>

PR #917, opened 2024-04-19
Added Hypervisor Serial Socket (#488)

baemyung:
This commit adds `wss://${window.location.host}/console1` WebSocket into bmcweb. it retrieves 'PHYP debug shell' info and displays WebSocket data into WebUI using SOL (Serial over LAN) console.

console1 reads data from UNIX domain socket '@obmc-console.hypervisor' and display it in WebUI.

Privilege is set to {"OemIBMPerformService"}.

Tested:
  Used webUI to open Hypervisor console. Console window was displayed.
  However, we are not getting the shell because the BMC goes to
  Quiecsed state and the Host does not move to 'Running' state.
  There is still some level of missing support on 1110-ghe, and this
  needs testing once we have all the missing support is enabled.

### DIFF
--- a/include/obmc_hypervisor.hpp
+++ b/include/obmc_hypervisor.hpp
@@ -1,0 +1,165 @@
+#pragma once
+
+#include "app.hpp"
+#include "async_resp.hpp"
+#include "websocket.hpp"
+
+#include <sys/socket.h>
+
+#include <boost/asio/local/stream_protocol.hpp>
+#include <boost/container/flat_map.hpp>
+#include <boost/container/flat_set.hpp>
+#include <boost/system/error_code.hpp>
+
+#include <array>
+#include <memory>
+#include <string>
+#include <string_view>
+
+namespace crow
+{
+namespace obmc_hypervisor
+{
+
+// NOLINTBEGIN(cppcoreguidelines-avoid-non-const-global-variables)
+static std::unique_ptr<boost::asio::local::stream_protocol::socket> hostSocket;
+
+static std::array<char, 4096> outputBuffer;
+static std::string inputBuffer;
+
+static boost::container::flat_set<crow::websocket::Connection*> sessions;
+
+static bool doingWrite = false;
+// NOLINTEND(cppcoreguidelines-avoid-non-const-global-variables)
+
+inline void doWrite()
+{
+    if (doingWrite)
+    {
+        BMCWEB_LOG_DEBUG("Already writing.  Bailing out");
+        return;
+    }
+
+    if (inputBuffer.empty())
+    {
+        BMCWEB_LOG_DEBUG("Outbuffer empty.  Bailing out");
+        return;
+    }
+
+    if (!hostSocket)
+    {
+        BMCWEB_LOG_ERROR("doWrite(): Socket closed.");
+        return;
+    }
+
+    doingWrite = true;
+    hostSocket->async_write_some(
+        boost::asio::buffer(inputBuffer.data(), inputBuffer.size()),
+        [](const boost::beast::error_code& ec, std::size_t bytesWritten) {
+        doingWrite = false;
+        inputBuffer.erase(0, bytesWritten);
+
+        if (ec == boost::asio::error::eof)
+        {
+            for (crow::websocket::Connection* session : sessions)
+            {
+                session->close("Error in reading to host port");
+            }
+            return;
+        }
+        if (ec)
+        {
+            BMCWEB_LOG_ERROR("Error in host serial write {}", ec);
+            return;
+        }
+        doWrite();
+    });
+}
+
+inline void doRead()
+{
+    if (!hostSocket)
+    {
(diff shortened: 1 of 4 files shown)


baemyung:
The tests are needed once 1110-ghe supports Hypervisor console.

include/obmc_hypervisor.hpp
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    doingWrite = true;
+    hostSocket->async_write_some(
+        boost::asio::buffer(inputBuffer.data(), inputBuffer.size()),
+        [](const boost::beast::error_code& ec, std::size_t bytesWritten) {
+        doingWrite = false;
+        inputBuffer.erase(0, bytesWritten);
+
+        if (ec == boost::asio::error::eof)
```
jeaaustx, line 58: I think this should be checking against ec.value().

include/obmc_hypervisor.hpp
```diff
@@ -0,0 +1,161 @@
...
+        {
+            for (crow::websocket::Connection* session : sessions)
+            {
+                session->close("Error in reading to host port");
+            }
+            return;
+        }
+        if (ec)
+        {
+            BMCWEB_LOG_ERROR("Error in host serial write {}", ec);
```
jeaaustx, line 68: Use ec.value() here.

include/obmc_hypervisor.hpp
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    BMCWEB_LOG_DEBUG("Reading from socket");
+    hostSocket->async_read_some(
+        boost::asio::buffer(outputBuffer.data(), outputBuffer.size()),
+        [](const boost::system::error_code& ec, std::size_t bytesRead) {
+        BMCWEB_LOG_DEBUG("read done.  Read {} bytes", bytesRead);
+        if (ec)
+        {
+            BMCWEB_LOG_ERROR("Couldn't read from host serial port: {}", ec);
```
jeaaustx, line 90: Should be using ec.value() here.

include/obmc_hypervisor.hpp
```diff
@@ -0,0 +1,161 @@
...
+        }
+        doRead();
+    });
+}
+
+inline void connectHandler(const boost::system::error_code& ec)
+{
+    if (ec)
+    {
+        BMCWEB_LOG_ERROR("Couldn't connect to host serial port: {}", ec);
```
jeaaustx, line 110: Should be using ec.value() here.

include/obmc_hypervisor.hpp
baemyung: I think - the current comparison (e.g. `if(ec.value() == EBADR)`) is actually semantically incorrect.

- `error_code` is a structure with `value` & `category`, and the value may potentially the same if the categories are different.
- In current bmcweb code (and most cases), it assumes they are in `system_category` and thus the comparison with `ec.value()` may be valid.

The code `if (ec == boost::asio::error::eof)` is translated to `if(ec == std::make_error_condition(boost::asio::error::eof)` and it is doing the comparison correctly.

https://www.boost.org/doc/libs/1_84_0/boost/system/detail/error_code.hpp
```
    template<class E, class N = typename detail::enable_if<std::is_error_condition_enum<E>::value>::type>
    inline friend bool operator==( error_code const & lhs, E rhs ) BOOST_NOEXCEPT
    {
        return lhs == make_error_condition( rhs );
    }
```

In case the rhs value is for system error (e.g. EBADR), the semantically-correct way would be

`if(ec.value() == EBADR)`
-->
`if (ec == boost::system::error_code{EBADR, boost::system::system_category()})`


Ideally, the use of the direct comparison of 'ec.value()` may need to be avoided unless the error category is known.

include/obmc_hypervisor.hpp
baemyung: I left this as is -- because the ec value here is not coming from dbus, but socket io which may have the different meaning of the value from dbus cases.

meson_options.txt
```diff
@@ -402,4 +402,12 @@ option('ibm-led-extensions',
...
     description : '''Enable the IBM LED extensions such as lamp test and system
                      attention indicators.'''
-)
\ No newline at end of file
+)
+
+option(
+      'hypervisor-serial-socket',
+      type: 'feature',
+      value: 'disabled',
```
gtmills, line 410: Can we go enabled here ?

baemyung: Enabled by default

gtmills (commented):
We aren't able to leverage https://github.com/openbmc/bmcweb/commit/052bcbf48802da1fa9583c8c0990378304e29903

include/obmc_hypervisor.hpp
jeaaustx: Thank you for the explanation. I see what you mean and agree with you.

gtmills (approved):
What we had before, seems reasonable

PR #921, opened 2024-04-22
Fix NTP set race condition

Pavithrab7:
There's currently a problem with phosphor-timesyncd, where enabling NTP doesn't immediately reflect in the system status on return[1].  To say it another way, NTP is not enabled/disabled atomically, which leads to the following problem.

// Disable NTP
PATCH /redfish/v1/Managers/bmc/NetworkProtocol
{"NTP":{"ProtocolEnabled": false}}
// Set the time manually
PATCH /redfish/v1/Managers/bmc {"DateTime": "<timestring"}

Doing this in rapid succession leads to a 500 error, which is obviously a bug.  In the prior commit, this error was changed to a PropertyValueConflict error, which is still incorrect, but at least informative of what's going on.  REST APIs are intended to have CRUD compliance.  The response should not be returned until the value has been accepted, and not doing that can lead to problems.

This commit changes the backend to use systemd directly, rather than routing through phosphor-settings, to avoid this race.

Quite possibly resolves #264 but haven't tested that.

Tested: The above procedure succeeds.

[1] https://github.com/systemd/systemd/pull/11424

Change-Id: Ie5777e95c4b43e823bcd67fcb8021feb1df64288
Signed-off-by: Pavithra Barithaya <pavithrabarithaya07@gmail.com>

### DIFF
--- a/redfish-core/lib/network_protocol.hpp
+++ b/redfish-core/lib/network_protocol.hpp
@@ -226,31 +226,29 @@ inline void getNetworkData(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
                          std::bind_front(afterNetworkPortRequest, asyncResp));
 } // namespace redfish
 
-inline void handleNTPProtocolEnabled(
-    const bool& ntpEnabled, const std::shared_ptr<bmcweb::AsyncResp>& asyncResp)
+inline void afterSetNTP(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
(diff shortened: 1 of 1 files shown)


gtmills:
add to approvelist

gtmills:
@Pavithrab7 https://gerrit.openbmc.org/c/openbmc/bmcweb/+/69948 is the upstream ?

gtmills (approved):
I am good here, just needs to pass CI

PR #922, opened 2024-04-22
Fix NTP set race condition

Pavithrab7:
There's currently a problem with phosphor-timesyncd, where enabling NTP doesn't immediately reflect in the system status on return[1].  To say it another way, NTP is not enabled/disabled atomically, which leads to the following problem.

// Disable NTP
PATCH /redfish/v1/Managers/bmc/NetworkProtocol
{"NTP":{"ProtocolEnabled": false}}
// Set the time manually
PATCH /redfish/v1/Managers/bmc {"DateTime": "<timestring"}

Doing this in rapid succession leads to a 500 error, which is obviously a bug.  In the prior commit, this error was changed to a PropertyValueConflict error, which is still incorrect, but at least informative of what's going on.  REST APIs are intended to have CRUD compliance.  The response should not be returned until the value has been accepted, and not doing that can lead to problems.

This commit changes the backend to use systemd directly, rather than routing through phosphor-settings, to avoid this race.

Quite possibly resolves #264 but haven't tested that.

Tested: The above procedure succeeds.

[1] https://github.com/systemd/systemd/pull/11424

Change-Id: Id53af9bf05bec052dd8013ae10c3c5e55f033038
Signed-off-by: Pavithra Barithaya <pavithrabarithaya07@gmail.com>

### DIFF
--- a/redfish-core/lib/network_protocol.hpp
+++ b/redfish-core/lib/network_protocol.hpp
@@ -238,30 +238,29 @@ inline void getNetworkData(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
                          std::bind_front(afterNetworkPortRequest, asyncResp));
 } // namespace redfish
 
-inline void handleNTPProtocolEnabled(
-    const bool& ntpEnabled, const std::shared_ptr<bmcweb::AsyncResp>& asyncResp)
+inline void afterSetNTP(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
+                        const boost::system::error_code& ec)
 {
-    std::string timeSyncMethod;
-    if (ntpEnabled)
+    if (ec)
     {
-        timeSyncMethod = "xyz.openbmc_project.Time.Synchronization.Method.NTP";
+        BMCWEB_LOG_ERROR("Failed to set elapsed time. DBUS response error {}",
+                         ec);
+        messages::internalError(asyncResp->res);
+        return;
     }
-    else
-    {
-        timeSyncMethod =
-            "xyz.openbmc_project.Time.Synchronization.Method.Manual";
-    }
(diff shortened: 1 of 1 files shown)


gtmills:
add to approvelist

Pavithrab7:
yes that's the upstream commit.

gtmills (approved):

redfish-core/lib/network_protocol.hpp
```diff
@@ -238,30 +238,29 @@ inline void getNetworkData(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
...
-    if (ntpEnabled)
-    {
-        timeSyncMethod = "xyz.openbmc_project.Time.Synchronization.Method.NTP";
-    }
-    else
+    if (ec)
     {
-        timeSyncMethod =
-            "xyz.openbmc_project.Time.Synchronization.Method.Manual";
+        BMCWEB_LOG_DEBUG("Failed to set elapsed time. DBUS response error {}",
```
baemyung, line 246: Would this is `BMCWEB_LOG_ERROR()`?

Pavithrab7: `BMCWEB_LOG_DEBUG() `would also be fine? - https://gerrit.openbmc.org/c/openbmc/bmcweb/+/69948/23/redfish-core/lib/network_protocol.hpp#246 ?

baemyung: @Pavithrab7 
I think - it needs to be updated to ERROR for 1110 and also for upstream, as it generates `internalError()`.

Pavithrab7: updated to ERROR for 1110.

PR #511, opened 2023-01-20
Remove VLAN support from Redfish interface

cm-jishnu:
This PR is part of the 1050 rebase.

It has the changes from the following commit in 1030
https://github.com/ibm-openbmc/bmcweb/commit/dc54f539f30a3e31b7598a7a36731bf44d32d5c4

This commit removes the VLAN related code from the bmcweb

PR #927, opened 2024-04-23
Remove VLAN support from Redfish interface (#511)

isha309:

Tested by:
 1. Verified VLAN resource is not listed while GET on the eth interfaces

### DIFF
--- a/redfish-core/lib/ethernet.hpp
+++ b/redfish-core/lib/ethernet.hpp
@@ -92,7 +92,6 @@ struct EthernetInterfaceData
     std::string defaultGateway;
     std::string ipv6DefaultGateway;
     std::string macAddress;
-    std::optional<uint32_t> vlanId;
     std::vector<std::string> nameServers;
     std::vector<std::string> staticNameServers;
     std::vector<std::string> domainnames;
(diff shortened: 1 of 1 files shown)


redfish-core/lib/ethernet.hpp
```diff
@@ -2115,27 +2079,6 @@ inline void requestEthernetInterfacesRoutes(App& app)
             }
             });
         });
-
-    BMCWEB_ROUTE(app, "/redfish/v1/Managers/bmc/EthernetInterfaces/<str>/")
```
gtmills, line 2119: This seems beyond VLAN.. We don't allow deleting ethernet interfaces?

raviteja-b: @isha309 can you please check why this route is getting deleted?

raviteja-b:
> Can we open a story here to drive something upstream? E.g. I assume upstream is some setting that you don't support VLAN

don't need to drive this upstream since other companies may be interested in this VLAN support.
for IBM systems, we have decided to not to support VLAN so this commit removes all VLAN resources

redfish-core/lib/ethernet.hpp
isha309: on 1060 as well on patch and get on eth0/eth1 is supported 
https://github.com/ibm-openbmc/bmcweb/blob/1060/redfish-core/lib/ethernet.hpp

raviteja-b (approved):

asmithakarun:
> > Can we open a story here to drive something upstream? E.g. I assume upstream is some setting that you don't support VLAN
> 
> don't need to drive this upstream since other companies may be interested in this VLAN support. for IBM systems, we have decided to not to support VLAN so this commit removes all VLAN resources

I agree

asmithakarun (approved):

gtmills:
> don't need to drive this upstream since other companies may be interested in this VLAN support.
for IBM systems, we have decided to not to support VLAN so this commit removes all VLAN resources

We would put a property out on D-Bus, that could be read to determine if you support VLAN support similar to what Andrew did for AllowedHostTransitions https://gerrit.openbmc.org/c/openbmc/bmcweb/+/69705

gtmills (approved):

PR #929, opened 2024-04-23
1030: Ignore if no ACF date

gtmills:
Like we have done in the past, ignore the error and just move on if can't get non-critical property, in this case the acf date.

This was bad we were clearing the response, remove that but go a step further and not throw an internalError.

This helps with 598833. Now the GUI logs in with an expired certificate.

Tested:
  "Oem": {
    "IBM": {
      "@odata.type": "#OemManagerAccount.v1_0_0.IBM",
      "ACF": {
        "@odata.type": "#OemManagerAccount.v1_0_0.ACF",
        "ACFFile": "",
        "ExpirationDate": "",
        "WarningLongDatedExpiration": null
      }
    }
  },

### DIFF
--- a/redfish-core/lib/account_service.hpp
+++ b/redfish-core/lib/account_service.hpp
@@ -1142,8 +1142,6 @@ inline void getAcfProperties(
         if (expirationDateCpy.length() != 10)
         {
             BMCWEB_LOG_ERROR << "expirationDate format invalid";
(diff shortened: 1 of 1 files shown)


baemyung (approved):
It looks good to me.

joseph-reynolds (commented):
Looks good to me.

PR #936, opened 2024-04-24
1110: Add BMC_SHELL web-base_shell (#483)(#473)

baemyung:

Path of WebSocket:
    wss://{hostname}/bmc-console

Implementation:

    like ls, cd, etc.

Authorize user for Shell :

Test:

    The essential operation can perform on this xterm console, Ex: ls,
    cd, grep, etc.

### DIFF
--- a/include/obmc_shell.hpp
+++ b/include/obmc_shell.hpp
@@ -1,0 +1,248 @@
+#pragma once
+
+#include "app.hpp"
+#include "logging.hpp"
+#include "websocket.hpp"
+
+#include <pty.h>
+#include <pwd.h>
+#include <sys/types.h>
+#include <sys/wait.h>
+#include <termios.h>
+
+#include <boost/asio/buffer.hpp>
+#include <boost/asio/posix/stream_descriptor.hpp>
+#include <boost/process/io.hpp>
+#include <boost/system/error_code.hpp>
+
+#include <array>
+#include <cstddef>
+#include <map>
+#include <memory>
+#include <string>
+#include <string_view>
+#include <variant>
+
+namespace crow
+{
+
+namespace obmc_shell
+{
+
+class Handler : public std::enable_shared_from_this<Handler>
+{
+  public:
+    explicit Handler(crow::websocket::Connection* conn) :
+        session(conn), streamFileDescriptor(conn->getIoContext())
+    {
+        outputBuffer.fill(0);
+    }
+
+    ~Handler() = default;
+
+    Handler(const Handler&) = delete;
+    Handler(Handler&&) = delete;
+    Handler& operator=(const Handler&) = delete;
+    Handler& operator=(Handler&&) = delete;
+
+    void doClose()
+    {
+        streamFileDescriptor.close();
+
+        // boost::process::child::terminate uses SIGKILL, need to send SIGTERM
+        int rc = kill(pid, SIGKILL);
+        session = nullptr;
+        if (rc != 0)
+        {
+            return;
+        }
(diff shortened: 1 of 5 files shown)


baemyung:
Will need to do the tests once GUI is available.

gtmills:
jenkins run tests please

meson_options.txt
```diff
@@ -389,6 +389,15 @@ option(
                     parameters such as only are not controlled by this option.'''
 )
 
+option(
+      'bmc-shell-socket',
+      type : 'feature',
+      value : 'disabled',
```
gtmills, line 395: let's enable this by default

baemyung: Changed to `enabled` by default

include/obmc_shell.hpp
```diff
@@ -0,0 +1,243 @@
...
+    {
+        if (session == nullptr)
+        {
+            BMCWEB_LOG_DEBUG("session is closed");
+            return;
+        }
+        streamFileDescriptor.async_read_some(
+            boost::asio::buffer(outputBuffer.data(), outputBuffer.size()),
+            [this, self(shared_from_this())](
+                const boost::system::error_code& ec, std::size_t bytesRead) {
```
gtmills, line 153: include error code

include/obmc_shell.hpp
```diff
@@ -0,0 +1,243 @@
...
+    std::string inputBuffer;
+
+  private:
+    crow::websocket::Connection* session;
+    boost::asio::posix::stream_descriptor streamFileDescriptor;
+    bool doingWrite{false};
+    int ttyFileDescriptor{0};
+    pid_t pid{0};
+
+    std::array<char, 4096> outputBuffer{};
```
gtmills, line 182: include array

include/obmc_shell.hpp
baemyung: Fixed

include/obmc_shell.hpp
baemyung: Fixed

include/obmc_shell.hpp
```diff
@@ -0,0 +1,248 @@
...
+            return;
+        }
+        waitpid(pid, nullptr, 0);
+    }
+
+    void connect()
+    {
+        pid = forkpty(&ttyFileDescriptor, nullptr, nullptr, nullptr);
+
+        if (pid == -1)
```
gtmills, line 66: We do something similar in  https://github.com/openbmc/phosphor-bmc-code-mgmt/blob/30352a66bed92bbc1b12a2468d3ccb5eea500004/image_manager.cpp#L282

include/obmc_shell.hpp
```diff
@@ -0,0 +1,248 @@
...
+{
+    auto it = mapHandler.find(&conn);
+    if (it != mapHandler.end())
+    {
+        it->second->inputBuffer += data;
+        it->second->doWrite();
+    }
+    else
+    {
+        BMCWEB_LOG_ERROR("connection to webScoket not found");
```
gtmills, line 233: socket

gtmills (approved):
It is what we had for 1020-1060

baemyung:
> It is what we had for 1020-1060

Yes, I also tested this with 1060-UI.  I'll request to merge.

(BTW, I'm separately investigating a failure related to Host Console/websocket under insufficient privilege)

include/obmc_shell.hpp
baemyung: Changed

PR #941, opened 2024-04-25
1110: Create Event Subscription meson build option

baemyung:
This is a PR of 1060 [c7592df96dbffe46702778375a0b67023c7ae428](https://github.com/ibm-openbmc/bmcweb/commit/c7592df96dbffe46702778375a0b67023c7ae428).

Create a Meson build option to enable and disable event subscription at /subscribe. 
By default, this is enabled.


gtmills:
Did we have a further commit to lock this down?

gtmills (approved):
It is what we had for 1020-1060

PR #950, opened 2024-04-29
1110:CI-only: create detect-secrets baseline file

geissonator:
This creates a baseline file (and initial responses to any potential secrets found).

After this file is merged, all CI jobs after will run the detect-secrets tool against incoming commits to verify no potential secrets are being shared. If one is found, CI will fail until the .secrets.baseline is updated for the new issue.

See the following for more info:
https://github.com/IBM/detect-secrets

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -1,0 +1,398 @@
+{
+  "exclude": {
+    "files": "^.secrets.baseline$",
+    "lines": null
+  },
+  "generated_at": "2024-04-29T18:24:07Z",
+  "plugins_used": [
+    {
+      "name": "AWSKeyDetector"
+    },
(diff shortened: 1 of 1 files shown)


gtmills (approved):
After this merges, should we rerun all commits in review to make sure they pass CI? I worry with the number of hits in bmcweb, some of our downstream only stuff is going to fail when we try to pull in and will need the exceptions

geissonator:
> After this merges, should we rerun all commits in review to make sure they pass CI? I worry with the number of hits in bmcweb, some of our downstream only stuff is going to fail when we try to pull in and will need the exceptions

I wouldn't worry about it. I can fix any fallout quickly if something hits.

PR #959, opened 2024-05-08
1110: redfish-core: Processor: Workaround to handle DCM

jeaaustx:
This is a rebase of a single commit from https://github.com/ibm-openbmc/bmcweb/pull/542:
https://github.com/ibm-openbmc/bmcweb/commit/a7a3704428676da0886f4b1bef30cbb78d8c41dd

Also changes `afterSetProcessorObject()` to call the new function `isProcObjectMatched()`. This is a rebase of the equivalent call made with https://github.com/ibm-openbmc/bmcweb/commit/4dbe5ec81f8e11ba5f5e8e942715d04842f40814

Cleanup changes were made to follow current upstream coding style.

redfish-core: Processor: Workaround to handle DCM

- The BMC inventory modeled processor object as "dcmN/cpuN" to support DCM (Dual-Chip Module) package (that contains two processors) based on system model.

- The Redfish does not support "Chip Module" (DCM or MCM (Multi-Chip Module)) but, due to inventory changes ("dcmN/cpuN") the existing Processor schema support won't work i.e we will see it as like below so we cannot locate the correct cpu object into inventory when want to pull it details or it subunits details.
```
  -> "@odata.id": "/redfish/v1/Systems/system/Processors/cpu0  <--
  -> "@odata.id": "/redfish/v1/Systems/system/Processors/cpu1
  -> "@odata.id": "/redfish/v1/Systems/system/Processors/cpu0  <--
  -> "@odata.id": "/redfish/v1/Systems/system/Processors/cpu1
```

- In this patch, added a workaround to support DCM in the Redfish model until getting some better way in Redfish to support DCM.
  - There is a DMTF forum thread to track this. https://redfishforum.com/thread/434/model-dual-chip-module

Tested:

- Tested in DCM system model (Rainer 2s2u)
```
curl -k -H "X-Auth-Token: $bmc_token" -X GET \
https://${bmc}/redfish/v1/Systems/system/Processors

{
  "@odata.id": "/redfish/v1/Systems/system/Processors",
  "@odata.type": "#ProcessorCollection.ProcessorCollection",
  "Members": [
    {
      "@odata.id": "/redfish/v1/Systems/system/Processors/dcm0-cpu0"
    },
    {
      "@odata.id": "/redfish/v1/Systems/system/Processors/dcm0-cpu1"
    },
    {
      "@odata.id": "/redfish/v1/Systems/system/Processors/dcm1-cpu0"
    },
    {
      "@odata.id": "/redfish/v1/Systems/system/Processors/dcm1-cpu1"
    }
  ],
  "Members@odata.count": 4,
  "Name": "Processor Collection"
}
```

- Tested in SCM system model (Witherspoon)
```
curl -k -H "X-Auth-Token: $bmc_token" -X GET \
https://${bmc}/redfish/v1/Systems/system/Processors
{
  "@odata.id": "/redfish/v1/Systems/system/Processors",
  "@odata.type": "#ProcessorCollection.ProcessorCollection",
  "Members": [
    {
      "@odata.id": "/redfish/v1/Systems/system/Processors/cpu0"
    },
    {
      "@odata.id": "/redfish/v1/Systems/system/Processors/cpu1"
    },
    {
      "@odata.id": "/redfish/v1/Systems/system/Processors/gv100card0"
    },
    {
      "@odata.id": "/redfish/v1/Systems/system/Processors/gv100card1"
    },
    {
      "@odata.id": "/redfish/v1/Systems/system/Processors/gv100card2"
    },
    {
      "@odata.id": "/redfish/v1/Systems/system/Processors/gv100card3"
    },
    {
      "@odata.id": "/redfish/v1/Systems/system/Processors/gv100card4"
    },
    {
      "@odata.id": "/redfish/v1/Systems/system/Processors/gv100card5"
    }
  ],
  "Members@odata.count": 8,
  "Name": "Processor Collection"
}
```

Signed-off-by: Ramesh Iyyar <rameshi1@in.ibm.com>
Change-Id: I45085491a439b598dde137d890d456c465a3e4d7

Conflicts:
  redfish-core/lib/processor.hpp

### DIFF
--- a/redfish-core/lib/processor.hpp
+++ b/redfish-core/lib/processor.hpp
@@ -49,6 +49,55 @@ constexpr std::array<std::string_view, 2> processorInterfaces = {
     "xyz.openbmc_project.Inventory.Item.Accelerator"};
 
 /**
+ * @brief Workaround to handle DCM (Dual-Chip Module) package for Redfish
+ *
+ * Make sure processor modeled as dual chip module ("dcmN-cpuN"),
+ * If yes then, replace Redfish processor id as "dcmN/cpuN" and check with
+ * given object path because Redfish does not support chip module concept.
+ *
+ * @param[in] processorId - The Redfish processor Id
+ * @param[in] objectPath  - The D-Bus object path that contain the processor
+ *                          segment
+ *
+ * @return true if matched with the given object path else false.
+ *
+ * @note Inventory modeled as "dcmN/cpuN" to support DCM so wherever using
+ *       Redfish processor id as "dcmN-cpuN" then this function (it support
+ *       both SCM and DCM) can be used for the inventory processor object
+ *       path validation.
+ */
+inline bool
+    isProcObjectMatched(const std::string& processorId,
+                        const sdbusplus::message::object_path& objectPath)
+{
+    bool isMatched = false;
+    if (processorId.find("dcm") != std::string::npos)
+    {
+        std::size_t delimiterPos = processorId.find('-');
+        if (delimiterPos != std::string::npos)
+        {
+            std::string procParent = processorId.substr(0, delimiterPos);
+            std::string procId = processorId.substr(delimiterPos + 1,
+                                                    processorId.length());
+
+            if ((objectPath.parent_path().filename() == procParent) &&
+                (objectPath.filename() == procId))
+            {
+                isMatched = true;
+            }
+        }
+    }
+    else
+    {
+        if (objectPath.filename() == processorId)
+        {
+            isMatched = true;
+        }
+    }
+    return isMatched;
+}
+
+/**
  * @brief Fill out uuid info of a processor by
  * requesting data from the given D-Bus object.
  *
@@ -805,9 +854,8 @@ inline void getProcessorObject(const std::shared_ptr<bmcweb::AsyncResp>& resp,
         for (const auto& [objectPath, serviceMap] : subtree)
         {
             // Ignore any objects which don't end with our desired cpu name
-            sdbusplus::message::object_path path(objectPath);
-            std::string name = path.filename();
-            if (name != processorId)
+            if (!isProcObjectMatched(
+                    processorId, sdbusplus::message::object_path(objectPath)))
             {
                 continue;
             }
@@ -1229,9 +1277,8 @@ inline void afterSetProcessorObject(
     for (const auto& [objectPath, serviceMap] : subtree)
     {
         // Ignore any objects which don't end with our desired cpu name
-        sdbusplus::message::object_path path(objectPath);
-        std::string name = path.filename();
-        if (name != processorId)
+        if (!isProcObjectMatched(processorId,
+                                 sdbusplus::message::object_path(objectPath)))
         {
             continue;
         }
@@ -1337,7 +1384,8 @@ inline void requestRoutesOperatingConfigCollection(App& app)
 
             for (const std::string& object : objects)
             {
-                if (!object.ends_with(cpuName))
+                if (!isProcObjectMatched(
+                        cpuName, sdbusplus::message::object_path(object)))
                 {
                     continue;
                 }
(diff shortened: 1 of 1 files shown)


redfish-core/lib/processor.hpp
```diff
@@ -48,6 +48,55 @@ constexpr std::array<std::string_view, 2> processorInterfaces = {
...
+{
+    bool isMatched = false;
+    if (processorId.find("dcm") != std::string::npos)
+    {
+        std::size_t delimiterPos = processorId.find('-');
+        if (delimiterPos != std::string::npos)
+        {
+            std::string procParent = processorId.substr(0, delimiterPos);
+            std::string procId = processorId.substr(delimiterPos + 1,
+                                                    processorId.length());
```
baemyung, line 81: I think this is tokenizing processorId by delimitting '-'.
It can be done using `bmcweb::split()` (which could be more readable).

```
     std::vector<std::string> substrs;
     bmcweb::split(substrs, processorId, '-');

     if(substrs.size() == 2)
     {
               const std::string& procParent = substrs[0];
              const std::string& procId = substrs[1];
             ...
```

redfish-core/lib/processor.hpp
```diff
@@ -1436,6 +1490,60 @@ inline void requestRoutesOperatingConfig(App& app)
...
+            leaf = path.filename();
+        }
+
+        if (leaf.empty())
+        {
+            continue;
+        }
+        std::string newPath = "/redfish/v1/Systems/system/Processors";
+        newPath += '/';
+        newPath += leaf;
```
baemyung, line 1541: As it is URL, 
```
        std::string newPath = boost::urls::format("/redfish/v1/Systems/system/Processors/{}", leaf);
```

redfish-core/lib/processor.hpp
```diff
@@ -1436,6 +1490,60 @@ inline void requestRoutesOperatingConfig(App& app)
     });
 }
 
+inline void
+    handleProcessors(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
+                     const boost::system::error_code& ec,
+                     const std::vector<std::string>& objects)
```
baemyung, line 1496: As this is invoked by `getSubTree()`,
```
const MapperGetSubTreePathsResponse& object)
```

redfish-core/lib/processor.hpp
```diff
@@ -1436,6 +1490,60 @@ inline void requestRoutesOperatingConfig(App& app)
...
+{
+    if (ec)
+    {
+        BMCWEB_LOG_DEBUG("DBUS response error {}", ec.value());
```
baemyung, line 1500: `BMCWEB_LOG_ERROR`

redfish-core/lib/processor.hpp
jeaaustx: Do you mean boost::split() here?
I can't find a bmcweb::split().

redfish-core/lib/processor.hpp
jeaaustx: Missed that one, corrected it.

redfish-core/lib/processor.hpp
jeaaustx: Good point. I've also changed leaf above to use std::format() for clarity as well.

redfish-core/lib/processor.hpp
jeaaustx: Good catch, corrected.

redfish-core/lib/processor.hpp
jeaaustx: I agree that is slightly easier to read. However it is not exactly the same thing semantically. 

With the current code if there are multiple `'-'` characters in the `processorId `then it will be handled with the split always into two elements. It will then do the following check to see if the names match or not.

But using bmcweb::split() a similar string would be split into three or more elements. Then it would fall out to the `!isMatched` case unless extra handling was put in to check for that case. The extra handling would make this harder to read than the original code.

I don't think it is likely there would be multiple `'-'` characters, but I'm hesitant to change the semantics.

Is that acceptable to you?

baemyung: Wouldn't this be correct only if it forms  like `"dcmN-cpuN"` (with it starts with "dcm")?
Or, if  the logic needs to allow more than 2 "-", it can be handled without checking 'substrs.size() == 2".

I'm okay to keep the original logic.

baemyung (approved):
It looks good to me

gtmills:
https://jsw.ibm.com/browse/PFEBMC-2356 will be the upstream here

PR #962, opened 2024-05-08
Check expand uris & limit to 3 

gtmills:
1) Limit Query Expand to 3 
 
2) Only allow expand on three parts of the tree:
/redfish/v1/Chassis
/redfish/v1/Systems/system
/redfish/v1/Cables

 These URIs are where the GUI's sensor page and PCIe topology page uses expand

### DIFF
--- a/redfish-core/include/query.hpp
+++ b/redfish-core/include/query.hpp
@@ -126,8 +126,8 @@ inline bool handleIfMatch(crow::App& app, const crow::Request& req,
 
     asyncResp->res.addHeader("OData-Version", "4.0");
 
-    std::optional<query_param::Query> queryOpt =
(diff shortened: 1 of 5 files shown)


baemyung (approved):
It looks good to me.

joseph-reynolds (approved):
Seems okay to me.
The code to limit expand to <= 3 levels looks good.
The code to limit expand to specific URIs looks good.
When either condition is not met, the new code gives QueryParameterValueFormatError.
The limit of 3 expand levels is given to users in URI /redfish/v1 property ExpandQuery.MaxLevels.
Users have no way to know ExpandQuery is limited to 3 specific URIs (needed by the web application).  I suppose the Redfish spec does not give a way to specify this?

PR #966, opened 2024-05-09
Added SubordinateOverrides & Fix Log_services privileges (#559)

baemyung:
SubordinateOverrides:
  This commit attempts to automate the creation of SubordinateOverrides
  privileges structures from the redfish privilege registry. In
  addition, it enhances the function of parse_registries.py.

  It read SubordinateOverrides privilege registry from DMTF and
  generates const defines SubordinateOverrides for all the privilege
  registry entries in the same format that the Privileges struct
  accepts.

  Moreover, it generates unique const defines for all
  SubordinateOverrides target levels.
  Ex: EthernetInterface SubordinateOverrides has two "Targets":
      ["Manager", "EthernetInterfaceCollection"]. So
      parse_registries.py generates two unique const

      1) Subordinate override for Manager -> EthernetInterface
      2) Subordinate override for Manager ->
         EthernetInterfaceCollection ->  EthernetInterface

  Note: if SubordinateOverrides privilege gets changed, then it
  automatically updates that route privilege, but if
  SubordinateOverrides target gets change, then the user needs to
  update that manually.

Fix Log_services privileges:
  In Log_services, some of the privileges not following the
  Redfish_1.1.0_PrivilegeRegistry registry.

  privilege Change this commit contains
  Get method:
    1) /redfish/v1/Systems/system/LogServices/Crashdump/
        ConfigureManager -> Login

    2)/redfish/v1/Systems/system/LogServices/Crashdump/Entries/
         ConfigureManager -> Login

    3) /redfish/v1/Systems/system/LogServices/Crashdump/Entries/<str>/
        ConfigureComponents -> Login

    Impact: ConfigureManager -> Login    and
          ConfigureComponents -> Login

    This change allows Admin, Operator, and Readonly users to access
    Crashdump data and related entries.

Tested: manually tested on Witherspoon system, there is no change in
        output. Run Redfish validator, with all different Privileges;
        Error Get: UUID: String '' does not match pattern ''
        this commit doesn't affect UUID

Email sent to openbmc list:
https://lists.ozlabs.org/pipermail/openbmc/2021-August/027232.html

It is also being pushed to upstream -
https://gerrit.openbmc.org/c/openbmc/bmcweb/+/45125.

Change-Id: I37d8a2882f1cfaa59a482083f180fdd0805e2e7d

### DIFF
--- a/redfish-core/include/registries/privilege_registry.hpp
+++ b/redfish-core/include/registries/privilege_registry.hpp
@@ -254,6 +254,15 @@ const static auto& putCertificate = privilegeSetConfigureManager;
 const static auto& deleteCertificate = privilegeSetConfigureManager;
 const static auto& postCertificate = privilegeSetConfigureManager;
 
+// Subordinate override for ComputerSystem -> Certificate
+const static auto& getCertificateSubOverComputerSystem =privilegeSetConfigureComponents;
+const static auto& headCertificateSubOverComputerSystem =privilegeSetConfigureComponents;
+const static auto& patchCertificateSubOverComputerSystem =privilegeSetConfigureComponents;
+const static auto& putCertificateSubOverComputerSystem =privilegeSetConfigureComponents;
+const static auto& deleteCertificateSubOverComputerSystem =privilegeSetConfigureComponents;
+const static auto& postCertificateSubOverComputerSystem =privilegeSetConfigureComponents;
+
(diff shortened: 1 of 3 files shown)


gtmills:
We should really try to get this upstream

gtmills (approved):

PR #967, opened 2024-05-10
http_client: ssl context optimization

abhilashraju:
ssl context are recreated whenever new ssl stream is created. We can reuse the already created ssl context object for any new ssl streams. This will reduce the number of ssl context object creation in the negative retry scenarios.
We have seen libcrypto crashes in the retry forever scenario. Reducing the number of ssl context object creation  may fix such crashes.

Tested By:
Following unit tests has been executed.
1) Basic sanity tests on eventing
2) Negative stale subscription cases

Need to execute all system level cases convering HMC/BMC connetion issues before merging

### DIFF
--- a/http/http_client.hpp
+++ b/http/http_client.hpp
@@ -152,7 +152,7 @@ class ConnectionInfo : public std::enable_shared_from_this<ConnectionInfo>
     std::optional<boost::beast::ssl_stream<boost::asio::ip::tcp::socket&>>
         sslConn;
     boost::asio::ip::tcp::socket conn;
-
+    std::optional<boost::asio::ssl::context> sslCtx;
     boost::asio::steady_timer timer;
 
     friend class ConnectionPool;
@@ -213,8 +213,7 @@ class ConnectionInfo : public std::enable_shared_from_this<ConnectionInfo>
         timer.cancel();
         if (ec)
         {
-            BMCWEB_LOG_ERROR << "Connect " << endpoint.address().to_string()
-                             << " : " << std::to_string(endpoint.port())
+            BMCWEB_LOG_ERROR << "Connect " << host << " : " << port
                              << ", id: " << std::to_string(connId)
                              << " failed: " << ec.message();
             state = ConnState::connectFailed;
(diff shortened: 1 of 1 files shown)


http/http_client.hpp
```diff
@@ -468,7 +467,7 @@ class ConnectionInfo : public std::enable_shared_from_this<ConnectionInfo>
             }
             if (connPolicy->retryPolicyAction == "RetryForever")
             {
-                state = ConnState::idle;
+                state = ConnState::closed;
```
gtmills, line 470: Why closed over terminate or suspend?

baemyung:
Isn't it good to go thru upstream?

>We have seen libcrypto crashes in the retry forever scenario. Reducing the number of ssl context object creation may fix such crashes.

Is there a root cause analysis what causes the crash?
If not, how would this fix the crashes although the crashes may be less often?

http/http_client.hpp
raviteja-b: @abhilashraju have we tested this change in retry forever scenario?
looks like this case needs to be tested.

abhilashraju: @raviteja-b yes I have done retry test . Doesnot show any issues in dev testing.
@gtmills  the  callback(false, connId, res); at 476  will set the connection state to closed anyway. do we really need this line of code. I don't know the history. Can we remove this code to simplify ?

abhilashraju:
> Isn't it good to go thru upstream?
> 
> > We have seen libcrypto crashes in the retry forever scenario. Reducing the number of ssl context object creation may fix such crashes.
> 
> Is there a root cause analysis what causes the crash? If not, how would this fix the crashes although the crashes may be less often?

We could reproduce it only once. The callstack points to libcrypto. At the moment we can only speculate the root cause. As I said in the commitlog, it need not fix the issue. But still it is an optimisation we can do. Yes we can propose the same to upstream. But the context creation is done a bit differently in upstream. We have commented it out in downstream. I don't know the reason.

gtmills:
What is our plan here? 
Should we close and bring back if we think we need in 1060?

PR #970, opened 2024-05-13
Don't internal error if no slot location code (#683)

gtmills:
1110 update: 
Cherry-pick of 1b105f0357
Upstream is https://gerrit.openbmc.org/c/openbmc/bmcweb/+/62219 

1050 PR:
https://github.com/ibm-openbmc/bmcweb/pull/650 added PcieSlots Location Code but it throws an internal error when there is no location code interface. Location code isn't required and logical slots don't have a location code so just leave off location code if the interface isn't there. This is what we had for 1030.

Archana was seeing "Base.1.13.0.InternalError", at https://${bmc}/redfish/v1/Chassis/chassis15364/PCIeSlots

Tested: Loaded onto the system and didn't see the error anymore.


baemyung (approved):
It looks good to me

PR #971, opened 2024-05-13
1110 check expand uris

gtmills:
1.  Limit Query Expand to 3

2.  Only allow expand on three parts of the tree:
    /redfish/v1/Chassis
    /redfish/v1/Systems/system
    /redfish/v1/Cables

These URIs are where the GUI's sensor page and PCIe topology page uses expand

https://github.com/ibm-openbmc/bmcweb/pull/962 was the 1050/1060 PR


baemyung (approved):
It looks good to me

PR #974, opened 2024-05-14
1110: Add LocationIndicatorActive for fabric adapters (#630)

baemyung:
This update adds the GET and PATCH support for LocationIndicatorActive to fabric adapters. Only some fabric adapters support this option.

I manually tested this using the following commands:

```
curl -k -H "X-Auth-Token: $token" -X  GET \
     https://${bmc}/redfish/v1/Systems/system/FabricAdapters/chassis-motherboard-disk_backplane0
```

and

```
curl -k -H "Content-Type: application/json" -X  PATCH -d '{"LocationIndicatorActive":true}' \
     https://${bmc}/redfish/v1/Systems/system/FabricAdapters/chassis-motherboard-disk_backplane0
```

I first ran the above GET command, which showed that LocationIndicatorActive was false. I then ran the above PATCH command, which also succeeded. Finally I ran the GET command which succeeded and now showed that LocationIndicatorActive was true.

Change-Id: I02349b68bbc23ffd98b195bc399a25f94790bc55

### DIFF
--- a/redfish-core/lib/fabric_adapters.hpp
+++ b/redfish-core/lib/fabric_adapters.hpp
@@ -2,6 +2,7 @@
 
 #include "app.hpp"
 #include "dbus_utility.hpp"
+#include "led.hpp"
 #include "query.hpp"
 #include "registries/privilege_registry.hpp"
 #include "utils/collection.hpp"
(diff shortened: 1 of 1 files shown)


baemyung:
Upstream: https://gerrit.openbmc.org/c/openbmc/bmcweb/+/71430

gtmills (approved):

PR #976, opened 2024-05-17
Systems: Add OEM safe mode to user interface

cjcain:
Tested: Forced safe mode and verified dbus was updated

Change-Id: Ia2e23154454025ca7a8107e54f2de3883363490a

### DIFF
--- a/redfish-core/lib/systems.hpp
+++ b/redfish-core/lib/systems.hpp
@@ -1583,9 +1583,10 @@ inline void
 
     std::string powerMode;
     const std::vector<std::string>* allowedModes = nullptr;
+    bool safeMode;
     const bool success = sdbusplus::unpackPropertiesNoThrow(
         dbus_utils::UnpackErrorPrinter(), properties, "PowerMode", powerMode,
-        "AllowedPowerModes", allowedModes);
+        "AllowedPowerModes", allowedModes, "SafeMode", safeMode);
 
     if (!success)
     {
@@ -1625,6 +1626,12 @@ inline void
         return;
     }
     asyncResp->res.jsonValue["PowerMode"] = modeValue;
+
+    BMCWEB_LOG_DEBUG("Safe mode: {}", safeMode);
+    nlohmann::json& oemSafeMode = asyncResp->res.jsonValue["Oem"];
+    oemSafeMode["@odata.type"] = "#OemComputerSystem.Oem";
+    oemSafeMode["IBM"]["@odata.type"] = "#OemComputerSystem.IBM";
+    oemSafeMode["IBM"]["SafeMode"] = safeMode;
 }
 
 /**
--- a/redfish-core/lib/systems.hpp
+++ b/redfish-core/lib/systems.hpp
@@ -1583,7 +1583,7 @@ inline void
 
     std::string powerMode;
     const std::vector<std::string>* allowedModes = nullptr;
-    bool safeMode;
+    std::optional<bool> safeMode;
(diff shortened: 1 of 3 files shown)


static/redfish/v1/schema/OemComputerSystem_v1.xml
```diff
@@ -95,6 +95,11 @@
                   <Annotation Term="OData.Description" String="An indication of PEL topology information is saves."/>
                   <Annotation Term="OData.LongDescription" String="This property when set to 'true' saves the topology information. The information is saved in the form of a PEL(Error Log)."/>
                 </Property>
+                <Property Name="SafeMode" Type="OemServiceRoot.IBM">
```
baemyung, line 98: Type would need to be `Edm.Boolean` to satisfy Redfish Service Validator.
The correct one is on 1060 branch
https://github.com/ibm-openbmc/bmcweb/blob/e01a12ccef4be156f0ac3383e267b77b1cf29dda/static/redfish/v1/schema/OemComputerSystem_v1.xml#L88

redfish-core/lib/systems.hpp
```diff
@@ -1583,9 +1583,10 @@ inline void
 
     std::string powerMode;
     const std::vector<std::string>* allowedModes = nullptr;
+    bool safeMode;
```
baemyung, line 1586: Should this be defined as optional (to check the existence of the field)?
```
const bool* safeMode = nullptr;
```
OR
```
std::optional<bool> safeMode = nullptr;
```

redfish-core/lib/systems.hpp
```diff
@@ -1625,6 +1626,12 @@ inline void
         return;
     }
     asyncResp->res.jsonValue["PowerMode"] = modeValue;
+
+    BMCWEB_LOG_DEBUG("Safe mode: {}", safeMode);
+    nlohmann::json& oemSafeMode = asyncResp->res.jsonValue["Oem"];
+    oemSafeMode["@odata.type"] = "#OemComputerSystem.Oem";
+    oemSafeMode["IBM"]["@odata.type"] = "#OemComputerSystem.IBM";
+    oemSafeMode["IBM"]["SafeMode"] = safeMode;
```
baemyung, line 1634: If `safeMode` is defined as optional, this would be inside the check of it.
Also, the type would be better to be specific as in 1060 branch.

```
if(safeMode)
{
   BMCWEB_LOG_DEBUG("Safe mode: {}", *safeMode);
    nlohmann::json& oemSafeMode = asyncResp->res.jsonValue["Oem"];
    oemSafeMode["@odata.type"] = "#OemComputerSystem.Oem";
    oemSafeMode["IBM"]["@odata.type"] = "#OemComputerSystem.v1_0_0.IBM";   <--
    oemSafeMode["IBM"]["SafeMode"] = *safeMode;
}
```

redfish-core/lib/systems.hpp
gtmills: I'm not sure.. If it doesn't exist it hits the success and throws an error. I think that is what we had for 1060 
https://github.com/ibm-openbmc/bmcweb/blob/1060/redfish-core/lib/systems.hpp#L1547


```
    if (!success)
    {
        return;
    }
```

gtmills: Oh well, I am good with the optional and just leave off if it doesn't exist

gtmills (approved):

baemyung (approved):

PR #980, opened 2024-05-20
1110: Fix core via upstream cherry-picks

baemyung:
This is another way of handling coredump from https://github.com/ibm-openbmc/bmcweb/pull/979

This PR is to address the coredump related to client ssl connection.

It is caused at SSL stream and it has been fixed via cherry-picks of upstream commits.
     - https://github.com/openbmc/bmcweb/commit/003301a2

One way is to cherry-pick to sync-up to upstream.
This approach helps us the future activities with  upstream.

This is also confirmed by testing upstream code.

Test method:
- On GUI, `https://rain100bmc.aus.stglabs.ibm.com/#/operations/host-console`
- Login as `service` and put `https://rain100bmc.aus.stglabs.ibm.com/#/operations/host-console` again

---
Following upstream commits are included:

- https://github.com/openbmc/bmcweb/commit/891eaa7c Use rvalue reference
- https://github.com/openbmc/bmcweb/commit/89144a3a Remove a copy
- https://github.com/openbmc/bmcweb/commit/dc8cfa66 Add nullptr check
- https://github.com/openbmc/bmcweb/commit/4c521c3c Fix file removal
- https://github.com/openbmc/bmcweb/commit/c09966bd Give static analysis some hints
- https://github.com/openbmc/bmcweb/commit/44106f34 Fix buffer_copy
- https://github.com/openbmc/bmcweb/commit/4a7fbefd Fix large copies with url_view and segments_view
- https://github.com/openbmc/bmcweb/commit/d9e89dfd Simplify router
- https://github.com/openbmc/bmcweb/commit/76b038f2 Fix object type in json utils
- https://github.com/openbmc/bmcweb/commit/f1a1e3dc Simplify query_param
- https://github.com/openbmc/bmcweb/commit/2b9c1dfe Fix account service
- https://github.com/openbmc/bmcweb/commit/3cd7072b Move run and redfish to compile units
- https://github.com/openbmc/bmcweb/commit/1919a03c Fix minor error in Request::method
- https://github.com/openbmc/bmcweb/commit/df254f2c Remove redundant static modifier
- https://github.com/openbmc/bmcweb/commit/8099c517 Allow parsing null or value classes
- https://github.com/openbmc/bmcweb/commit/8e5cc7bd Fix http2 payloads
- https://github.com/openbmc/bmcweb/commit/296579be Call dump() less
- https://github.com/openbmc/bmcweb/commit/60e299bd meson: Disable tests build for tinyxml2 dependency
[...]

PR #979, opened 2024-05-20
1110: Fix core with upstream ssl fix

baemyung:
This PR is to address the coredump related to client ssl connection.

It is caused at SSL stream and it has been fixed via cherry-picks of upstream commits
     - https://github.com/openbmc/bmcweb/commit/003301a2
     - https://github.com/openbmc/bmcweb/commit/8db83747

This is also confirmed by trying upstream code.

Test method:
- On GUI, `https://rain100bmc.aus.stglabs.ibm.com/#/operations/host-console`
- Login as `service` and put `https://rain100bmc.aus.stglabs.ibm.com/#/operations/host-console` again

PR #982, opened 2024-05-21
1110: Check nullptr before using var at log-services

baemyung:
requestRoutesDBusCELogEntryCollection() handling hits a core due to a missing nullptr of variable `hidden` in log_services.hpp.
This is introduced by a mistake during https://github.com/ibm-openbmc/bmcweb/commit/a325bc8c07cac21c02df624298b4f3b94864e42f

```
Core was generated by `/usr/bin/bmcweb'.
Program terminated with signal SIGSEGV, Segmentation fault.
...
   std::shared_ptr<bmcweb::AsyncResp> const&, std::__cxx11::basic_string<char, std::char_traits<char>,
   std::allocator<char> > const&)#1}::operator()(crow::Request const&, std::shared_ptr<bmcweb::AsyncResp> const&, std::__cxx11::basic_string<char,
...
    std::shared_ptr<bmcweb::AsyncResp> const&, std::__cxx11::basic_string<char, std::char_traits<char>,
    std::allocator<char> > const&)#1}::operator()(crow::Request const&, std::shared_ptr<bmcweb::AsyncResp> const&,
   std::__cxx11::basic_string<char,
....
long> > > > > > > > > > > > > const&) const ()
at /usr/src/debug/bmcweb/1.0+git/redfish-core/lib/log_services.hpp:2618

2618	   if (!(*hidden))
(gdb)

```

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "^.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2024-05-08T21:40:58Z",
+  "generated_at": "2024-05-21T20:15:05Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -210,7 +210,7 @@
         "hashed_secret": "6e069300b0db722cc2ad4b1425a30ff5a8ae20cf",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 2282,
+        "line_number": 2356,
         "type": "Secret Keyword",
         "verified_result": null
       },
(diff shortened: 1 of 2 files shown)


spinler (approved):
The change is fine, though I'm worried how this even occurred as that interface/property should always be there.

baemyung:
> The change is fine, though I'm worried how this even occurred as that interface/property should always be there.
> 
> Maybe that's why I never saw it in my testing, because that property was always there at the time.

This can be seen after the other bmcweb core profile is fixed.

- Have the system with the fix of https://github.com/ibm-openbmc/bmcweb/pull/979 or https://github.com/ibm-openbmc/bmcweb/pull/980.
- Do reboot or other jobs.  I think this problem may occur this time as it attempts to see PEL and cause bmcweb to coredump.

gtmills (approved):

PR #983, opened 2024-05-23
RedfishEvents GHE: Enable TLS1.2 in http client

raviteja-b:
This is downstream only patch  to enable TLS1.2 support
Currently TLS1.3 is hardcoded in http client which restricts redfish events path to work with TLS1.3
and fails with TLS1.2

This commit enables TLS1.2 along with TLS1.3 in http_client and make sure redfish events work with HCM for both TLS1.2 and TLS1.3

Tested by:
Verified redfish events on HMC with TLS1.2 and TLS1.3

### DIFF
--- a/include/ssl_key_handler.hpp
+++ b/include/ssl_key_handler.hpp
@@ -494,6 +494,13 @@ inline std::optional<boost::asio::ssl::context> getSSLClientContext()
         return std::nullopt;
     }
 
+    // Currently remote server's certificate verfication fails for
+    // HMC provided self-signed certificates,
+    // so skip remote server's certificate verfication.
+
+#if 0 // TODO: certificate verfication can be enabled after supporting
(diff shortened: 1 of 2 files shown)


include/ssl_key_handler.hpp
```diff
@@ -494,6 +494,13 @@ inline std::optional<boost::asio::ssl::context> getSSLClientContext()
         return std::nullopt;
     }
 
+    // Currently remote server's certificate verfication fails for
+    // HMC provided self-signed certificates,
+    // so skip remote server's certificate verfication.
+
+#if 0 // TODO: certificate verfication can be enabled after supporting
```
abhilashraju, line 501: did you try  mSslContext->add_verify_path(trustStorePath); to check if the verification works before disabling this?

raviteja-b: yes @abhilashraju  I have tried, there is no difference

PR #984, opened 2024-05-24
1110: Revert "Add topology links for Chassis"

baemyung:
https://gerrit.openbmc.org/c/openbmc/bmcweb/+/60914 implements topology links for chassis using `getAssociationEndPoints()` for `containing/contained_by` associations.

However, the same associations can also be used for the other cases which may also be used for the other types.

For example, https://gerrit.openbmc.org/c/openbmc/openbmc/+/70372 adds the associations between chassis and the other resources.

In that case, Chassis Links gives the incorrect result including the non-chassis resources in `Contains` collection..

```
$ curl -k -X GET https://${bmc}/redfish/v1/Chassis/chassis
{
  "@odata.id": "/redfish/v1/Chassis/chassis",
  "@odata.type": "#Chassis.v1_22_0.Chassis",

  "Links": {
    "Contains": [
      ...
      {
        "@odata.id": "/redfish/v1/Chassis/connector4"
      },
...
```

This causes Redfish Service Validator failures.  So, we've decided
to review this commit https://github.com/ibm-openbmc/bmcweb/commit/a56174963e1d8bd9fc90cdeb3aa7f1b7f2e0dca4.

Tested:

- Check Chassi/Links collection to check non-chassis resources.
```
curl -k -X GET https://${bmc}/redfish/v1/Chassis/chassis
```

Change-Id: I472fc12379694acc35055965400141dbb1b33bfc

### DIFF
--- a/redfish-core/lib/chassis.hpp
+++ b/redfish-core/lib/chassis.hpp
@@ -204,103 +204,6 @@ inline void handleChassisCollectionGet(
         "/xyz/openbmc_project/inventory");
 }
 
-inline void getChassisContainedBy(
-    const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
-    const std::string& chassisId, const boost::system::error_code& ec,
-    const dbus::utility::MapperEndPoints& upstreamChassisPaths)
-{
-    if (ec)
-    {
-        if (ec.value() != EBADR)
-        {
-            BMCWEB_LOG_ERROR("DBUS response error {}", ec);
-            messages::internalError(asyncResp->res);
-        }
-        return;
-    }
-    if (upstreamChassisPaths.empty())
-    {
-        return;
-    }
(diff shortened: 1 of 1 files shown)


baemyung:
Upstream: https://gerrit.openbmc.org/c/openbmc/bmcweb/+/71612

baemyung:
I incorrectly read the goal of https://gerrit.openbmc.org/c/openbmc/bmcweb/+/60914

It is to show any resources that are contained by the chassis.
This may need to be done by fixing the association json file, or perhaps use `getSubTree()` to see the inner resources??

At this time, perhaps, the fix should be done at json association file - https://gerrit.openbmc.org/c/openbmc/openbmc/+/70372?

gtmills:
One choice would be just remove this..

jeaaustx (approved):

gtmills (approved):

PR #991, opened 2024-06-06
Consume availability interface for cores (#254)

baemyung:
In the current state, the logic is bmcweb for displaying cores is: It checks the Present property on core dbus object to update the [Status][State] as Enabled/Absent.

In addition to Present property, we should also use the Available property , if the Available is false, then set the [Status][State] should be set as "UnavailableOffline".

Refer: ibm-openbmc/dev#3481

### DIFF
--- a/redfish-core/lib/processor.hpp
+++ b/redfish-core/lib/processor.hpp
@@ -1075,6 +1075,7 @@ inline void afterGetCpuCoreDataByService(
 
         bool present = true;
         bool functional = true;
+        bool available = true;
 
         for (const auto& [interface, properties] : interfaces)
         {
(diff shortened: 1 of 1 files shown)


gtmills (approved):

gtmills:
This should be an easy one to get upstream

baemyung:
Upstream: https://gerrit.openbmc.org/c/openbmc/bmcweb/+/38570
After that, this may need to be added.

redfish-core/lib/processor.hpp
```diff
@@ -1111,18 +1112,37 @@ inline void afterGetCpuCoreDataByService(
                     }
                 }
             }
+            else if (interface ==
+                     "xyz.openbmc_project.State.Decorator.Availability")
+            {
+                for (const auto& [proName, proValue] : properties)
```
jeaaustx, line 1118: nit: Variable name was changed to "propName" and "propValue" for similar checks above. Nice if this used the same names.

jeaaustx (commented):
One minor nit comment.

redfish-core/lib/processor.hpp
baemyung: Good catch.
I've changed them.

jeaaustx (approved):

PR #992, opened 2024-06-06
Logging invalid service name for debug (#291) (#626)

baemyung:
The method getPrettyName() may be called more than one time.
This will generate journal logging messages for the required Service name
if there are unexpected service names.

### DIFF
--- a/redfish-core/include/utils/name_utils.hpp
+++ b/redfish-core/include/utils/name_utils.hpp
@@ -37,7 +37,11 @@ inline void getPrettyName(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
     // Ensure we only got one service back
     if (services.size() != 1)
     {
-        BMCWEB_LOG_ERROR("Invalid Service Size  {}", services.size());
+        BMCWEB_LOG_ERROR("Invalid Service Size {}", services.size());
+        for (const auto& service : services)
+        {
+            BMCWEB_LOG_ERROR("Invalid Service Name: {}", service.first);
+        }
         if (asyncResp)
         {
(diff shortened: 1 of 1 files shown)


baemyung:
Upstream: https://gerrit.openbmc.org/c/openbmc/bmcweb/+/43967
However, the community decided to abandon.

---

I think the backend team is pursuing another attempt to upstream getPrettyName??

gtmills (approved):

redfish-core/include/utils/name_utils.hpp
```diff
@@ -37,7 +37,11 @@ inline void getPrettyName(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
     // Ensure we only got one service back
     if (services.size() != 1)
     {
-        BMCWEB_LOG_ERROR("Invalid Service Size  {}", services.size());
+        BMCWEB_LOG_ERROR("Invalid Service Size {}", services.size());
+        for (const auto& service : services)
+        {
+            BMCWEB_LOG_ERROR("Invalid Service Name: {}", service.first);
```
jeaaustx, line 43: Would this be better as BMCWEB_LOG_DEBUG()? Seems like it could flood the log if we have a coding problem that is calling this with multiple services regularly.

baemyung: This section is a part `internalError()` [which is enabled if `asyncResp` is defined].
So, it makes sense to show it as an error.   (I think it was also the original intention )

jeaaustx: Agreed that makes sense.

jeaaustx (approved):

PR #993, opened 2024-06-06
RedfishEvents: Disable peer certificate verification

raviteja-b:
This is downstream only commit to disable peer certificate verification in redfish events path.

Currently http_client code verifies ssl certificate provided by HMC during SSL handshake
certificate validation fails with unknown CA, since BMC can't trust HMC provided certificates
HMC-BMC certificate flows needs to be enabled to pass this certificate validations.
 
 This commit disable peer certificate verification during redfish events ssl handshake

Tested by:
Verified redfish events on HMC

### DIFF
--- a/include/ssl_key_handler.hpp
+++ b/include/ssl_key_handler.hpp
@@ -540,9 +540,16 @@ inline std::optional<boost::asio::ssl::context> getSSLClientContext()
         return std::nullopt;
     }
 
+    // Currently remote server's certificate verfication fails for
+    // HMC provided self-signed certificates,
+    // so skip remote server's certificate verfication.
+
+    // TODO: certificate verfication can be enabled after supporting
+    // HMC-BMC connection certificate flows on both HMC and BMC.
+
     // Add a directory containing certificate authority files to be used
     // for performing verification.
-    sslCtx.set_default_verify_paths(ec);
+    /* sslCtx.set_default_verify_paths(ec);
     if (ec)
     {
         BMCWEB_LOG_ERROR("SSL context set_default_verify failed");
@@ -555,7 +562,7 @@ inline std::optional<boost::asio::ssl::context> getSSLClientContext()
     {
         BMCWEB_LOG_ERROR("SSL context set_verify_mode failed");
         return std::nullopt;
-    }
+    } */
 
     // All cipher suites are set as per OWASP datasheet.
     // https://cheatsheetseries.owasp.org/cheatsheets/TLS_Cipher_String_Cheat_Sheet.html


raviteja-b:
> > This is downstream only commit to disable peer certificate verfication in redfish events path.
> 
> > Currently HMC-BMC certificate flows are not supported, so this commit disable peer certificate verfication during ssl handshake.
> 
> Why aren't they supported? Why do we want this change? What is our future plan here?

Certificate validation is disabled on 1050/1060 builds as well
Actually this upstream issue which needs to fixed, there should conditional verification of certificates need to supported

raviteja-b:
> > This is downstream only commit to disable peer certificate verfication in redfish events path.
> 
> > Currently HMC-BMC certificate flows are not supported, so this commit disable peer certificate verfication during ssl handshake.
> 
> Why aren't they supported? Why do we want this change? What is our future plan here?

Changes involved in both BMC and HMC to support HMC-BMC certificate flows, we have already discussed with HMC team on this which may be considered in PFuture

include/ssl_key_handler.hpp
```diff
@@ -556,7 +563,7 @@ inline std::optional<boost::asio::ssl::context> getSSLClientContext()
         BMCWEB_LOG_ERROR("SSL context set_verify_mode failed");
         return std::nullopt;
     }
-
+#endif
```
baemyung, line 566: 1) Instead using `#if 0`, I think it is better to add an option (`bmcweb-peer-cert-verify`) and make it as `disabled` by default?

After that, upstream it (with enabling it).

Wouldn't this be a better way to handle this?

2) 1060, this part of code is not disabled.  
     Why would need to be disabled in 1110?

https://github.com/ibm-openbmc/bmcweb/blob/a33fa9a4d1528126ba92b9290915811ef1132201/include/ssl_key_handler.hpp#L497

abhilashraju:
> > > This is downstream only commit to disable peer certificate verfication in redfish events path.
> > 
> > 
> > > Currently HMC-BMC certificate flows are not supported, so this commit disable peer certificate verfication during ssl handshake.
> > 
> > 
> > Why aren't they supported? Why do we want this change? What is our future plan here?
> 
> Changes involved in both BMC and HMC to support HMC-BMC certificate flows, we have already discussed with HMC team on this which may be considered for PFuture

I would suggest to make changes in HMC. Any way hmc is sending self signed entity certificate now. Why can't HMC certify the entity ceritificate with  CA certificate and upload the CA to bmc trust store?

gtmills:
Spelling in the PR title 
> verfication
verification

include/ssl_key_handler.hpp
raviteja-b: we are already doing the same for 1050 and 1060 code base and as per our discussions with HMC team, looks like HMC can't make changes in their certificate flows for p11.
we have already discussed this upstream, for now we should disable  server verification

raviteja-b:
> > > > This is downstream only commit to disable peer certificate verfication in redfish events path.
> > > 
> > > 
> > > > Currently HMC-BMC certificate flows are not supported, so this commit disable peer certificate verfication during ssl handshake.
> > > 
> > > 
> > > Why aren't they supported? Why do we want this change? What is our future plan here?
> > 
> > 
> > Changes involved in both BMC and HMC to support HMC-BMC certificate flows, we have already discussed with HMC team on this which may be considered for PFuture
> 
> I would suggest to make changes in HMC. Any way hmc is sending self signed entity certificate now. Why can't HMC certify the entity ceritificate with CA certificate and upload the CA to bmc trust store?

As per understanding after discussing with HMC, HMC can't make any changes in certificate flows for p11 now. this can be discussed as pfuture item.

include/ssl_key_handler.hpp
raviteja-b: > 1. Instead using `#if 0`, I think it is better to add an option (`bmcweb-peer-cert-verify`) and make it as `disabled` by default?
> 
> After that, upstream it (with enabling it).
> 
> Wouldn't this be a better way to handle this?
> 
> 2. 1060, this part of code is not disabled.
>    Why would need to be disabled in 1110?
> 
> https://github.com/ibm-openbmc/bmcweb/blob/a33fa9a4d1528126ba92b9290915811ef1132201/include/ssl_key_handler.hpp#L497

1060 does not use this method at all

raviteja-b: Till now HMC certificate verification was never supported on BMC systems, never heard of any security concern as such till now on any , especially this server certificate verification part which hit in redfish event flow

abhilashraju: " never heard of any security concern" does not mean that there is no vulnerability exists. It is a potential threat and now any one can subscribe to redfish events which probably has some secure data. Yes it is an inconvenience during development, but we can solve it by uploading self signed CA to bmc.

abhilashraju: > we are already doing the same for 1050 and 1060 code base and as per our discussions with HMC team, looks like HMC can't make changes in their certificate flows for p11. we have already discussed this upstream, for now we should disable server verification

Yes ,we have already discussed about this in our scrum. I pointed out this in this commit to make the note on where the actual fix supposed to go.

baemyung: My previous comment is mainly about the code management  so that we can potentially avoid any more downstream-only patches in the future by doing;
- upstream bmcweb - Add a meson option to enable/disable this part
- upstream bmcweb - make it as enabled (`meson_options.txt`)
- downstream openbmc - make it as disabled (`bmcweb_%.bbappend`)

In this way, as long as this part needs to be different between upstream and downstream, we can keep using the same code in both upstream & downstream.

gtmills:
What is 1060 PR that turned this off?
Have the link to the discord work?
Have an issue to address this long term?

abhilashraju:
> > This is downstream only commit to disable peer certificate verfication in redfish events path.
> 
> > Currently HMC-BMC certificate flows are not supported, so this commit disable peer certificate verfication during ssl handshake.
> 
> Why aren't they supported? Why do we want this change? What is our future plan here?

https://github.com/ibm-openbmc/bmcweb/blob/1060/http/http_client.hpp#L651 this is the 1060 code where peer verification is disabled

raviteja-b:
> What is 1060 PR that turned this off? Have the link to the discord work? Have an issue to address this long term?

This the PR which disabled verify_peer certificate in http_client 1050/1060 builds 
https://github.com/ibm-openbmc/bmcweb/pull/613

here discord thread 
https://discord.com/channels/775381525260664832/1242416923484880907

will create story to track this work for PFuture

include/ssl_key_handler.hpp
```diff
@@ -540,6 +540,13 @@ inline std::optional<boost::asio::ssl::context> getSSLClientContext()
         return std::nullopt;
     }
 
+    // Currently remote server's certificate verfication fails for
+    // HMC provided self-signed certificates,
+    // so skip remote server's certificate verfication.
+
+#if 0 // TODO: certificate verfication can be enabled after supporting
```
gtmills, line 547: I never see the #if 0 upstream. Can you just comment this code out? Or the option like @baemyung suggested

baemyung: I also agree to do either - commenting out the code section, or having an option.
It is much better to read the code - especially when we do `blame` or search.

raviteja-b: done

include/ssl_key_handler.hpp
```diff
@@ -540,6 +540,13 @@ inline std::optional<boost::asio::ssl::context> getSSLClientContext()
         return std::nullopt;
     }
 
+    // Currently remote server's certificate verfication fails for
+    // HMC provided self-signed certificates,
+    // so skip remote server's certificate verfication.
+
```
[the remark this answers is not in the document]
raviteja-b: I dont think we need to add stories in code.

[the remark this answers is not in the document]
gtmills: Upstream we do. 
I think it would help but oh well I don't feel that strongly about it

include/ssl_key_handler.hpp
```diff
@@ -540,9 +540,16 @@ inline std::optional<boost::asio::ssl::context> getSSLClientContext()
...
+    // TODO: certificate verfication can be enabled after supporting
+    // HMC-BMC connection certificate flows on both HMC and BMC.
+
     // Add a directory containing certificate authority files to be used
     // for performing verification.
-    sslCtx.set_default_verify_paths(ec);
+    /* sslCtx.set_default_verify_paths(ec);
```
gtmills, line 552: I think commenting out every line would be better for the git blame / or searching

baemyung: Yeah, like

```
//sslCtx.set_default_verify_paths(ec);
//if (ec)
//...
```

gtmills (approved):

baemyung (approved):

PR #1011, opened 2024-06-19
1110: sync-up upstream: generate metadata at runtime

baemyung:
The upstream commits are:

https://github.com/openbmc/bmcweb/commit/38221509 Fix regression in metadata
https://github.com/openbmc/bmcweb/commit/090ab8e1 Generate metadata at runtime
https://github.com/openbmc/bmcweb/commit/40a5fb27 Let OEM schemas be formatted by prettier
https://github.com/openbmc/bmcweb/commit/1fc21359 Add symlinks to folder
https://github.com/openbmc/bmcweb/commit/2bde4061 Add all schemas to schema folder
https://github.com/openbmc/bmcweb/commit/720c9898 Move existing schemas into folders

Tested:
- Redfish service validator runs in the same way as before
- SchemaFiles are the same before and after.


gtmills:
Testing here?

baemyung:
> Testing here?

Yes, I have tested (around a week ago).  I'll test it again as I based on the latest 1110 and add the test statement.

gtmills (approved):
If the validator passes, and can see the schemas, I am good. I think this is moving in the right direction.

baemyung:
> If the validator passes, and can see the schemas, I am good. I think this is moving in the right direction.

I've verified the redfish service validator and works the same way before and after. 
I also verified the content of SchemaFiles which are the same.

Once clang-format for schema is fixed, I'll request to merge.

PR #1014, opened 2024-06-24
Dump offload refactoring as per phosphor-debug-collector

asmithakarun:
phosphor-debug-collector has introduced new design changes to dumps. The backend has the following dbus structure to hold bmc and host dumps:

/xyz/openbmc_project/dump/bmc/entry - stores all bmc dump entries /xyz/openbmc_project/dump/system/entry - stores all host dump entries

All types of host dumps are listed under the same dbus object (mentioned above). Each dump types are differentiated with their IDs.

This commit contains changes to the offload method of various host dumps, with system and resource dumps following the unix socket streaming way (current offload method) and all other bmc stored host dumps following the file body approach that is already used to offload bmc dump dumps.

Tested By:
[1] GET
https://${bmc}/redfish/v1/Systems/system/LogServices/Dump/Entries/0000000D/attachment
> attchment_000D

The same command is used to test for all the other host dumps as well.

### DIFF
--- a/http/http_connection.hpp
+++ b/http/http_connection.hpp
@@ -330,9 +330,15 @@ class Connection :
         }
 
         std::string url(req->target());
-        if (boost::algorithm::contains(url,
-                                       "/system/LogServices/Dump/Entries/") &&
-            boost::algorithm::ends_with(url, "/attachment"))
+        // Check if the offload URL has system/resource dump entry ID
+        // System dumps ID starts with 'A<>'
+        // Resource dumps ID starts with 'B<>'
+        if ((boost::algorithm::contains(url,
+                                        "/system/LogServices/Dump/Entries/A") &&
+             boost::algorithm::ends_with(url, "/attachment")) ||
+            (boost::algorithm::contains(url,
+                                        "/system/LogServices/Dump/Entries/B") &&
+             boost::algorithm::ends_with(url, "/attachment")))
         {
             asyncResp->res.setCompleteRequestHandler(
                 [self(shared_from_this())](crow::Response& thisRes) {
--- a/include/dump_offload.hpp
+++ b/include/dump_offload.hpp
@@ -341,76 +341,101 @@ inline void resetHandlers()
     }
 }
 
+inline void dumpOffloadStreamingOnOpenCallback(
+    crow::streaming_response::Connection& conn)
+{
+    if (!systemHandlers.empty())
+    {
+        BMCWEB_LOG_WARNING("Can't allow dump offload opertaion, one "
+                           "of the host dump is already offloading");
+        conn.sendStreamErrorStatus(
+            boost::beast::http::status::service_unavailable);
+        conn.close();
+        return;
+    }
+
+    std::string url(conn.req.target());
+    std::string startDelimiter = "Entries/";
+    std::size_t pos1 = url.rfind(startDelimiter);
+    std::size_t pos2 = url.rfind("/attachment");
+    if (pos1 == std::string::npos || pos2 == std::string::npos)
+    {
+        BMCWEB_LOG_WARNING("Unable to extract the dump id");
+        conn.sendStreamErrorStatus(boost::beast::http::status::not_found);
+        conn.close();
+        return;
+    }
+    std::string dumpId = url.substr(pos1 + startDelimiter.length(),
+                                    pos2 - pos1 - startDelimiter.length());
+    std::string dumpType = "system";
+
+    boost::asio::io_context* ioCon = conn.getIoContext();
+
+    // Generating random id to create unique socket file
+    // for each dump offload request
+    std::random_device rd;
+    std::default_random_engine gen(rd());
+    std::uniform_int_distribution<> dist{0, 1024};
+    std::string unixSocketPath = std::string(unixSocketPathDir) + dumpType +
+                                 "_dump_" + std::to_string(dist(gen));
+    systemHandlers[&conn] = std::make_shared<Handler>(*ioCon, dumpId, dumpType,
+                                                      unixSocketPath);
+    systemHandlers[&conn]->connection = &conn;
+
+    if (!crow::ibm_utils::createDirectory(unixSocketPathDir))
+    {
+        systemHandlers[&conn]->connection->sendStreamErrorStatus(
+            boost::beast::http::status::not_found);
+        systemHandlers[&conn]->connection->close();
+        return;
+    }
+    BMCWEB_LOG_CRITICAL("INFO: {}  dump id {} offload initiated by: ", dumpType,
+                        dumpId, conn.req.session->clientIp);
+    systemHandlers[&conn]->getDumpSize(dumpId, dumpType);
+}
+
+inline void dumpOffloadStreamingOnCloseCallback(
+    crow::streaming_response::Connection& conn, bool& status)
+{
+    auto handler = systemHandlers.find(&conn);
+    if (handler == systemHandlers.end())
+    {
+        BMCWEB_LOG_DEBUG("No handler to cleanup");
+        return;
+    }
+    handler->second->cleanupSocketFiles();
+    if (!status)
+    {
+        handler->second->resetOffloadURI();
+    }
+    handler->second->outputBuffer.clear();
+    systemHandlers.clear();
+}
+
 inline void requestRoutes(App& app)
 {
+    // System Dumps
     BMCWEB_ROUTE(
         app,
-        "/redfish/v1/Systems/system/LogServices/Dump/Entries/<str>/attachment/")
+        "/redfish/v1/Systems/system/LogServices/Dump/Entries/A<str>/attachment/")
         .privileges({{"ConfigureComponents", "ConfigureManager"}})
         .streamingResponse()
         .onopen([](crow::streaming_response::Connection& conn) {
-        if (!systemHandlers.empty())
-        {
-            BMCWEB_LOG_WARNING("Can't allow dump offload opertaion, one "
-                               "of the host dump is already offloading");
-            conn.sendStreamErrorStatus(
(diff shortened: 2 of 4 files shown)


raviteja-b (approved):

raviteja-b:
@asmithakarun please fix CI failures

redfish-core/lib/log_services.hpp
```diff
@@ -4082,6 +4070,30 @@ inline void handleLogServicesDumpEntryComputerSystemDelete(
...
+inline void handleLogServicesDumpEntryComputerSystemDownloadGet(
+    crow::App& app, const std::string& dumpType, const crow::Request& req,
+    const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
+    const std::string& /*dumpId*/)
+{
+    if (!redfish::setUpRedfishRoute(app, req, asyncResp))
+    {
+        return;
+    }
+    std::string url(req.target());
```
abhilashraju, line 4082: look like this code can be taken out in to a function that returns optional value containing the url segment between Entries and attachement

redfish-core/lib/log_services.hpp
```diff
@@ -4082,6 +4070,30 @@ inline void handleLogServicesDumpEntryComputerSystemDelete(
...
+    std::string startDelimiter = "Entries/";
+    std::size_t pos1 = url.rfind(startDelimiter);
```
abhilashraju, line 4084: can we use url.segments() to split urls in to array of string and use std::find to find segment after the Entries segment?

redfish-core/lib/log_services.hpp
asmithakarun: Its just at this place we are having this requirement. 
I guess this should still be fine.

Do you still think if another function is required?

redfish-core/lib/log_services.hpp
```diff
@@ -4339,6 +4351,41 @@ inline void requestRoutesSystemDumpEntryCollection(App& app)
             std::ref(app)));
 }
 
+inline void requestRoutesSystemDumpEntryDownload(App& app)
+{
+    // Hardware dump
+    BMCWEB_ROUTE(
+        app,
+        "/redfish/v1/Systems/system/LogServices/Dump/Entries/0<str>/attachment/")
```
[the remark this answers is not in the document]
asmithakarun: Its because we will need to have two paths: System & Resource dumps and all other bmc stored host dumps.

I tested with A<str>, B<str> and others as just /<str>/ in the URI. But I remember getting errors for URL/route not found when I used it this way. Hence I defined separate routes for these bmc stored host dumps (IBM specific)

http/http_connection.hpp
```diff
@@ -330,9 +330,12 @@ class Connection :
         }
 
         std::string url(req->target());
-        if (boost::algorithm::contains(url,
-                                       "/system/LogServices/Dump/Entries/") &&
-            boost::algorithm::ends_with(url, "/attachment"))
+        if ((boost::algorithm::contains(url,
```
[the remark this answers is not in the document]
asmithakarun: As per debug-collector design, all the system dumps' ID starts with 'A' and resource dumps' ID with 'B'
I will add them in the code as a comment.

include/dump_offload.hpp
```diff
@@ -341,76 +341,100 @@ inline void resetHandlers()
...
+                                 "_dump_" + std::to_string(dist(gen));
+    systemHandlers[&conn] = std::make_shared<Handler>(*ioCon, dumpId, dumpType,
+                                                      unixSocketPath);
+    systemHandlers[&conn]->connection = &conn;
+
+    if (!crow::ibm_utils::createDirectory(unixSocketPathDir))
+    {
+        systemHandlers[&conn]->connection->sendStreamErrorStatus(
+            boost::beast::http::status::not_found);
+        systemHandlers[&conn]->connection->close();
```
abhilashraju, line 389: what happens to object created at line 381? Since we are using the connection address as the key to the map there is not guarantee that the object will be reinitialised at the same map location, hence there is a possibility of memory leak.

include/dump_offload.hpp
```diff
@@ -341,76 +341,100 @@ inline void resetHandlers()
...
+    std::random_device rd;
+    std::default_random_engine gen(rd());
+    std::uniform_int_distribution<> dist{0, 1024};
+    std::string unixSocketPath = std::string(unixSocketPathDir) + dumpType +
+                                 "_dump_" + std::to_string(dist(gen));
+    systemHandlers[&conn] = std::make_shared<Handler>(*ioCon, dumpId, dumpType,
+                                                      unixSocketPath);
+    systemHandlers[&conn]->connection = &conn;
+
+    if (!crow::ibm_utils::createDirectory(unixSocketPathDir))
```
abhilashraju, line 385: can't you do this check before creating the object at 381?

include/dump_offload.hpp
```diff
@@ -341,76 +341,100 @@ inline void resetHandlers()
...
+        BMCWEB_LOG_DEBUG("No handler to cleanup");
+        return;
+    }
+    handler->second->cleanupSocketFiles();
+    if (!status)
+    {
+        handler->second->resetOffloadURI();
+    }
+    handler->second->outputBuffer.clear();
+    systemHandlers.clear();
```
abhilashraju, line 412: better change the data structure to single value container such as std::optional

redfish-core/lib/log_services.hpp
asmithakarun: Done

redfish-core/lib/log_services.hpp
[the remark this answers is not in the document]
abhilashraju: but these are not valid redfish urls right?  If that is the case why don't your try something like
/redfish/v1/Systems/system/LogServices/Dump/Entries/A/<str>/attachment/
/redfish/v1/Systems/system/LogServices/Dump/Entries/B/<str>/attachment/
/redfish/v1/Systems/system/LogServices/Dump/Entries/<str>/attachment/
better to replace A & B with better readable name

http/http_connection.hpp
```diff
@@ -330,9 +330,15 @@ class Connection :
...
 
         std::string url(req->target());
-        if (boost::algorithm::contains(url,
-                                       "/system/LogServices/Dump/Entries/") &&
-            boost::algorithm::ends_with(url, "/attachment"))
+        // Check if the offload URL has system/resource dump entry ID
+        // System dumps ID starts with 'A<>'
+        // Resource dumps ID starts with 'B<>'
+        if ((boost::algorithm::contains(url,
+                                        "/system/LogServices/Dump/Entries/A") &&
```
gtmills, line 337: How does a Redfish user know A is system and B is resource?
Does an ID that gives some more information help? Can we fill out the Name/Description fields?

include/dump_offload.hpp
```diff
@@ -341,76 +341,101 @@ inline void resetHandlers()
...
+    systemHandlers[&conn]->connection = &conn;
+
+    if (!crow::ibm_utils::createDirectory(unixSocketPathDir))
+    {
+        systemHandlers[&conn]->connection->sendStreamErrorStatus(
+            boost::beast::http::status::not_found);
+        systemHandlers[&conn]->connection->close();
+        return;
+    }
+    BMCWEB_LOG_CRITICAL("INFO: {}  dump id {} offload initiated by: ", dumpType,
```
gtmills, line 392: Can we get rid of this? PE opened a defect on these

gtmills: https://jazz07.rchland.ibm.com:13443/jazz/web/projects/CSSD#action=com.ibm.team.workitem.viewWorkItem&id=613773

redfish-core/lib/log_services.hpp
```diff
@@ -4339,6 +4345,45 @@ inline void requestRoutesSystemDumpEntryCollection(App& app)
...
+        "/redfish/v1/Systems/system/LogServices/Dump/Entries/3<str>/attachment/")
+        .privileges(redfish::privileges::getLogEntry)
+        .methods(boost::beast::http::verb::get)(
+            std::bind_front(handleLogServicesDumpEntryComputerSystemDownloadGet,
+                            std::ref(app), "System"));
+
+    // OCMB SBE dump
+    BMCWEB_ROUTE(
+        app,
+        "/redfish/v1/Systems/system/LogServices/Dump/Entries/4<str>/attachment/")
```
gtmills, line 4380: Instead of these 0-4 which have some special meaning, can we use the name here?
"/redfish/v1/Systems/system/LogServices/Dump/Entries/ocmb<str>/attachment/"

Even better though I feel is 

"/redfish/v1/Systems/system/LogServices/OCMBDump/Entries/<str>/attachment/"

Can we move to that?

redfish-core/lib/log_services.hpp
```diff
@@ -883,18 +883,6 @@ inline void
...
-    if (downloadEntryType == "System")
-    {
-        if (!asyncResp->res.openFd(fd, bmcweb::EncodingType::Base64))
-        {
-            messages::internalError(asyncResp->res);
-            close(fd);
-            return;
-        }
-        asyncResp->res.addHeader(
-            boost::beast::http::field::content_transfer_encoding, "Base64");
```
gtmills, line 895: Don't we still want this header ?

redfish-core/lib/log_services.hpp
abhilashraju: but it is doing potentially reusable function of extracting the needed url segment from url. This has nothing to do actual function. So separate the concerns, which will improve readability and future reusability.

asmithakarun:
NO changes are needed for offload as the old "InitiateOffload" method itself works for all the host dumps.

The decision is to have this way of offload for now until there is a "GetFileHandle" method implemented for system and resource dumps.

PR #1015, opened 2024-06-25
Revert "Fix regression in metadata"

gtmills:
This reverts commit 093eebecf9c395aff3543b8252e2794e51530e59 to fix defect 629044.

Defect 629044 reported an endless redirect loop of `/?next=/`

This was found using git bisect. 

Tested: A 2nd test with and without this commit showed, the endless redirect without this commit but fixed by this commit.


baemyung (approved):
Let's move with this in and continue to investigate the upstream part.

PR #1020, opened 2024-07-11
Audit Log: Use named fields for log entries

jeaaustx:
The JSON file created by phosphor-auditlog now uses named fields for the expected message parameters. It also no longer knows about the message registry message. Adjust bmcweb to build the message from the expected JSON fields.

Addresses [comment from review](https://github.com/ibm-openbmc/phosphor-logging/pull/41#discussion_r1442173025) of original implementation.

Must be merged with dependent change: https://github.com/ibm-openbmc/phosphor-logging/pull/64

Tested:
 - Verified Redfish response same after change.
 - Redfish Validator passed.

### DIFF
--- a/redfish-core/lib/log_services.hpp
+++ b/redfish-core/lib/log_services.hpp
@@ -54,6 +54,7 @@
 #include <cstdio>
 #include <filesystem>
 #include <iterator>
+#include <map>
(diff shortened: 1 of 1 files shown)


baemyung (approved):
It looks good to me

PR #1021, opened 2024-07-12
Show hypervisor path on GET of Systems collection

asmithakarun:
Redfish GET on `/redfish/v1/Systems` is missing to show the hypervisor resource in its members list, even though hypervisor is available. This commit fixes missing hypervisor resource in a GET response on Systems collection. The `Members@odata.count` is retrieved as `uint64_t` pointer. But the Json library might store this value as "int" or "size_t". Therefore, directly casting it to uint64_t* might not work as expected.

Upstream commit: https://gerrit.openbmc.org/c/openbmc/bmcweb/+/72734

Tested By:
Verified "GET https://${bmc}/redfish/v1/Systems" returns the hypervisor resource

Change-Id: If8713fc70663cf72cc555f695b8f4ec6547215a2


redfish-core/lib/systems.hpp
```diff
@@ -2638,18 +2638,35 @@ inline void handleComputerSystemCollectionGet(
             BMCWEB_LOG_CRITICAL("Count wasn't found??");
             return;
         }
-        uint64_t* count = val->get_ptr<uint64_t*>();
-        if (count == nullptr)
+        nlohmann::json& countRef = *val;
```
gtmills, line 2641: Why is this needed?

redfish-core/lib/systems.hpp
```diff
@@ -2623,8 +2623,8 @@ inline void handleComputerSystemCollectionGet(
     system["@odata.id"] = "/redfish/v1/Systems/system";
     ifaceArray.emplace_back(std::move(system));
     sdbusplus::asio::getProperty<std::string>(
-        *crow::connections::systemBus, "xyz.openbmc_project.Settings",
```
gtmills, line 2626: This is upstream.. Can you also fix it there?

redfish-core/lib/systems.hpp
asmithakarun: this is to count the number of members in the list. like:
`"Members@odata.count": 2,`

raviteja-b (approved):

raviteja-b (approved):

redfish-core/lib/systems.hpp
```diff
@@ -2623,8 +2623,8 @@ inline void handleComputerSystemCollectionGet(
...
-        "/xyz/openbmc_project/network/hypervisor",
+        *crow::connections::systemBus, "xyz.openbmc_project.Network.Hypervisor",
```
gtmills, line 2626: This part isn't upstream. Why do we need this change?

asmithakarun: this is a downstream only change. We are enabling hypervisor with IBM flag in the networkd.
So this bmcweb part of code was never changed..

baemyung:
>But the Json library might store this value as "int" or "size_t". Therefore, directly casting it to uint64_t* might not work as expected.

Then, should we do the cast of `size_t`?

asmithakarun:
> > But the Json library might store this value as "int" or "size_t". Therefore, directly casting it to uint64_t* might not work as expected.
> 
> Then, should we do the cast of `size_t`?

But here, in our usecase int64_t works, so I just used that.

baemyung:
>But here, in our usecase int64_t works, so I just used that.

Yes, it turns out "int64_t" seems appropriate as it is Json integer type.

baemyung (approved):
Besides Gunnar's comment, it looks good to me

gtmills (approved):

PR #1027, opened 2024-07-18
Change getPrettyName to take specific service name

cjcain:
Rebasing part of the existing change from
ibm-openbmc#698 The portion related to
pretty name is addressed here

Change-Id: Ieb1bc828a825609297151635db8d8cb53d1d4294


gtmills:
I think what we have works for 1110?  So closing.

PR #1028, opened 2024-07-19
RedfishEvents GHE: Enable TLS1.2 in http client

raviteja-b:
This is downstream only patch to enable TLS1.2 support
Currently TLS1.3 is hardcoded in http client which restricts redfish events path to work with TLS1.3
and fails with TLS1.2

This commit enables TLS1.2 along with TLS1.3 in http_client and make sure redfish events work with HMC for both TLS1.2 and TLS1.3

Tested by:
Verified redfish events on HMC with TLS1.2 and TLS1.3


baemyung (approved):
It looks good to me

PR #1031, opened 2024-07-24
1110: Fix PCIeSlots when empty association

baemyung:
In case when there is no association between chassis and pcieslots, `getAssociatedSubTreePaths()` may give the empty Paths rather than the error return.

This is to skip the non-associated pcieslots rather than generating an internal error.

This happens when PLDM discovers the additional slots which may not be associated with the other inventory chassis.

```
curl -k -X GET https://${bmc}/redfish/v1/Chassis/chassis
...
> debug code:
pcieSlotPath=/xyz/openbmc_project/inventory/system/chassis15363/logical_slot1, chassisID=chassis
```

Tested:
- curl -k -X GET https://${bmc}/redfish/v1/Chassis/chassis

### DIFF
--- a/redfish-core/lib/pcie_slots.hpp
+++ b/redfish-core/lib/pcie_slots.hpp
@@ -27,6 +27,7 @@
 #include <optional>
 #include <string>
 #include <string_view>
+#include <utility>
 #include <vector>
 
 namespace redfish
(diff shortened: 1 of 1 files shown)


gtmills:
Is there a defect here or anything?

gtmills (approved):

baemyung:
> Is there a defect here or anything?

I found it while working with @ArchanaKakani during PLDM development for MEX on rain1118bmc.

baemyung:
Perhaps, this may not be needed if  the associations (containing/contained_by) are created for all pcieslots - including MEX logical_slots.

The code skips if the associated chassis are different, but requires all pcieslots to have chassis association 

https://github.com/ibm-openbmc/bmcweb/blob/c4b1a3f3fa017947160bf2b1e7c6fd1ec7056f81/redfish-core/lib/pcie_slots.hpp#L454

So, as long as the associations are created, this may not be needed.

gtmills:
> I found it while working with @ArchanaKakani during PLDM development for MEX on rain1118bmc.

Ack.

redfish-core/lib/pcie_slots.hpp
```diff
@@ -611,10 +481,81 @@ inline void getValidPCIeSlotList(
...
+    }
+
+    BMCWEB_LOG_DEBUG("Get properties for PCIeSlots associated to chassis = {}",
+                     chassisID);
+
+    asyncResp->res.jsonValue["@odata.type"] = "#PCIeSlots.v1_5_0.PCIeSlots";
+    asyncResp->res.jsonValue["Name"] = "PCIe Slot Information";
+    asyncResp->res.jsonValue["@odata.id"] =
+        boost::urls::format("/redfish/v1/Chassis/{}/PCIeSlots", chassisID);
+    asyncResp->res.jsonValue["Id"] = "1";
```
gtmills, line 507: why is Id 1 instead of PCIeSlots?

gtmills: Do collections normally have IDs?

gtmills: Oh well, it was there before

redfish-core/lib/pcie_slots.hpp
```diff
@@ -571,27 +449,19 @@ inline void getValidPCIeSlotList(
...
 
     dbus::utility::getAssociatedSubTree(
-        chassisPath + "/inventory",
+        chassisPath + "/containing",
         sdbusplus::message::object_path("/xyz/openbmc_project/inventory"), 0,
         pcieSlotIntf,
         [asyncResp, chassisID, chassisPath, callback{std::move(callback)}](
             const boost::system::error_code& ec,
             const dbus::utility::MapperGetSubTreeResponse& subtree) {
+        std::vector<std::pair<std::string, std::string>> slotPathConnNames;
```
gtmills, line 458: nitpick: include utility for std::pair

gtmills (approved):

redfish-core/lib/pcie_slots.hpp
baemyung: Upstream and 1060 all have the same Id = 1.
What other value may be better as Id?  Or, would it better to remove it?

redfish-core/lib/pcie_slots.hpp
baemyung: Add <utility>

redfish-core/lib/pcie_slots.hpp
baemyung: Would it be better to change to "PCIeSlots"

baemyung: >Gunnar help to show that "Id" is not used for the other collections and also dmtf mockup does not use it.
>https://github.com/openbmc/bmcweb/blob/master/redfish-core/lib/processor.hpp#L1291
>Yeah, there mockups don't have it https://redfish.dmtf.org/redfish/mockups/v1/1448

So, I'll remove it

jeaaustx (approved):

PR #1034, opened 2024-07-28
Fix RemoteRoleMap PATCH operation

raviteja-b:
This commit fixes RemoteRoleMap Patch operation

currently setting LocalRole property fails which requires converting LocalRole 
Redfish to D-bus property values.

Tested By:
Verified Patch operation on RemoteRoleMap


raviteja-b:
Upstream commit: https://gerrit.openbmc.org/c/openbmc/bmcweb/+/73090

gtmills (approved):

PR #1037, opened 2024-07-30
Add new Message registry with ASCII post code place holder

manojkiraneda:
- In the current state, the BIOS Post code message registry has a place holder to only display a hex based post code.

- Since Power 5 series of servers, IBM progress codes are ASCII based character codes as mentioned in the below link: (https://www.ibm.com/support/knowledgecenter/POWER5/ipha6_p5/progcodesmain.htm)

- These progress codes are displayed in multiple places like BMC GUI, Operator Panel , Hardware Management Console e.t.c, and most of the customers of power are well accustomed to seeing these character based codes. And in the current infrastructure we display the hex form of uint64 value, which can lead to misses where some customer does not know how to look up the code & translate it to ASCII and last thing we would want is for a service rep to do the translation.

- For most of the systems where we don't have the ASCII convertible progress codes, it should not be an issue as we still display the hex codes, but in addition we just display empty string in the ASCII place holder.

- Clients like management console, GUI(probably needs an OEM code), Operator Panel can probably do this conversion of hex to ASCII, but its the customers that we are really concerned about & also by adding this additional field in a new message registry we can probably share a common GUI code (that displays both the hex codes as well as ASCII codes).

- Redfish experts suggestion is, instead of changing the existing message registry which can break clients, we have to add a new message registry with almost the similar name like BIOSPOSTCodeASCII with the new argument set and bump up the minor version of the Message registry.

(https://redfishforum.com/thread/447/bumping-message-registry-backward-compatability)


baemyung (approved):
The logic seems good to me, but CI failure would need to be fixed.

http/utility.hpp
```diff
@@ -310,6 +310,30 @@ inline bool base64Decode(std::string_view input, std::string& output)
...
+ * @return ascii converted post code string
+ */
+
+inline std::string convertToAscii(const uint64_t& element)
+{
+    uint64_t tmpelement = element;
+    uint8_t* p = static_cast<uint8_t*>(static_cast<void*>(&tmpelement));
+    std::span<unsigned char> bytearray{p, 8};
+
+    if (std::count_if(bytearray.begin(), bytearray.end(), [](unsigned char c) {
```
jeaaustx, line 327: ```#include <algorithm>```

http/utility.hpp
```diff
@@ -310,6 +310,30 @@ inline bool base64Decode(std::string_view input, std::string& output)
...
+        return (std::isprint(c) == 0);
```
jeaaustx, line 328: ```#include <cctype>```

redfish-core/include/registries/openbmc_message_registry.hpp
```diff
@@ -241,6 +241,16 @@ constexpr std::array registry =
             },
             "None.",
         }},
+    MessageEntry{
```
jeaaustx, line 244: Need to actually make the change in openbmc.json in the same directory and then build the openbmc_message_registry.hpp from that.

(See L2-L10 of this file.)

Also, the subminor version is supposed to be updated when a new message is added.

http/utility.hpp
```diff
@@ -310,6 +310,30 @@ inline bool base64Decode(std::string_view input, std::string& output)
...
 
+/**
+ * Method returns 64bit post code into ASCII string
+ *
+ * @param[in] postcode value
+ *
+ * @return ascii converted post code string
+ */
+
+inline std::string convertToAscii(const uint64_t& element)
```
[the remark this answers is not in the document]
manojkiraneda: done.

http/utility.hpp
manojkiraneda: done.

http/utility.hpp
manojkiraneda: done

redfish-core/include/registries/openbmc_message_registry.hpp
manojkiraneda: done, and bumped the registry subminor version to 0.6

jeaaustx (approved):
Looks good to me!

include/str_utility.hpp
```diff
@@ -41,4 +43,28 @@ inline bool asciiIEquals(std::string_view left, std::string_view right)
...
+ * @param[in] 64 bit value
+ *
+ * @return ascii converted string
+ */
+
+inline std::string convertToAscii(const uint64_t& element)
+{
+    uint64_t tmpelement = element;
+    uint8_t* p = static_cast<uint8_t*>(static_cast<void*>(&tmpelement));
+    std::span<unsigned char> bytearray{p, 8};
```
gtmills, line 58: include span, include algorithm

gtmills (approved):

include/str_utility.hpp
manojkiraneda: Included span, already included algorithm.

PR #1040, opened 2024-07-30
1110: Pick Upstream Commits for new option handling 

baemyung:
The key commit to pick up is https://gerrit.openbmc.org/c/openbmc/bmcweb/+/70915

Bring consistency to config options

The configuration options that exist in bmcweb are an amalgimation of
CROW options, CMAKE options using #define, pre-bmcweb ifdef mechanisms
and meson options using a config file.  This history has led to a lot of
different ways to configure code in the codebase itself, which has led
to problems, and issues in consistency.

In addition to options in upstream, this also handles IBM options
- audit-events
- bmc-shell-socket
- hypervisor-serial-socket
- ibm-management-console
- redfish-license
- ibm-led-extensions
- redfish-health-populate
- hw-isolation

Tested:
- unittest-run succeeds
- Simics bring up succeeds
- Redfish Service validator passes


gtmills (approved):
Seems reasonable to me

PR #1041, opened 2024-07-30
Hypervisor fix validator errors

asmithakarun:
Currently, Redfish validator fails for two things when doing a GET on hypervisor ethernet interfaces:
1. Empty mac address getting displayed, and
2. Empty Address origin

This commit ensures that the redfish validator passes by adding appropriate checks in the GET handler method.

Upsrteam: https://gerrit.openbmc.org/c/openbmc/bmcweb/+/73196

Fixes: [
](https://jazz07.rchland.ibm.com:13443/jazz/web/projects/CSSD#action=com.ibm.team.workitem.viewWorkItem&id=637212)

Tested By:
Validator Passed.

### DIFF
--- a/redfish-core/lib/hypervisor_system.hpp
+++ b/redfish-core/lib/hypervisor_system.hpp
@@ -506,7 +506,10 @@ inline void parseInterfaceData(nlohmann::json& jsonResponse,
     jsonResponse["Id"] = ifaceId;
     jsonResponse["@odata.id"] = boost::urls::format(
         "/redfish/v1/Systems/hypervisor/EthernetInterfaces/{}", ifaceId);
-    jsonResponse["MACAddress"] = ethData.macAddress;
+    if (!ethData.macAddress.empty())
+    {
+        jsonResponse["MACAddress"] = ethData.macAddress;
+    }
 
     jsonResponse["HostName"] = ethData.hostName;
     jsonResponse["DHCPv4"]["DHCPEnabled"] =
@@ -528,26 +531,48 @@ inline void parseInterfaceData(nlohmann::json& jsonResponse,
     ipv4Array = nlohmann::json::array();
     ipv4StaticArray = nlohmann::json::array();
     bool ipv4IsActive = false;
+    std::string v4Origin;
+    std::string v4Gateway;
+    std::string v4Netmask;
+    std::string v4Address;
+
     for (const auto& ipv4Config : ipv4Data)
     {
         if (ipv4Config.isActive)
         {
             ipv4IsActive = ipv4Config.isActive;
         }
+        if (!ipv4Config.origin.empty())
+        {
+            v4Origin = ipv4Config.origin;
+        }
+        if (!ipv4Config.gateway.empty())
+        {
+            v4Gateway = ethData.defaultGateway;
+        }
+        if (!ipv4Config.netmask.empty())
+        {
+            v4Netmask = ipv4Config.netmask;
+        }
         if (!ipv4Config.address.empty())
         {
-            nlohmann::json::object_t ipv4;
-            ipv4["AddressOrigin"] = ipv4Config.origin;
-            ipv4["SubnetMask"] = ipv4Config.netmask;
-            ipv4["Address"] = ipv4Config.address;
-            ipv4["Gateway"] = ipv4Config.gateway;
-
-            if (ipv4Config.origin == "Static")
-            {
-                ipv4StaticArray.emplace_back(ipv4);
-            }
-            ipv4Array.emplace_back(std::move(ipv4));
+            v4Address = ipv4Config.address;
         }
+    }
+    nlohmann::json::object_t ipv4;
+    ipv4["AddressOrigin"] = v4Origin;
+    ipv4["SubnetMask"] = v4Netmask;
+    ipv4["Address"] = v4Address;
+    ipv4["Gateway"] = v4Gateway;
+    if (v4Origin == "Static")
+    {
+        ipv4StaticArray.push_back(ipv4);
+    }
+    ipv4Array.emplace_back(std::move(ipv4));
+
+    if (ipv4IsActive)
+    {
+        jsonResponse["InterfaceEnabled"] = true;
     }
 
     std::string ipv6GatewayStr = ethData.ipv6DefaultGateway;
--- a/redfish-core/lib/hypervisor_system.hpp
+++ b/redfish-core/lib/hypervisor_system.hpp
@@ -281,9 +281,18 @@ inline bool extractHypervisorInterfaceData(
 
                         if (origin != nullptr)
                         {
-                            ipv4Address.origin =
-                                translateAddressOriginDbusToRedfish(*origin,
-                                                                    true);
+                            if (protocol == "ipv4")
+                            {
+                                ipv4Address.origin =
+                                    translateAddressOriginDbusToRedfish(*origin,
+                                                                        true);
+                            }
+                            if (protocol == "ipv6")
+                            {
+                                ipv6Address.origin =
+                                    translateAddressOriginDbusToRedfish(*origin,
+                                                                        false);
+                            }
                         }
                     }
 
@@ -531,10 +540,6 @@ inline void parseInterfaceData(nlohmann::json& jsonResponse,
     ipv4Array = nlohmann::json::array();
     ipv4StaticArray = nlohmann::json::array();
     bool ipv4IsActive = false;
-    std::string v4Origin;
-    std::string v4Gateway;
-    std::string v4Netmask;
-    std::string v4Address;
 
     for (const auto& ipv4Config : ipv4Data)
     {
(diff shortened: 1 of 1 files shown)


redfish-core/lib/hypervisor_system.hpp
```diff
@@ -527,27 +530,44 @@ inline void parseInterfaceData(nlohmann::json& jsonResponse,
...
+    nlohmann::json::object_t ipv4;
+    ipv4["AddressOrigin"] = v4Origin;
+    ipv4["SubnetMask"] = v4Netmask;
+    ipv4["Address"] = v4Address;
+    ipv4["Gateway"] = v4Gateway;
+    if (v4Origin == "Static")
+    {
+        ipv4StaticArray.push_back(ipv4);
+    }
+    ipv4Array.emplace_back(std::move(ipv4));
```
baemyung, line 571: I think the logic is not correct.
Please see my comment - https://gerrit.openbmc.org/c/openbmc/bmcweb/+/73196/comment/2def91f4_7fe4e03d/

asmithakarun:
jenkins run tests please

redfish-core/lib/hypervisor_system.hpp
```diff
@@ -527,25 +530,40 @@ inline void parseInterfaceData(nlohmann::json& jsonResponse,
...
+        }
+        if (ipv4.size() == ipv4ConfigFieldsRequired)
+        {
             if (ipv4Config.origin == "Static")
             {
-                ipv4StaticArray.emplace_back(ipv4);
+                ipv4StaticArray.push_back(ipv4);
             }
             ipv4Array.emplace_back(std::move(ipv4));
+            ipv4 = nlohmann::json::object();
```
baemyung, line 566: L565 stmt should be moved to upper scope as each loop will need a new clean `ipv4`.
```
        if (ipv4.size() == 4)
        {
           ...
           ipv4Array.emplace_back(std::move(ipv4));
          ipv4 = nlohmann::json::object();
        }
      <-- Here the older info in ipv4 may split over to the next loop

==>

        if (ipv4.size() == 4)
        {
           ...
           ipv4Array.emplace_back(std::move(ipv4));
        }
        ipv4 = nlohmann::json::object();
```

In fact, `ipv4` is for each loop, and L534 is better moved to the inside the loop.

```
--- a/redfish-core/lib/hypervisor_system.hpp
+++ b/redfish-core/lib/hypervisor_system.hpp
@@ -531,9 +531,9 @@ inline void parseInterfaceData(nlohmann::json& jsonResponse,
     ipv4StaticArray = nlohmann::json::array();
     bool ipv4IsActive = false;

-    nlohmann::json::object_t ipv4;
     for (const auto& ipv4Config : ipv4Data)
     {
+        nlohmann::json::object_t ipv4;
         if (ipv4Config.isActive)
         {
             ipv4IsActive = ipv4Config.isActive;
@@ -562,7 +562,6 @@ inline void parseInterfaceData(nlohmann::json& jsonResponse,
                 ipv4StaticArray.push_back(ipv4);
             }
             ipv4Array.emplace_back(std::move(ipv4));
-            ipv4 =  nlohmann::json::object();
         }
     }
```

redfish-core/lib/hypervisor_system.hpp
```diff
@@ -591,9 +621,10 @@ inline void parseInterfaceData(nlohmann::json& jsonResponse,
                 }
             }
             ipv6Array.emplace_back(std::move(ipv6));
+            ipv6 = nlohmann::json::object();
```
baemyung, line 624: Similar comment like ipv4 case

asmithakarun: I observed that, there is Origin captured first and in the next loop, Origin is null, and in a later loop, the address is captured and in a later loop prefix length is captured.

I initially had the change you are suggesting, but the problem is that I want to push ipv6 details into the ipv6 array only when all the required fields are there in the json object. So I added a check for ipv6.size().
But the size is becoming 0 on every loop and it never reached the check in line 608.

redfish-core/lib/hypervisor_system.hpp
asmithakarun: Somehow the below problem didnot occur for ipv4. At one point in loop, I was able to see all the fields captured and then size becomes 4, and it will be pushed to ipv4Array.
I made the change here for ipv4 just to keep both ipv4 and ipv6 logics same.

baemyung: The current code may potentially spill ipv4Data to the next element if fields may be different between them.
For example,
 ipv4Data[0] has 3 fields on, and ipv4data[1] has 3 different fields on.

- [0] AddressOrigin1, SubnetMask1, Address1
- [1] AddressOrigin2, SubnetMask2,                 Gateway2

If so, the current code will create an entry with { AddressOrigin2, SubnetMask2,    Address1,   Gateway2}

Each loop must need to clear `ipv4`.  
This means that, `ipv4` is better to be defined at the top of each loop.

redfish-core/lib/hypervisor_system.hpp
baemyung: Question.

How/what would `ipv6Data` be provided - in case for one or more IP configurations?
Would each ip config field is separated each array element for the given IP, or each array element contain all fields?

You may add some debug statement to show the content of `ipv6Config` in each loop.

redfish-core/lib/hypervisor_system.hpp
```diff
@@ -213,41 +213,38 @@ inline bool extractHypervisorInterfaceData(
...
+                ethIfaceId, protocol);
+            IPv4AddressData ipv4Address;
+            IPv6AddressData ipv6Address;
+            if (objpath.first == ipAddrObj)
             {
-                std::string ipAddrObj = std::format(
-                    "/xyz/openbmc_project/network/hypervisor/{}/{}/addr0",
-                    ethIfaceId, protocol);
-                if (objpath.first == ipAddrObj)
+                for (const auto& ifacePair : objpath.second)
```
baemyung, line 228: Why is this loop reorder needed?  
The new loop order seems assuming there is only IPv4 & IPv6 address.
Even if it may be true in the current usage, I think  I think the origin loop order seems correct as it can also deal with multiple IP addressses.

```
        for (const auto& ifacePair : objpath.second)
        {
            IPv4AddressData& ipv4Address = ipv4Config.emplace_back();
            IPv6AddressData& ipv6Address = ipv6Config.emplace_back();

            for (std::string protocol : {"ipv4", "ipv6"})
            {
```

I  think what may be needed is the new L276-L289 for origin, and L409-L502 for MacAddress, as the validator is reporting.

Wouldn't this be sufficient?


L276:
```
               if (origin != nullptr && !(origin->empty()))
                            if (protocol == "ipv4")
                                ipv4Address.origin =  translateAddressOriginDbusToRedfish(*origin, true);
                            }
                            if (protocol == "ipv6")
                            {
                                ipv6Address.origin = translateAddressOriginDbusToRedfish(*origin, false);
                            }
```

L502:
 
    if (!ethData.macAddress.empty())
    {
        jsonResponse["MACAddress"] = ethData.macAddress;
    }

redfish-core/lib/hypervisor_system.hpp
asmithakarun: Done. have made the change.

redfish-core/lib/hypervisor_system.hpp
asmithakarun: Pushed the new changes

redfish-core/lib/hypervisor_system.hpp
asmithakarun: So, there is a problem with the original code (as I mentioned in slack thread)

The outer for loop will loop through the getManaged Objects and the second for loop will loop through each interface implemented by that dbus object..
We will need IP interface only for the ipv4Data and IPv6Data to be filled. Instead, empty values are being pushed to the ipv4 and ipv6Data on every loop.

With my previous change in this PR, I can see just the following:
```

Aug 02 18:56:48 rain111bmc bmcweb[5264]: [DEBUG hypervisor_system.hpp:418] **** ipv4 size: 1, ipv6 size: 1
Aug 02 18:56:48 rain111bmc bmcweb[5264]: [DEBUG hypervisor_system.hpp:534] *** Inside for v4: isActive: true, origin: Static, netmask: 0.0.0.0, address: 0.0.0.0, ipv4Config.gateway: 0.0.0.0, ipv4.size(): 0
Aug 02 18:56:48 rain111bmc bmcweb[5264]: [DEBUG hypervisor_system.hpp:589] *** Inside for v6: isActive: false, origin: , address: ::, prefixLength: 128, size(): 0

```

But, with the original code, I can see the following (example ipv4):
```
Aug 02 18:03:14 rain111bmc bmcweb[3525]: [DEBUG hypervisor_system.hpp:769] *** Inside for v4: isActive: false, origin: , netmask: , address: , ipv4Config.gateway: , ipv4.size(): 0
Aug 02 18:03:14 rain111bmc bmcweb[3525]: [DEBUG hypervisor_system.hpp:769] *** Inside for v4: isActive: false, origin: , netmask: , address: , ipv4Config.gateway: , ipv4.size(): 0
Aug 02 18:03:14 rain111bmc bmcweb[3525]: [DEBUG hypervisor_system.hpp:769] *** Inside for v4: isActive: false, origin: , netmask: , address: , ipv4Config.gateway: , ipv4.size(): 0
Aug 02 18:03:14 rain111bmc bmcweb[3525]: [DEBUG hypervisor_system.hpp:769] *** Inside for v4: isActive: false, origin: , netmask: , address: , ipv4Config.gateway: , ipv4.size(): 0
Aug 02 18:03:14 rain111bmc bmcweb[3525]: [DEBUG hypervisor_system.hpp:769] *** Inside for v4: isActive: false, origin: Static, netmask: , address: , ipv4Config.gateway: , ipv4.size(): 0
Aug 02 18:03:14 rain111bmc bmcweb[3525]: [DEBUG hypervisor_system.hpp:769] *** Inside for v4: isActive: false, origin: Static, netmask: , address: , ipv4Config.gateway: , ipv4.size(): 1
Aug 02 18:03:14 rain111bmc bmcweb[3525]: [DEBUG hypervisor_system.hpp:769] *** Inside for v4: isActive: true, origin: Static, netmask: 0.0.0.0, address: 0.0.0.0, ipv4Config.gateway: 0.0.0.0, ipv4.size(): 1
```

The above trace will be repeated for almost 33 times with empty value in most of the iterations.

This refactoring can be taken as a separate work/story I guess.

redfish-core/lib/hypervisor_system.hpp
asmithakarun: Done

baemyung (approved):
It looks good to me.

PR #1016, opened 2024-06-27
1110: Fix Handling of BiosAttributeRegistry

baemyung:
Redfish Service Validator is failing like this.
```
1 failGet errors in /redfish/v1/Registries/BiosAttributeRegistry/BiosAttributeRegistry
1 fails errors in /redfish/v1/Registries/BiosAttributeRegistry/BiosAttributeRegistry
ERROR - URI did not return resource /redfish/v1/Registries/BiosAttributeRegistry/BiosAttributeRegistry
```

It can also be verified by doing this.
```
curl -k -X GET https://${bmc}/redfish/v1/Registries/BiosAttributeRegistry/BiosAttributeRegistry
{
  "error": {
    "code": "Base.1.16.0.ResourceNotFound",
    "message": "The requested resource of type MessageRegistryFile named 'BiosAttributeRegistry' was not found."
```

It is because bmcweb does not handle the overlapping routes with
- `/redfish/v1/Registries/<str>/<str>`
- `/redfish/v1/Registries/BiosAttributeRegistry/BiosAttributeRegistry`

In fact, such routing logic worked before although it is not by intentional design, but it is no longer supported [1].

So, I've modified the logic to handle `BiosAttributeRegistry` so that it is routed from `/redfish/v1/Registries/<str>/<str>`

Tested:
- Success on `curl -k -X GET https://${bmc}/redfish/v1/Registries/BiosAttributeRegistry/BiosAttributeRegistry`

- Redfish Service Validaor passes on /redfish/v1/Registries/

[1] https://discord.com/channels/775381525260664832/1255570841568153702

PR #1042, opened 2024-07-30
Fix 404 on Hypervisor Reset

gtmills:
/redfish/v1/Systems/hypervisor/Actions/ComputerSystem.Reset was returning a 404. On closer inspection, it was due to hitting the System .Reset Action and not making it to the Hypervisor specific action. [1].

Remove the Hypervisor specific action and call the method from the System .Reset Action. This is how we do the handleHypervisorSystemGet already. [2]

Fixes defects 637587 and 632508.

https://gerrit.openbmc.org/c/openbmc/bmcweb/+/73203 is upstream 

[1]: https://github.com/ibm-openbmc/bmcweb/blob/1110/redfish-core/lib/systems.hpp#L3335
[2]: https://github.com/ibm-openbmc/bmcweb/blob/1110/redfish-core/lib/systems.hpp#L2906C9-L2906C34

Tested: Loaded this on a system and was able to "Continue to OS running".

### DIFF
--- a/redfish-core/lib/hypervisor_system.hpp
+++ b/redfish-core/lib/hypervisor_system.hpp
@@ -1440,13 +1440,9 @@ inline void handleHypervisorResetActionGet(
 }
 
 inline void handleHypervisorSystemResetPost(
-    App& app, const crow::Request& req,
+    const crow::Request& req,
     const std::shared_ptr<bmcweb::AsyncResp>& asyncResp)
 {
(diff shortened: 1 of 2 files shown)


baemyung (approved):
This is an overlapping route issue similar  to https://github.com/ibm-openbmc/bmcweb/pull/1016.
It looks good to me.

asmithakarun (approved):

PR #654, opened 2023-04-04
PowerSubsystem: Add Allocation properties

smccarney:
Add the Allocation properties to the PowerSubsystem output.

Set the Redfish properties based on properties from the D-Bus xyz.openbmc_project.Control.Power.Cap interface:
* Redfish AllocatedWatts property:
  * Set to PowerCap property if PowerCapEnable is true
  * Set to MaxPowerCapValue property if PowerCapEnable is false
* Redfish RequestedWatts property:
  * Set to MaxPowerCapValue property

Tested:
* Passed the Redfish Validator
* For detailed test plan, see https://gist.github.com/smccarney/613d4f2fd65ed7cb6973fe1512ac327a

Example Output:
$ curl -k https://${bmc}/redfish/v1/Chassis/chassis/PowerSubsystem {
  "@odata.id": "/redfish/v1/Chassis/chassis/PowerSubsystem",
  "@odata.type": "#PowerSubsystem.v1_1_0.PowerSubsystem",
  "Allocation": {
    "AllocatedWatts": 1233,
    "RequestedWatts": 3982
  },
  "Id": "PowerSubsystem",
  "Name": "Power Subsystem",
  "PowerSupplies": {
    "@odata.id": "/redfish/v1/Chassis/chassis/PowerSubsystem/PowerSupplies"
  },
  "Status": {
    "Health": "OK",
    "State": "Enabled"
  }
}

Change-Id: Ie7cce11246f0e7affbd6370e36737e06fab5a577

PR #1044, opened 2024-07-31
PowerSubsystem: Add Allocation properties

smccarney:
Add the Allocation properties to the PowerSubsystem output.

Set the Redfish properties based on properties from the D-Bus xyz.openbmc_project.Control.Power.Cap interface:
* Redfish AllocatedWatts property:
  * Set to PowerCap property if PowerCapEnable is true
  * Set to MaxPowerCapValue property if PowerCapEnable is false
* Redfish RequestedWatts property:
  * Set to MaxPowerCapValue property

Tested:
* Ran the Redfish Validator; no new errors/warnings
* For detailed test plan, see https://gist.github.com/smccarney/1fc1426e51d0fd9ffc715b4d522f4cf9

Example Output:
```
$ curl -k https://${bmc}/redfish/v1/Chassis/chassis/PowerSubsystem
{
  "@odata.id": "/redfish/v1/Chassis/chassis/PowerSubsystem",
  "@odata.type": "#PowerSubsystem.v1_1_0.PowerSubsystem",
  "Allocation": {
    "AllocatedWatts": 1233,
    "RequestedWatts": 3982
  },
  "Id": "PowerSubsystem",
  "Name": "Power Subsystem",
  "PowerSupplies": {
    "@odata.id": "/redfish/v1/Chassis/chassis/PowerSubsystem/PowerSupplies"
  },
  "Status": {
    "Health": "OK",
    "State": "Enabled"
  }
}
```

Change-Id: I0d5e6dc251a66de1c1207257100e13df6f35047f


smccarney:
This is the 1110 port of the following PR from 1050: https://github.com/ibm-openbmc/bmcweb/pull/654

smccarney:
I am going to force-push an amended commit that is rebased on the latest 1110 and resolves the merge conflicts.  There are no functional changes in the commit.

test/redfish-core/lib/power_subsystem_test.cpp
```diff
@@ -1,45 +0,0 @@
-#include "async_resp.hpp"
```
gtmills, line 1: What in the test fails with your code?

smccarney: The test exits due to a SIGSEGV:

```
  =================================== 42/43 ====================================
  test:         power_subsystem_test
  start time:   23:25:27
  duration:     0.01s
  result:       killed by signal 11 SIGSEGV
  command:
  UBSAN_OPTIONS=halt_on_error=1:abort_on_error=1:print_summary=1:print_stacktrace=1
  ASAN_OPTIONS=halt_on_error=1:abort_on_error=1:print_summary=1                                                                                                                                                                               MALLOC_PERTURB_=242
  /home/shawnmm/ci_test_area/bmcweb/build/power_subsystem_test
  ----------------------------------- stdout -----------------------------------
  Running main() from /googletest-1.14.0/googletest/src/gtest_main.cc
  [==========] Running 1 test from 1 test suite.
  [----------] Global test environment set-up.
  [----------] 1 test from PowerSubsystemCollectionTest
  [ RUN      ]                                                                                                                                                                                                                                PowerSubsystemCollectionTest.PowerSubsystemCollectionStaticAttributesAreExpected
  ==============================================================================
```

We discussed this test over Slack in the last day or so.  It is a very trivial test, and it dies due to this signal when I added the new properties.  You had suggested I just delete it again like we did in 1050.

redfish-core/lib/power_subsystem.hpp
```diff
@@ -1,20 +1,78 @@
...
+            {
+                asyncResp->res.jsonValue["Allocation"]["AllocatedWatts"] =
+                    powerCap;
+            }
+            else
+            {
+                asyncResp->res.jsonValue["Allocation"]["AllocatedWatts"] =
+                    maxPowerCapValue;
+            }
+            asyncResp->res.jsonValue["Allocation"]["RequestedWatts"] =
```
gtmills, line 70: Can you explain this mapping to me? By the PDI definition and the Redfish definition, isn't clear to me why these map together?
```
            description: The Maximum supported PowerCap setting.
```


                "RequestedWatts": {
                    "description": "The potential power, in watt units, that the subsystem requests, which might be higher than the current level being consumed because the requested power includes a budget that the subsystem wants for future use.",

gtmills:
Is there a story or a commit for going upstream with this?

redfish-core/lib/power_subsystem.hpp
smccarney: This is the same mapping that we had in 1050/1060 and is what was requested by Derek Howard (Power team lead) and Martha Broyles (OCC and Power Management team lead):

Set the Redfish properties based on properties from the D-Bus xyz.openbmc_project.Control.Power.Cap interface:
* Redfish AllocatedWatts property:
  * Set to PowerCap property if PowerCapEnable is true
  * Set to MaxPowerCapValue property if PowerCapEnable is false
* Redfish RequestedWatts property:
  * Set to MaxPowerCapValue property

test/redfish-core/lib/power_subsystem_test.cpp
gtmills: It is trivial. Sure, it can be fixed next time.

redfish-core/lib/power_subsystem.hpp
smccarney: I think the idea is that MaxPowerCapValue (if it has a valid value) represents the max power that could be allocated.  If the PowerCap is enabled, then that is the amout of power that is actually available/allocated.  If the PowerCap is not enabled, then the max power is available/allocated.  RequestedWatts is set to the max power, I guess assuming we always want as much as possible.

smccarney:
> Is there a story or a commit for going upstream with this?

There is not a plan for an upstream commit in P11.  I think it would make sense to work on this for P12.  I think this would take some serious work because we'd need to get community agreement on the interface(s) to use and how to map them to Redfish.

redfish-core/lib/power_subsystem.hpp
gtmills: Ack.

redfish-core/lib/power_subsystem.hpp
```diff
@@ -1,20 +1,78 @@
...
 {
 
+inline void getPowerSubsystemAllocation(
+    const std::shared_ptr<bmcweb::AsyncResp>& asyncResp)
+{
+    sdbusplus::asio::getAllProperties(
+        *crow::connections::systemBus, "xyz.openbmc_project.Settings",
+        "/xyz/openbmc_project/control/host0/power_cap",
+        "xyz.openbmc_project.Control.Power.Cap",
+        [asyncResp](const boost::system::error_code& ec,
```
gtmills, line 31: #include <boost/system/error_code.hpp>

gtmills (approved):

redfish-core/lib/power_subsystem.hpp
smccarney: Good catch I missed that include.  It builds OK, so I think given that it is downstream and testing has been done I'm going to leave as is.  However if I have to make any other changes I'll fix that too.

baemyung (approved):
Based on 1050 code - https://github.com/ibm-openbmc/bmcweb/pull/654, the code for 1110 seems good to me.

PR #1059, opened 2024-09-05
Ignore all ChapData errors

gtmills:
Have seen pldm send back errors in cases like it is powering off. Just ignore all ChapData errors. The HMC doesn't use ChapData. We will still return a 500 here in 1050/1060 but for 1110, let's just ignore all errors.

Fixes defect 642324

### DIFF
--- a/redfish-core/lib/systems.hpp
+++ b/redfish-core/lib/systems.hpp
@@ -2281,11 +2281,8 @@ inline void getChapData(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp)
                     const dbus::utility::DBusPropertiesMap& propertiesList) {
         if (ec)
         {
-            if (ec.value() != EBADR)
-            {
-                BMCWEB_LOG_ERROR("Get ChapData D-bus error: {}", ec.value());
-                messages::internalError(asyncResp->res);
-            }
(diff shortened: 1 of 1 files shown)


redfish-core/lib/systems.hpp
```diff
@@ -2281,11 +2281,7 @@ inline void getChapData(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp)
                     const dbus::utility::DBusPropertiesMap& propertiesList) {
         if (ec)
         {
-            if (ec.value() != EBADR)
-            {
-                BMCWEB_LOG_ERROR("Get ChapData D-bus error: {}", ec.value());
-                messages::internalError(asyncResp->res);
-            }
+            // ChapData isn't that important. Ignore all errors.
```
baemyung, line 2284: Can we leave a DEBUG message here?
In the future (when bmcweb dynamic logging is included), this part can be debugged if needed.

baemyung (approved):
It looks like `clang-format-18` would need to be applied from now on.

Other than a minor comment, the change looks good to me.

redfish-core/lib/systems.hpp
gtmills: Done. I left the logging in but moved it to DEBUG

PR #1063, opened 2024-09-10
Fix RemoteRoleMap PATCH operation

gtmills:
RemoteRoleMap "LocalRole" property update fails since there was conversion missing from redfish privilege to D-bus values

Looks like this commit dropped this change
https://gerrit.openbmc.org/c/openbmc/bmcweb/+/64325/

This commit fixes this issue

Tested by:
Verified patch operation on RemoteRoleMap

Change-Id: Ic05aa3457a45e98ea5dc8e9dd83e0f1a42772070


raviteja-b (approved):

baemyung (approved):
It looks good

PR #1066, opened 2024-09-12
Downgrade INFO logs to just that

gtmills:
BMCWEB_LOG_INFO vs BMCWEB_LOG_CRITICAL.

Fixes 613773

### DIFF
--- a/include/dump_offload.hpp
+++ b/include/dump_offload.hpp
@@ -185,8 +185,8 @@ class Handler : public std::enable_shared_from_this<Handler>
                 }
                 return;
             }
-            BMCWEB_LOG_CRITICAL("INFO: Reset OffloadUri of {} dump id {}",
-                                dumpType, entryID);
+            BMCWEB_LOG_INFO("Reset OffloadUri of {} dump id {}", dumpType,
+                            entryID);
         },
             "xyz.openbmc_project.Dump.Manager",
             "/xyz/openbmc_project/dump/" + dumpType + "/entry/" + entryID,
@@ -287,7 +287,7 @@ class Handler : public std::enable_shared_from_this<Handler>
                     this->connection->close();
                     return;
                 }
-                BMCWEB_LOG_CRITICAL("INFO: Hit Dump end of file");
+                BMCWEB_LOG_INFO("Hit Dump end of file");
                 this->connection->completionStatus = true;
                 this->connection->close();
                 return;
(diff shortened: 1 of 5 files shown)


baemyung (approved):
I agree to change those logs as INFO.

include/dump_offload.hpp
```diff
@@ -185,8 +185,8 @@ class Handler : public std::enable_shared_from_this<Handler>
                 }
                 return;
             }
-            BMCWEB_LOG_CRITICAL("INFO: Reset OffloadUri of {} dump id {}",
-                                dumpType, entryID);
+            BMCWEB_LOG_INFO("INFO: Reset OffloadUri of {} dump id {}", dumpType,
```
jeaaustx, line 188: Doesn't the BMCWEB_LOG_INFO() already insert "INFO" into the message in the log? Seems like the "INFO:" can be removed from these messages now that they are not output as CRITICAL any more. (I.e. I think it was part of the string before to clarify that it wasn't actually a critical error.)

gtmills: Yeah, that is fair. 
I'll remove the INFO:

raviteja-b:
I dont agree to remove these traces, all these traces are required to debug various issues which we have discussed earlier multiple times.

include/dump_offload.hpp
raviteja-b: INFO logs can't be seen in bmc journal, there is no point in logging INFO traces

gtmills: INFO logs can be seen if you increase bmcweb logging level

gtmills: I added this to the agenda for Monday. 
I am going to wait here to remove the "INFO:" until we discuss

gtmills:
Slack discussion at https://ibm-systems-power.slack.com/archives/G3WAVQYTH/p1726246893628949?thread_ts=1726239559.931879&cid=G3WAVQYTH

include/dump_offload.hpp
gtmills: As discussed in slack, going to go with this

gtmills:
> I dont agree to remove these traces, all these traces are required to debug various issues which we have discussed earlier multiple times.

Discussed in slack, we are going to go with this to fix this defect.

raviteja-b (approved):

PR #1071, opened 2024-09-18
Make everything 0.6. for the OpenBMC message registry

gtmills:
https://github.com/ibm-openbmc/bmcweb/commit/a09d6d12e1c499672c4feb314ecbd5bef3d06840 pushed several things to 0.6. but forgot a few. We will go with 0.6. for 1110.00, even if we add another message. We can control when we release so we don't need to bump for every message like upstream. Upstream is 0.5 so it makes sense we are 1 version above since we added a few messages.


jeaaustx (approved):
Looks good to me!

baemyung (approved):
It looks good to me

PR #1077, opened 2024-10-09
Audit Log: Use HttpBody to download FullAudit attachment

jeaaustx:
HttpBody has added support to use unix_fd to read the data to be downloaded. Update the handling of FullAudit attachment download to use this approach.

```
curl -k -H "X-Auth-Token: ${token}" -X GET https://${bmc}/redfish/v1/Systems/system/LogServices/AuditLog/Entries/FullAudit/attachment
```
Upstream commit which added the unix_fd support and modified the BMC dump to use it: [67667](https://gerrit.openbmc.org/c/openbmc/bmcweb/+/67667)

Tested:
 - Used version of phosphor-auditlog to always return the same full audit log file, size 10MB. Ran curl command to download the file in a loop 500 times on hardware and confirmed downloaded file matched expected data on all iterations.

### DIFF
--- a/redfish-core/lib/log_services.hpp
+++ b/redfish-core/lib/log_services.hpp
@@ -6275,10 +6275,19 @@ inline void getFullAuditLogAttachment(
         return;
     }
 
-    long long int size = lseek(fd, 0, SEEK_END);
+    off_t size = lseek(fd, 0, SEEK_END);
     if (size == -1)
     {
         BMCWEB_LOG_ERROR("Failed to get size of fd {}", fd);
+        messages::internalError(asyncResp->res);
+        close(fd);
+        return;
+    }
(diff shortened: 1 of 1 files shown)


redfish-core/lib/log_services.hpp
```diff
@@ -6275,14 +6275,23 @@ inline void getFullAuditLogAttachment(
         return;
     }
 
-    long long int size = lseek(fd, 0, SEEK_END);
+    off_t size = lseek(fd, 0, SEEK_END);
     if (size == -1)
```
baemyung, line 6279: This new logic seems appropriate and correct - https://pubs.opengroup.org/onlinepubs/9699919799.2016edition/functions/lseek.html

baemyung (approved):
It looks good to me

PR #1081, opened 2024-10-17
Delete IPv4 default gateway when deleting an IPv4 static address

raviteja-b:
The Redfish schema for creating static IPv4 addresses requires the IP address, the netmask, and a gateway IP address. There's an issue inherent with this method. A network interface is only permitted a single IPv4 default gateway. If more than one IPv4 static address is assigned to the NIC each entry is processed, and potentially conflicting default gateways may be assigned. The last entry processed assigns the IPv4 default gateway. This behavior will cause unexpected results. It is necessary to prevent assigning mismatched default gateway values.

The IPv4 address removal process requires additional work also. The default gateway value is left in place even after the final static IPv4 address is removed. It is necessary to perform an additional action to clear the gateway address. Without explicit removal the network is left in a condition that may prevent IP traffic from being able to be sent from the BMC. This even in the event that the NIC is actively being managed via DHCPv4.

Tested:
Disabled DHCPv4 on a secondary NIC (eth1)
Assigned a static IPv4 address.
Inspected the systemd-networkd config file in order to confirm the Gateway entry is added. This is done to be explicitly sure the network.config file has the Gateway entry.
Sent a Redfish PATCH command to delete the static IPv4 address. Confirmed that the systemd-networkd config file no longer contained a Gateway entry. This is done to be explicitly sure the network.config file no longer contains the Gateway entry.
Created a PATCH containing multiple IPv4 static addresses all with different Gateway values. Confirmed an error is returned when a mismatch occurs in the Gateway values.
Assigned a new static address, and then restored DHCPv4. Confirmed that the default gateway entry in the config file is removed. Submitted a delete request for the remaining static IPv4 address that is now orphaned by re-enabling DHCPv4. This removed the static IPv4 address.

Change-Id: Ia12cf2a38ba86266ce71dc28475b0d07b7e09ebc

### DIFF
--- a/redfish-core/lib/ethernet.hpp
+++ b/redfish-core/lib/ethernet.hpp
@@ -45,6 +45,12 @@ enum class LinkType
     Global
 };
 
+enum class IpVersion
+{
+    IpV4,
+    IpV6
+};
+
 /**
  * Structure for keeping IPv4 data required by Redfish
  */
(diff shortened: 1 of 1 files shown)


gtmills:
Are there upstream commits here?

baemyung (approved):
It seems this is a cherry-pick of upstream commits

- https://gerrit.openbmc.org/c/openbmc/bmcweb/+/70577
- https://gerrit.openbmc.org/c/openbmc/bmcweb/+/71611

Please add this into PR commit message for the future references.

Other than that & if CI passes, it looks good to me.

raviteja-b:
> It seems this is a cherry-pick of upstream commits
> 
> * https://gerrit.openbmc.org/c/openbmc/bmcweb/+/70577
> * https://gerrit.openbmc.org/c/openbmc/bmcweb/+/71611
> 
> Please add this into PR commit message for the future references.
> 
> Other than that & if CI passes, it looks good to me.

pulling these commits is not working for 1110, I am still working on this PR

gtmills:
This is just pulling in 

    https://gerrit.openbmc.org/c/openbmc/bmcweb/+/70577
    https://gerrit.openbmc.org/c/openbmc/bmcweb/+/71611

and making work with downstream?

gtmills (approved):

PR #1086, opened 2024-10-28
mfa: Add right AccountService schema version

asmithakarun:
This commit adds the correct AccountService schema version (AccountService_v1_15_0.*) and fixes the redfish validator failure. It also fixes a bug where "MFABypass" user-level option was not displayed in the redfish response.

Tested By:
[1] GET https://${bmc}/redfish/v1/AccountService/Accounts/testmfaadminA

Redfish Validator passed.

Change-Id: I406e7113c3b6ceae9396bb5d9f0d9249299d6267

### DIFF
--- a/redfish-core/lib/account_service.hpp
+++ b/redfish-core/lib/account_service.hpp
@@ -1662,8 +1662,7 @@ inline void
 
     nlohmann::json& json = asyncResp->res.jsonValue;
     json["@odata.id"] = "/redfish/v1/AccountService";
-    json["@odata.type"] = "#AccountService."
-                          "v1_10_0.AccountService";
+    json["@odata.type"] = "#AccountService.v1_15_0.AccountService";
     json["Id"] = "AccountService";
     json["Name"] = "Account Service";
     json["Description"] = "Account Service";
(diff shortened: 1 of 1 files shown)


raviteja-b (approved):

baemyung:
Is this a cherry-pick of  https://gerrit.openbmc.org/c/openbmc/bmcweb/+/74667?
If so, would you add "Change-Id: I406e7113c3b6ceae9396bb5d9f0d9249299d6267" to the commit message so that we can associate this later?

redfish-core/lib/account_service.hpp
```diff
@@ -1663,7 +1663,7 @@ inline void
     nlohmann::json& json = asyncResp->res.jsonValue;
     json["@odata.id"] = "/redfish/v1/AccountService";
     json["@odata.type"] = "#AccountService."
-                          "v1_10_0.AccountService";
+                          "v1_15_0.AccountService";
```
baemyung, line 1666: This would be better to be combined in a string
```
    json["@odata.type"] = "#AccountService.v1_15_0.AccountService";
```

baemyung (commented):
Basically, this is to update "@odata.type" to "#AccountService.v1_15_0.AccountService"?
I had a comment to make it as a single string.

redfish-core/lib/account_service.hpp
asmithakarun: done

baemyung (approved):
It looks good to me, once CI passes.

PR #1093, opened 2024-11-11
MFA-Bypass: Disable service user for MFA bypass

abhilashraju:
MFA-Bypass option is disabled for service user.
Service user will be always bypassed for MFA and the patch operation should be disabled from  altering it.

### DIFF
--- a/redfish-core/lib/account_service.hpp
+++ b/redfish-core/lib/account_service.hpp
@@ -1349,6 +1349,11 @@ inline void updateUserProperties(
             // Check if the input vector contains just one bypass type:
             // as there are just two defined in the backend:
             // GoogleAuthenticator  and None
+            if (username == "service")
+            {
+                messages::operationNotAllowed(asyncResp->res);
+                return;
+            }
             if (mfaBypass->size() > 1)
             {
                 messages::propertyNotWritable(asyncResp->res,


redfish-core/lib/account_service.hpp
```diff
@@ -1349,6 +1349,11 @@ inline void updateUserProperties(
             // Check if the input vector contains just one bypass type:
             // as there are just two defined in the backend:
             // GoogleAuthenticator  and None
+            if (username == "service")
```
raviteja-b, line 1352: can you please add service user checks to GenerateSecretKey and VerifyTOTP redfish operations??

abhilashraju: why? will not it be automatically covered by the bypass property? If bypass enabled generate secret key will not work right?

raviteja-b: what's error returned for generated secret key and verify totp operation for service ??

abhilashraju: But we should have same error for any user who are bypassed and yet try to create secret key right? Why special error for service user?

raviteja-b: then its going to return internal server error

gtmills (approved):

PR #1101, opened 2024-11-25
1110: Rebase: json utility: fixed core dump during sensor load

jeaaustx:
Pulling in an upstream commit to fix a problem with objectKeyCmp() which causes bmcweb to core dump. Haven't seen this core dump downstream but thought it would be safer to pull in the fix since we are using this function:
 - https://gerrit.openbmc.org/c/openbmc/bmcweb/+/75867

bmcweb replaces underscores with spaces in sensor names for better readability. The existing objectKeyCmp function did not handle this case, leading to core dumps in the sensor load path.

Error details are provided below.

```
bmcwebd[1368]: [DEBUG sensors.hpp:507] Added sensor P0_NS_VR_FAN_2
bmcwebd[1368]: terminate called after throwing an instance of
       'boost::detail::with_throw_location<boost::system::system_error>'
bmcwebd[1368]:   what():  leftover [boost.url.grammar:4]
```

Implemented a new algorithm that alphabetically sorts non-URL keys and retains the existing logic for URL-type keys.

Tested: Updated and verified the test cases.

Change-Id: I39c3f7cc54dec5e7cf9658977e1078acb827afb2

### DIFF
--- a/redfish-core/include/utils/json_utils.hpp
+++ b/redfish-core/include/utils/json_utils.hpp
@@ -22,6 +22,9 @@
 #include "human_sort.hpp"
 #include "logging.hpp"
 
+#include <boost/system/result.hpp>
+#include <boost/url/parse.hpp>
+#include <boost/url/url_view.hpp>
 #include <nlohmann/json.hpp>
 
 #include <algorithm>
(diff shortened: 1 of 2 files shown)


baemyung (approved):
Yes, I agree to pull this

PR #1109, opened 2024-12-05
Correct errors for spare core overwrite

gtmills:
Three commits here to return a better error when resolve/delete error log is unavailable due to active gard record. This is part of the spare cores overwrite story. 

The 1st commit is  [Switch resolved to use SetProperty](https://github.com/ibm-openbmc/bmcweb/commit/ad5b2a67cd6032189c065e597ed044d7c8f66b87). This follows patterns elsewhere and is already done upstream at https://github.com/openbmc/bmcweb/blob/master/redfish-core/lib/log_services.hpp#L1806. There is a lot of downstream commits in this area so pulling upstream wasn't possible 
2nd commit is [Map Error.Unavailable to something better](https://github.com/ibm-openbmc/bmcweb/commit/1525466aaf0629f9f609c4105176b7aa0d4107cf). Upstream at https://gerrit.openbmc.org/c/openbmc/bmcweb/+/75570 
3rd commit is [Add specific error on Delete](https://github.com/ibm-openbmc/bmcweb/commit/e97723a0a8a73428d63799ee6cd1c3407c2a118f). Haven't seen us do this for detete, pending 75570 

Tested: Nikhil tested on the GUI and said this worked

### DIFF
--- a/redfish-core/lib/log_services.hpp
+++ b/redfish-core/lib/log_services.hpp
@@ -2149,7 +2149,8 @@ inline void
 {
     // Process response from Logging service.
     auto respHandler = [asyncResp,
-                        entryId](const boost::system::error_code& ec) {
+                        entryId](const boost::system::error_code& ec,
+                                 const sdbusplus::message::message& msg) {
         BMCWEB_LOG_DEBUG("EventLogEntry (DBus) doDelete callback: Done");
         if (ec)
         {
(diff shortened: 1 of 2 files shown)


redfish-core/lib/log_services.hpp
```diff
@@ -2167,10 +2159,26 @@ inline void
...
+            if (std::string_view(
+                    "xyz.openbmc_project.Common.Error.Unavailable") ==
+                dbusError->name)
+            {
+                messages::propertyValueExternalConflict(asyncResp->res,
+                                                        "LogEntry", "Delete");
+                return;
+            }
+
             BMCWEB_LOG_ERROR("EventLogEntry (DBus) doDelete "
```
jeaaustx, line 2178: Do we want to move the BMCWEB_LOG_ERROR() up so we will see it regardless of if the dbusError() provided anything else? (Or at add another BMCWEB_LOG_ERROR() if dbusError == nullptr?

jeaaustx (commented):
Just one minor comment

baemyung (approved):
Based on upstreaming commits, this PR seems good for 1110 and to me.

redfish-core/lib/log_services.hpp
gtmills: So, `if (dbusError == nullptr)` hits, we have the internalError, which is already in the journal. I feel that adding an additional trace there doesn't buy us anything. 

`if (ec.value() == EBADR)` or ` if (std::string_view("xyz.openbmc_project.Common.Error.Unavailable") == dbusError->name)` hit, they are 400 errors / client errors and based on https://github.com/openbmc/bmcweb/blob/06b71375b21fedf20c189965cb4e6925dfcbb044/DEVELOPING.md?plain=1#L227 
if there is an trace there, it should be a warning, I don't see a warning trace there helps us because we already have a unique return code

gtmills: Let me know if you want to discuss more

PR #1113, opened 2024-12-11
Update to use power cap limits from CapLimits instead of Cap

cjcain:
The power cap limits were previously hosted by Settings, but those are system settings and the limits are not really settings (users can not adjust these).

A new dbus interface will host these power cap limits:
  xyz.openbmc_project.Control.Power.CapLimits

Tested on Rainier:
'''
=> GET /redfish/v1/Chassis/chassis/EnvironmentMetrics ...
    "PowerLimitWatts": {
        "AllowableMax": 2777,
        "AllowableMin": 1286,
        "ControlMode": "Disabled",
        "SetPoint": 2500
    },

==> GET /redfish/v1/Chassis/chassis/PowerSubsystem (no powercap enabled) ...
    "Allocation": {
        "AllocatedWatts": 2777,
        "RequestedWatts": 2777
    },

== Enable 2121W power cap ==

=> GET /redfish/v1/Chassis/chassis/EnvironmentMetrics ...
    "PowerLimitWatts": {
        "AllowableMax": 2777,
        "AllowableMin": 1279,
        "ControlMode": "Automatic",
        "SetPoint": 2121
    },

==> GET /redfish/v1/Chassis/chassis/PowerSubsystem ...
    "Allocation": {
        "AllocatedWatts": 2121,
        "RequestedWatts": 2777
    },
'''
Ran validator and no failures for PowerSubsystem or EnvironmentMetrics

Change-Id: Ib501b3511421fafd4cae28126705400704f07a72

### DIFF
--- a/redfish-core/lib/environment_metrics.hpp
+++ b/redfish-core/lib/environment_metrics.hpp
@@ -308,6 +308,45 @@ inline void
     getPowerLimitWatts(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp)
 {
     sdbusplus::asio::getAllProperties(
+        *crow::connections::systemBus, "org.open_power.OCC.Control",
+        "/xyz/openbmc_project/control/host0/power_cap_limits",
+        "xyz.openbmc_project.Control.Power.CapLimits",
+        [asyncResp](const boost::system::error_code& ec,
+                    const dbus::utility::DBusPropertiesMap& propertiesList) {
+            if (ec)
+            {
+                if (ec.value() != EBADR)
+                {
+                    BMCWEB_LOG_ERROR("DBUS response error: {}", ec.value());
+                    messages::internalError(asyncResp->res);
+                }
+                return;
+            }
+
+            const uint32_t* minCap = nullptr;
+            const uint32_t* maxCap = nullptr;
+            const bool success = sdbusplus::unpackPropertiesNoThrow(
+                dbus_utils::UnpackErrorPrinter(), propertiesList,
+                "MinPowerCapValue", minCap, "MaxPowerCapValue", maxCap);
+            if (!success)
+            {
+                messages::internalError(asyncResp->res);
+                return;
+            }
+
+            if (minCap != nullptr)
+            {
+                asyncResp->res.jsonValue["PowerLimitWatts"]["AllowableMin"] =
+                    *minCap;
+            }
+            if (maxCap != nullptr)
+            {
+                asyncResp->res.jsonValue["PowerLimitWatts"]["AllowableMax"] =
+                    *maxCap;
+            }
+        });
+
+    sdbusplus::asio::getAllProperties(
         *crow::connections::systemBus, "xyz.openbmc_project.Settings",
         "/xyz/openbmc_project/control/host0/power_cap",
         "xyz.openbmc_project.Control.Power.Cap",
@@ -329,13 +368,10 @@ inline void
 
             const uint32_t* powerCap = nullptr;
             const bool* powerCapEnable = nullptr;
-            const uint32_t* minCap = nullptr;
-            const uint32_t* maxCap = nullptr;
 
             const bool success = sdbusplus::unpackPropertiesNoThrow(
                 dbus_utils::UnpackErrorPrinter(), propertiesList, "PowerCap",
-                powerCap, "PowerCapEnable", powerCapEnable, "MinPowerCapValue",
-                minCap, "MaxPowerCapValue", maxCap);
+                powerCap, "PowerCapEnable", powerCapEnable);
 
             if (!success)
             {
@@ -353,18 +389,6 @@ inline void
             {
                 asyncResp->res.jsonValue["PowerLimitWatts"]["ControlMode"] =
                     "Disabled";
-            }
-
-            if (minCap != nullptr)
-            {
-                asyncResp->res.jsonValue["PowerLimitWatts"]["AllowableMin"] =
-                    *minCap;
-            }
-
-            if (maxCap != nullptr)
-            {
-                asyncResp->res.jsonValue["PowerLimitWatts"]["AllowableMax"] =
-                    *maxCap;
             }
         });
 }
--- a/redfish-core/lib/power_subsystem.hpp
+++ b/redfish-core/lib/power_subsystem.hpp
@@ -25,6 +25,26 @@ namespace redfish
 inline void getPowerSubsystemAllocation(
     const std::shared_ptr<bmcweb::AsyncResp>& asyncResp)
 {
+    // Get max from CapLimit
+    sdbusplus::asio::getProperty<uint32_t>(
+        *crow::connections::systemBus, "org.open_power.OCC.Control",
+        "/xyz/openbmc_project/control/host0/power_cap_limits",
+        "xyz.openbmc_project.Control.Power.CapLimits", "MaxPowerCapValue",
+        [asyncResp](const boost::system::error_code& ec, uint32_t maxCap) {
+            if (ec)
+            {
+                if (ec.value() != EBADR)
+                {
+                    BMCWEB_LOG_ERROR("DBUS response error: {}", ec.value());
+                    messages::internalError(asyncResp->res);
+                }
+                return;
+            }
+            asyncResp->res.jsonValue["Allocation"]["AllocatedWatts"] = maxCap;
+            asyncResp->res.jsonValue["Allocation"]["RequestedWatts"] = maxCap;
+        });
+
+    // Get power cap enable status and the cap (if set)
     sdbusplus::asio::getAllProperties(
         *crow::connections::systemBus, "xyz.openbmc_project.Settings",
         "/xyz/openbmc_project/control/host0/power_cap",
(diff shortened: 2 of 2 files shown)


cjcain:
This commit goes with the occ-control change to add CapLimits:
Add CapLimits dbus interface to occ-control app - https://gerrit.openbmc.org/c/openbmc/openpower-occ-control/+/58066
Update persistent power cap data filename - https://gerrit.openbmc.org/c/openbmc/openpower-occ-control/+/76437

redfish-core/lib/power_subsystem.hpp
```diff
@@ -25,6 +25,47 @@ namespace redfish
 inline void getPowerSubsystemAllocation(
     const std::shared_ptr<bmcweb::AsyncResp>& asyncResp)
 {
+    // Get max from CapLimits
+    sdbusplus::asio::getAllProperties(
+        *crow::connections::systemBus, "org.open_power.OCC.Control",
+        "/xyz/openbmc_project/control/host0/power_cap_limits",
+        "xyz.openbmc_project.Control.Power.CapLimits",
+        [asyncResp](const boost::system::error_code& ec,
+                    const dbus::utility::DBusPropertiesMap& propertiesList) {
```
jeaaustx, line 34: Similar to above the long lambda and hard-coded path will likely not be accepted upstream.

But right now this is not yet implemented upstream. There is an abandoned commit that started trying to do this: https://gerrit.openbmc.org/c/openbmc/bmcweb/+/68442

If we revive that commit we will have to restructure this code to merge upstream. I'm okay with addressing both of these if/when that is done.

redfish-core/lib/power_subsystem.hpp
```diff
@@ -25,6 +25,47 @@ namespace redfish
...
+            {
+                if (ec.value() != EBADR)
+                {
+                    BMCWEB_LOG_ERROR("DBUS response error: {}", ec.value());
+                    messages::internalError(asyncResp->res);
+                }
+                return;
+            }
+
+            const uint32_t* minCap = nullptr;
```
jeaaustx, line 45: minCap is not being used here, please remove.

redfish-core/lib/power_subsystem.hpp
```diff
@@ -25,6 +25,47 @@ namespace redfish
...
+            const uint32_t* maxPowerCapValue = nullptr;
+            const bool success = sdbusplus::unpackPropertiesNoThrow(
```
jeaaustx, line 47: nit: Could just get the single MaxPowerCapValue here instead of all properties if no need of minCap

redfish-core/lib/power_subsystem.hpp
```diff
@@ -25,6 +25,47 @@ namespace redfish
...
+                messages::internalError(asyncResp->res);
+                return;
+            }
+
+            // If MaxPowerCapValue valid, store Allocation properties in JSON
+            if ((maxPowerCapValue != nullptr) &&
+                (*maxPowerCapValue < UINT32_MAX))
+            {
+                asyncResp->res.jsonValue["Allocation"]["AllocatedWatts"] =
+                    *maxPowerCapValue;
```
jeaaustx, line 62: If the following getAllProperties() below fails we will be returning these results now where we used to not return them since we only had a single call.

For the "AllocatedWatts" entry this could be an incorrect value if we actually do have PowerCapEnable on. From [Redfish schema](https://redfish.dmtf.org/schemas/v1/PowerSubsystem.v1_1_2.json):
`This property shall represent the total power currently allocated or budgeted to this subsystem.`

Seems like we might need to chain these getProperty calls together and fill in the response in a single function.

redfish-core/lib/power_subsystem.hpp
```diff
@@ -41,35 +82,26 @@ inline void getPowerSubsystemAllocation(
...
                 return;
             }
 
-            // If MaxPowerCapValue valid, store Allocation properties in JSON
-            if ((maxPowerCapValue > 0) && (maxPowerCapValue < UINT32_MAX))
+            if ((powerCap != nullptr) && (powerCapEnable != nullptr))
             {
-                if (powerCapEnable)
+                if ((*powerCapEnable) && (*powerCap > 0) &&
+                    (*powerCap < UINT32_MAX))
```
jeaaustx, line 100: Since this is GET are the powerCap value checks necessary here? Should we just be returning what we got from D-Bus?

Since the variable itself is a uint32 it cannot be larger than UINT32_MAX can it?

Also the [D-Bus interface](https://github.com/openbmc/phosphor-dbus-interfaces/blob/bfcb5d0c39571cd5f0215d22e9f7fee0c23dfd59/yaml/xyz/openbmc_project/Control/Power/Cap.interface.yaml#L5) doesn't say anything about a minimum value.

jeaaustx (commented):
Thank you for updating these downstream paths to use the new interface. I have a couple of comments.

redfish-core/lib/power_subsystem.hpp
```diff
@@ -25,6 +25,47 @@ namespace redfish
...
+                maxPowerCapValue);
+            if (!success)
+            {
+                messages::internalError(asyncResp->res);
+                return;
+            }
+
+            // If MaxPowerCapValue valid, store Allocation properties in JSON
+            if ((maxPowerCapValue != nullptr) &&
+                (*maxPowerCapValue < UINT32_MAX))
```
baemyung, line 59: I think this is better to use numeric_limits like
```
(*maxPowerCapValue < std::numeric_limits<uint32_t>::max()))
```

redfish-core/lib/environment_metrics.hpp
```diff
@@ -307,6 +307,45 @@ inline void
 inline void
     getPowerLimitWatts(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp)
 {
+    sdbusplus::asio::getAllProperties(
+        *crow::connections::systemBus, "org.open_power.OCC.Control",
+        "/xyz/openbmc_project/control/host0/power_cap_limits",
+        "xyz.openbmc_project.Control.Power.CapLimits",
+        [asyncResp](const boost::system::error_code& ec,
+                    const dbus::utility::DBusPropertiesMap& propertiesList) {
```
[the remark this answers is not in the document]
cjcain: Agree. This change was just to get functionality updated for 110.

redfish-core/lib/power_subsystem.hpp
cjcain: done

redfish-core/lib/power_subsystem.hpp
cjcain: I tried to chain them, but was not able to get it working.

redfish-core/lib/power_subsystem.hpp
cjcain: Done

redfish-core/lib/power_subsystem.hpp
cjcain: Good point on the variable checks.  Done.

redfish-core/lib/power_subsystem.hpp
cjcain: Since this is already defined as uint32, check is not really needed. Done.

redfish-core/lib/power_subsystem.hpp
cjcain: Since that is an error case, we would be allocating the full amount which is safer (system will not be consuming more than that). Seems better to overreport power allocation vs not reporting anything. If it triggers customer warning that is probably fine, because there is a problem reading the current power cap / enable status.

jeaaustx: I'm okay with this for downstream. I'm sure for upstream we will have to figure out how to chain these together for it to get merged there.

I'd like @gtmills to weigh in on the partial reporting aspect.

jeaaustx (commented):
I'm okay with this for downstream. There will be changes necessary for upstream to get it merged I think. It does have different semantic behavior on error. But I think the need to get this working downstream with the new interface outweighs that.

cjcain:
FYI, This PR will need to go with the occ-control bump request (commits above)

baemyung:
What is the bmcweb upstream commit?

baemyung (approved):
I also think it seems okay for downstream, esp, to move forward.
Upstream may keep working on.

redfish-core/lib/power_subsystem.hpp
```diff
@@ -41,35 +61,25 @@ inline void getPowerSubsystemAllocation(
...
-                {
-                    asyncResp->res.jsonValue["Allocation"]["AllocatedWatts"] =
-                        powerCap;
-                }
-                else
+                if (*powerCapEnable)
                 {
                     asyncResp->res.jsonValue["Allocation"]["AllocatedWatts"] =
-                        maxPowerCapValue;
+                        *powerCap;
```
jeaaustx, line 81: On further reflection I realize this won't be returning a consistent value for "AllocatedWatts". Both of the places which are filling in the value are asynchronous handlers. So there is no guarantee on the order that the value will be filled in.

Before I was thinking this was just an issue for the error path. But I actually believe it is an issue for the normal path as well. I do think these calls are going to have to be coordinated to assure a consistent value is returned.

redfish-core/lib/power_subsystem.hpp
```diff
@@ -25,6 +25,26 @@ namespace redfish
...
+            if (ec)
+            {
+                if (ec.value() != EBADR)
+                {
+                    BMCWEB_LOG_ERROR("DBUS response error: {}", ec.value());
+                    messages::internalError(asyncResp->res);
+                }
+                return;
+            }
+            asyncResp->res.jsonValue["Allocation"]["AllocatedWatts"] = maxCap;
```
jeaaustx, line 43: On further reflection I realize this won't be returning a consistent value for "AllocatedWatts". Both of the places which are filling in the value are asynchronous handlers. So there is no guarantee on the order that the value will be filled in.

Before I was thinking this was just an issue for the error path. But I actually believe it is an issue for the normal path as well. I do think these calls are going to have to be coordinated to assure a consistent value is returned.

jeaaustx (changes requested):
I've had some reflection on the setting of "AllocatedWatts". I do think this should be addressed downstream.

redfish-core/lib/power_subsystem.hpp
cjcain: Would moving that 2nd async request inside the first resolve this?

PR #1116, opened 2024-12-19
network: fix dhcp values in redfish response

asmithakarun:
Whenever IPv6 SLAAC is enabled and DHCPv4 is enabled, the "DHCPEnabled" value in the backend will be "v4v6stateless". Currently in bmcweb, the check to translate this value to bool for ipv4 is not present and will return false. That means, "DHCPEnabled" is wrongly displayed 'false' while the correct value is 'true'.

Upstream: https://gerrit.openbmc.org/c/openbmc/bmcweb/+/76560

Change-Id: I8713d73727c6a382f06b7bf0d598ab61a757e1e3


raviteja-b (approved):

baemyung (approved):
It looks good to me

PR #1118, opened 2025-01-01
Fix setting gateways

cmjishnu:
There's a number of conditions in setting gateways that don't work properly.  Specifically, one of the issues is setting a gateway on an address that already exists.  It returns a PropertyValueConflict error on Ipv4Addresses/1/Gateway with Ipv4Addresses/1/Gateway

Obviously an address can't conflict with itself, so this is wrong.

To address this, move the gateway setting and selection code into a routine outside of the main loop, after all the gateways are accounted for, and so we can treat them separately.

Tested;
PATCH to an existing ip address works, and no longer returns the error.

Fixes : https://jazz07.rchland.ibm.com:13443/jazz/web/projects/CSSD#action=com.ibm.team.workitem.viewWorkItem&id=635827
Upstream : https://gerrit.openbmc.org/c/openbmc/bmcweb/+/75859
Change-Id: I0339e02fc27164337416637153d0b0f744b64ad8

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -412,6 +412,7 @@ srcfiles_unittest = files(
     'test/redfish-core/include/utils/stl_utils_test.cpp',
     'test/redfish-core/include/utils/time_utils_test.cpp',
     'test/redfish-core/lib/chassis_test.cpp',
+    'test/redfish-core/lib/ethernet_test.cpp',
     'test/redfish-core/lib/log_services_dump_test.cpp',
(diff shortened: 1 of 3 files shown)


baemyung (approved):
Looking at upstream https://gerrit.openbmc.org/c/openbmc/bmcweb/+/75859, this looks good to me

raviteja-b (approved):

PR #1120, opened 2025-01-06
Fix race condition updating AllocatedWatts from CapLimits

cjcain:
The existing code had 2 async calls updating the "AllocatedWatts" parameter. There is no guarantee that the values would be updated in same order everytime.

This change will ensure the 2nd async call happens after the first and the correct value will be returned for AllocatedWatts.

Change-Id: I3cf10e6ecb2ab3719034521f1e9d05dbe1c72936

### DIFF
--- a/redfish-core/lib/power_subsystem.hpp
+++ b/redfish-core/lib/power_subsystem.hpp
@@ -42,45 +42,48 @@ inline void getPowerSubsystemAllocation(
             }
             asyncResp->res.jsonValue["Allocation"]["AllocatedWatts"] = maxCap;
             asyncResp->res.jsonValue["Allocation"]["RequestedWatts"] = maxCap;
-        });
 
-    // Get power cap enable status and the cap (if set)
-    sdbusplus::asio::getAllProperties(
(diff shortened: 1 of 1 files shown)


redfish-core/lib/power_subsystem.hpp
```diff
@@ -42,45 +42,48 @@ inline void getPowerSubsystemAllocation(
...
+                        messages::internalError(asyncResp->res);
+                        return;
+                    }
+
+                    if ((powerCap != nullptr) && (powerCapEnable != nullptr))
+                    {
+                        if (*powerCapEnable)
+                        {
+                            asyncResp->res
+                                .jsonValue["Allocation"]["AllocatedWatts"] =
```
jeaaustx, line 82: For upstream I'm pretty sure the maintainer will want this only updated in a single location. Making note of it here so when I make these changes upstream I address it. No need to change it for downstream.

jeaaustx (approved):
Thank you for addressing the race condition. I've noted a change I'll need to make when I try to get this upstream, but I'm okay with this as-is for downstream.

gtmills (approved):

PR #1111, opened 2024-12-09
Dimm: Look at Pretty Name 

gtmills:
https://github.com/ibm-openbmc/bmcweb/pull/542 (1050/1060) https://github.com/ibm-openbmc/bmcweb/pull/918 (1110) added PrettyName for Processors, Assembly, Chassis, Storage but not Memory. I believe this was due to not needed it for Memory at the time. We want it now so the GUI can use on the inventory page.  (#973) pulled this in for 1060 but was missed for 1110. 
Fixes 667557

No upstream here, https://jsw.ibm.com/browse/PFEBMC-2871 tracks getting prettyname upstream 

Tested: 
```
curl -k https://$bmc/redfish/v1/Systems/system/Memory/dimm0
...
  "Name": "Memory module 0",
```

PR #918, opened 2024-04-20
Rebase Pretty-name

deepakala-k:
Cherry picked and resolved conflicts for the following commit.

[util: Add pretty name for resources](https://github.com/ibm-openbmc/bmcweb/pull/542/commits/39d15319d1ec2a4fa835c32b5e919c826c6b7913)

PR #1125, opened 2025-01-14
Power Supply, Fan: Look at PrettyName

gtmills:
Use getPrettyName like we do for Processors, Assembly, Chassis, Storage, Memory.

https://github.com/ibm-openbmc/bmcweb/pull/542 (1050/1060) https://github.com/ibm-openbmc/bmcweb/pull/918 (1110) added PrettyName for Processors, Assembly, Chassis, Storage but not PowerSupplies/Fan. I believe this was due to not needed at the time. https://github.com/ibm-openbmc/bmcweb/pull/1111 added PrettyName for Dimm. 

Fixes 668395

No upstream here, https://jsw.ibm.com/browse/PFEBMC-2871 tracks getting prettyname upstream

Tested: 
Before 
`"Name": "Power Supply",`
and 
`"Name": "Fan",`

After:
`"Name": "powersupply0",`
and 
`"Name": "fan1",`

### DIFF
--- a/redfish-core/lib/fan.hpp
+++ b/redfish-core/lib/fan.hpp
@@ -10,6 +10,7 @@
 #include "utils/chassis_utils.hpp"
 #include "utils/fan_utils.hpp"
 #include "utils/json_utils.hpp"
+#include "utils/name_utils.hpp"
 
 #include <boost/system/error_code.hpp>
 #include <boost/url/format.hpp>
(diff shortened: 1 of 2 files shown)


jinuthomas (approved):
LGTM

deepakala-k (approved):

redfish-core/lib/fan.hpp
```diff
@@ -349,6 +349,9 @@ inline void
     getFanAsset(asyncResp, fanPath, service);
     getFanLocation(asyncResp, fanPath, service);
     getLocationIndicatorActive(asyncResp, fanPath);
+    const dbus::utility::MapperServiceMap& serviceMatch = {{service, {""}}};
```
jeaaustx, line 352: I see that this will work, but it does with the understanding that getPrettyName() only uses the first element of the first pair.

For upstream it seems like handleFanPath() would need to be altered to pass the object to the callback function. Then the callback function would pull the fanPath for those related functions and pass the object to getPrettyName(). (Could also leave the setting of Name in addFanCommonProperties() if the object is passed on to it.

This is probably more change than we want in 1110?

jeaaustx (approved):
I have one comment on how I think this would have to be done upstream. But looks okay to me for 1110.

PR #1132, opened 2025-01-28
Fix log entry

deepakala-k:
Resolve the issue where LogEntry and Severity key-value pairs are being treated as separate elements instead of being structured under the status/Conditions hierarchy.

### DIFF
--- a/redfish-core/include/utils/error_log_utils.hpp
+++ b/redfish-core/include/utils/error_log_utils.hpp
@@ -5,6 +5,28 @@ namespace redfish
 namespace error_log_utils
 {
 
+static void getHiddenPropertyValue(
+    const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
+    const std::string& entryId, std::function<void(bool hidden)>&& callback)
+{
+    sdbusplus::asio::getProperty<bool>(
+        *crow::connections::systemBus, "xyz.openbmc_project.Logging",
+        "/xyz/openbmc_project/logging/entry/" + entryId,
+        "org.open_power.Logging.PEL.Entry", "Hidden",
+        [callback = std::move(callback), asyncResp,
+         entryId](const boost::system::error_code& ec, bool hidden) {
+            if (ec)
+            {
+                BMCWEB_LOG_ERROR(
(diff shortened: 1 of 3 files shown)


redfish-core/include/utils/error_log_utils.hpp
```diff
@@ -5,6 +5,26 @@ namespace redfish
...
+    sdbusplus::asio::getProperty<bool>(
+        *crow::connections::systemBus, "xyz.openbmc_project.Logging",
+        "/xyz/openbmc_project/logging/entry/" + entryId,
+        "org.open_power.Logging.PEL.Entry", "Hidden",
+        [callback{std::forward<Callback>(callback)},
+         asyncResp](const boost::system::error_code& ec, bool hidden) {
+            if (ec)
+            {
+                BMCWEB_LOG_ERROR("DBUS response error: {}", ec);
+                messages::internalError(asyncResp->res);
```
baemyung, line 21: Should this be causing an internal error - even if `Hidden` field is not defined?

redfish-core/include/utils/error_log_utils.hpp
```diff
@@ -5,6 +5,26 @@ namespace redfish
 namespace error_log_utils
 {
 
+template <typename Callback>
+void getHiddenPropertyValue(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
+                            const std::string& entryId, Callback&& callback)
```
baemyung, line 10: Instead of using `template<>`, I think it is better to use the explicit function proprotype  like
```
void getHiddenPropertyValue(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
                            const std::string& entryId,  std::function<void(bool hidden)>&& callback)
```

deepakala-k: addressed this comment

redfish-core/include/utils/error_log_utils.hpp
deepakala-k: I would leave it as is for now. Because that was the original implementation. May be I will discuss and see the implication and address it later

baemyung (approved):
It looks good now to me

dhruvibm (approved):

PR #1135, opened 2025-01-29
Fix 670598: dbus_util changes to fix response message

abhilashraju:
Taking latest upstream changes to fix multiple success message seen in response. Now the behaviour is similar to what upstream does, which mean there is no message in the response. The response is set with no_content flag.

Upstream change : https://github.com/openbmc/bmcweb/blob/master/redfish-core/src/utils/dbus_utils.cpp#L91

### DIFF
--- a/redfish-core/src/utils/dbus_utils.cpp
+++ b/redfish-core/src/utils/dbus_utils.cpp
@@ -10,6 +10,7 @@
 #include <boost/asio/error.hpp>
 #include <boost/beast/http/status.hpp>
 #include <boost/system/error_code.hpp>
+#include <nlohmann/json.hpp>
 #include <sdbusplus/message.hpp>
 
 #include <memory>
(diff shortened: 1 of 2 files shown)


redfish-core/src/utils/dbus_utils.cpp
```diff
@@ -73,8 +74,7 @@ void afterSetProperty(
             }
             if (errorName == "xyz.openbmc_project.Common.Error.Unavailable")
             {
-                messages::propertyValueExternalConflict(
-                    asyncResp->res, redfishPropertyName, propertyValue);
+                messages::resourceInStandby(asyncResp->res);
```
gtmills, line 77: This conflicts with another downstream patch

gtmills: This should stay this way

abhilashraju: I can put back. But before that, is it a recommended practise  to have down stream change in generic utility functionality ? can't we move the downstream changes to where it really  needed?

gtmills: There is a story to get this upstream; let's focus on that. Before this goes though, let's not overwrite the downstream patch here

abhilashraju: done

raviteja-b (approved):

raviteja-b:
please share upstream commit/code link

abhilashraju:
> please share upstream commit/code link

done

gtmills (approved):

PR #1139, opened 2025-02-03
MFA: Disallow bypass option to Readonly user

asmithakarun:
Currently, a readonly user is able to bypass MFA for self, which allows the user to continue with the session unauthorized.

This commit disallows a readonly user to bypass MFA when global MFA is enabled. Only users with admin privilege will be able to bypass MFA for self and for other users.

Tested By:

'''
 PATCH -D patch.txt -d '{"MFABypass":{"BypassTypes":["GoogleAuthenticator"]}}' https://${bmc}/redfish/v1/AccountService/Accounts/test_readonly -H "Content-Type: application/json"

This throws an "InsufficientPrivilege" error for readonly user. '''

### DIFF
--- a/redfish-core/lib/account_service.hpp
+++ b/redfish-core/lib/account_service.hpp
@@ -2574,6 +2574,14 @@ inline void
             return;
         }
 
+        // Read only users should not be allowed to bypass self
+        if (mfaBypass)
+        {
+            BMCWEB_LOG_ERROR("User has insufficient privilege to bypass self");
+            messages::insufficientPrivilege(asyncResp->res);
+            return;
+        }
+
         // ConfigureSelf accounts can only modify their password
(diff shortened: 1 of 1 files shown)


asmithakarun:
Fixes: 661774

redfish-core/lib/account_service.hpp
```diff
@@ -2574,6 +2574,14 @@ inline void
             return;
         }
 
+        // Read only users should not be allowed to bypass self
+        if (mfaBypass)
```
raviteja-b, line 2578: how does this check decides if its read only privilege or admin privilege user or not?
this check can hit with configure self privilege also right?

raviteja-b (approved):

baemyung (approved):
It looks good to me

PR #1141, opened 2025-02-05
MFA: Remove the session only when MFA is enabled

cmjishnu:
This commit includes the changes to remove other sessions of the same user on successful verification of TOTP only when MFA is enabled. 

Tested By:

1. Created multiple redfish sessions of the same user
2. Verified other sessions are cleared only when MFA is enabled.
3. Verified when MFA is disabled, other sessions of the same user is not cleared

Change-Id: I95dbecc107258836f3ec8f8f06b80cd6f91b6f79
Upstream : https://gerrit.openbmc.org/c/openbmc/bmcweb/+/74807

### DIFF
--- a/redfish-core/lib/account_service.hpp
+++ b/redfish-core/lib/account_service.hpp
@@ -2812,32 +2812,64 @@ inline void
 inline void
     verifyTotpDbusUtil(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
                        const std::string& totp, const std::string& userPath,
-                       const std::function<void(bool)>& callback)
+                       std::function<void(bool)>&& callback)
 {
-    crow::connections::systemBus->async_method_call(
-        [asyncResp,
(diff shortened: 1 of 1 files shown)


redfish-core/lib/account_service.hpp
```diff
@@ -2812,32 +2812,65 @@ inline void
...
+                    if (googleAuthEnabled)
+                    {
+                        messages::success(asyncResp->res);
+                        callback(true);
+                    }
+                    else
+                    {
+                        messages::success(asyncResp->res);
+                        callback(false);
+                    }
```
baemyung, line 2868: I think this is better to be written to shortcut `false` case like

```
                    if (!googleAuthEnabled)
                    {
                        messages::success(asyncResp->res);
                        callback(false);
                        return;
                    }
                    messages::success(asyncResp->res);
                     callback(true);
```

asmithakarun (approved):

PR #1143, opened 2025-02-07
Updates to resolve getting a response for wrong URI in Update

abiasojo:
Updates to resolve getting a response for wrong URI in Update
Service URI (/redfish/v1/UpdateService/FirmwareInventory/)

Tested with the correct URI and results are correct.
Tested with wrong URI and 404 is returned

### DIFF
--- a/redfish-core/lib/update_service.hpp
+++ b/redfish-core/lib/update_service.hpp
@@ -1227,13 +1227,10 @@ inline void requestRoutesSoftwareInventory(App& app)
             std::shared_ptr<std::string> swId =
                 std::make_shared<std::string>(param);
 
-            asyncResp->res.jsonValue["@odata.id"] = boost::urls::format(
-                "/redfish/v1/UpdateService/FirmwareInventory/{}", *swId);
-
             constexpr std::array<std::string_view, 1> interfaces = {
                 "xyz.openbmc_project.Software.Version"};
             dbus::utility::getSubTree(
-                "/", 0, interfaces,
+                "/xyz/openbmc_project/software/", 0, interfaces,
                 [asyncResp,
                  swId](const boost::system::error_code& ec,
                        const dbus::utility::MapperGetSubTreeResponse& subtree) {
@@ -1252,16 +1249,22 @@ inline void requestRoutesSoftwareInventory(App& app)
                                  std::string, std::vector<std::string>>>>& obj :
                          subtree)
                     {
-                        if (!obj.first.ends_with(*swId))
+                        sdbusplus::message::object_path path(obj.first);
+                        std::string id = path.filename();
+                        if (id.empty())
+                        {
+                            BMCWEB_LOG_DEBUG("Failed to find software id in {}",
+                                             obj.first);
+                            continue;
+                        }
+                        if (id != *swId)
                         {
                             continue;
                         }
-
                         if (obj.second.empty())
                         {
                             continue;
                         }
-
                         found = true;
                         sw_util::getSwStatus(asyncResp, swId,
                                              obj.second[0].first);
@@ -1281,6 +1284,9 @@ inline void requestRoutesSoftwareInventory(App& app)
                                 *swId));
                         return;
                     }
+                    asyncResp->res.jsonValue["@odata.id"] = boost::urls::format(
+                        "/redfish/v1/UpdateService/FirmwareInventory/{}",
+                        *swId);
                     asyncResp->res.jsonValue["@odata.type"] =
                         "#SoftwareInventory.v1_1_0.SoftwareInventory";
                     asyncResp->res.jsonValue["Name"] = "Software Inventory";


baemyung (commented):
It seems also needed to be upstreamed.

redfish-core/lib/update_service.hpp
```diff
@@ -1262,7 +1262,15 @@ inline void requestRoutesSoftwareInventory(App& app)
                             continue;
                         }
```
gtmills, line 1264: Instead of line 1255 - 1265 could we go with object_path as in: 
    
    sdbusplus::message::object_path path(obj.first);
    std::string id = path.filename(); 
    
    if (id.empty())
    {
         BMCWEB_LOG_DEBUG("Failed to find software id in {}", obj.first);
         continue;
    }
               
    if (id != *swId)
    {
      continue;
    }
    if (obj.second.empty())
    {
        continue; 
    }
    found = true;


https://github.com/openbmc/bmcweb/blob/80d2ef31c343ec9c562778848b9d62fb6ce57c07/redfish-core/include/utils/chassis_utils.hpp#L57

https://github.com/openbmc/bmcweb/blob/80d2ef31c343ec9c562778848b9d62fb6ce57c07/redfish-core/src/dbus_log_watcher.cpp#L118

redfish-core/lib/update_service.hpp
```diff
@@ -1262,7 +1262,15 @@ inline void requestRoutesSoftwareInventory(App& app)
...
 
-                        found = true;
+                        std::string swIdpath = "/xyz/openbmc_project/software/";
+                        if (obj.first == (swIdpath + *swId))
```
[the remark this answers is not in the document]
gtmills: dbus::utility::getSubTree( "/xyz/openbmc_project/software/" 
would help with performance, it is worth looking into

redfish-core/lib/update_service.hpp
baemyung: Reading README (https://github.com/openbmc/phosphor-dbus-interfaces/blob/master/yaml/xyz/openbmc_project/Software/README.md, or https://gerrit.openbmc.org/plugins/gitiles/openbmc/docs/+/refs/changes/80/24980/2/REST+REDFISH-cheatsheet.md) , the dbus path name for image seems like

```
/xyz/openbmc_project/software/<swId>/
```
Then, can we use `dbus::utility::checkDbusPathExists()` instead of `getSubTree()?

redfish-core/lib/update_service.hpp
```diff
@@ -1262,7 +1262,15 @@ inline void requestRoutesSoftwareInventory(App& app)
                             continue;
                         }
 
-                        found = true;
+                        std::string swIdpath = "/xyz/openbmc_project/software/";
```
jeaaustx, line 1265: Above on L1230-1231 we are setting the @odata.id value before doing the async D-Bus call. On failure that will be part of the response. I think for Redfish that is not supposed to be in the response on failure. Seems like it shouldn't be set until we have confirmed the swId is valid.

@gtmills Please confirm if my understanding is correct or not.

redfish-core/lib/update_service.hpp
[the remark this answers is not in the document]
jeaaustx: Example of using object_path type to match just the last part of a path:
[getValidChassisPath](https://github.com/ibm-openbmc/bmcweb/blob/7d56dde64a459a2d8a5a68e11638112787e4a682/redfish-core/include/utils/chassis_utils.hpp#L45). That function is handling just a vector of paths, but same idea can be done here just using the obj.first instead of `chassis`.

jeaaustx (commented):
I think it will be helpful to include the command output or more details of the problem being fixed by this change in the commit message. It will help in later releases when this patch is being rebased to assure the change is correctly folded forward.

(Might also be helpful to include the defect number.)

redfish-core/lib/update_service.hpp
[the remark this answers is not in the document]
abiasojo: I updated and tested to this 

dbus::utility::getSubTree( "/xyz/openbmc_project/software/"

redfish-core/lib/update_service.hpp
abiasojo: I updated to use obj_path and tested.

redfish-core/lib/update_service.hpp
```diff
@@ -1252,16 +1252,22 @@ inline void requestRoutesSoftwareInventory(App& app)
...
                     {
-                        if (!obj.first.ends_with(*swId))
+                        sdbusplus::message::object_path path(obj.first);
+                        std::string id = path.filename();
+                        if (id.empty())
+                        {
+                            BMCWEB_LOG_DEBUG("Failed to find software id in {}",
+                                             obj.first);
+                            continue;
+                        }
```
baemyung, line 1262: I think L1255-L1262 is not needed, as L1263 `if(id != *swId)` will cover this part.

baemyung (approved):
Other than a minor comment, I think it seems good.

redfish-core/lib/update_service.hpp
gtmills: I am fine with it since 
https://github.com/ibm-openbmc/bmcweb/blob/5f88ee7e7129137bedd208b323093125c6d96e52/redfish-core/include/utils/chassis_utils.hpp#L49 

has it

redfish-core/lib/update_service.hpp
abiasojo: Updated to not fill in "@odata.id" on a 404

gtmills (approved):

rfrandse:
Request from [@abiola.asojo](https://ibm-systems-power.slack.com/team/WJ5RY4RKR)

Pull Request: https://github.com/ibm-openbmc/bmcweb/pull/1143
Upstream: https://gerrit.openbmc.org/c/openbmc/bmcweb/+/79006
Fixes: 60993
Releases: 1110

PR #1144, opened 2025-02-10
MFA Bypass: Change the error returned

asmithakarun:
When any user tries to bypass MFA for self, before setting up the secret key, currently "Operation Not Allowed" error is thrown. But the right error here should be "GenerateSecretKeyRequired". All other operaions was also returning the latter.

Tested By:
'''
PATCH -d '{"MFABypass":{"BypassTypes":["GoogleAuthenticator"]}}' https://${bmc}/redfish/v1/AccountService/Accounts/<username> -H "Content-Type: application/json" '''


asmithakarun:

raviteja-b (approved):

asmithakarun:
Fixes: https://jazz07.rchland.ibm.com:13443/jazz/web/projects/CSSD#action=com.ibm.team.workitem.viewWorkItem&id=676685

baemyung (approved):
It looks good

PR #1166, opened 2025-03-04
1120: Remove Inventory.Item.Board interface from Chassis (#845)

baemyung:
Remove Inventory.Item.Board interface from Chassis (#845)

IBM does not use the xyz.openbmc_project.Inventory.Item.Board interface
for chassis objects. Remove that interfaces so that only objects
implements xyz.openbmc_project.Inventory.Item.Chassis are returned for
`getValidChassisPath()`.

---

Refactor chassisInterfaces into chassis_utils

Some (e.g. IBM) do not use the
`xyz.openbmc_project.Inventory.Item.Board` interface for chassis
objects. To handle the use pattern easier, this refactors the Chassis
interface into one location and it is referenced from the needed places
(e.g. `getValidChassisPath()`).

Tested:
- GET Chassis related API and check they are the same as before

Change-Id: Id4a51986262892c5dc81b1a3bc46fa5be7c0e9da

### DIFF
--- a/redfish-core/include/utils/chassis_utils.hpp
+++ b/redfish-core/include/utils/chassis_utils.hpp
@@ -26,6 +26,9 @@
 namespace redfish
 {
 
+static constexpr std::array<std::string_view, 1> chassisInterfaces = {
+    "xyz.openbmc_project.Inventory.Item.Chassis"};
+
 static constexpr std::array<std::string_view, 9> chassisAssemblyInterfaces = {
     "xyz.openbmc_project.Inventory.Item.Vrm",
     "xyz.openbmc_project.Inventory.Item.Tpm",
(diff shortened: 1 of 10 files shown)


redfish-core/include/utils/chassis_utils.hpp
```diff
@@ -30,8 +30,7 @@ void getValidChassisPath(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
                          const std::string& chassisId, Callback&& callback)
 {
     BMCWEB_LOG_DEBUG("checkChassisId enter");
-    constexpr std::array<std::string_view, 2> interfaces = {
-        "xyz.openbmc_project.Inventory.Item.Board",
+    constexpr std::array<std::string_view, 1> interfaces = {
```
gtmills, line 33: Maybe just make sure you don't need to remove it else where https://github.com/search?q=repo%3Aibm-openbmc%2Fbmcweb%20xyz.openbmc_project.Inventory.Item.Board&type=code but looks good

gtmills: Maybe combine with this commit https://github.com/ibm-openbmc/bmcweb/commit/112698a436a4c6d7226f43c38c8cc085fb66d14f ?

baemyung:
upstream: https://gerrit.openbmc.org/c/openbmc/bmcweb/+/79119

redfish-core/include/utils/chassis_utils.hpp
baemyung: I've refactored this and upstreamed to https://gerrit.openbmc.org/c/openbmc/bmcweb/+/79119.

One question: 
Would it be better to remove it from chassisAssemblyInterfaces too?
 - https://github.com/ibm-openbmc/bmcweb/blob/d1f1721fdfe447621b568d4a05f7226f19b4c6c1/redfish-core/include/utils/chassis_utils.hpp#L38

baemyung:
Another question regarding https://gerrit.openbmc.org/c/openbmc/bmcweb/+/79119.

- Should we introduce a meson.options so that Item.Board can be selectively used (for the upstream)?
For example,  `use-board-as-chassis`?

gtmills:
> Should we introduce a meson.options so that Item.Board can be selectively used (for the upstream)?
For example, use-board-as-chassis?

I think Ed is going to say we just need to get assembly in

gtmills (approved):

PR #1088, opened 2024-10-29
1110: Add support for MetricDefinition scheme

rfrandse:
#### Add support for MetricDefinition scheme
```
Added MetricDefinition node to Redfish code. Now user is able to list
all available metrics in OpenBMC that are supported by Telemetry
service. Metrics are grouped by reading type.

MetricDefinitions contains all physical sensors supported by redfish,
algorithm iterates through all chassis and collects results for each
node available in that chassis (Power, Thermal, Sensors).

When BMCWEB_NEW_POWERSUBSYSTEM_THERMALSUBSYSTEM will be enabled by
default (meson option redfish-new-powersubsystem-thermalsubsystem) it
will be possible to optimize this algorithm to only get sensors from
Sensors node. Currently Sensors node doesn't contain all available
sensors.

Removal of ResourceNotFound in sensors.hpp to fix MetricDefinition
Collection REST Get call.

Tested:
 - MetricDefinitions response is filled with existing sensors, it works
   with and without Telemetry service
 - Validated a presence of MetricDefinition members and its attributes
 - Successfully passed RedfishServiceValidator.py using witherspoon
   image on QEMU
 - Tested using following GET,POST requests

GET /redfish/v1/TelemetryService/MetricDefinitions
{
  "@odata.id": "/redfish/v1/TelemetryService/MetricDefinitions",
  "@odata.type": "#MetricDefinitionCollection.MetricDefinitionCollection",
  "Members": [
    {
      "@odata.id": "/redfish/v1/TelemetryService/MetricDefinitions/Fan_Pwm"
    },
    {
      "@odata.id": "/redfish/v1/TelemetryService/MetricDefinitions/Fan_Tach"
    },
    {
      "@odata.id": "/redfish/v1/TelemetryService/MetricDefinitions/HostCpuUtilization"
    },
    {
      "@odata.id": "/redfish/v1/TelemetryService/MetricDefinitions/HostMemoryBandwidthUtilization"
    },
    {
      "@odata.id": "/redfish/v1/TelemetryService/MetricDefinitions/HostPciBandwidthUtilization"
    },
    {
      "@odata.id": "/redfish/v1/TelemetryService/MetricDefinitions/Inlet_BRD_Temp"
    },
    {
      "@odata.id": "/redfish/v1/TelemetryService/MetricDefinitions/Left_Rear_Board_Temp"
    }
  ],
  "Members@odata.count": 7,
  "Name": "Metric Definition Collection"
}

[...]

PR #1168, opened 2025-03-05
WIP: Revert "1110: Add support for MetricDefinition scheme (#1088)"

gtmills:
This reverts commit 31338dcce5462b2e9d3204867576ccf1c1cfa6d3.

Reverting due to a rash of defects in this area:
682422, 682838, 684542, 682505.

This endpoint is used to get the supported Metric Definitions (sensors), the Metric Definitions are like the available materials for the blueprint (the blueprint would be the Metric Report Definitions) and the house/finished product would be the Metric Reports.

The HMC is calling this endpoint several at a time and the BMC is timing out.

The BMC journal is filled with:
Feb 24 08:21:25 xxx bmcweb[14780]: [ERROR sensors.hpp:420] getChassis respHandler DBUS error: Connection timed out [generic:110]

Upstream we don't support this.
This API takes 2 seconds to complete and is very resource intensive.

The Sensor API /redfish/v1/Chassis/chassis/Sensors can be used to create the telemetry Metric Report Definitions.


rfrandse (approved):

jeaaustx (approved):

baemyung (approved):
It looks good to me

gtmills:
Going with the HMC's solution here

PR #1170, opened 2025-03-06
Reduce memory usage during Firmware update (#739)

jeaaustx:
Cherry-pick and rebase PR https://github.com/ibm-openbmc/bmcweb/pull/739

Tested:
- Verififed detail data for non-update actions properly logged.
- Verified data limit for update caught 
```
[WARNING audit_events.hpp:230] Audit buffer too small for data: bufLeft=215 detailLen=257
[DEBUG audit_events.hpp:268] auditEvent: bufLeft=201  opPathLen=41 detailLen=257 userLen=14
root@p10bmc:/tmp# ausearch -i | grep UpdateService
type=USYS_CONFIG msg=audit(01/22/25 22:17:21.909:34) : pid=1522 uid=root auid=unset ses=unset msg='op=POST:/redfish/v1/UpdateService/update acct=service exe=/tmp/bmcweb hostname=p10bmc addr=10.0.2.1 terminal=? res=success' 
```

bmcweb currently consumes the excessive memory during firmware update. It is due to the replications of the req.body (which is the full content of firmware).

- audit copies the entire req.body before truncating it to the buffer.

Tested:

- Redfish validator passes
- Monitor memory consumption of bmcweb process periodically. e.g., top -b -d 0.5 | grep /usr/bin/bmcweb
- With the fix, the max memory usage should only be increased about the size of the image.

### DIFF
--- a/include/audit_events.hpp
+++ b/include/audit_events.hpp
@@ -215,7 +215,8 @@ inline void auditEvent(const crow::Request& req, const std::string& userName,
     std::string detail;
     if (wantDetail(req))
     {
-        detail = req.body() + " ";
+        detail = req.body().substr(0, maxBuf);
+        detail += " ";
     }
 
     if (!detail.empty())


include/audit_events.hpp
```diff
@@ -215,7 +215,8 @@ inline void auditEvent(const crow::Request& req, const std::string& userName,
     std::string detail;
     if (wantDetail(req))
     {
-        detail = req.body() + " ";
+        detail = req.body().substr(0, maxBuf);
+        detail += " ";
```
jeaaustx, line 219: Note: The blank is needed here since cnfgBuf will be appended with detail and then the username string as well. (If it fits.) Need the blank between them.

baemyung: In fact, it should also be good with one line as `substr()` is small.
```
detail = req.body().substr(0, maxBuf) + " ";
```
So, either way is good.

baemyung (approved):
It looks good

baemyung:
I think we can modify to remove this line from the commit message and PR abstraction
> req.body is captured by-value instead of by-reference in a lambda when the user-validation.

gtmills (approved):

gtmills:
> I think we can modify to remove this line from the commit message and PR abstraction
>    req.body is captured by-value instead of by-reference in a lambda when the user-validation.

Removed from the PR msg.

PR #1180, opened 2025-03-11
1120: Add Link PCIeSlot from PCIeDevices (#592)(#961)(#1100)

baemyung:
This commit allows us to associate the PCIe device with PCIe Slot. PCIe Slot doesn't have an individual route path, which provides a unique association with the PCIe device. As a solution, bmcweb includes a stander route path with all PCIe Slot information, and based on Device naming information users can determine which slot it is.

For example, "Links|Oem|IBM|PCIeSlot" property will be added like

Tested:

- GET PCIeDevice and check whether it has a link to PCIeSlot
```
curl -k -X GET https://${bmc}:18080/redfish/v1/Systems/system/PCIeDevices/pcie_card1
{
  "@odata.id": "/redfish/v1/Systems/system/PCIeDevices/pcie_card1",
  "Id": "pcie_card1",
  "Links": {
    "Oem": {
      "IBM": {
        "@odata.type": "#IBMPCIeDevice.v1_0_0.PCIeLinks",
        "PCIeSlot": {
          "@odata.id": "/redfish/v1/Chassis/chassis/PCIeSlots"
        }
      }
    }
  },
...
}
```

### DIFF
--- a/redfish-core/lib/pcie.hpp
+++ b/redfish-core/lib/pcie.hpp
@@ -165,6 +165,52 @@ inline void requestRoutesSystemPCIeDeviceCollection(App& app)
             std::bind_front(handlePCIeDeviceCollectionGet, std::ref(app)));
 }
 
+inline void afterAddLinkToPCIeSlot(
+    const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
+    const std::string& pcieDeviceSlot, const boost::system::error_code& ec,
+    const dbus::utility::MapperEndPoints& chassisPaths)
+{
+    if (ec)
+    {
+        if (ec.value() == EBADR)
+        {
+            // This PCIeSlot has no chassis association.
+            return;
+        }
+        BMCWEB_LOG_ERROR("DBUS response ec={} for slot path={}", ec.value(),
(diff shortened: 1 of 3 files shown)


redfish-core/lib/pcie.hpp
```diff
@@ -165,6 +165,50 @@ inline void requestRoutesSystemPCIeDeviceCollection(App& app)
...
+            // This PCIeSlot has no chassis association.
+            return;
+        }
+        BMCWEB_LOG_ERROR("DBUS response error {}", ec.value());
+        messages::internalError(asyncResp->res);
+        return;
+    }
+    if (chassisPaths.size() != 1)
+    {
+        BMCWEB_LOG_ERROR("PCIe Slot association error! ");
```
gtmills, line 186: printing the paths here would make this more useful, oh well

gtmills (approved):

redfish-core/lib/pcie.hpp
baemyung: I added the pcieslot for the references (instead of list of chassispaths) for the simplicity of the code.
I think it would still help.

PR #1182, opened 2025-03-12
redfish: bump Role to v1_3_0 (#831)

gtmills:
This is required for the new "Restricted" property.

Seeing Validator errors because of this.

This restricted role is downstream


baemyung (approved):
It looks good to me.

BTW, should this also be upstreamed?

gtmills:
> BTW, should this also be upstreamed?

It can't until we get service upstream

PR #1183, opened 2025-03-12
Restrict the Operator role (#319) (#526) (#954)

gtmills:
* Restrict the Operator role (#319)

This is downstream only due to upstream having the operator role

Tested:
2. POST /redfish/v1/AccountService/Accounts/new with Role:Operator Ensure it failed with message Base.1.9.0.RestrictedRole. Posting a new Administrator user is successful.

* Fixup clang boolean expression

---------


baemyung (approved):
It looks good to me

PR #1184, opened 2025-03-12
Limit Query Expand to 3

gtmills:
This was supposed to be done for 1050 but somehow was missed.

upstream this is https://gerrit.openbmc.org/c/openbmc/bmcweb/+/78192 

Tested: Any expand greater than 3 rejects.


baemyung (approved):
It looks good to me

PR #1185, opened 2025-03-12
WIP: Only allow expand from certain URIs

gtmills:
We only allow expand on three parts of the tree:
/redfish/v1/Chassis
/redfish/v1/Systems/system
/redfish/v1/Cables

The GUI uses expand in about 8 places on the Sensor and PCIe Topology page. https://github.com/ibm-openbmc/webui-vue/blob/3d2d537afc5f338e4afb88e2a458b784d4c55193/src/store/modules/HardwareStatus/SensorsStore.js#L34 https://github.com/ibm-openbmc/webui-vue/blob/3d2d537afc5f338e4afb88e2a458b784d4c55193/src/store/modules/HardwareStatus/PcieTopologyStore.js#L85

These 8 calls are at or under the URIs above.

Tested: Only these URIs allow expand.


baemyung (approved):
It looks good to me

PR #1138, opened 2025-01-30
Revert "Change Session Cookie name (#1130)"

gtmills:
This reverts commit 05b61c64f39e58e36cebfca10d52dbb1f6791f64.

In DVT, we are hitting an error 
```
Keyword 'Upload Image To BMC' failed after retrying 1 time. The last error was: SSLError: HTTPSConnectionPool(host='xxxxxx', port=443): Max retries exceeded with url: /redfish/v1/UpdateService/update (Caused by SSLError(SSLEOFError(8, 'EOF occurred in violation of protocol (_ssl.c:2427)')))
```

I can't explain it but reverting this commit appears to fix it

PR #1190, opened 2025-03-13
Change cookie name

gtmills:
1) revert the hack at https://github.com/ibm-openbmc/bmcweb/commit/b90a2c3beb6e93a007a8839608f948d9ba95a4d7 
It doesn't work for sarafi browser because for some reason that hmc cookie is coming in 2nd 
686862 

2) Change Session Cookie name
Upstream is https://gerrit.openbmc.org/c/openbmc/bmcweb/+/77539
This was added at https://github.com/ibm-openbmc/bmcweb/pull/1130 but
had to be reverted at https://github.com/ibm-openbmc/bmcweb/pull/1138
due to DVT failing.

We have a use case where the GUI sits behind a Apache Tomcat proxy[1].

In this environment the cookie looks like:
```
en-US,en;q=0.9cookie:
JSESSIONIDSSO=4E999D77EF4E01CB72DE63949D5FF830;
CCFWSESSION=48A66EB93C00AD4F6327FB3FC2A338FC; LOGIN_MODE=Dashboard;
XSRF-TOKEN=Ue1La3Ik48Bn5NosyLnJ; SESSION=pCAdqApWt4Kb4IUV9vh8dnt:
```

The bmcweb code thinks the CCFWSESSION= is the SESSION. The bmcweb code
could be made smarter to differentiate "CCFWSESSION" and "SESSION" but
reading SESSION seems too generic of a name and something like
"BMCWEB-SESSION" better matches [2], [3], and [4].

[1]: https://tomcat.apache.org/tomcat-9.0-doc/proxy-howto.html
[2]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#session_cookie
[3]: https://http.dev/set-cookie
[4]: https://www.geeksforgeeks.org/http-headers-set-cookie/

Tested: The GUI works and this proxy environment now works.

### DIFF
--- a/include/authentication.hpp
+++ b/include/authentication.hpp
@@ -127,7 +127,7 @@ inline std::shared_ptr<persistent_data::UserSession>
     {
         std::string_view cookieValue = it->value();
         BMCWEB_LOG_DEBUG("Checking cookie {}", cookieValue);
-        auto startIndex = cookieValue.find("SESSION=");
+        auto startIndex = cookieValue.find("BMCWEB-SESSION=");
         if (startIndex == std::string::npos)
         {
             BMCWEB_LOG_DEBUG(
@@ -135,7 +135,7 @@ inline std::shared_ptr<persistent_data::UserSession>
                 cookieValue);
             continue;
         }
-        startIndex += sizeof("SESSION=") - 1;
+        startIndex += sizeof("BMCWEB-SESSION=") - 1;
         auto endIndex = cookieValue.find(';', startIndex);
         if (endIndex == std::string::npos)
         {
--- a/include/authentication.hpp
+++ b/include/authentication.hpp
@@ -127,8 +127,7 @@ inline std::shared_ptr<persistent_data::UserSession>
     {
         std::string_view cookieValue = it->value();
         BMCWEB_LOG_DEBUG("Checking cookie {}", cookieValue);
-        // TODO (Gunnar): Flipping to rfind is a hack, fix this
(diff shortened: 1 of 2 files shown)


baemyung (approved):
Yes, this seems more making sense.

joseph-reynolds (approved):
Seems okay to me.
Note that users will need to login again across this fw update.

joseph-reynolds:
Should we also change the `XSRF-TOKEN` cookie name, for the same reason?

joseph-reynolds (commented):
Should we also change the XSRF-TOKEN name (to BMCWEB-XSRF-TOKEN) for the same reason (to avoid name collisions)?

gtmills:
> Should we also change the XSRF-TOKEN name (to BMCWEB-XSRF-TOKEN) for the same reason (to avoid name collisions)?
 
 No, XSRF-TOKEN has to be that name 
 https://www.google.com/search?client=firefox-b-1-d&sca_esv=895494ce73fddbb7&channel=entpr&q=XSRF-TOKEN+Angular&sa=X&ved=2ahUKEwjsuoqr1oqMAxWjIUQIHY-2JqMQ1QJ6BAg9EAE&biw=1920&bih=919&dpr=1
 ```
 Angular provides built-in support to mitigate Cross-Site Request Forgery (CSRF) attacks, using the "Double Submit Cookie" pattern. This defense mechanism relies on the server setting a cookie named XSRF-TOKEN upon user authentication. 
 ```
 
 This comes from back in the day but I believe Vue works just the same

PR #1196, opened 2025-03-17
AccountService: Restrict self-configure privilege

abhilashraju:
Restrict the admin user's privileges to only self-configuration when the session state is in self-configure mode. Currently, the admin can modify other users' information even when the admin session is in isConfigureSelfOnly mode, which appears to be incorrect. Eg:
1) Expire the admin user password
2) Login bmc with admin credentials
3) Password change required error is generated for the user

Expected after step 3

Admin should not be able to change any user information belonging to other users until they have changed their own password.

Observed after step 3

Admin was able to change other user's information before changing own password

Tested By:

1) Expire the admin password using passwd --expire 2) Change user name of other user  as below
```
patch /redfish/v1/AccountService/Accounts/newuser1 -d '{"UserName": "newuser2"}'
```
3) Observed response
```
{
  "error": {
    "@Message.ExtendedInfo": [
      {
        "MessageArgs": [],
      }
    ],
  }
}
```
4) Change the admin password
5) Do the step 2 above
6) Operation Resonded with Ok

upstream commit https://gerrit.openbmc.org/c/openbmc/bmcweb/+/78854/3

### DIFF
--- a/redfish-core/lib/account_service.hpp
+++ b/redfish-core/lib/account_service.hpp
@@ -2574,7 +2574,8 @@ inline void
         redfish::getUserPrivileges(*req.session);
     Privileges configureUsers = {"ConfigureUsers"};
     bool userHasConfigureUsers =
-        effectiveUserPrivileges.isSupersetOf(configureUsers);
+        effectiveUserPrivileges.isSupersetOf(configureUsers) &&
+        !req.session->isConfigureSelfOnly;
     if (!userHasConfigureUsers)
     {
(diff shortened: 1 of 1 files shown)


gtmills:
Can you get CI to pass upstream? 
This would be one, I would like to wait a few days to get upstream's opinion here

baemyung (approved):
It looks good to me

joseph-reynolds (commented):
Seems okay.  This fixes the security bug.

PR #869, opened 2024-03-25
Add pretty name for fabric adapters

PriyangaRamasamy:
The commit implement changes to make use of util function to
populate name for fabric adapters.

Validator was executed and no new error was found.

Test output:
curl -k -H "X-Auth-Token: $bmc_token" -X GET "https://${BMC_IP}/redfish/v1/Systems/system/FabricAdapters/motherboard-pcieslot8-pcie_card8"
{
  "@odata.id": "/redfish/v1/Systems/system/FabricAdapters/motherboard-pcieslot8-pcie_card8",
  "@odata.type": "#FabricAdapter.v1_4_0.FabricAdapter",
  "Id": "motherboard-pcieslot8-pcie_card8",
  "Links": {
    "PCIeDevices": [
      {
        "@odata.id": "/redfish/v1/Systems/system/PCIeDevices/chassis_motherboard_pcieslot8_pcie_card8"
      }
    ],
    "PCIeDevices@odata.count": 1
  },
  "Location": {
    "PartLocation": {
      "ServiceLabel": "U780C.ND0.RCH0011-P0-C8"
    }
  },
  "Model": "",
  "Name": "PCIe4 x16 or PCIe5 x8 adapter",
  "PartNumber": "",
  "Ports": {
    "@odata.id": "/redfish/v1/Systems/system/FabricAdapters/motherboard-pcieslot8-pcie_card8/Ports"
  },
  "SerialNumber": "",
  "Status": {
    "Health": "OK",
    "State": "Absent"
  }
}

PR #858, opened 2024-03-19
Util api to fetch pretty name for assemblies

PriyangaRamasamy:
This commit implements change to use util api to fetch prettyName property for any given assembly and publish it as "Name" in assembly schema.

In case prettyName is blank for that FRU, then last part of the object path will be used to populate assembly name.

Validator executed with no new errors.

PR #1201, opened 2025-03-19
util: Add pretty name for resources (#918)

deepakala-k:
Created a helper function to fetch pretty name and added for some resources.

- Chassis
- FirmwareInventory
- Storage
- StorageController
- System
- Processor

Implementation Details:
The helper function makes a call to ObjectMapper to get the services that support the object path with the interfaces we want. It expects only one service is exporting the object path.

Then it will try to get the `PrettyName` property. If it is found and is valid, it will set the `Name` to the value. If not, it will use the default value that is set before the function call.

This will affect all systems and replace the Name with the PrettyName if it exist within the resource's object path.

Tested:
Passed Redfish Validator

The output names release the existing `Name` property only if the PrettyName property exists. If not, it will use the default value.

Output with PrettyName
```
$ curl  -u root:0penBmc -X GET http://${bmc}/redfish/v1/Chassis/chassis0
{
  "@odata.id": "/redfish/v1/Chassis/chassis0",
  "@odata.type": "#Chassis.v1_14_0.Chassis",
  "ChassisType": "RackMount",
  "Id": "chassis0",
  "Links": {
    "ComputerSystems": [
      {
      }
    ],
    "ManagedBy": [
      {
      }
    ]
  },
  "Name": "Test Chassis",
  ...
}
```

Without PrettyName
```
$ curl  -u root:0penBmc -X GET http://${bmc}/redfish/v1/Chassis/chassis0
{
  "@odata.id": "/redfish/v1/Chassis/chassis0",
  "@odata.type": "#Chassis.v1_14_0.Chassis",
  "ChassisType": "RackMount",
  "Id": "chassis0",
  "Links": {
    "ComputerSystems": [
      {
      }
    ],
    "ManagedBy": [
      {
      }
    ]
  },
  "Name": "chassis0",
  ...
}
```

Performance Diff:
 - Before:
```
$ for i in {1..5}; do
    time curl -su root:0penBmc -X GET http://${bmc}/redfish/v1/Chassis/warthog_1 > /dev/null
  done

real    0m0.314s
user    0m0.006s
sys     0m0.003s

real    0m0.346s
user    0m0.003s
sys     0m0.009s

real    0m0.369s
user    0m0.001s
sys     0m0.013s

real    0m0.343s
user    0m0.005s
sys     0m0.010s

real    0m0.356s
user    0m0.006s
sys     0m0.010s
```
 - After:
```
$ for i in {1..5}; do
    time curl -su root:0penBmc -X GET http://${bmc}/redfish/v1/Chassis/warthog_1 > /dev/null;
  done

real    0m0.322s
user    0m0.009s
sys     0m0.005s

real    0m0.343s
user    0m0.007s
sys     0m0.007s

real    0m0.337s
user    0m0.000s
sys     0m0.009s

real    0m0.348s
user    0m0.011s
sys     0m0.005s

real    0m0.379s
user    0m0.010s
sys     0m0.005s

Change-Id: I76e65f4aae979365266ebb38525c4ac45298737b

### DIFF
--- a/redfish-core/include/utils/name_utils.hpp
+++ b/redfish-core/include/utils/name_utils.hpp
@@ -1,0 +1,94 @@
+
+#pragma once
+#include "dbus_utility.hpp"
+#include "error_messages.hpp"
+#include "logging.hpp"
+
+#include <async_resp.hpp>
+#include <boost/system/error_code.hpp>
+
+#include <memory>
+#include <string>
+
+namespace redfish
+{
+namespace name_util
+{
+
+/**
+ * @brief Populate the collection "Members" from a GetSubTreePaths search of
+ *        inventory
+ *
+ * @param[i,o] asyncResp  Async response object
+ * @param[i]   path       D-bus object path to find the pretty name
+ * @param[i]   services   Service to exporting the D-bus object path
+ * @param[i]   namePath   Json pointer to the name field to update.
+ *
+ * @return void
+ */
+inline void getPrettyName(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
+                          const std::string& path,
+                          const std::string& serviceName,
+                          const nlohmann::json::json_pointer& namePath)
+{
+    BMCWEB_LOG_DEBUG("Get PrettyName for: {}", path);
+
+    dbus::utility::getProperty<std::string>(
+        serviceName, path, "xyz.openbmc_project.Inventory.Item", "PrettyName",
+        [asyncResp, path, namePath](const boost::system::error_code& ec,
+                                    const std::string& prettyName) {
+            if (ec)
+            {
+                BMCWEB_LOG_DEBUG("DBUS response error : {}", ec.value());
+                return;
+            }
+
+            if (prettyName.empty())
+            {
+                return;
+            }
+
+            BMCWEB_LOG_DEBUG("Pretty Name: {}", prettyName);
+
+            asyncResp->res.jsonValue[namePath] = prettyName;
+        });
+}
+
+/**
+ * @brief Populate the collection "Members" from a GetSubTreePaths search of
+ *        inventory
+ *
+ * @param[i,o] asyncResp  Async response object
+ * @param[i]   path       D-bus object path to find the pretty name
+ * @param[i]   services   List of services to exporting the D-bus object path
+ * @param[i]   namePath   Json pointer to the name field to update.
+ *
+ * @return void
+ */
+inline void getPrettyName(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
+                          const std::string& path,
+                          const dbus::utility::MapperServiceMap& services,
+                          const nlohmann::json::json_pointer& namePath)
+{
+    BMCWEB_LOG_DEBUG("Get PrettyName with MapperServiceMap for: {}", path);
+
+    // Ensure we only got one service back
+    if (services.size() != 1)
+    {
+        BMCWEB_LOG_ERROR("Invalid Service Size {}", services.size());
+        for (const auto& service : services)
+        {
+            BMCWEB_LOG_ERROR("Invalid Service Name: {}", service.first);
+        }
+        if (asyncResp)
+        {
+            messages::internalError(asyncResp->res);
+        }
+        return;
+    }
+
+    getPrettyName(asyncResp, path, services[0].first, namePath);
+}
+
+} // namespace name_util
+} // namespace redfish
--- a/redfish-core/lib/assembly.hpp
+++ b/redfish-core/lib/assembly.hpp
@@ -14,6 +14,7 @@
 #include "utils/chassis_utils.hpp"
 #include "utils/dbus_utils.hpp"
 #include "utils/json_utils.hpp"
+#include "utils/name_utils.hpp"
 
 #include <boost/beast/http/verb.hpp>
 #include <boost/system/error_code.hpp>
@@ -231,6 +232,10 @@ inline void getAssemblyProperties(
                     messages::internalError(asyncResp->res);
                     return;
                 }
+
+                nlohmann::json::json_pointer ptr(
+                    "/Assemblies/" + std::to_string(assemblyIndex) + "/Name");
+                name_util::getPrettyName(asyncResp, assembly, object, ptr);
 
                 for (const auto& [serviceName, interfaceList] : object)
                 {
--- a/redfish-core/lib/chassis.hpp
+++ b/redfish-core/lib/chassis.hpp
@@ -22,6 +22,7 @@
 #include "utils/collection.hpp"
 #include "utils/dbus_utils.hpp"
 #include "utils/json_utils.hpp"
+#include "utils/name_utils.hpp"
 
 #include <boost/beast/http/field.hpp>
 #include <boost/beast/http/verb.hpp>
@@ -478,7 +479,8 @@ inline void handleChassisGetSubTree(
         asyncResp->res.jsonValue["@odata.type"] = "#Chassis.v1_22_0.Chassis";
         asyncResp->res.jsonValue["@odata.id"] =
             boost::urls::format("/redfish/v1/Chassis/{}", chassisId);
-        asyncResp->res.jsonValue["Name"] = "Chassis Collection";
+        name_util::getPrettyName(asyncResp, path, connectionNames,
+                                 "/Name"_json_pointer);
         asyncResp->res.jsonValue["Actions"]["#Chassis.Reset"]["target"] =
             boost::urls::format("/redfish/v1/Chassis/{}/Actions/Chassis.Reset",
                                 chassisId);
--- a/redfish-core/lib/fabric_adapters.hpp
+++ b/redfish-core/lib/fabric_adapters.hpp
@@ -18,6 +18,7 @@
 #include "utils/collection.hpp"
 #include "utils/dbus_utils.hpp"
 #include "utils/json_utils.hpp"
+#include "utils/name_utils.hpp"
 
 #include <asm-generic/errno.h>
 
@@ -309,7 +310,10 @@ inline void doAdapterGet(
         "</redfish/v1/JsonSchemas/FabricAdapter/FabricAdapter.json>; rel=describedby");
     asyncResp->res.jsonValue["@odata.type"] =
         "#FabricAdapter.v1_4_0.FabricAdapter";
-    asyncResp->res.jsonValue["Name"] = "Fabric Adapter";
+
+    nlohmann::json::json_pointer ptr("/Name");
+    name_util::getPrettyName(asyncResp, fabricAdapterPath, serviceName, ptr);
+
     asyncResp->res.jsonValue["Id"] = adapterId;
     asyncResp->res.jsonValue["@odata.id"] = boost::urls::format(
         "/redfish/v1/Systems/{}/FabricAdapters/{}", systemName, adapterId);
--- a/redfish-core/lib/fabric_ports.hpp
+++ b/redfish-core/lib/fabric_ports.hpp
@@ -17,6 +17,7 @@
 #include "query.hpp"
 #include "registries/privilege_registry.hpp"
 #include "utils/json_utils.hpp"
+#include "utils/name_utils.hpp"
 
 #include <asm-generic/errno.h>
 
@@ -156,7 +157,10 @@ inline void getFabricPortProperties(
         boost::urls::format("/redfish/v1/Systems/{}/FabricAdapters/{}/Ports/{}",
                             systemName, adapterId, portId);
     asyncResp->res.jsonValue["Id"] = portId;
-    asyncResp->res.jsonValue["Name"] = "Fabric Port";
+
+    nlohmann::json::json_pointer ptr("/Name");
+    name_util::getPrettyName(asyncResp, portPath, serviceName, ptr);
+
     asyncResp->res.jsonValue["Status"]["State"] = resource::State::Enabled;
     asyncResp->res.jsonValue["Status"]["Health"] = resource::Health::OK;
 
--- a/redfish-core/lib/memory.hpp
+++ b/redfish-core/lib/memory.hpp
@@ -20,6 +20,7 @@
 #include "utils/dbus_utils.hpp"
 #include "utils/hex_utils.hpp"
 #include "utils/json_utils.hpp"
+#include "utils/name_utils.hpp"
 
 #include <asm-generic/errno.h>
 
@@ -408,7 +409,6 @@ inline void assembleDimmProperties(
     const nlohmann::json::json_pointer& jsonPtr)
 {
     asyncResp->res.jsonValue[jsonPtr]["Id"] = dimmId;
-    asyncResp->res.jsonValue[jsonPtr]["Name"] = "DIMM Slot";
     asyncResp->res.jsonValue[jsonPtr]["Status"]["State"] =
         resource::State::Enabled;
     asyncResp->res.jsonValue[jsonPtr]["Status"]["Health"] =
@@ -623,6 +623,7 @@ inline void getDimmDataByService(
     const std::string& service, const std::string& objPath)
 {
     BMCWEB_LOG_DEBUG("Get available system components.");
+    name_util::getPrettyName(asyncResp, objPath, service, "/Name"_json_pointer);
     dbus::utility::getAllProperties(
         service, objPath, "",
         [dimmId, asyncResp{std::move(asyncResp)}](
--- a/redfish-core/lib/processor.hpp
+++ b/redfish-core/lib/processor.hpp
@@ -20,11 +20,13 @@
 #include "utils/dbus_utils.hpp"
 #include "utils/hex_utils.hpp"
 #include "utils/json_utils.hpp"
+#include "utils/name_utils.hpp"
 
 #include <boost/beast/http/field.hpp>
 #include <boost/beast/http/verb.hpp>
 #include <boost/system/error_code.hpp>
 #include <boost/url/format.hpp>
+#include <nlohmann/json.hpp>
 #include <sdbusplus/message/native_types.hpp>
 #include <sdbusplus/unpack_properties.hpp>
 
@@ -292,7 +294,8 @@ inline void getCpuDataByService(
                 return;
             }
             asyncResp->res.jsonValue["Id"] = cpuId;
-            asyncResp->res.jsonValue["Name"] = "Processor";
+            name_util::getPrettyName(asyncResp, objPath, service,
+                                     "/Name"_json_pointer);
             asyncResp->res.jsonValue["ProcessorType"] =
                 processor::ProcessorType::CPU;
 
--- a/redfish-core/lib/storage.hpp
+++ b/redfish-core/lib/storage.hpp
@@ -21,10 +21,12 @@
 #include "utils/chassis_utils.hpp"
 #include "utils/collection.hpp"
 #include "utils/dbus_utils.hpp"
+#include "utils/name_utils.hpp"
 
 #include <boost/beast/http/verb.hpp>
 #include <boost/system/error_code.hpp>
 #include <boost/url/format.hpp>
+#include <nlohmann/json.hpp>
 #include <sdbusplus/message/native_types.hpp>
 #include <sdbusplus/unpack_properties.hpp>
 
@@ -664,7 +666,8 @@ inline void afterGetSubtreeSystemsStorageDrive(
     asyncResp->res.jsonValue["@odata.id"] =
         boost::urls::format("/redfish/v1/Systems/{}/Storage/1/Drives/{}",
                             BMCWEB_REDFISH_SYSTEM_URI_NAME, driveId);
-    asyncResp->res.jsonValue["Name"] = driveId;
+    name_util::getPrettyName(asyncResp, path, connectionNames,
+                             "/Name"_json_pointer);
     asyncResp->res.jsonValue["Id"] = driveId;
 
     if (connectionNames.size() != 1)
--- a/redfish-core/lib/fan.hpp
+++ b/redfish-core/lib/fan.hpp
@@ -14,6 +14,7 @@
 #include "registries/privilege_registry.hpp"
 #include "utils/chassis_utils.hpp"
 #include "utils/json_utils.hpp"
+#include "utils/name_utils.hpp"
 
 #include <asm-generic/errno.h>
 
@@ -235,7 +236,6 @@ inline void addFanCommonProperties(crow::Response& resp,
     resp.addHeader(boost::beast::http::field::link,
                    "</redfish/v1/JsonSchemas/Fan/Fan.json>; rel=describedby");
     resp.jsonValue["@odata.type"] = "#Fan.v1_3_0.Fan";
-    resp.jsonValue["Name"] = "Fan";
     resp.jsonValue["Id"] = fanId;
     resp.jsonValue["@odata.id"] = boost::urls::format(
         "/redfish/v1/Chassis/{}/ThermalSubsystem/Fans/{}", chassisId, fanId);
@@ -387,6 +387,7 @@ inline void afterGetValidFanPath(
     getFanAsset(asyncResp, fanPath, service);
     getFanLocation(asyncResp, fanPath, service);
     getLocationIndicatorActive(asyncResp, fanPath);
+    name_util::getPrettyName(asyncResp, fanPath, service, "/Name"_json_pointer);
 }
 
 inline void doFanGet(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,

### CODE REVIEW

**Comment by gtmills** [comment]
Hi @deepakala-k Can you also add to Power Supply, Fan, Dimm  and squash to 1 commit like you have done? 
All the PrettyNames together would be helpful 

https://github.com/ibm-openbmc/bmcweb/commit/d9bf7d7ad14c2703bf663394b813fc7e45bf2435
https://github.com/ibm-openbmc/bmcweb/commit/5983f0e46cd636a6bc378a106d039fced8865332

**Comment by gtmills** [comment]
Also don't you need it for Fabric Adapters https://github.com/ibm-openbmc/bmcweb/pull/869

File: redfish-core/include/utils/name_utils.hpp [hunk 1]
```diff
@@ -0,0 +1,81 @@
...
+                          const dbus::utility::MapperServiceMap& services,
+                          const nlohmann::json::json_pointer& namePath)
+{
+    BMCWEB_LOG_DEBUG("Get PrettyName for: {}", path);
+
+    // Ensure we only got one service back
+    if (services.size() != 1)
+    {
+        BMCWEB_LOG_ERROR("Invalid Service Size  {}", services.size());
+        if (asyncResp)
```
jeaaustx, line 41: In current 1110 code we are also adding all of the services to the debug log. That was added by commit https://github.com/ibm-openbmc/bmcweb/commit/74a0b78d51b3d1f7c5d7ea90b13e13140efe596a

Can that one be included here and squashed to put all the pretty name changes together?

redfish-core/include/utils/name_utils.hpp
```diff
@@ -0,0 +1,81 @@
...
+        {
+            messages::internalError(asyncResp->res);
+        }
+        return;
+    }
+
+    crow::connections::systemBus->async_method_call(
```
jeaaustx, line 48: Current 1110 code uses getProperty() instead which also simplifies the checking of prettyName returned. That change was part of a different commit in 1110 (https://github.com/ibm-openbmc/bmcweb/commit/3bd202e4723e1a4a0a94f2062fd0cc25e82b32e9) but seems like it should really be part of this basic support for prettyName.

redfish-core/lib/chassis.hpp
```diff
@@ -583,7 +584,9 @@ inline void handleChassisGetSubTree(
         asyncResp->res.jsonValue["@odata.type"] = "#Chassis.v1_22_0.Chassis";
         asyncResp->res.jsonValue["@odata.id"] =
             boost::urls::format("/redfish/v1/Chassis/{}", chassisId);
-        asyncResp->res.jsonValue["Name"] = "Chassis Collection";
+        name_util::getPrettyName(asyncResp, path, connectionNames,
+                                 "/Name"_json_pointer);
+        asyncResp->res.jsonValue["ChassisType"] = "RackMount";
```
jeaaustx, line 589: This should be using the enum for setting ChassisType: chassis::ChassisType::RackMount

redfish-core/lib/storage.hpp
```diff
@@ -660,10 +661,10 @@ inline void afterGetSubtreeSystemsStorageDrive(
     const dbus::utility::MapperServiceMap& connectionNames = drive->second;
 
     asyncResp->res.jsonValue["@odata.type"] = "#Drive.v1_7_0.Drive";
-    asyncResp->res.jsonValue["@odata.id"] =
-        boost::urls::format("/redfish/v1/Systems/{}/Storage/1/Drives/{}",
-                            BMCWEB_REDFISH_SYSTEM_URI_NAME, driveId);
-    asyncResp->res.jsonValue["Name"] = driveId;
+    asyncResp->res.jsonValue["@odata.id"] = boost::urls::format(
+        "/redfish/v1/Systems/system/Storage/1/Drives/{}", driveId);
```
jeaaustx, line 665: This should be using the BMCWEB_REDFISH_SYSTEM_URI_NAME setting to match upstream.

redfish-core/include/utils/name_utils.hpp
deepakala-k: sure. I will pull it

redfish-core/include/utils/name_utils.hpp
deepakala-k: sure

redfish-core/lib/chassis.hpp
deepakala-k: I will remove this line from the commit? As it is not part of prettyName. I believe subsequent commits will get that code in.

redfish-core/lib/storage.hpp
deepakala-k: I will revert the odata.id change. Should the pretty name be used for the name property or should that also be reverted to 
asyncResp->res.jsonValue["Name"] = driveId; I see that the "Id" is anyways populated with driveId

deepakala-k:
> And assemblies #858 ?

I could not get that as the assembly.hpp file is not yet in 1120

deepakala-k:
> Also don't you need it for Fabric Adapters #869

I am unable to get this as redfish-core/lib/fabric_ports.hpp is not present in 1120.

baemyung:
> I could not get that as the assembly.hpp file is not yet in 1120
> I am unable to get this as redfish-core/lib/fabric_ports.hpp is not present in 1120.

I think we can make this PR as WIP and wait for a while (so that the others can be rebased).

deepakala-k:
> > I think we can make this PR as WIP and wait for a while (so that the others can be rebased).
> > For example,
> > Assembly - #1205 is being reviewed
> > Ports - it is being worked (also for openbmc backend). So this may take some more time
> 
> I am okay either way. I am okay with this going first then being added to assembly and ports later

I can wait, so it becomes easier to handle in the future

baemyung:
> > > I think we can make this PR as WIP and wait for a while (so that the others can be rebased).
> > > For example,
> > > Assembly - #1205 is being reviewed
> > > Ports - it is being worked (also for openbmc backend). So this may take some more time
> > 
> > 
> > I am okay either way. I am okay with this going first then being added to assembly and ports later
> 
> I can wait, so it becomes easier to handle in the future

All dependent PRs are merged.  So, please go ahead to complete this PR.
- Assembly - #1205 
- Port - https://github.com/ibm-openbmc/bmcweb/pull/1215

baemyung:
> @baemyung In the fabric_ports.hpp It also needs some of the changes from this commit. [6d5d94a](https://github.com/ibm-openbmc/bmcweb/commit/6d5d94a8d6f8b935793181df8a3718a33269c02d)
> 
> e.g. The pretty name PR alters afterGetAssociatedFabricPortSubTree method, which is not yet added

The current `fabric_ports.hpp` is not using `afterGetAssociatedFabricPortSubTree` but it uses `getAssociatedSubTreePathsById` (by `getValidFabricPortPath`).   

BTW, is this PrettyName replacing `Name`? 
 If so, wouldn't it be to change this - https://github.com/ibm-openbmc/bmcweb/blob/fc1dcb8f6057b7f1105be7aa7bb69a8b02a4c29f/redfish-core/lib/fabric_ports.hpp#L159

baemyung:
Comparing with 1110, `fabric_adapters.hpp`  seems not done for 1120.

baemyung (commented):
In addition to a comment about `fabric_ports.hpp` & comparing with 1110, `fabric_adapters.hpp` seems not done for 1120.

deepakala-k:
> In addition to a comment about `fabric_ports.hpp` & comparing with 1110, `fabric_adapters.hpp` seems not done for 1120.

So, should i wait for all of those commits to be in?

baemyung:
> > In addition to a comment about `fabric_ports.hpp` & comparing with 1110, `fabric_adapters.hpp` seems not done for 1120.
> 
> So, should i wait for all of those commits to be in?

The dependent codes are currently ready.
Please do this prettyName parts (e.g. fabric_ports.hpp & fabric_adapters.hpp) with the latest 1120 code.

redfish-core/lib/fabric_adapters.hpp
```diff
@@ -376,6 +377,9 @@ inline void afterGetValidFabricAdapterPath(
             fabricAdapterPath = adapterPath;
             serviceName = serviceMap.begin()->first;
             interfaces = serviceMap.begin()->second;
+
+            nlohmann::json::json_pointer ptr("/Name");
+            name_util::getPrettyName(asyncResp, adapterPath, serviceMap, ptr);
             break;
```
baemyung, line 383: I think this part should be moved out of `getValidFabricAdapterPath()` and `afterGetValidFabricAdapterPath()`, as they are used not only for GET /HEAD and also PATCH.

PrettyName should be done in a path of GET - i.e.
- handleFabricAdapterGet() ->  afterHandleFabricAdapterGet() -> doAdapterGet()

redfish-core/include/utils/name_utils.hpp
```diff
@@ -0,0 +1,74 @@
...
+ *        inventory
+ *
+ * @param[i,o] asyncResp  Async response object
+ * @param[i]   path       D-bus object path to find the pretty name
+ * @param[i]   services   List of services to exporting the D-bus object path
+ * @param[i]   namePath   Json pointer to the name field to update.
+ *
+ * @return void
+ */
+inline void getPrettyName(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
```
baemyung, line 29: - I think it is better to have a function of `getPrettyName()` with just `service` (not needing `ServiceMap'), as the current getPrettyName() takes only the first service.
- This approach would make the caller function simpler - e.g. Ports Name.

- I also think the error handling of `services.size()`  may be handled separately from this function.  If the caller needs to take care of services.size(), it is the one to deal with such error case.

redfish-core/lib/fabric_adapters.hpp
baemyung: So either introduce a simpler `getPrettyName()` only with service (not serviceMap), or use the similar approach like what's in for fabric_ports

```

inline void doAdapterGet(
    const std::string& systemName, const std::string& adapterId,
    const std::string& fabricAdapterPath, const std::string& serviceName,
    const dbus::utility::InterfaceList& interfaces)
{
...

DEL>   asyncResp->res.jsonValue["Name"] = "Fabric Adapter";

ADD BEGIN>
    // Pretty name expects a services map
    const dbus::utility::MapperServiceMap& serviceMap = {{serviceName, {""}}};
    nlohmann::json::json_pointer ptr("/Name");
    name_util::getPrettyName(asyncResp, fabricAdapterPath, serviceMap, ptr);
ADD-END>
..
}
```

baemyung (changes requested):
fabric_adapter prettyName handling needs to be changed.

redfish-core/include/utils/name_utils.hpp
```diff
@@ -0,0 +1,113 @@
...
+                return;
+            }
+
+            BMCWEB_LOG_DEBUG("Pretty Name: {}", prettyName);
+
+            asyncResp->res.jsonValue[namePath] = prettyName;
+        });
+}
+
+} // namespace name_util
```
baemyung, line 112: I think we can rearrange the codes 
- getPrettyName( serviceName, namePath)
- getPrettyName(serviceMap...)

Then, we can make the first one from the second one.
This way, we can manage the code better later when needed.

```
inline void getPrettyName(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
                          const std::string& path,
                          const dbus::utility::MapperServiceMap& services,
                          const nlohmann::json::json_pointer& namePath)
{
    BMCWEB_LOG_DEBUG("Get PrettyName for: {}", path);

    // Ensure we only got one service back
    if (services.size() != 1)
    {
......<same from second one>...
        return;
    }

     getPrettyName( asyncResp, path,
                                services[0].first, 
                                namePath);

}
```

redfish-core/lib/storage.hpp
baemyung: As @jeaaustx pointed out, this would be -- as in https://github.com/openbmc/bmcweb/blob/193582b6c4acc1eb60eb8ca1f445e3af4f3fe6d8/redfish-core/lib/storage.hpp#L664

```
    asyncResp->res.jsonValue["@odata.id"] =
        boost::urls::format("/redfish/v1/Systems/{}/Storage/1/Drives/{}",
                            BMCWEB_REDFISH_SYSTEM_URI_NAME, driveId);
```

redfish-core/lib/assembly.hpp
```diff
@@ -232,6 +233,11 @@ inline void getAssemblyProperties(
                     return;
                 }
 
+                nlohmann::json::json_pointer ptr(
+                    "/Assemblies/" + std::to_string(assemblyIndex) + "/Name");
+
```
baemyung, line 238: I think it is better to remove the blank line here

redfish-core/lib/chassis.hpp
baemyung: Yes, please remove the line of this.
I think ChassisType is added to each chassis (not chassis collection) as in upstream, and also it is not a part of this prettyName.

```
DEL> asyncResp->res.jsonValue["ChassisType"] = "RackMount";
```

redfish-core/lib/memory.hpp
```diff
@@ -754,6 +754,8 @@ inline void afterGetDimmData(
                                          objectPath);
                     dimmInterface = true;
                     found = true;
+                    name_util::getPrettyName(asyncResp, objectPath, serviceMap,
+                                             "/Name"_json_pointer);
```
baemyung, line 758: Why should this be here?  
Because getPrettyName() expects only one service, I don't think it needs to be in a loop.
Should be this moved to L412?

redfish-core/lib/processor.hpp
```diff
@@ -813,6 +815,8 @@ inline void handleProcessorSubtree(
                     // process must be on the same object path.
 
                     callback(objectPath, serviceMap);
+                    name_util::getPrettyName(asyncResp, objectPath, serviceMap,
+                                             "/Name"_json_pointer);
```
baemyung, line 819: I also think this is not a right place.
This function `handleProcessorSubtree()` is an intermediate callback of `getProcessorObject()` (which is not a direct GET function, and it is also used by PATCH).

getPrettyName() should be done for GET - i.e. replacing "Name".
So, the right place to do is L297 - in `getCpuDataByService()` which is a part of `getProcessorData()`.

I also think 1110 code also needs to be changed.

baemyung (commented):
There is also a few piece.
- pcie.hpp. 
    In fact,  1110 code also seems incorrect (similar processor.hpp).  I think getPrettyName() should be moved into `addPCIeDeviceCommonProperties`

redfish-core/lib/memory.hpp
deepakala-k: doing it here as compared to Ln412 is almost the same. If we called at Ln.412, we got to pass the service name as assembleDimmProperties does not have service name. That can be done. But otherwise I do not see it getting called in a loop.

afterGetDimmData -> getDimmDataByService -> assembleDimmProperties ->Ln412

redfish-core/lib/processor.hpp
deepakala-k: The problem everywhere is that we need to find the service name, which we get in one of these methods. I know we could do the right way. But just trying to ensure that we do not break something that is already working. Will it affect if we update the pretty name in the patch function?

baemyung:
> * addPCIeDeviceCommonProperties

For 1110 pcie.hpp, I may misread the code. It is done for GET only.
No need to change for pcie.hpp / 1110.

redfish-core/lib/processor.hpp
baemyung: Have you tried PATCH for Processor and see what's returned?

redfish-core/lib/memory.hpp
baemyung: I meant -- more put it inside `getDimmDataByService(service)` or more like inside ` assembleDimmProperties() by passing 'service', as Name is one of properties...

redfish-core/lib/processor.hpp
baemyung: For example, this PATCH is issued, it goes thru `getPrettyName()` which is not needed.

- Enable debug level
```
root@p10bmc:~# bmcweb loglevel debug
<6>[webserver_cli.cpp:88] logging level changed to: DEBUG

root@p10bmc:~# journalctl -f
```

- Run PATCH
`curl -k -H "Content-Type: application/json" -X PATCH https://${bmc}/redfish/v1/Systems/system/Processors/dcm0-cpu0 -d '{"LocationIndicatorActive":true}'`

Then we can see

```
Apr 05 13:32:48 p10bmc bmcwebd[920]: [routing.hpp:652] Matched rule '/redfish/v1/Systems/<str>/Processors/<str>/' PATCH / 16
Apr 05 13:32:48 p10bmc bmcwebd[920]: [http_connection.hpp:906] 0x2708bc0 Timer canceled
Apr 05 13:32:48 p10bmc bmcwebd[920]: [dbus_privileges.hpp:59] userName = service userRole =
Apr 05 13:32:48 p10bmc bmcwebd[920]: [query.hpp:125] setup redfish route
Apr 05 13:32:48 p10bmc bmcwebd[920]: [processor.hpp:845] Get available system processor resources.
Apr 05 13:32:48 p10bmc bmcwebd[920]: [led.hpp:427] Set LocationIndicatorActive for /xyz/openbmc_project/inventory/system/chassis/motherboard/dcm0/cpu0
Apr 05 13:32:48 p10bmc bmcwebd[920]: [name_utils.hpp:73] Get PrettyName with MapperServiceMap for: /xyz/openbmc_project/inventory/system/chassis/motherboard/dcm0/cpu0
Apr 05 13:32:48 p10bmc bmcwebd[920]: [name_utils.hpp:34] Get PrettyName for: /xyz/openbmc_project/inventory/system/chassis/motherboard/dcm0/cpu0
Apr 05 13:32:48 p10bmc bmcwebd[920]: [name_utils.hpp:51] Pretty Name: System processor module 0
Apr 05 13:32:48 p10bmc bmcwebd[920]: [http_response.hpp:259] 0x26d3620 calling completion handler
Apr 05 13:32:48 p10bmc bmcwebd[920]: [http_response.hpp:262] 0x26d3620 completion handler was valid
Apr 05 13:32:48 p10bmc bmcwebd[920]: [http_response.hpp:97] Moving response containers; this: 0x270ae08; other: 0x26d3620
Apr 05 13:32:48 p10bmc bmcwebd[920]: [audit_events.cpp:250] auditEvent: bufLeft=152  opPathLen=57 detailLen=33 userLen=14
Apr 05 13:32:48 p10bmc bmcwebd[920]: [complete_response_fields.hpp:25] Response: 204
Apr 05 13:32:48 p10bmc bmcwebd[920]: [http_connection.hpp:851] 0x2708bc0 doWrite
Apr 05 13:32:48 p10bmc bmcwebd[920]: [http_response.hpp:213] 0x270ae08 Response content provided but code was no-content or not_modified, which aren't allowed to have a body
Apr 05 13:32:48 p10bmc bmcwebd[920]: [http_connection.hpp:934] 0x2708bc0 timer started
Apr 05 13:32:48 p10bmc bmcwebd[920]: [http_body.hpp:198] Returning 41 bytes more=false
Apr 05 13:32:48 p10bmc bmcwebd[920]: [http_response.hpp:269] 0x270ae08 setting completion handler
Apr 05 13:32:48 p10bmc bmcwebd[920]: [http_connection.hpp:799] 0x2708bc0 async_write wrote 352 bytes, ec=Success [system:0]
```

deepakala-k: Thanks fir checking. I will try to fix it

deepakala-k: done

redfish-core/lib/memory.hpp
deepakala-k: Addressed it. I didnt have to call getPrettyName as all the properties are anyways read and so I added pretty name to it. And tested the same. It works as expected.

redfish-core/lib/memory.hpp
```diff
@@ -614,6 +614,10 @@ inline void assembleDimmProperties(
             .jsonValue[jsonPtr]["Location"]["PartLocation"]["ServiceLabel"] =
             *locationCode;
     }
+    if (prettyName != nullptr)
+    {
+        asyncResp->res.jsonValue[jsonPtr]["Name"] = *prettyName;
+    }
```
baemyung, line 620: Perhaps, for the consistency, would it be better to use `getPrettyName()` function by either passing `serviceName` to `assembleDimmProperties()`, or doing it like this?

```
inline void getDimmDataByService(
    std::shared_ptr<bmcweb::AsyncResp> asyncResp, const std::string& dimmId,
    const std::string& service, const std::string& objPath)
{
   ...
  name_util::getPrettyName(asyncResp, objPath, service, "/Name"_json_pointer);
  ...
  dbus::utility::getAllProperties(
        ....
        });
}
```

deepakala-k: I will better do it here, because in the future if people look into assembleDimmProperties and start questioning why we are again making separate call for pretty name, it might be confusing.

redfish-core/lib/assembly.hpp
```diff
@@ -21,6 +21,7 @@
 #include <nlohmann/json.hpp>
 #include <sdbusplus/asio/property.hpp>
 #include <sdbusplus/unpack_properties.hpp>
+#include <utils/name_utils.hpp>
```
baemyung, line 24: This would be at  L16 
```
#include "utils/name_utils.hpp"
```

redfish-core/lib/fabric_adapters.hpp
```diff
@@ -27,6 +27,7 @@
 #include <boost/url/format.hpp>
 #include <sdbusplus/message/native_types.hpp>
 #include <sdbusplus/unpack_properties.hpp>
+#include <utils/name_utils.hpp>
```
baemyung, line 30: Move to L20
```
#include "utils/name_utils.hpp"
```

redfish-core/lib/fabric_ports.hpp
```diff
@@ -25,6 +25,7 @@
 #include <boost/system/error_code.hpp>
 #include <boost/url/format.hpp>
 #include <sdbusplus/asio/property.hpp>
+#include <utils/name_utils.hpp>
```
baemyung, line 28: Move to L20
```
#include "utils/name_utils.hpp"
```

redfish-core/lib/memory.hpp
```diff
@@ -623,6 +622,7 @@ inline void getDimmDataByService(
     const std::string& service, const std::string& objPath)
 {
     BMCWEB_LOG_DEBUG("Get available system components.");
+    name_util::getPrettyName(asyncResp, objPath, service, "/Name"_json_pointer);
```
baemyung, line 625: Missing 
```
#include "utils/name_utils.hpp"
```

redfish-core/lib/processor.hpp
baemyung: The fix seems good for 1120.
Would you also create a PR for 1110?

redfish-core/lib/memory.hpp
```diff
@@ -19,6 +19,7 @@
 #include "utils/collection.hpp"
 #include "utils/dbus_utils.hpp"
 #include "utils/hex_utils.hpp"
+#include "utils/name_utils.hpp"
 #include "utils/json_utils.hpp"
```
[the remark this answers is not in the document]
deepakala-k: sorry, my docker does not work. And I have to use the CI to look for the formatting changes. Thanks a lot! I will fix my docker soon!

redfish-core/lib/storage.hpp
```diff
@@ -664,7 +666,8 @@ inline void afterGetSubtreeSystemsStorageDrive(
     asyncResp->res.jsonValue["@odata.id"] =
         boost::urls::format("/redfish/v1/Systems/{}/Storage/1/Drives/{}",
                             BMCWEB_REDFISH_SYSTEM_URI_NAME, driveId);
-    asyncResp->res.jsonValue["Name"] = driveId;
+    name_util::getPrettyName(asyncResp, path, drive->second,
```
baemyung, line 669: L663 sets `connectionNames` for `drive->second`.
So, this may be better to be like
```
name_util::getPrettyName(asyncResp, path, connectionNames,
```

baemyung (commented):
One more comment. I think it is getting close.

redfish-core/lib/storage.hpp
deepakala-k: done :)

deepakala-k:
I will get the 1110 PR by tomorrow.

deepakala-k:
Just started it. https://github.com/ibm-openbmc/bmcweb/pull/1234
I need to work further on it

baemyung (approved):
Looks good to me

baemyung:
> > There is also a few piece.
> > 
> > * pcie.hpp.
> >   In fact,  1110 code also seems incorrect (similar processor.hpp).  I think getPrettyName() should be moved into `addPCIeDeviceCommonProperties`
> 
> The problem is the same. We do not have the service name and probably the best way would be move those into a function and call it both from get and patch. I guess handle was created with that intent. But we need to create another level of method may be!

I thought getPrettyName for pcie.hpp is handled by this PR....but it turns out it is missed.

baemyung:
> > > There is also a few piece.
> > > 
> > > * pcie.hpp.
> > >   In fact,  1110 code also seems incorrect (similar processor.hpp).  I think getPrettyName() should be moved into `addPCIeDeviceCommonProperties`
> > 
> > 
> > The problem is the same. We do not have the service name and probably the best way would be move those into a function and call it both from get and patch. I guess handle was created with that intent. But we need to create another level of method may be!
> 
> I thought getPrettyName for pcie.hpp is handled by this PR....but it turns out it is missed.

It looks like another PR https://github.com/ibm-openbmc/bmcweb/pull/1201 is trying to handle for pcie.hpp.

PR #1218, opened 2025-03-27
Audit Log: Rebase service audit log

jeaaustx:
Rebase and squashed all 1110 bmcweb commits to support service audit log. 1110 commits included:
```
957cc7990c Audit Log: Rebase service audit log (#820)
5fd2e37e03 Audit Log: Correct logging of sessions (#1129)
2cc4e01f4c Reduce memory usage during Firmware update (#739)
```

Includes restructure of the code to split into audit-events.hpp and audit-events.cpp files. Only the .cpp file will include libaudit.h and will only be built if audit-events is enabled. This allows use of constexpr for calls to auditEvent().

Tested on hardware simulator. Built version with audit-events enabled and confirmed audit entries are recorded as expected. Note recipe changes are needed to enable linux audit service which will be part of an openbmc/openbmc PR.

This introduces a new bmcweb meson option, audit-events, which is defaulted to disabled. Recipe changes in GHE openbmc/openbmc enable audit logging.

When audit-events is enabled all bmcweb PATCH, POST, PUT, and DELETE events are logged using the Linux kernel auditd subsystem. Additionally, login events coming through bmcweb are also recorded.

The body of the events is recorded as well except in the following cases:
  - /redfish/v1/AccountService/Accounts PATCH/POST events - body is not recorded as it may contain clear text password
  - /ibm/v1 PATCH/POST/PUT events - body is not recorded as it contains HMC config file binary data
  - Limit size of body recorded to avoid flooding log or using too much memory.

Events are recorded in /var/log/audit/. User type dreport will gather these log files.

Tested:
 - Enabled auditing then initiated a variety of Redfish events using curl.
 - Confirmed the events were recorded and confirmed the data recorded was accurate.
 - Confirmed password was not included in audit data recorded.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -396,6 +396,12 @@ srcfiles_bmcweb = files(
     'src/webserver_run.cpp',
 )
 
+if (get_option('audit-events').allowed())
+    audit_dep = dependency('audit')
+    bmcweb_dependencies += audit_dep
+    srcfiles_bmcweb += files('src/audit_events.cpp')
+endif
+
 bmcweblib = static_library(
     'bmcweblib',
     srcfiles_bmcweb,
@@ -482,6 +488,10 @@ srcfiles_unittest = files(
     'test/redfish-core/lib/thermal_subsystem_test.cpp',
     'test/redfish-core/lib/update_service_test.cpp',
 )
+
+if (get_option('audit-events').allowed())
+    srcfiles_unittest += files('test/include/audit_events_test.cpp')
+endif
 
 if (get_option('tests').allowed())
     gtest = dependency(
--- a/meson.options
+++ b/meson.options
@@ -461,6 +461,14 @@ option(
                     behavior changes or be removed at any time.''',
 )
 
+# BMCWEB_AUDIT_EVENTS
+option(
+    'audit-events',
+    type: 'feature',
+    value: 'disabled',
+    description: 'Enable audit events support for bmcweb',
+)
+
 # Insecure options. Every option that starts with a `insecure` flag should
 # not be enabled by default for any platform, unless the author fully comprehends
 # the implications of doing so.In general, enabling these options will cause security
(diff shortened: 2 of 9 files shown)


meson.options
```diff
@@ -461,6 +461,14 @@ option(
                     behavior changes or be removed at any time.''',
 )
 
+# BMCWEB_AUDIT_EVENTS
+option(
+    'audit-events',
+    type: 'feature',
+    value: 'disabled',
```
gtmills, line 468: Want to enable this by default? There was a commit on the spreadsheet that does this

gtmills (commented):
Overall seems reasonable

meson.build
```diff
@@ -396,6 +401,10 @@ srcfiles_bmcweb = files(
     'src/webserver_run.cpp',
 )
 
+if (get_option('audit-events').allowed())
+    srcfiles_bmcweb += files('src/audit_events.cpp')
+endif
+
```
baemyung, line 407: Wouldn't it be better to combine L248-L252 and L404-407 into here for bmcweblib ?

```
if (get_option('audit-events').allowed())
    audit_dep = dependency('audit')
    bmcweb_dependencies += audit_dep
    srcfiles_bmcweb += files('src/audit_events.cpp')
endif
```

baemyung (approved):
It looks good to me - with a minor comment

meson.options
jeaaustx: Yes this is wanted. Without the recipe changes the libaudit is not being found. I think a wrap file is needed to fix that. I looked a bit at that but it doesn't have a yocto package and its download looks different from others. I couldn't get the wrap file to work.

So, my plan is to add the openbmc/openbmc recipe updates separately. Then when I add the bmcweb commit which adds the customer audit log support I'll enable the option by default. Eventually I will have to figure out the wrap file.

meson.build
jeaaustx: That makes sense. I've moved them together as you suggested.

jeaaustx:
Last push after Myung's approval was just a rebase to get CI to pass.

PR #1061, opened 2024-09-06
Fix logic excluding groups

gtmills:
https://github.com/ibm-openbmc/bmcweb/commit/b9468b98a916eb80b133d24b3565f6ff5fc9433a came in but it didn't actually set the group it passed to modGroupsList https://github.com/ibm-openbmc/bmcweb/commit/b9468b98a916eb80b133d24b3565f6ff5fc9433a#diff-a6441dc86bbef5cbe2192c66717e8b938f9ee3a02670ff54c702338095a0d660R2200 

We also already have a spot where we are excluding groups so just move this logic to exclude ipmi and ssh (when user is not admin) to that loop https://github.com/ibm-openbmc/bmcweb/blob/1110/redfish-core/lib/account_service.hpp#L1927

```
Sep 06 18:19:40 xx useradd[1674]: add 'testadmin1' to group 'priv-admin'
Sep 06 18:19:40 xx useradd[1674]: add 'testadmin1' to group 'web'
Sep 06 18:19:40 xx useradd[1674]: add 'testadmin1' to group 'redfish'
Sep 06 18:19:40 xx useradd[1674]: add 'testadmin1' to group 'hostconsole'

Sep 06 18:21:03 xx useradd[1722]: add 'readonly1' to group 'priv-user'
Sep 06 18:21:03 xx useradd[1722]: add 'readonly1' to group 'web'
Sep 06 18:21:03 xx useradd[1722]: add 'readonly1' to group 'redfish'
```
Matches the traces 
```
Sep 06 18:19:40 xxx bmcweb[1115]: [ERROR account_service.hpp:1938] group added hostconsole
Sep 06 18:19:40 xx bmcweb[1115]: [ERROR account_service.hpp:1945] group skipped ipmi
Sep 06 18:19:40 xx bmcweb[1115]: [ERROR account_service.hpp:1938] group added redfish
Sep 06 18:19:40 xx bmcweb[1115]: [ERROR account_service.hpp:1938] group added ssh
Sep 06 18:19:40 xx bmcweb[1115]: [ERROR account_service.hpp:1938] group added web


Sep 06 18:21:03 xx bmcweb[1115]: [ERROR account_service.hpp:1945] group skipped ipmi
Sep 06 18:21:03 xx bmcweb[1115]: [ERROR account_service.hpp:1938] group added redfish
Sep 06 18:21:03 xx bmcweb[1115]: [ERROR account_service.hpp:1945] group skipped ssh
[...]

PR #1221, opened 2025-04-01
1120: ReadOnly users should not have ssh (#890)(#1061)

baemyung:
ReadOnly users should not have ssh #890)((#1061)

1. ReadOnly users should not have ssh (#890)

This changes the Redfish create new user API (POST
/redfish/v1/AccountService/Accounts/) so only Administrator users
will have the HostConsole AccountType value.  This is the same as
the "ssh" phosphor privilege group.

Note that ReadOnly user created prior to this change will have the
HostConsole AccountType privilege.  A BMC admin can PATCH a user's
AccountTypes property to add or remove the Hostconsole value.

Tested: Yes

2. Fix logic excluding groups (#1061)

https://github.com/ibm-openbmc/bmcweb/commit/b9468b9 came in but it
didn't actually set the group but it passed to modGroupsList [1].

We also already have a spot where we are excluding groups so just move
this logic to exclude ipmi and ssh (when user is not admin) to that
loop [2].

[1] https://github.com/ibm-openbmc/bmcweb/commit/b9468b9#diff-a6441dcR2200
[2] https://github.com/ibm-openbmc/bmcweb/blob/1110/redfish-core/lib/account_service.hpp#L1927

### DIFF
--- a/redfish-core/lib/account_service.hpp
+++ b/redfish-core/lib/account_service.hpp
@@ -2112,6 +2112,16 @@ inline void processAfterGetAllGroups(
             }
             continue;
         }
+
+        // Remove the ipmi group.  Also Remove "ssh" if the new
+        // user is not an Administrator.
+        if ((grp == "ipmi") || ((grp == "ssh") && (roleId != "priv-admin")))
+
+        {
+            BMCWEB_LOG_DEBUG("group skipped {}", grp);
+            continue;
+        }
+
         userGroups.emplace_back(grp);
     }
 


gtmills:
Does this include https://github.com/ibm-openbmc/bmcweb/commit/717d0166a4ebd45c7c9dc2c3120b44221cadf691 ?

baemyung:
> Does this include [717d016](https://github.com/ibm-openbmc/bmcweb/commit/717d0166a4ebd45c7c9dc2c3120b44221cadf691) ?

Thanks for pointing out it.
I included and squashed as one commit.

redfish-core/lib/account_service.hpp
```diff
@@ -2186,7 +2196,7 @@ inline void handleAccountCollectionPost(
                        const std::vector<std::string>& allGroupsList) {
             if (ec)
             {
-                BMCWEB_LOG_ERROR("D-Bus response error {}", ec);
+                BMCWEB_LOG_ERROR("ERROR with async_method_call {}", ec);
```
gtmills, line 2199: Is this still needed?

gtmills (approved):

redfish-core/lib/account_service.hpp
baemyung: I think the previous one seems good enough and it seems consistent with the other places.
```
BMCWEB_LOG_ERROR("D-Bus response error {}", ec);
```

gtmills: Let's revert this one line and then this can go

PR #920, opened 2024-04-20
1110: Add ACFWindowActive and AllowUnauthACFUpload (#525)(#678)

baemyung:
1. Add ACFWindowActive to /redfish/v1 (#331) (#525)

* Add ACFWindowActive to /redfish/v1 (#331)

This adds the Oem.IBM.ACFWindowActive boolean property to the Redfish service root at URI /redfish/v1.

Tested:
Via curl https://${BMC}/redfish/v1
Tested values true, false, and panel app not running. Simulated pressing function 74 via busctl set-property "com.ibm.PanelApp"
  "/com/ibm/panel_app" "com.ibm.panel" "ACFWindowActive" b true

* Add ACFWindowActive to /redfish/v1

This restructures the ACFWindowActive code to BMCWeb coding style. This defaults to ACFWindowActive=false when the value is not available. This add the XML schema.

Tested:
Tested both true and false values via:
curl -k https://127.0.0.1:2643/redfish/v1
and simulating button press via
busctl set-property "com.ibm.PanelApp" "/com/ibm/panel_app" \
  "com.ibm.panel" "ACFWindowActive" b true

2. New property to always AllowUnauthACFUpload (#466) (#678)

* New property to always AllowUnauthACFUpload

This adds a new BMC security setting as a Redfish OEM property: URI /redfish/v1/AccountService/Accounts/service property Oem.IBM.ACF.AllowUnauthACFUpload can be set by the admin.  When true, any network agent is allowed to upload an ACF file to the BMC. The default value is false.  This capability is included in the existing URI /redfish/v1 property Oem.IBM.ACFWindowActive.  For example, when AllowUnauthACFUpload=true, ACFWindowActive is true.

Tested: via Redfish scenarios
1. Ensure AllowUnauthACFUpload=false by default.
2. Ensure a non admin user is not allowed to GET or PATCH AllowUnauthACFUpload.
3. Ensure the admin can PATCH AllowUnauthACFUpload true and false, and GET the correct values, and these values survive BMC reboot.
4. With (OpPanel fn 74 is not active) and AllowUnauthACFUpload=false, ensure ACFWindowActive=false.  And unauth ACF upload is not allowed.
5. With (OpPanel fn 74 is not active) and AllowUnauthACFUpload=true,

Cheatsheet:
```
$ busctl set-property com.ibm.PanelApp /com/ibm/panel_app \
  com.ibm.panel ACFWindowActive b true

[...]

PR #1224, opened 2025-04-03
1110: Bypass uploadACF on PATCH if no ACF is provided

baemyung:
https://github.com/ibm-openbmc/bmcweb/pull/920 mistakenly continues to perform uploadACF even if ACF is not provided on PATCH.

For example, initially ACF is cleared.

```
acf=""
curl -k -H "Content-Type: application/json"  -X PATCH \
    -d '{"Oem":{"IBM":{"ACF":{"ACFFile":"'${acf}'"}}}}' \
    https://${bmc}/redfish/v1/AccountService/Accounts/service
```

And then, the following PATCH may still go through uploadACF.
```
curl -k -H "Content-Type: application/json"  -X PATCH \
    -d '{"Oem":{}}'  \
    https://${bmc}/redfish/v1/AccountService/Accounts/service
...
Apr 03 14:30:45 p10bmc bmcweb[1043]: [ERROR account_service.hpp:1608] ACF upload not allowed
```

This fixes to skip uploadACF if PATCH input does not contain `ACF`.

Tested:
- Verify whether uploadACF is not performed if PATCH input does not contain `ACF`. For example, `PATCH -d '{"Oem":{}}'`
- Redfish Service Validator passes.

### DIFF
--- a/redfish-core/lib/account_service.hpp
+++ b/redfish-core/lib/account_service.hpp
@@ -1507,6 +1507,60 @@ inline void uploadACF(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
         "InstallACF", decodedAcf);
 }
 
+// Allow ACF upload when D-Bus property allow_unauth_upload is true (aka Redfish
+// property AllowUnauthACFUpload).
+inline void doUnauthenticatedACFUpload(
+    const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
+    const std::vector<uint8_t>& decodedAcf)
+{
+    sdbusplus::asio::getProperty<bool>(
+        *crow::connections::systemBus, "xyz.openbmc_project.Settings",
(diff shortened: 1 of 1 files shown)


redfish-core/lib/account_service.hpp
```diff
@@ -1564,56 +1618,8 @@ inline void triggerUnauthenticatedACFUpload(
             messages::internalError(asyncResp->res);
             return;
         }
+        doUnauthenticatedACFUpload(asyncResp, decodedAcf);
```
gtmills, line 1621: This moves it in the 
```
    if (acf)
{
```
and moves it to a function? If so, this looks correct

baemyung: Yes, it is just moved to a function

jeaaustx (approved):

gtmills (approved):

PR #1225, opened 2025-04-04
1120: Refresh of PCIe Topology (#593)

baemyung:
This implements PATCH of PCIeTopologyRefresh & SavePCIeTopologyInfo.

This pulls in the following 1030 PRs:

- https://github.com/ibm-openbmc/bmcweb/pull/377
- https://github.com/ibm-openbmc/bmcweb/pull/382
- https://github.com/ibm-openbmc/bmcweb/pull/404
- https://github.com/ibm-openbmc/bmcweb/pull/406

This commit also adds functionality, where bmcweb creates a watchdog (using timer), where timer waits for one second and then checks if PCIe Topology gets refreshed or not. If not, then wait for one more second and repeat this process eight times, and then then watchdog expires in between if it gets refreshed then unblock the request.

Tested:

- Successful PATCH:

```
curl -k -H "Content-Type: application/json" -X PATCH https://${bmc}/redfish/v1/Systems/system/ -d '{"Oem":{"IBM":{"PCIeTopologyRefresh":true}}}'
curl -k -H "Content-Type: application/json" -X PATCH https://${bmc}/redfish/v1/Systems/system/ -d '{"Oem":{"IBM":{"SavePCIeTopologyInfo":true}}}'
```

### DIFF
--- a/redfish-core/lib/oem/ibm/pcie_topology_refresh.hpp
+++ b/redfish-core/lib/oem/ibm/pcie_topology_refresh.hpp
@@ -1,0 +1,166 @@
+#pragma once
+
+#include "async_resp.hpp"
+#include "dbus_singleton.hpp"
+#include "error_messages.hpp"
+#include "http_request.hpp"
+#include "io_context_singleton.hpp"
+#include "logging.hpp"
+
+#include <sys/types.h>
+
+#include <boost/asio/steady_timer.hpp>
+#include <boost/system/error_code.hpp>
+#include <sdbusplus/asio/property.hpp>
+
+#include <chrono>
+#include <memory>
+#include <string>
+
+namespace redfish
+{
+// NOLINTBEGIN(cppcoreguidelines-avoid-non-const-global-variables)
+// Timer for PCIe Topology Refresh
+static std::unique_ptr<boost::asio::steady_timer> pcieTopologyRefreshTimer;
+static uint countPCIeTopologyRefresh = 0;
+// NOLINTEND(cppcoreguidelines-avoid-non-const-global-variables)
+
+/**
+ * @brief PCIe Topology Refresh monitor. which block incoming request
+ *
+ * @param[in] ec          error code to be handled
+ * @param[in] timer       pointer to steady timer which block call
+ * @param[in] asyncResp   Shared pointer for generating response message.
+ * @param[in] countPtr   how many time this method get called.
+ *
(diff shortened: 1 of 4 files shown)


redfish-core/lib/oem/ibm/pcie_topology_refresh.hpp
```diff
@@ -0,0 +1,166 @@
...
+#include <boost/system/error_code.hpp>
+#include <sdbusplus/asio/property.hpp>
+
+#include <chrono>
+#include <memory>
+#include <string>
+
+namespace redfish
+{
+// NOLINTBEGIN(cppcoreguidelines-avoid-non-const-global-variables)
```
gtmills, line 22: I wish we could redesign this.. 
Maybe something we could discuss for PFuture

redfish-core/lib/systems.hpp
```diff
@@ -3461,7 +3464,9 @@ inline void handleComputerSystemPatch(
             "PowerMode", powerMode,                                        //
             "PowerRestorePolicy", powerRestorePolicy,                      //
             "Oem/IBM/ChapData/ChapName", chapName,                         //
-            "Oem/IBM/ChapData/ChapSecret", chapSecret                      //
+            "Oem/IBM/ChapData/ChapSecret", chapSecret,                     //
+            "Oem/IBM/PCIeTopologyRefresh", pcieTopologyRefresh,            //
```
gtmills, line 3468: This is all the PATCH, we didn't have something on the GET here?

baemyung: PCIeTopologyRefresh is only for PATCH, not GET.
I think it is to trigger PLDM PCIeTopologyRefresh action.

redfish-core/lib/oem/ibm/pcie_topology_refresh.hpp
baemyung: I also agree...maybe for PFuture, and perhaps, put those static variables onto `App`?

redfish-core/lib/systems.hpp
gtmills: It should be there for GET too.. Because otherwise how would client discover it.. That said, it was this way before

gtmills (approved):
Seems reasonable for what was there.. Maybe let's open a story to refactor this for PFuture

PR #1230, opened 2025-04-07
1120: Create Event Subscription meson build option (#941)

baemyung:
Create a Meson build option to enable and disable event subscription at /subscribe. 
By default, this is enabled

### DIFF
--- a/meson.options
+++ b/meson.options
@@ -440,6 +440,14 @@ option(
     resources such as EthernetInterfaces and ComputerSystem.Reset.''',
 )
 
+# BMCWEB_EVENT_SUBSCRIPTION
+option(
+    'event-subscription',
+    type: 'feature',
+    value: 'enabled',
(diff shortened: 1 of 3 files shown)


meson.options
```diff
@@ -441,6 +441,13 @@ option(
 )
 
 # BMCWEB_EXPERIMENTAL_REDFISH_MULTI_COMPUTER_SYSTEM
```
gtmills, line 443: This looks incorrect. Can you move this down and then also add a comment for this option ?

baemyung: Right.
Fixed

gtmills (approved):
Do we limit what routes can be accessed anywhere?

baemyung:
> Do we limit what routes can be accessed anywhere?

I don't know whether there is.

gtmills:
> I don't know whether there is.

We can handle that later

jeaaustx (approved):

PR #1231, opened 2025-04-07
1120: Add BMC_SHELL web-base_shell (#483)(#473) (#936)

baemyung:

Path of WebSocket:
    wss://{hostname}/bmc-console

Implementation:

    like ls, cd, etc.

Authorize user for Shell :

Test:

    The essential operation can perform on this xterm console, Ex: ls,
    cd, grep, etc.

### DIFF
--- a/meson.options
+++ b/meson.options
@@ -543,3 +543,13 @@ option(
                     enable on production systems at this time.  Other query
                     parameters such as only are not controlled by this option.''',
 )
+
+# BMCWEB_BMC_SHELL_SOCKET
+option(
+    'bmc-shell-socket',
+    type: 'feature',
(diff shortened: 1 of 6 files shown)


meson.options
```diff
@@ -543,3 +543,12 @@ option(
                     enable on production systems at this time.  Other query
                     parameters such as only are not controlled by this option.''',
 )
+
```
gtmills, line 546: Do you want to add a comment here?

gtmills (approved):
Just continueing this forward

meson.options
baemyung: Added

gtmills (approved):

PR #1246, opened 2025-04-16
WIP: Request body serializer

abhilashraju:
bmcweb changes for serialising request body into file
Signed-off-by: Abhilash Raju <abhilash.kollam@gmail.com>

### DIFF
--- a/http/app.hpp
+++ b/http/app.hpp
@@ -182,6 +182,20 @@ class App
     {
         return router.getRoutes(parent);
     }
+    void prepareRequestBody(boost::beast::http::request<bmcweb::HttpBody>& req)
+    {
+        auto result = boost::urls::parse_relative_ref(req.target());
+        if (!result)
+        {
+            return;
+        }
+        auto url = *result;
+        auto routeRes = router.findRoute(url, req.method());
+        if (routeRes.route.rule)
+        {
+            routeRes.route.rule->prepareBody(req);
+        }
+    }
 
     std::optional<server_type> server;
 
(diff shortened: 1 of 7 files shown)


baemyung:
This seems putting the request body into a file - which seems good for the given purpose.
However, I'm wondering how `body()` function will be handled as it becomes empty for this route.

bmcweb uses `body()` in many places.  
How can we make sure not to use it in this path?

Perhaps, can this feature be enabled by an meson option, and/or by a dynamic option (as a part of Request header)?
Also, going through upstream community first?

abhilashraju:
> This seems putting the request body into a file - which seems good for the given purpose. However, I'm wondering how `body()` function will be handled as it becomes empty for this route.
> 
> bmcweb uses `body()` in many places. How can we make sure not to use it in this path?
> 
> Perhaps, can this feature be enabled by an meson option, and/or by a dynamic option (as a part of Request header)? Also, going through upstream community first?

Serialization is not applied by default. You can refer to the update service route registration handler to see how serialization is enabled for a specific URL. In these cases, the handler chooses to stream the request body directly into a file, so it does not use the body() call on the request object. By default, all other handlers read the body into a string as usual.

gtmills:
@abhilashraju Do we still want this? Do we want to open a story to explore this for PFuture ?
I know this was originally opened for file body code update.. Relating to running out of memory. 
https://github.com/ibm-openbmc/bmcweb/pull/1240 
https://github.com/ibm-openbmc/bmcweb/pull/1248 
was some commits we did around this area. Also a lot of commits in the telemetry area. 

I am trying to figure out what we have left for 1120 bmcweb, trying to lock it down a bit.

abhilashraju:
I think I can close this . Yes this was done as part of POC for out memory issue. If there is a requirement to serialize large body content to filesytem, then we can rework on this

PR #1254, opened 2025-04-21
http_client: fixing ip displayed in error_log

abhilashraju:
The IP address shown in the error log on connection failure was incorrect because the endpoint received in the connection callback is invalid when the connection fails. The code has been updated to log the host and port used for the connection attempt instead.


baemyung:
I think this deserves the upstream commit.

raviteja-b (approved):

abhilashraju:
> I think this deserves the upstream commit.

Done

baemyung (approved):
Looks good to me

PR #1261, opened 2025-04-23
1120: Redfish : Look for OperationalStatus for the memory Health

abiasojo:

Tested: Validator passes.
```
{
  "Status": {
    "Health": "Critical",
    "State": "Enabled"
  }
}
```

### DIFF
--- a/redfish-core/lib/memory.hpp
+++ b/redfish-core/lib/memory.hpp
@@ -435,6 +435,7 @@ inline void assembleDimmProperties(
     const std::string* sparePartNumber = nullptr;
     const std::string* model = nullptr;
     const std::string* locationCode = nullptr;
+    const bool* functional = nullptr;
 
     const bool success = sdbusplus::unpackPropertiesNoThrow(
         dbus_utils::UnpackErrorPrinter(), properties, "MemoryDataWidth",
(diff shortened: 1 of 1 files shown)


gtmills:
https://gerrit.openbmc.org/c/openbmc/bmcweb/+/69859 is upstream

gtmills (approved):

redfish-core/lib/memory.hpp
```diff
@@ -615,6 +616,14 @@ inline void assembleDimmProperties(
             *locationCode;
     }
 
+    if (functional != nullptr)
+    {
+        if (!*functional)
+        {
+            asyncResp->res.jsonValue[jsonPtr]["Status"]["Health"] = "Critical";
```
gtmills, line 623: resource::Health::Critical here please :)

gtmills (changes requested):

jeaaustx (commented):
Just the same comment as Gunnar, but otherwise looks good to me.

redfish-core/lib/memory.hpp
abiasojo: > resource::Health::Critical here please :)

Done

abiasojo:
> Just the same comment as Gunnar, but otherwise looks good to me.

Done.

gtmills (approved):

PR #1262, opened 2025-04-24
Restrict Resource Dump Creation to service user

asmithakarun:
This change ensures that resource dump creation requests are permitted only when initiated by a service user — specifically, a user with the "priv-oemibmserviceagent" privilege.

Requests from any other users will be denied with an "Insufficient Privilege" error response from bmcweb.

The rationale behind this restriction is that resource dumps are intended solely for debugging issues in resources of the host, which are to be collected exclusively by service user. As such, access is limited to service users and explicitly restricted for all others.

Fixes: https://jazz07.rchland.ibm.com:13443/jazz/web/projects/CSSD#action=com.ibm.team.workitem.viewWorkItem&id=681755

Tested By:
[1] POST https://${bmc}/redfish/v1/Systems/system/LogServices/Dump/Actions/LogService.CollectDiagnosticData -d '{"DiagnosticDataType":"OEM","OEMDiagnosticDataType":"Resource_<string>_<pwd>"}'

Performed above operation for service user and admin user. The request will be successful for service user and errors out for admin user.

### DIFF
--- a/redfish-core/lib/log_services.hpp
+++ b/redfish-core/lib/log_services.hpp
@@ -15,6 +15,7 @@
 */
 #pragma once
 
+#include "account_service.hpp"
 #include "app.hpp"
 #include "assembly.hpp"
 #include "dbus_utility.hpp"
@@ -1311,6 +1312,22 @@ inline void createDump(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
             }
             if (resourceDumpParams.size() == 3)
             {
+                // Check if admin user is initiating resource dump with
+                // non-empty pwd
+                std::string roleId =
+                    getRoleIdFromPrivilege(req.session->userRole);
+                if (roleId == "Administrator")
+                {
+                    if (std::holds_alternative<std::string>(
+                            resourceDumpParams[2]) &&
+                        !std::get<std::string>(resourceDumpParams[2]).empty())
+                    {
+                        BMCWEB_LOG_WARNING(
+                            "Request to create resource dump failed. User has insufficient privilege.");
+                        messages::insufficientPrivilege(asyncResp->res);
+                        return;
+                    }
+                }
                 createDumpParams.emplace_back(
                     "com.ibm.Dump.Create.CreateParameters.Password",
                     resourceDumpParams[2]);
@@ -1325,6 +1342,7 @@ inline void createDump(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
                         "/redfish/v1/Systems/system/LogServices/Dump/Actions/LogService.CollectDiagnosticData"));
                 return;
             }
+
             createDumpParams.emplace_back(
                 "com.ibm.Dump.Create.CreateParameters.DumpType",
                 "com.ibm.Dump.Create.DumpType.Resource");


raviteja-b (approved):

joseph-reynolds (changes requested):
The design calls for checking: "a Resource dump password is only allowed for users who have service privilege."
Please note this function's authority check is unusually complex:
1. The BMC "Initiate Resource Dump" function is allowed for role=Administrator and for the service user.  Admin users must be allowed to issue resource dumps, and PHYP checks their authority based on the exact macro they invoked.  This current authority check for this is correct, and it must not change.
2. The missing authority check was only for the "third" part of the resource dump parameter "OEMDiagnosticDataType".  This parameter has three parts, separated by the underscore character ("_"): (1) the literal string "Resource", (2) the dump invocation string, and (3) the service user password (see the invocation below).

The check required is: if (this password field is not empty) AND (the user is not the service user) THEN issue an insufficient privilege error.

The new or modified test cases are for the Redfish API
> POST https://$bmc/redfish/v1/Systems/system/LogServices/Dump/Actions/LogService.CollectDiagnosticData -d '{"DiagnosticDataType":"OEM", "OEMDiagnosticDataType":"**Resource**\_**macro-invocation**\_**optional-password**"}'

Note about test validation steps: For these tests, we don't care what PHYP does.  The important part is whether (A) the BMC passes the macro into HMC, or (B) the BMC gives an insufficient privilege error message.

1. From a session with Role=Administrator, initiate a resource dump and do not supply the service user password (the password field is omitted or is the empty string).  Verify the BMC passes the dump into PHYP.
2.  From a session with Role=Administrator, initiate a resource dump with a non-empty service user password.  Verify the BMC stops with insufficient privilege and does not pass the dump into PHYP.
3. From a session with the service user, initiate a resource dump with a non-empty service user password.  Verify the BMC passes the dump to PHYP.

Bonus: The following existing test case must continue to work: the service user initiates a resource dump with no password (the password field is omitted or is the empty string).  Verify the BMC passes the dump into PHYP.  <-- This tests that the service user is allowed to issue dumps which only require admin authority.

redfish-core/lib/log_services.hpp
```diff
@@ -1281,6 +1281,16 @@ inline void createDump(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
         }
         if ((*oemDiagnosticDataType).starts_with("Resource"))
         {
+            // Check if service user is requesting the dumps
+            // and then proceed with dump collection
+            if (req.session &&
+                req.session->userRole != "priv-oemibmserviceagent")
```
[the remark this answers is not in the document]
asmithakarun: 2 will take both "operator and service" as restricted roles (which, cannot be used in this use case, where we check for admin and service user privileges)

I think 3 seems to be better. I have pushed it in the latest PS.

redfish-core/lib/log_services.hpp
```diff
@@ -1281,6 +1282,16 @@ inline void createDump(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
...
+            // Check if service user is requesting the dumps
+            // and then proceed with dump collection
+            std::string roleId = getRoleIdFromPrivilege(req.session->userRole);
+            if (roleId != "OemIBMServiceAgent" && roleId != "Administrator")
+            {
+                BMCWEB_LOG_WARNING(
+                    "Request to create resource dump failed. User is not a service/admin user");
+                messages::insufficientPrivilege(asyncResp->res);
+                return;
+            }
```
joseph-reynolds, line 1294: This check is redundant.  The Admin/service privilege was already checked via
requestRoutesBMCDumpCreate() BMCWEB_ROUTE(...).privileges(redfish::privileges::postLogService).
You will need the RoleId for the check below.
The code is okay, but technically this should check if the session has privileges({{"OemIBMPerformService"}}),
example code in https://github.com/ibm-openbmc/bmcweb/blob/1110/redfish-core/lib/account_service.hpp#L1712

asmithakarun: > This check is redundant.

Agree. Will remove it.

> but technically this should check if the session has privileges({{"OemIBMPerformService"}}),

But this. Why is it any different from what I have used?
Basically, roleId is assigned based on the privilege (priv-admin or priv-oemserviceagent). Just thinking why this shouldn't be used.

joseph-reynolds (approved):

redfish-core/lib/log_services.hpp
joseph-reynolds: > > but technically this should check if the session has privileges({{"OemIBMPerformService"}}),
> 
> But this. Why is it any different from what I have used? Basically, roleId is assigned based on the privilege (priv-admin or priv-oemserviceagent). Just thinking why this shouldn't be used.

The Redfish authority model maps user accounts to Roles, maps Roles to Redfish Privileges, and maps Operations to Privileges.  Ideally all authority checks would be checking if the user's Privileges are enough to perform the privileges required by the Operation.  But the service user is very special, has the Restricted ServiceAgent  Role which has a special privilege OemIBMPerformService... so the check for Role has the same outcome as the check for privilege.

asmithakarun: So, the current change that I have still should be okay I believe.

joseph-reynolds (approved):

PR #1265, opened 2025-04-25
PowerSubsystem: Add Allocation properties (#1044)

cjcain:
Add the Allocation properties to the PowerSubsystem output.

Set the Redfish properties based on properties from the D-Bus xyz.openbmc_project.Control.Power.Cap interface:
* Redfish AllocatedWatts property:
  * Set to PowerCap property if PowerCapEnable is true
  * Set to MaxPowerCapValue property if PowerCapEnable is false
* Redfish RequestedWatts property:
  * Set to MaxPowerCapValue property

Tested:
* Ran the Redfish Validator; no new errors/warnings
* For detailed test plan, see https://gist.github.com/smccarney/1fc1426e51d0fd9ffc715b4d522f4cf9

Example Output:
```
$ curl -k https://${bmc}/redfish/v1/Chassis/chassis/PowerSubsystem
{
  "@odata.id": "/redfish/v1/Chassis/chassis/PowerSubsystem",
  "@odata.type": "#PowerSubsystem.v1_1_0.PowerSubsystem",
  "Allocation": {
    "AllocatedWatts": 1233,
    "RequestedWatts": 3982
  },
  "Id": "PowerSubsystem",
  "Name": "Power Subsystem",
  "PowerSupplies": {
    "@odata.id": "/redfish/v1/Chassis/chassis/PowerSubsystem/PowerSupplies"
  },
  "Status": {
    "Health": "OK",
    "State": "Enabled"
  }
}
```

Change-Id: I0d5e6dc251a66de1c1207257100e13df6f35047f


test/redfish-core/lib/power_subsystem_test.cpp
```diff
@@ -1,47 +0,0 @@
-// SPDX-License-Identifier: Apache-2.0
```
gtmills, line 1: We can't fix this to pass?

cjcain: @smccarney

smccarney: I'm not sure what is required to fix this.  At the time in 1050 we made the decision to remove the test case because it was very minimal anyway (wasn't doing much).

gtmills: If someone gets time, it would be good to look at this but I am okay with staying with what we have

gtmills (approved):

test/redfish-core/lib/power_subsystem_test.cpp
smccarney: The reason the test crashes is that tons of bmcweb code including this code uses `crow::connections::systemBus`. 

That is declared in include/dbus_singleton.hpp
```
extern sdbusplus::asio::connection* systemBus;
```

and defined in src/dbus_singleton.cpp
```
sdbusplus::asio::connection* systemBus = nullptr;
```

In the real bmcweb application, this gets set to a real sdbusplus::asio::connection.  But in test code (at least in this test file), that is still nullptr.  I added a std::cout to the code to verify this:

```
{
    std::cout << "crow::connections::systemBus is nullptr? "
              << (crow::connections::systemBus == nullptr) << std::endl;
    ...
```

Here is the output of the gtest:

```
Running main() from /googletest-1.15.2/googletest/src/gtest_main.cc                                                                                                                                                                         [==========] Running 1 test from 1 test suite.                                                                                                                                                                                              [----------] Global test environment set-up.                                                                                                                                                                                                [----------] 1 test from PowerSubsystemCollectionTest                                                                                                                                                                                       [ RUN      ] PowerSubsystemCollectionTest.PowerSubsystemCollectionStaticAttributesAreExpected                                                                                                                                               crow::connections::systemBus is nullptr? 1 
```

And the test fails with a SIGSEGV, which is often a nullptr issue.

Let me know if you need me to do anything else.

 That is a global

gtmills: I am good here. Hopefully something we can look at for PFuture

PR #1268, opened 2025-04-28
Update to use power cap limits from CapLimits instead of Cap  (#1113)

cjcain:
Update to use power cap limits from CapLimits instead of Cap
 (#1113)

The power cap limits were previously hosted by Settings, but those are
system settings and the limits are not really settings (users can not
adjust these).

A new dbus interface will host these power cap limits:
  xyz.openbmc_project.Control.Power.CapLimits

Tested on Rainier:
'''
==> GET /redfish/v1/Chassis/chassis/PowerSubsystem (no powercap enabled)
...
    "Allocation": {
        "AllocatedWatts": 2777,
        "RequestedWatts": 2777
    },

== Enable 2121W power cap ==

==> GET /redfish/v1/Chassis/chassis/PowerSubsystem
...
    "Allocation": {
        "AllocatedWatts": 2121,
        "RequestedWatts": 2777
    },
'''
Ran validator and no failures for PowerSubsystem or EnvironmentMetrics

Signed-off-by: Chris Cain <cjcain@us.ibm.com>
Change-Id: Ib501b3511421fafd4cae28126705400704f07a72

### DIFF
--- a/redfish-core/lib/power_subsystem.hpp
+++ b/redfish-core/lib/power_subsystem.hpp
@@ -20,7 +20,6 @@
 #include <boost/beast/http/field.hpp>
 #include <boost/beast/http/verb.hpp>
 #include <boost/url/format.hpp>
-#include <sdbusplus/asio/property.hpp>
 #include <sdbusplus/unpack_properties.hpp>
 
 #include <cstdint>
@@ -36,12 +35,12 @@ namespace redfish
 inline void getPowerSubsystemAllocation(
     const std::shared_ptr<bmcweb::AsyncResp>& asyncResp)
 {
-    sdbusplus::asio::getAllProperties(
-        *crow::connections::systemBus, "xyz.openbmc_project.Settings",
-        "/xyz/openbmc_project/control/host0/power_cap",
-        "xyz.openbmc_project.Control.Power.Cap",
-        [asyncResp](const boost::system::error_code& ec,
-                    const dbus::utility::DBusPropertiesMap& propertiesList) {
+    // Get max from CapLimit
+    dbus::utility::getProperty<uint32_t>(
+        *crow::connections::systemBus, "org.open_power.OCC.Control",
+        "/xyz/openbmc_project/control/host0/power_cap_limits",
+        "xyz.openbmc_project.Control.Power.CapLimits", "MaxPowerCapValue",
+        [asyncResp](const boost::system::error_code& ec, uint32_t maxCap) {
             if (ec)
             {
                 if (ec.value() != EBADR)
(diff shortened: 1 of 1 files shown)


meson.build
```diff
@@ -479,7 +479,6 @@ srcfiles_unittest = files(
     'test/redfish-core/lib/log_services_dump_test.cpp',
     'test/redfish-core/lib/manager_diagnostic_data_test.cpp',
     'test/redfish-core/lib/metadata_test.cpp',
-    'test/redfish-core/lib/power_subsystem_test.cpp',
```
gtmills, line 482: Why are we removing this?

redfish-core/lib/power_subsystem.hpp
```diff
@@ -24,6 +33,58 @@
 namespace redfish
 {
 
+inline void getPowerSubsystemAllocation(
+    const std::shared_ptr<bmcweb::AsyncResp>& asyncResp)
+{
+    sdbusplus::asio::getAllProperties(
```
baemyung, line 59: Please use ` dbus::utility::getAllProperties("xyz.openbmc_project.Settings", ...)`

redfish-core/lib/power_subsystem.hpp
```diff
@@ -24,6 +33,58 @@
...
+                return;
+            }
+
+            uint32_t powerCap{0};
+            bool powerCapEnable{false};
+            uint32_t maxPowerCapValue{0};
+            const bool success = sdbusplus::unpackPropertiesNoThrow(
+                dbus_utils::UnpackErrorPrinter(), propertiesList, "PowerCap",
+                powerCap, "PowerCapEnable", powerCapEnable, "MaxPowerCapValue",
+                maxPowerCapValue);
```
baemyung: It would be better to use a separate line for each field (with sorted field names).
```
    const bool success = sdbusplus::unpackPropertiesNoThrow(
        dbus_utils::UnpackErrorPrinter(), propertiesList, //
        "MaxPowerCapValue", maxPowerCapValue,             //
        "PowerCap", powerCap,                             //
        "PowerCapEnable", powerCapEnable);
```

meson.build
cjcain: Original change was done by Shawn who said I think it was removed because it was failing CI downstream.  Not 100% sure, but that is my recollection.

redfish-core/lib/power_subsystem.hpp
```diff
@@ -24,6 +33,58 @@
...
+
+            if (!success)
+            {
+                messages::internalError(asyncResp->res);
+                return;
+            }
+
+            // If MaxPowerCapValue valid, store Allocation properties in JSON
+            if ((maxPowerCapValue > 0) && (maxPowerCapValue < UINT32_MAX))
```
baemyung: This seems not proper.
maxPowerCapValue is defined as uint32_t and thus its value can only be `{0, ..., UINT32_MAX}`.

I think dbus internal property is uint64_t and the logic is intended to check the range of uint32_t.
So, the code may need to be like

```

    uint64_t powerCap = 0;
    bool powerCapEnable = false;
    uint64_t maxPowerCapValue = 0;

    const bool success = sdbusplus::unpackPropertiesNoThrow(
        dbus_utils::UnpackErrorPrinter(), propertiesList, //
        "MaxPowerCapValue", maxPowerCapValue,             //
        "PowerCap", powerCap,                             //
        "PowerCapEnable", powerCapEnable);
```

And then
```
    // If MaxPowerCapValue valid, store Allocation properties in JSON
    if ((maxPowerCapValue < std::numeric_limits<uint32_t>::lowest()) &&
        (maxPowerCapValue < std::numeric_limits<uint32_t>::max()))
    {
        ...
    }
```

meson.build
smccarney: When I did this back in 1050, I had originally commented out the test because I believe it was failing CI.  Gunnar suggested removing the file instead.

https://github.com/ibm-openbmc/bmcweb/pull/654

If I remember correctly, CI wasn't failing because my change was incorrect.  It was more because the test case didn't set up a real environment, and then things failed when it tried to do real things.

smccarney (approved):

redfish-core/lib/power_subsystem.hpp
cjcain: done

redfish-core/lib/power_subsystem.hpp
```diff
@@ -52,35 +72,25 @@ inline void getPowerSubsystemAllocation(
...
-            uint32_t powerCap{0};
-            bool powerCapEnable{false};
-            uint32_t maxPowerCapValue{0};
+            const uint32_t* powerCap = nullptr;
+            const bool* powerCapEnable = nullptr;
             const bool success = sdbusplus::unpackPropertiesNoThrow(
                 dbus_utils::UnpackErrorPrinter(), propertiesList, "PowerCap",
-                powerCap, "PowerCapEnable", powerCapEnable, "MaxPowerCapValue",
-                maxPowerCapValue);
+                powerCap, "PowerCapEnable", powerCapEnable);
```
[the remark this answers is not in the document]
cjcain: I agree, but the lint/prettier tools do not.

redfish-core/lib/power_subsystem.hpp
```diff
@@ -36,6 +36,26 @@ namespace redfish
 inline void getPowerSubsystemAllocation(
     const std::shared_ptr<bmcweb::AsyncResp>& asyncResp)
 {
+    // Get max from CapLimit
+    sdbusplus::asio::getProperty<uint32_t>(
```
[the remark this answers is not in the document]
cjcain: done

baemyung (approved):

redfish-core/lib/power_subsystem.hpp
```diff
@@ -4,17 +4,26 @@
...
 #include "registries/privilege_registry.hpp"
 #include "utils/chassis_utils.hpp"
+#include "utils/dbus_utils.hpp"
+
+#include <asm-generic/errno.h>
 
 #include <boost/beast/http/field.hpp>
 #include <boost/beast/http/verb.hpp>
 #include <boost/url/format.hpp>
+#include <sdbusplus/asio/property.hpp>
```
baemyung, line 23: This will need to be removed to satisfy CI.
```
- #include <sdbusplus/asio/property.hpp>
```

meson.build
gtmills: I don't see what in the test would fail. Can we try with this test in ?

gtmills (commented):
I am good with this, just want to see if we can get away without removing the test

cjcain:
Holding off on this PR until dependent PR is merged: https://github.com/ibm-openbmc/bmcweb/pull/1265

smccarney:
I answered the question about why the test fails here: https://github.com/ibm-openbmc/bmcweb/pull/1265#discussion_r2069510859

Basically, there is a global variable `extern sdbusplus::asio::connection* systemBus;` that tons of bmcweb code uses including power_subsystem.hpp.  That variable is nullptr in the testcase (and I would guess other testcases?).

PR #1278, opened 2025-05-06
1120: Additional updates for chassis HealthRollup 

abiasojo:
The health status of the current CPUs and powersupplies will not rollup to chassis. This submission implements rollup of the health status of CPUs and powersupplies to chassis.

Tested:
1. add the xyz.openbmc_project.Inventory.Item.Global interface to the chassis.
I create a critical association of powersupply1 using the same method as phosphor-logging[1].

[1] https://github.com/openbmc/phosphor-logging/blob/76198a2ea2e3deb14078e0d3dd3932b657ddd9b8/extensions/openpower-pels/service_indicators.cpp#L83

```
curl -k -H "X-Auth-Token: $token" -X GET https://${bmc}/redfish/v1/Chassis/chassis
"Status": {
    "Health": "OK",
    "HealthRollup": "Critical",
    "State": "Enabled"
  },
```

Change-Id: Iafacbfacff081d09c38405a83b7fabb7074f61a1

### DIFF
--- a/redfish-core/lib/chassis.hpp
+++ b/redfish-core/lib/chassis.hpp
@@ -484,6 +484,37 @@ inline void handleChassisGetSubTree(
                         return; // no sensors = no failures
                     }
                     health->inventory = resp;
+
+                    constexpr std::array<std::string_view, 13> interfaces = {
+                        "xyz.openbmc_project.Inventory.Item.Dimm",
+                        "xyz.openbmc_project.Inventory.Item.Cpu",
+                        "xyz.openbmc_project.Inventory.Item.PowerSupply",
+                        "xyz.openbmc_project.Inventory.Item.Fan",
+                        "xyz.openbmc_project.Inventory.Item.PCIeSlot",
+                        "xyz.openbmc_project.Inventory.Item.Vrm",
+                        "xyz.openbmc_project.Inventory.Item.Tpm",
+                        "xyz.openbmc_project.Inventory.Item.Panel",
+                        "xyz.openbmc_project.Inventory.Item.Battery",
+                        "xyz.openbmc_project.Inventory.Item.DiskBackplane",
+                        "xyz.openbmc_project.Inventory.Item.Board",
+                        "xyz.openbmc_project.Inventory.Item.Board.Motherboard",
+                        "xyz.openbmc_project.Inventory.Item.Connector"};
+                    dbus::utility::getSubTreePaths(
+                        "/", 0, interfaces,
+                        [health](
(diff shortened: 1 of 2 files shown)


jeaaustx (commented):
This is a rebase correct?

If so, the commit should keep any signoffs from the earlier commit.

Code changes look fine although it is quite ugly the way it was originally done. I'm a bit surprised the work isn't inside of health->populate(). (I'm not asking for that to change as part of the rebase.)

abiasojo:
> This is a rebase correct?
> 
> If so, the commit should keep any signoffs from the earlier commit.
> 
> Code changes look fine although it is quite ugly the way it was originally done. I'm a bit surprised the work isn't inside of health->populate(). (I'm not asking for that to change as part of the rebase.)

Yes it is rebase. I had worked on another previous commit message with Gunnar and he mentioned that I remove the previous signoff. I can add back the previous signoff for this.

jeaaustx:
> Yes it is rebase. I had worked on another previous commit message with Gunnar and he mentioned that I remove the previous signoff. I can add back the previous signoff for this.

Oh, well I would go with what Gunnar says instead. I guess I don't understand when we should keep them.

abiasojo:
> > Yes it is rebase. I had worked on another previous commit message with Gunnar and he mentioned that I remove the previous signoff. I can add back the previous signoff for this.
> 
> Oh, well I would go with what Gunnar says instead. I guess I don't understand when we should keep them.
I updated it already :)

baemyung (commented):
Since this is addressing the missing pieces from the previous commit 97da18e6c6643c8f5f1ffe249fb4352be7117e5f ( PR - https://github.com/ibm-openbmc/bmcweb/pull/1271),  it is better to show the prev commit info in this commit msg.

Also maybe better to use the different subject of commit message and PR from the previous one.
Perhaps something like `Fix the missing chassis HealthRollup`?

abiasojo:
> Since this is addressing the missing pieces from the previous commit [97da18e](https://github.com/ibm-openbmc/bmcweb/commit/97da18e6c6643c8f5f1ffe249fb4352be7117e5f) ( PR - #1271), it is better to show the prev commit info in this commit msg.
> 
> Also maybe better to use the different subject of commit message and PR from the previous one. Perhaps something like `Fix the missing chassis HealthRollup`?

I have updated.

baemyung (approved):
It looks good to me

abiasojo:
> It looks good to me

Thank you.

gtmills (approved):

PR #1281, opened 2025-05-08
1120: Check PowerSupply availability prop for State/Hlth

baemyung:
This commit makes use of the Available property on the power supply D-Bus object when determining the power supply state and health in the PowerSupply response.  The power supply monitor code from phosphor-power will set the property to false if it can determine that the power supply isn't able to provide power due to certain faults.

If the PS isn't available, then the Health property will be set to Critical, and the State will be set to UnavailableOffline (assuming it shouldn't be Absent instead).

This is necessary because on IBM systems the Functional property determines the fault LED state as well as the health value, and in certain cases the fault LED needs to stay off even though the PS health is not OK.  A specific example of this is when the PS cord is unplugged: no fault LED is desired but it is desired for the Redfish output to show that there is an issue with the PS.

If the Available property isn't present on D-Bus, it acts as if it had a value of true and behaves the same as it does today.

Tested:
Available = false:
    "Health": "Critical",
    "State": "UnavailableOffline"

    "Health": "Critical",
    "State": "Enabled"

PS missing:
    "Health": "Critical",
    "State": "Absent"

Everything OK:
    "Health": "OK",
    "State": "Enabled"

    "Health": "OK",
    "State": "Enabled"

### DIFF
--- a/redfish-core/lib/power_supply.hpp
+++ b/redfish-core/lib/power_supply.hpp
@@ -208,11 +208,12 @@ inline void getValidPowerSupplyPath(
 
 inline void getPowerSupplyState(
     const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
-    const std::string& service, const std::string& path)
+    const std::string& service, const std::string& path, bool available)
 {
     dbus::utility::getProperty<bool>(
         service, path, "xyz.openbmc_project.Inventory.Item", "Present",
-        [asyncResp](const boost::system::error_code& ec, const bool value) {
+        [asyncResp,
+         available](const boost::system::error_code& ec, const bool value) {
             if (ec)
             {
                 if (ec.value() != EBADR)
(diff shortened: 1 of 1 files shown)


gtmills (approved):

jeaaustx (approved):

baemyung:
upstream: https://gerrit.openbmc.org/c/openbmc/bmcweb/+/51336
Jira: https://jsw.ibm.com/browse/PFEBMC-4200

gtmills:
jenkins run tests please

baemyung:
Currently all BMCWEB PRs are failing because of IBM systems routing issue - which blocks to access 
```
wget https://redfish.dmtf.org/
```

The related threads are 
- https://ibm-systems-power.slack.com/archives/C9X7P1PGX/p1747054262417169
- https://ibm-systems-power.slack.com/archives/C26MX8XFW/p1747059638902019
-

baemyung:
Once  https://github.com/ibm-openbmc/bmcweb/pull/1285 is merged, this will be rebased

PR #1293, opened 2025-05-19
Audit Log: Skip detail for UpdateService URIs

jeaaustx:
The UpdateService Redfish URIs may contain huge binary data in their body. Add the URI to the list of URIs to skip adding detail when auditing.

Tested:
 - Issued update with DEBUG log messages turned on. Confirmed no detail was attempted to be added to audit record. Confirmed audit record was recorded as expected.
 - Confirmed other audit records with details were still recording detail as expected.


gtmills (approved):

baemyung (approved):
It looks good to me

PR #1296, opened 2025-05-21
Remove VLAN support from Redfish interface

ishasahni2000:

Tested by:

1. Verified VLAN resource is not listed while GET on the eth interfaces

### DIFF
--- a/redfish-core/lib/ethernet.hpp
+++ b/redfish-core/lib/ethernet.hpp
@@ -134,7 +134,6 @@ struct EthernetInterfaceData
     std::string ipv6DefaultGateway;
     std::string ipv6StaticDefaultGateway;
     std::optional<std::string> macAddress;
-    std::optional<uint32_t> vlanId;
     std::vector<std::string> nameServers;
     std::vector<std::string> staticNameServers;
     std::vector<std::string> domainnames;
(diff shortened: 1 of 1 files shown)


isha309 (approved):
adding comment so ci could start

redfish-core/lib/ethernet.hpp
```diff
@@ -2628,33 +2589,6 @@ inline void requestEthernetInterfacesRoutes(App& app)
                     });
             });
 
-    BMCWEB_ROUTE(app, "/redfish/v1/Managers/<str>/EthernetInterfaces/<str>/")
```
abhilashraju, line 2631: why you want to remove this?

ishasahni2000: physical interfaces like eth0/eth1 we don't support delete operation on interfaces but this delete operation route was added for only VLAN resources.
 you can see similar discussion which happend for 1110 PR https://github.com/ibm-openbmc/bmcweb/pull/927

abhilashraju (approved):

redfish-core/lib/ethernet.hpp
gtmills: From #927 
```
@gtmills I think physical interfaces like eth0/eth1 we don't support delete operation on interfaces
but this delete operation route added for only VLAN resources.

on 1060 as well on patch and get on eth0/eth1 is supported
https://github.com/ibm-openbmc/bmcweb/blob/1060/redfish-core/lib/ethernet.hpp

delete not supported
```

gtmills:
add to approvelist

gtmills:
This would be a good one to go upstream.. Like an option

gtmills (approved):

baemyung (approved):

PR #907, opened 2024-04-17
1110 ghe : Add BIOS support

cmjishnu:
This commit adds:
1. redfish interface "/Registries/BiosAttributeRegistry" for BIOS Attribute Registry
2. redfish interface "/Systems/system/Bios/Settings" for BIOS settings
3. support for PATCH operation of the bios variables that updates
   the BaseBiosTable.
   
   
 Previous downstream PR : https://github.com/ibm-openbmc/bmcweb/commit/7863e92e690ea4a4752f0bbd22d0cc75f9fbe704

PR #1298, opened 2025-05-27
 Add BIOS support (#907)

asmithakarun:
This commit adds:

redfish interface "/Registries/Bios" for BIOS Attribute Registry
redfish interface "/Systems/system/Bios/Settings" for BIOS settings
support for PATCH operation of the bios variables that updates the BaseBiosTable.

Part of this change has an upstream commit:
https://gerrit.openbmc.org/c/openbmc/bmcweb/+/79698

Tested By:
[1] GET https://<bmc_ip>/redfish/v1/Systems/system/Bios
[2] GET https://<bmc_ip>/redfish/v1/Registries/BiosAttributeRegistry
[3] POST https://ip_addr/redfish/v1/Systems/system/Bios/Actions/Bios.ResetBios
[4] Single/Multiple Attributes:
PATCH '{"Attributes":{"<attr name>":"<attr value>"}}' https://${bmc}/redfish/v1/Systems/system/Bios/Settings

This makes use of the "Set" of "PendingAttributes" in the backend and that updates the BaseBiosTable.

Redfish validator passed.

### DIFF
--- a/redfish-core/include/registries_selector.hpp
+++ b/redfish-core/include/registries_selector.hpp
@@ -3,6 +3,7 @@
 #pragma once
 #include "registries.hpp"
 #include "registries/base_message_registry.hpp"
+#include "registries/bios_message_registry.hpp"
 #include "registries/heartbeat_event_message_registry.hpp"
 #include "registries/openbmc_message_registry.hpp"
 #include "registries/resource_event_message_registry.hpp"
@@ -27,6 +28,10 @@ inline std::optional<registries::HeaderAndUrl>
     if (base::header.registryPrefix == registryName)
     {
         return HeaderAndUrl{base::header, base::url};
+    }
+    if (bios::header.registryPrefix == registryName)
+    {
+        return HeaderAndUrl{bios::header, bios::url};
     }
     if (heartbeat_event::header.registryPrefix == registryName)
     {
--- a/redfish-core/include/utils/bios_utils.hpp
+++ b/redfish-core/include/utils/bios_utils.hpp
@@ -1,0 +1,137 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright OpenBMC Authors
+#pragma once
+
+#include "async_resp.hpp"
+#include "dbus_singleton.hpp"
+#include "dbus_utility.hpp"
+#include "error_messages.hpp"
+#include "logging.hpp"
+
+#include <array>
+#include <cstdint>
+#include <functional>
+#include <memory>
+#include <string>
+#include <string_view>
+#include <unordered_map>
+#include <variant>
+
+namespace redfish
+{
+namespace bios_utils
+{
+static constexpr std::string_view biosConfigManagerPath =
+    "/xyz/openbmc_project/bios_config/manager";
+static constexpr std::string_view biosConfigManagerInterface =
+    "xyz.openbmc_project.BIOSConfig.Manager";
+
+template <typename Type>
+inline void extractValue(nlohmann::json& attributes, const std::string& name,
+                         const dbus::utility::DbusVariantType& value)
+{
+    const Type* tValue = std::get_if<Type>(&value);
+    if (tValue != nullptr)
+    {
+        attributes[name] = *tValue;
+        return;
+    }
+    attributes[name] = Type{};
+}
+
+template <>
+inline void extractValue<bool>(nlohmann::json& attributes,
+                               const std::string& name,
+                               const dbus::utility::DbusVariantType& value)
+{
+    const int64_t* tValue = std::get_if<int64_t>(&value);
+    if (tValue != nullptr)
+    {
+        attributes[name] = (*tValue != 0);
+        return;
+    }
+    attributes[name] = false;
+}
+
+using HandlerType = std::function<void(nlohmann::json&, const std::string&,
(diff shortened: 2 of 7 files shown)


redfish-core/include/registries/bios_registry.hpp
```diff
@@ -0,0 +1,20 @@
...
+    1,
+    0,
+    0,
+    "Bios Attribute Registry",
+    "en",
+    "This registry defines the messages for bios attribute registry.",
+    "BiosAttributeRegistry",
+    "DMTF",
+};
+constexpr const char* url = "";
```
abhilashraju: if you initialise it will nullptr ,can we avoid extrac check at line 125 in message registries.hpp?

raviteja-b (approved):

redfish-core/include/registries/bios_registry.hpp
```diff
@@ -0,0 +1,20 @@
+#pragma once
```
baemyung: For the consistency with the other file names, would it be better to name this file as `bios_message_registry.hpp`?

redfish-core/include/registries_selector.hpp
```diff
@@ -28,6 +29,10 @@ inline std::optional<registries::HeaderAndUrl>
     {
         return HeaderAndUrl{base::header, base::url};
     }
+    if (bios::header.registryPrefix == registryName)
+    {
+        return HeaderAndUrl{bios::header, bios::url};
+    }
```
baemyung, line 35: I think we should also have the selection for bios for `getRegistryFromPrefix()`?

redfish-core/include/utils/bios_utils.hpp
```diff
@@ -0,0 +1,137 @@
...
+                return;
+            }
+            handler(value);
+        });
+}
+
+template <typename CallbackFunc>
+inline void getBIOSManagerObject(
+    const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
+    CallbackFunc&& callback)
```
baemyung, line 112: It would be better to avoid template callback if possible.
Here, it can be 
```
inline void getBIOSManagerObject(
    std::function<void(const std::string& objectPath)>&& callback)
```

redfish-core/include/utils/bios_utils.hpp
```diff
@@ -0,0 +1,137 @@
...
+        auto it = typeMap.find(*typeStr);
+        if (it != typeMap.end())
+        {
+            it->second(attributes, name, value);
+        }
+    }
+}
+
+template <typename T>
+inline void getBIOSManagerProperty(
```
baemyung, line 87: Wouldn't this be for `BaseTable` only?
Or do we expect the other types too for this function?
If not, can we make this explicitly for BaseTable without using template type?

```
inline void getBIOSManagerProperty(
    const std::string& property, const std::string& objectPath,
    std::function<void(const BaseTable&)> handler)
```

redfish-core/lib/message_registries.hpp
```diff
@@ -145,6 +146,11 @@ inline void handleMessageRegistryGet(
         return;
     }
 
+    if (registry == "BiosAttributeRegistry")
```
baemyung: I think registryMatch also needs to be checked.
```
if (registry == "BiosAttributeRegistry" && registryMatch == "BiosAttributeRegistry")
```

For example, GET `/redfish/v1/Registries/BiosAttributeRegistry/BAD/ `would need to return `NotFound`.

redfish-core/include/registries/bios_registry.hpp
asmithakarun: But, this is not a message registry, but an attribute registry. Hence this file name "bios_registry"

redfish-core/include/utils/bios_utils.hpp
asmithakarun: I have just cherry-picked the upstream commit here. I can try making this change and push it to gerrit in parallel and pull it downstream in a new commit.

redfish-core/include/utils/bios_utils.hpp
asmithakarun: Yes, it can be removed. Again, i will try to modify the upstream commit and pull it downstream in a new commit.

redfish-core/include/registries_selector.hpp
asmithakarun: We will never hit "getRegistryFromPrefix" path for bios, as we would check and call the bios handler method way before we call this method.

redfish-core/include/registries/bios_registry.hpp
asmithakarun: done

redfish-core/lib/message_registries.hpp
asmithakarun: done

redfish-core/lib/bios.hpp
```diff
@@ -57,12 +238,415 @@ inline void handleBiosServiceGet(
     asyncResp->res.jsonValue["Actions"]["#Bios.ResetBios"]["target"] =
         std::format("/redfish/v1/Systems/{}/Bios/Actions/Bios.ResetBios",
```
baemyung, line 239: boost::urls::format()

baemyung (approved):
I think this can be merged into 1120.
More enhancements can be done later as needed

PR #1303, opened 2025-05-30
WIP: Bmcshell acf-task tracker apis.

abhilashraju:
For tracking the active bmc-shell acf tasks.
/redfish/v1/AccountService/acf/scripts/ lists all active bmc-shell acf task
/redfish/v1/AccountService/acf/<str>/cancel/ cancel the bmc-shell task corresponding to the parameter

### DIFF
--- a/redfish-core/lib/acf_service.hpp
+++ b/redfish-core/lib/acf_service.hpp
@@ -1,0 +1,89 @@
+#pragma once
+
+#include "privileges.hpp"
+#include "registries/privilege_registry.hpp"
+#include "utils/dbus_utils.hpp"
+#include "utils/json_utils.hpp"
+
+#include <app.hpp>
+#include <boost/beast/http/status.hpp>
+#include <nlohmann/json.hpp>
+
+#include <string>
+
+namespace redfish
+{
+static void handleAcfScriptsGet(
+    App& /*unused*/, const crow::Request& /*unused*/,
+    const std::shared_ptr<bmcweb::AsyncResp>& asyncResp)
+{
+    BMCWEB_LOG_DEBUG("Handling GET request for ACF scripts");
+    crow::connections::systemBus->async_method_call(
+        [asyncResp](const boost::system::error_code& ec,
+                    const std::vector<std::string>& msg) {
+            if (ec)
+            {
+                messages::internalError(asyncResp->res);
+                return;
+            }
+            nlohmann::json& response = asyncResp->res.jsonValue;
+            response["@odata.context"] =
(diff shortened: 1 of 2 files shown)


gtmills (commented):
Why can't we use standard Redfish tasks here?

abhilashraju:
> Why can't we use standard Redfish tasks here?

Are you referring to a Redfish task? In this context, the "task" refers to a shell script job executed by acfshell. There can be multiple jobs running simultaneously on the system. Our goal is to list all currently running jobs upon user request and provide a way to cancel them. I don't believe Redfish tasks offer this functionality.

redfish-core/lib/acf_service.hpp
```diff
@@ -0,0 +1,89 @@
...
+                msg.begin(), msg.end(), std::back_inserter(scripts),
+                [](const std::string& script) {
+                    nlohmann::json scriptJson;
+                    scriptJson["Id"] = script; // Use the script ID
+                    scriptJson["Actions"] = {
+                        {"#AccountService.AcfScripts.Cancel",
+                         {{"target",
+                           std::format(
+                               "/redfish/v1/AccountService/acf/{}/cancel",
+                               script)}}}};
```
baemyung, line 48: I think this can be explicitly specified (refer to https://github.com/openbmc/bmcweb/blob/master/COMMON_ERRORS.md#13-complete-replacement-of-the-response-object), and also use `boost:urls::format()`

```
              scriptJson["Actions"]["#AccountService.AcfScripts.Cancel"]["target"] =
                           boost::urls::format(
                               "/redfish/v1/AccountService/acf/{}/cancel",  script);
```

redfish-core/lib/acf_service.hpp
```diff
@@ -0,0 +1,89 @@
+#pragma once
```
baemyung, line 1: License statement?
```
// SPDX-License-Identifier: Apache-2.0
// SPDX-FileCopyrightText: Copyright OpenBMC Authors
```

gtmills:
> Our goal is to list all currently running jobs upon

All current running ACF jobs right? Why can't each of those be a task ?

> and provide a way to cancel them

Redfish tasks can be canceled

gtmills:
@abhilashraju Are we still planning on doing this for 1120 ?

abhilashraju:
> @abhilashraju Are we still planning on doing this for 1120 ?

The requirement is not accepted. So we are not planning this for 1120. If there is new requirement to support multiple ACF , then we may need this in future

PR #1304, opened 2025-05-30
1120: Consume availability interface for cores (#991)

baemyung:
In the current state, the logic is bmcweb for displaying cores is: It checks the Present property on core dbus object to update the [Status][State] as Enabled/Absent.

In addition to Present property, we should also use the Available property , if the Available is false, then set the [Status][State] should be set as "UnavailableOffline".

Refer: ibm-openbmc/dev#3481

### DIFF
--- a/redfish-core/lib/processor.hpp
+++ b/redfish-core/lib/processor.hpp
@@ -1000,14 +1000,15 @@ inline void getProcessorPaths(
         });
 }
 
-inline void getOperationalStatusData(
+inline void getSubProcessorsCoreHealth(
     const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
-    const std::string& service, const std::string& objPath,
-    const std::string& interface)
+    const std::string& service, const std::string& objPath, bool available)
 {
     dbus::utility::getProperty<bool>(
-        service, objPath, interface, "Functional",
-        [asyncResp](const boost::system::error_code& ec, bool value) {
+        service, objPath,
+        "xyz.openbmc_project.State.Decorator.OperationalStatus", "Functional",
+        [asyncResp,
+         available](const boost::system::error_code& ec, bool functional) {
             if (ec)
             {
                 if (ec.value() != EBADR)
@@ -1018,7 +1019,7 @@ inline void getOperationalStatusData(
                 return;
             }
 
-            if (!value)
+            if (!functional || !available)
             {
                 asyncResp->res.jsonValue["Status"]["Health"] =
                     resource::Health::Critical;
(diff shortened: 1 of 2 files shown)


gtmills (approved):
https://github.com/ibm-openbmc/bmcweb/pull/254 was 1020 
https://github.com/ibm-openbmc/bmcweb/pull/991 was this for 1110

baemyung:
Upstream is not started.  The dependent upstream is still ongoing - https://gerrit.openbmc.org/c/openbmc/bmcweb/+/64982

redfish-core/lib/processor.hpp
```diff
@@ -1018,21 +1019,22 @@ inline void getOperationalStatusData(
                 return;
             }
 
-            if (!value)
+            if (!functional || !available)
```
jeaaustx, line 1022: In 1110 we were only setting Critical for !functional. Here we are also setting if !available; this looks like a semantic change from 1110. Is this desired? Maybe call it out in the commit message.

https://github.com/ibm-openbmc/bmcweb/blob/45efd00c470ad056baafab0e5032ecad0c000ad8/redfish-core/lib/processor.hpp#L1183

redfish-core/lib/processor.hpp
```diff
@@ -1043,11 +1045,40 @@ inline void getItemData(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
                 return;
             }
 
-            if (!value)
+            if (!present)
```
jeaaustx, line 1048: In 1110 we were checking for !available before !present. So we would favor UnavailableOffline over Absent. This looks like a semantic change from 1110. Is that desired? If so maybe call it out in the commit message.

https://github.com/ibm-openbmc/bmcweb/blob/45efd00c470ad056baafab0e5032ecad0c000ad8/redfish-core/lib/processor.hpp#L1174

redfish-core/lib/processor.hpp
baemyung: I think 1110's code is partly correct when dbus "Availability"  is set to false and it produced like
```
  "Status": {
    "Health": "OK",
    "State": "UnavailableOffline"
  }
```

I think the proper Health state for the case would be similar to how power_suppy is handled - https://github.com/ibm-openbmc/bmcweb/blame/1120/redfish-core/lib/power_supply.hpp#L261
```
  "Status": {
    "Health": "Critical",
    "State": "UnavailableOffline"
  }
```

I'll update the commit message.

redfish-core/lib/processor.hpp
baemyung: I think - the way of power_supply's handling is appropriate - https://github.com/ibm-openbmc/bmcweb/blob/45efd00c470ad056baafab0e5032ecad0c000ad8/redfish-core/lib/power_supply.hpp#L229C4-L229C5

The case where  both Present and Availability are false,  I think it is better to set as Absent (rather than UnavailableOffline) as it is not present.

```
Case Present=Availability=false:
"State": "Absent"
```

PowerSupply state/Hlth PR - https://github.com/ibm-openbmc/bmcweb/pull/1281 and its upstream is https://gerrit.openbmc.org/c/openbmc/bmcweb/+/51336 (although it is abandoned).

The comment (Matt's)  on upstream - https://gerrit.openbmc.org/c/openbmc/bmcweb/+/51336/comment/ea8257ab_9718e604/ says:

```
What I am trying to do is support the following:
1) Health can be Critical even though Functional is true
and
2) State can be UnavailableOffline separately from Absent

If the PSU daemon defines the Available property to mean 'present but can't supply good power', then the code I have here accomplishes that.
```

I think this makes sense even for processor/core. If so, "Present=false" would take precedence over "Availability".

I'll modify the commit message to add those (similar to https://github.com/ibm-openbmc/bmcweb/pull/1281)

PR #1310, opened 2025-06-05
1120: Help recover in DOS attack

gtmills:
Upstream is https://gerrit.openbmc.org/c/openbmc/bmcweb/+/80819 and https://gerrit.openbmc.org/c/openbmc/bmcweb/+/80839 and has merged. The DOS attack we are helping clear up is coming from the HMC.  2 commits: 

1) Fix DOS attack scenario

Problem : When 201 connections made in parallel to BMC and closed them immediately without properly sending a close_notify alert it was observed that Bmcweb server was taking several minutes to close the sockets. All the 200 TCP sockets were in CLOSE_WAIT state.

Journal log shows below line
[CRITICAL http_connection.hpp:213] 0x29d1ef0Max connection count exceeded.

Not able to make new connection

```
$ curl -k -H "X-Auth-Token:$bmc_token"  -X GET https://${BMC_IP}/redfish
/v1/AccountService/Accounts
curl: (35) OpenSSL SSL_connect: SSL_ERROR_SYSCALL in connection to
127.0.0.1:2443
```

Fix : The bmcweb server failed to identify the end of stream at TCP (SSL
/TLS) layer, therefore added check to identify the end of stream which
closes the connection and socket

Change-Id: I1c277db0b774d33c656b4a2b1bd14f3575535bec

2) Do hard close if client disobeys protocol 

Do hard close if client disobeys protocol

There are cases bmcweb might close the connection due to a violation of
the protocol.  Currently these are done gracefully, under the assumption
that a client might attempt to recover.  But this opens us up to
potentially leaving sockets open for far longer than we intend if the
client is completely gone, due to a disconnect or explicitly closing the
socket hard.

In cases where we get a protocol error, shutdown the socket hard, rather
than attempt to do things "correctly".

Tested:

I tested this MR using a script that simulated 5,000 parallel
connections simultaneously to BMC and closed them immediately without
properly sending a close_notify alert

Observations:

The BMC became unresponsive for 30-40 seconds before recovering.

After recovery, it took approximately 90 seconds to close all
connections in QEMU.
On real hardware, connection closure times may be slightly higher
(though still within expected parameters).

Conclusion:
This behavior aligns with expectations.

After 90 seconds observed that

1) No sockets in CLOSE_WAIT state

2) Able to make new connection.

```

curl -k -H "X-Auth-Token:$bmc_token" https://${ip}/redfish/v1/AccountService/Accounts
{
  "@odata.id": "/redfish/v1/AccountService/Accounts",
  "@odata.type": "#ManagerAccountCollection.ManagerAccountCollection",
  "Description": "BMC User Accounts",
  "Members": [
    {
      "@odata.id": "/redfish/v1/AccountService/Accounts/root"
    }
  ],
  "Members@odata.count": 1,
  "Name": "Accounts Collection"
}

```
Change-Id: I6ab4347efd8fda9ae86bfbb8575666ad3eabe88c
Signed-off-by: Ed Tanous <etanous@nvidia.com>
Signed-off-by: Chandramohan Harkude <chandramohan.harkude@gmail.com>

### DIFF
--- a/http/http_connection.hpp
+++ b/http/http_connection.hpp
@@ -637,19 +637,9 @@ class Connection :
                 doWrite();
                 return;
             }
-            if (ec == boost::beast::http::error::end_of_stream ||
-                ec == boost::asio::ssl::error::stream_truncated)
-            {
-                BMCWEB_LOG_WARNING("{} End of stream, closing {}", logPtr(this),
-                                   ec);
-                hardClose();
-                return;
-            }
-
-            BMCWEB_LOG_DEBUG("{} Closing socket due to read error {}",
-                             logPtr(this), ec.message());
-            gracefulClose();
-
+            BMCWEB_LOG_WARNING("{} End of stream, closing {}", logPtr(this),
+                               ec);
+            hardClose();
             return;
         }
 
@@ -741,17 +731,10 @@ class Connection :
                 }
                 return;
             }
+            BMCWEB_LOG_WARNING("{} End of stream, closing {}", logPtr(this),
+                               ec);
+            hardClose();
 
-            if (ec == boost::beast::http::error::end_of_stream ||
-                ec == boost::asio::ssl::error::stream_truncated)
-            {
-                BMCWEB_LOG_WARNING("{} End of stream, closing {}", logPtr(this),
-                                   ec);
-                hardClose();
-                return;
-            }
-
-            gracefulClose();
             return;
         }
 
--- a/http/http_connection.hpp
+++ b/http/http_connection.hpp
@@ -22,6 +22,7 @@
 
 #include <boost/asio/error.hpp>
 #include <boost/asio/ip/tcp.hpp>
+#include <boost/asio/ssl/error.hpp>
 #include <boost/asio/ssl/stream.hpp>
 #include <boost/asio/ssl/stream_base.hpp>
 #include <boost/asio/ssl/verify_context.hpp>
@@ -636,7 +637,8 @@ class Connection :
                 doWrite();
                 return;
             }
-            if (ec == boost::beast::http::error::end_of_stream)
+            if (ec == boost::beast::http::error::end_of_stream ||
+                ec == boost::asio::ssl::error::stream_truncated)
             {
                 BMCWEB_LOG_WARNING("{} End of stream, closing {}", logPtr(this),
                                    ec);
@@ -740,6 +742,15 @@ class Connection :
                 return;
             }
 
+            if (ec == boost::beast::http::error::end_of_stream ||
+                ec == boost::asio::ssl::error::stream_truncated)
+            {
+                BMCWEB_LOG_WARNING("{} End of stream, closing {}", logPtr(this),
+                                   ec);
+                hardClose();
+                return;
+            }
+
             gracefulClose();
             return;
         }
@@ -807,6 +818,16 @@ class Connection :
             doWrite();
             return;
         }
+
+        if (ec == boost::beast::http::error::end_of_stream ||
+            ec == boost::asio::ssl::error::stream_truncated)
+        {
+            BMCWEB_LOG_WARNING("{} End of stream, closing {}", logPtr(this),
+                               ec);
+            hardClose();
+            return;
+        }
+
         if (ec)
         {
             BMCWEB_LOG_DEBUG("{} from write(2)", logPtr(this));


baemyung (approved):
It looks good to me

http/http_connection.hpp
```diff
@@ -807,6 +801,16 @@ class Connection :
             doWrite();
             return;
         }
+
+        if (ec == boost::beast::http::error::end_of_stream ||
```
jeaaustx, line 805: Should this be comparing to ec.value()?

gtmills: It isn't needed/I think this is more natural. Under the covers, it is an int compare already (I think?).  Getting this from upstream. 

ec.value() would return an int, end of stream looks to be 1. 
`ec.value() == 1` or `ec.value == boost::beast::http::error::end_of_stream` I think would work the same here but I think is extra code. 

https://live.boost.org/doc/libs/1_67_0/boost/beast/http/error.hpp
https://www.boost.org/doc/libs/1_74_0/boost/system/error_code.hpp 

Some silly code I wrote and ran at https://www.programiz.com/cpp-programming/online-compiler/

```
enum Hello
{
    Hello1,
    Hello2,
    Hello3
};

int main() {
    // Write C++ code here
    Hello hello = Hello2;
    if (hello == Hello1)
    {
        std::cout << "here in if checking for Hello1\n";
    }
    if (hello == Hello2)
    {
        std::cout << "here in if checking for Hello2\n";
    }
    if (hello == 0)
    {
        std::cout << "here in if checking for 0\n";
    }
    
    if (hello == 1)
    {
        std::cout << "here in if checking for 1\n";
    }  
    return 0;
}
```

It outputs
```
here in if checking for Hello2
here in if checking for 1
```

jeaaustx: I remember seeing a commit where checking against ec.value() was required. I guess that was because it was a linux error instead?

https://gerrit.openbmc.org/c/openbmc/bmcweb/+/69947

jeaaustx (approved):

http/http_connection.hpp
baemyung: Similar to my comment on a different PR - https://github.com/ibm-openbmc/bmcweb/pull/1313#discussion_r2136318519, boost::error_code has 2 fields - category and value. 
For 2 ec's, their value() can be same if they are in different categories.

For example, the values of the followings are the same (i.e. 1).
- `boost::beast::websocket::closed`  - https://live.boost.org/doc/libs/1_67_0/boost/beast/websocket/error.hpp)
- `boost::beast::http::end_of_stream` - https://live.boost.org/doc/libs/1_67_0/boost/beast/http/error.hpp

baemyung: Regarding https://gerrit.openbmc.org/c/openbmc/bmcweb/+/69947,  I think the underneath library (`sdplusplus` or `beast`) seems returning ec with the different category (although the value may be same).
If we see `ec` (or both category and value), we may be able to know which full error name can be used??

```

 if ((ec.value() ==
             boost::system::linux_error::bad_request_descriptor) ||
            (ec.value() == boost::asio::error::basic_errors::host_unreachable))

```

- linux_error category is defined - https://www.boost.org/doc/libs/1_61_0/boost/system/linux_error.hpp
- asio::basic_errors is defined - https://www.boost.org/doc/libs/1_84_0/boost/asio/error.hpp

I think the underneath libraries perhaps encapsulates `EBADR` under  the standard system error (`boost::system::errc` https://www.boost.org/doc/libs/1_84_0/boost/system/detail/errc.hpp), although `EBADR` is an extension under `linux_error`??

That may be the reason why it does not properly compare the case of `EBADR` (if the full error name is used)??

gtmills: https://github.com/search?q=repo%3Aopenbmc%2Fbmcweb+boost%3A%3Abeast%3A%3Ahttp%3A%3Aerror&type=code

vs

https://github.com/search?q=repo%3Aopenbmc%2Fbmcweb%20bad_request_descriptor&type=code

I give. This is a good comments but unless someone objects, I vote we move forward with this / if we want any changes do them upstream

baemyung: I would agree to move forward with this, as it has been working okay.
(Later / when possible, we can revisit ec comparison)

gtmills:
https://github.com/ibm-openbmc/bmcweb/pull/1311 is 1110 but going to run with this for a while first

PR #1320, opened 2025-06-13
Fix IPv4 Static IP config with same DHCP gateway

raviteja-b:
This commit avoids same gateway check while setting gateway. While DHCP gateway already configured on the interface,then moving to static configuration with the same DHCP gateway fails to set.

This is commit needed to support switching from DHCP to static

Tested By:
Configure DHCP on interface
Configure Static IPv4 address with same dhcp gateway

### DIFF
--- a/redfish-core/lib/ethernet.hpp
+++ b/redfish-core/lib/ethernet.hpp
@@ -1726,12 +1726,7 @@ inline void handleIPv4StaticPatch(
         }
     }
 
-    // now update to the new gateway.
-    // Default gateway is already empty, so no need to update if we're clearing
-    if (!gatewayOut.empty() && ethData.defaultGateway != gatewayOut)
-    {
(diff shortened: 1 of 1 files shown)


raviteja-b:
@gtmills This PR needed to unblock network test suite failures

gtmills:
Can we push upstream ?

raviteja-b:
> Can we push upstream ?

1120 downstream networkd has this patch to avoid Static and DHCP ip address co-existence
https://github.com/ibm-openbmc/phosphor-networkd/commit/55e8b2e1b6b0a8a4a16add69836ccc9f1009baf8
which is needed for our customer setups to switch between static and DHCP configuration without network interruption.(without additional DHCP disable operation)

upstream does not have these changes, as per upstream to move to DHCP to static config then first disable DHCP and then configure static config

so we need to keep this patch in downstream for now

baemyung (approved):
It looks good to me, provided that I think this needs to upstream.

PR #1325, opened 2025-06-19
Support host dump offload

asmithakarun:
This PR pulls the commits required to support host dump offload.

### DIFF
--- a/http/http_connection.hpp
+++ b/http/http_connection.hpp
@@ -437,7 +437,6 @@ class Connection :
                 self->completeRequest(thisRes);
             });
         std::string url(req->target());
-//	boost::urls::url_view urlView(req->target());
         if (boost::algorithm::contains(url,
                                        "/system/LogServices/Dump/Entries/") &&
             boost::algorithm::ends_with(url, "/attachment"))
@@ -458,7 +457,7 @@ class Connection :
             });
 
             redfish::dump_utils::getValidDumpEntryForAttachment(
-                asyncResp, boost::urls::url_view(req->target()),//urlView,
+                asyncResp, boost::urls::url_view(req->target()),
                 [asyncResp, this, self(shared_from_this())](
                     [[maybe_unused]] const std::string& objectPath,
                     [[maybe_unused]] const std::string& entryID,
@@ -474,7 +473,6 @@ class Connection :
         }
         if (doUpgrade(asyncResp))
         {
-            BMCWEB_LOG_ERROR("**** doUpgrade(asyncResp) return;");
             return;
         }
         std::string_view expected =
--- a/http/http_connection.hpp
+++ b/http/http_connection.hpp
@@ -7,12 +7,12 @@
 #include "audit_events.hpp"
 #include "authentication.hpp"
 #include "complete_response_fields.hpp"
+#include "dump_utils.hpp"
 #include "forward_unauthorized.hpp"
 #include "http2_connection.hpp"
 #include "http_body.hpp"
 #include "http_connect_types.hpp"
 #include "http_request.hpp"
-#include "dump_utils.hpp"
 #include "http_response.hpp"
 #include "http_utility.hpp"
 #include "logging.hpp"
@@ -44,6 +44,7 @@
 #include <boost/beast/http/verb.hpp>
 #include <boost/none.hpp>
 #include <boost/optional/optional.hpp>
+#include <boost/url/url_view.hpp>
 
 #include <bit>
 #include <chrono>
@@ -57,7 +58,6 @@
 #include <system_error>
 #include <type_traits>
 #include <utility>
-#include <vector>
 
 namespace crow
 {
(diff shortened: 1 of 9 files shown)


http/http_response.hpp
```diff
@@ -349,4 +351,114 @@ struct Response
     bool completed = false;
     std::function<void(Response&)> completeRequestHandler;
 };
+
+struct DynamicResponse
+{
+    using response_type = boost::beast::http::response<
+        boost::beast::http::basic_dynamic_body<boost::beast::flat_static_buffer<
+            static_cast<std::size_t>(1024 * 1024)>>>;
```
nkotania, line 360: why 1024*1024?

http/http_connection.hpp
```diff
@@ -433,6 +436,41 @@ class Connection :
             [self(shared_from_this())](crow::Response& thisRes) {
                 self->completeRequest(thisRes);
             });
+        std::string url(req->target());
```
abhilashraju, line 439: can you put this entire code in to a function in another header and make a single line call to it?Will be beneficial for future refactor

http/http_request.hpp
```diff
@@ -33,6 +34,7 @@ struct Request : std::enable_shared_from_this<Request>
     boost::urls::url urlBase;
 
   public:
+    boost::asio::io_context* ioService = nullptr;
```
abhilashraju: can you remove it from request class?

http/http_response.hpp
```diff
@@ -349,4 +351,114 @@ struct Response
     bool completed = false;
     std::function<void(Response&)> completeRequestHandler;
 };
+
+struct DynamicResponse
```
abhilashraju, line 356: move this entire code in to new file and just include it wherever needed. ease of future refactoring

http/http_stream.hpp
```diff
@@ -0,0 +1,168 @@
...
+#include <memory>
+#include <string>
+
+namespace crow
+{
+
+namespace streaming_response
+{
+
+struct Connection : std::enable_shared_from_this<Connection>
```
abhilashraju, line 31: why do you need this class?

http/http_stream.hpp
```diff
@@ -0,0 +1,168 @@
...
+            });
+    }
+
+    void sendMessage(const boost::asio::mutable_buffer& buffer,
+                     std::function<void()> handler) override
+    {
+        if (buffer.size() != 0)
+        {
+            this->handlerFunc = handler;
+            auto bytes = boost::asio::buffer_copy(
```
abhilashraju, line 136: think about avoiding this extra copy. may be a story for different commit.

http/http_request.hpp
```diff
@@ -72,6 +74,7 @@ struct Request : std::enable_shared_from_this<Request>
     {
         req.clear();
         urlBase.clear();
+        ioService = nullptr;
```
baemyung: This is no longer used in 1120 / upstream and replaced by `getIoContext()`

http/http_request.hpp
asmithakarun: Done

asmithakarun:
Will address all the other comments in a new PR.

include/dump_offload.hpp
```diff
@@ -0,0 +1,444 @@
...
+                                 const boost::system::error_code& ec) {
+            if (ec)
+            {
+                BMCWEB_LOG_ERROR("Async_wait failed {}", ec);
+                return;
+            }
+
+            if (connectRetryCount < maxConnectRetryCount)
+            {
+                BMCWEB_LOG_ERROR("Failed to connect, retrying...",
```
gtmills: You need a {} here for the count

gtmills:
> Will address all the other comments in a new PR.
Let's address the ones here you can. 

Any that are to be addressed later, let's open an issue for that and add that issue here. 
Also why can't we use the upstream implementation? NVIDIA switched to it.

asmithakarun:
> > Will address all the other comments in a new PR.
> > Let's address the ones here you can.
> 
> Any that are to be addressed later, let's open an issue for that and add that issue here. Also why can't we use the upstream implementation? NVIDIA switched to it.

Well, for pFuture, we would be able to use the file body approach for host dump offload (just like bmc dump offload)

asmithakarun:
https://jsw.ibm.com/browse/PFEBMC-4570 - Jira story to track the rework in this PR

gtmills:
> Well, for pFuture, we would be able to use the file body approach for host dump offload (just like bmc dump offload)

So this commit is going away for PFuture ?

asmithakarun:
> > Well, for pFuture, we would be able to use the file body approach for host dump offload (just like bmc dump offload)
> 
> So this commit is going away for PFuture ?

yes

raviteja-b (approved):
changes looks good, we can get this merge after CI passes

http/http_response.hpp
asmithakarun: 1024 * 1024 sets the static buffer to 1 MB - which is reasonable for the redfish response.

Increasing this buffer size would consume more memory and the response will be truncated more when we reduce it further.

http/http_stream.hpp
```diff
@@ -0,0 +1,179 @@
...
+        openHandler(std::move(openHandlerIn)),
+        messageHandler(std::move(messageHandlerIn)),
+        closeHandler(std::move(closeHandlerIn)),
+        errorHandler(std::move(errorHandlerIn))
+    {}
+
+    boost::asio::io_context* getIoContext() override
+    {
+        return static_cast<boost::asio::io_context*>(
+            &adaptor.get_executor().context());
```
baemyung, line 76: Should this be used from the existing one - https://github.com/ibm-openbmc/bmcweb/blob/4fe08a67a30ce801cbb643dde190655e3277c1ea/include/io_context_singleton.hpp#L7

```
#include "io_context_singleton.hpp"


==> Remove this, but use io_context_singleton's getIoContext()?

    boost::asio::io_context* getIoContext() override
    {
        return static_cast<boost::asio::io_context*>(
            &adaptor.get_executor().context());
   }

```

http/http_stream.hpp
```diff
@@ -0,0 +1,179 @@
...
+                buffer);
+            streamres.getBufferResponse().body().commit(bytes);
+            doWrite();
+        }
+    }
+
+    void close() override
+    {
+        streamres.end();
+        adaptor.lowest_layer().close();
```
baemyung, line 147: Why is this needed to change?

```
 boost::beast::get_lowest_layer(adaptor).close();
-->
adaptor.lowest_layer().close();
```

Or, should it be `adaptor.next_layer().close()`?

raviteja-b: this change mainly needed to make sure closing underlying socket which is associated the stream response.
I don't see any problem with this change. both next_layer() or lowest_layer() do the same job in this case.

http/http_stream.hpp
raviteja-b: this change was there right from 1120 bmcweb rebase (jan31). dump offload works with July 1120 builds
I dont think this commit impacts dump offload function but we can give a try.

PR #1326, opened 2025-06-20
1110: Enhance FFDC log traces related to PCIe associations

baemyung:
PCIe topology may show an empty page or entry due to the association error between pcie slot or device to chassis.

However, bmcweb log entries do not show the related dbus object paths or it shows only the error number.
This causes the difficulity of knowing the incorrect associations.

For example, journal shows like

```
Feb 21 15:14:13 p10bmc bmcweb[1135]: [ERROR processor.hpp:1074] DBUS response error, ec: 110
```
This may be ETIMEDOUT but it would be clearer if the log says `generic:110`.

```
Feb 21 15:14:30 p10bmc bmcweb[1135]: [ERROR pcie.hpp:224] PCIe Slot association error!
Feb 21 15:14:30 p10bmc bmcweb[1135]: [CRITICAL error_messages.cpp:287] Internal Error /usr/src/debug/bmcweb/1.0+git/redfish-core/lib/pcie.hpp(225:32) `void redfish::afterAddLinkToPCIeSlot(const std::shared_ptr<bmcweb::AsyncResp>&, const boost::system::error_code&, const dbus::utility::MapperEndPoints&)`:
```

This commit is to enhance ffdc log traces to show dbus object paths if the association is in error.

The areas of the changes are:
- the association error between pcie device to chassis
- the association error between fabric adapter to chassis
- the association error between pcieslot to chassis

Tested:
- Redfish Service Validator runs and passes

### DIFF
--- a/redfish-core/lib/pcie.hpp
+++ b/redfish-core/lib/pcie.hpp
@@ -203,10 +203,10 @@ inline void fillPcieDeviceStatus(crow::Response& resp,
     }
 }
 
-inline void
-    afterAddLinkToPCIeSlot(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
-                           const boost::system::error_code& ec,
-                           const dbus::utility::MapperEndPoints& chassisPaths)
+inline void afterAddLinkToPCIeSlot(
+    const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
+    const std::string& pcieDeviceSlot, const boost::system::error_code& ec,
+    const dbus::utility::MapperEndPoints& chassisPaths)
 {
     if (ec)
     {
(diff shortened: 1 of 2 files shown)


baemyung:
STG defect: https://jazz07.rchland.ibm.com:13443/jazz/web/projects/CSSD#action=com.ibm.team.workitem.viewWorkItem&id=682247

redfish-core/lib/fabric_adapters.hpp
```diff
@@ -182,7 +183,13 @@ inline void afterDoCheckFabricAdapterChassis(
     }
     if (chassisPaths.size() != 1)
     {
-        BMCWEB_LOG_ERROR("PCIe Slot association error! ");
+        BMCWEB_LOG_ERROR(
+            "Association error for fabricAdapter:{} to chassis. Its association count:{} is not equal to 1.",
+            fabricAdapterPath, chassisPaths.size());
+        for (const auto& chassisPath : chassisPaths)
+        {
+            BMCWEB_LOG_ERROR("Invalid chassisPath: {}", chassisPath);
```
jeaaustx: Would too many associations be a common error? I'm wondering if the list of all the paths should be under DEBUG instead to avoid flooding the journal. (Same question with the others below.)

jeaaustx (approved):
Looks good, I just have the one question on DEBUG vs. ERROR for the detail logging. But I'm okay with as-is if the too many paths is not common.

redfish-core/lib/fabric_adapters.hpp
baemyung: Too many associations would be a critical problem in the system configuration (e.g. via inventory or pldm) and such situation should not be expected in the production environment.
If it happens (esp at the customer systems - in addition to the test environment), it would be difficult to debug without the information at hand (as it may not be easy or possible to reproduce). 
That's why it is appropriate to show it as the default log level.

baemyung:
STG defect: https://jazz07.rchland.ibm.com:13443/jazz/web/projects/CSSD#action=com.ibm.team.workitem.viewWorkItem&id=682247
1110:  https://github.com/ibm-openbmc/bmcweb/pull/1326
1120:  https://github.com/ibm-openbmc/bmcweb/pull/1327

gtmills (approved):

PR #1332, opened 2025-07-17
1120: Skip log entry if Hidden dbus property is missing

baemyung:
There may be a timing window that a PEL dbus entry may not have a field `Hidden`, as a PEL dbus entry is handled in a 2 steps
 - a dbus log entry without Hidden field is created first
 - PEL interface reads a dbus log entry and adds `Hidden` field.

During the period, a dbus log entry could have no `Hidden` field.

This commit is to skip those intermediate dbus log entries for LogEntry GET collection handling.

For example,

```
curl -k -X GET https://${bmc}/redfish/v1/Systems/system/LogServices/CELog/Entries/
```

Then, bmc journal may show like this
```
Jul 17 11:12:54 p10bmc bmcwebd[1168]: [log_services.hpp:1756] Skip the log entry /xyz/openbmc_project/logging/entry/2 as it has no Hidden property
```

Tested:
- Redfish Service Validator passes

PR #1333, opened 2025-07-17
1110: Skip log entry if Hidden dbus property is missing

baemyung:
There may be a timing window that a PEL dbus entry may not have a field `Hidden`, as a PEL dbus entry is handled in a 2 steps
 - a dbus log entry without Hidden field is created first
 - PEL interface reads a dbus log entry and adds `Hidden` field.

During the period, a dbus log entry could have no `Hidden` field.

This commit is to skip those intermediate dbus log entries for LogEntry GET collection handling.

For example,

```
curl -k -X GET https://${bmc}/redfish/v1/Systems/system/LogServices/CELog/Entries/
```

Then, bmc journal may show like this
```
Jul 17 11:12:54 p10bmc bmcwebd[1168]: [log_services.hpp:1756] Skip the log entry /xyz/openbmc_project/logging/entry/2 as it has no Hidden property
```

Tested:

### DIFF
--- a/redfish-core/include/utils/error_log_utils.hpp
+++ b/redfish-core/include/utils/error_log_utils.hpp
@@ -7,7 +7,8 @@ namespace error_log_utils
 
 static void getHiddenPropertyValue(
     const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
-    const std::string& entryId, std::function<void(bool hidden)>&& callback)
+    const std::string& entryId,
+    std::function<void(const std::optional<bool>& hidden)>&& callback)
 {
     sdbusplus::asio::getProperty<bool>(
         *crow::connections::systemBus, "xyz.openbmc_project.Logging",
@@ -17,6 +18,15 @@ static void getHiddenPropertyValue(
          entryId](const boost::system::error_code& ec, bool hidden) {
             if (ec)
             {
+                if (ec.value() == EBADR)
+                {
+                    // Put this log trace to journal by default.
+                    BMCWEB_LOG_ERROR(
+                        "DBUS property 'Hidden' for entry {} does not exist",
+                        entryId);
+                    callback(std::nullopt);
+                    return;
+                }
                 BMCWEB_LOG_ERROR(
                     "Failed to get DBUS property 'Hidden' for entry {}: {}",
                     entryId, ec);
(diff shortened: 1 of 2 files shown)


jeaaustx (approved):
Changes look okay to me.

Is there a defect or something that shows the problem? Might be helpful to add to the PR message.

redfish-core/lib/log_services.hpp
```diff
@@ -3243,8 +3258,8 @@ inline void requestRoutesDBusCELogEntry(App& app)
 
                 auto updatePropertyCallback =
                     [resolved, managementSystemAck, asyncResp,
-                     entryId](bool hiddenPropVal) {
-                        if (!hiddenPropVal)
+                     entryId](const std::optional<bool>& hiddenPropVal) {
+                        if (!hiddenPropVal.has_value() || !*hiddenPropVal)
```
spinler: if you wanted, on all of these you could also use the hiddenPropValue.value_or() to do it all in one shot like if (!hiddenPropValue.value_or(true))

spinler (approved):

redfish-core/include/utils/error_log_utils.hpp
```diff
@@ -7,16 +7,26 @@ namespace error_log_utils
...
         "org.open_power.Logging.PEL.Entry", "Hidden",
         [callback = std::move(callback), asyncResp,
-         entryId](const boost::system::error_code& ec, bool hidden) {
+         entryId](const boost::system::error_code& ec,
+                  const std::optional<bool>& hidden) {
             if (ec)
             {
+                if (ec.value() == EBADR)
+                {
+                    BMCWEB_LOG_ERROR(
```
jeaaustx, line 24: Sorry, missed this the first time. Should the caller be displaying this message based on their handling? (Some should be BMCWEB_LOG_WARNING, some seem to just skip it and so should be BMCWEB_LOG_INFO or BMCWEB_LOG_DEBUG instead.)

redfish-core/lib/log_services.hpp
baemyung: Thanks for the tip.  I'll update the code as you suggested.

This case would be
```
 if (!hiddenPropVal.has_value() || !*hiddenPropVal)
-->
 if (!hiddenPropValue.value_or(false))
```

baemyung:
STG: https://jazz07.rchland.ibm.com:13443/jazz/web/projects/CSSD#action=com.ibm.team.workitem.viewWorkItem&id=722338

baemyung:
> Changes look okay to me.
> 
> Is there a defect or something that shows the problem? Might be helpful to add to the PR message.

The defect is https://jazz07.rchland.ibm.com:13443/jazz/web/projects/CSSD#action=com.ibm.team.workitem.viewWorkItem&id=722338

redfish-core/include/utils/error_log_utils.hpp
baemyung: The goal is to show the dbus path on a journal by default, but bmcweb's default log level is ERROR.
So, to make it happen, we would need to mark it as either ERROR or CRITICAL. 
(I chose as ERROR, but not sure which one is better).  Suggestion?

baemyung:
Related links are:
- STG defect [722338](https://jazz07.rchland.ibm.com:13443/jazz/web/projects/CSSD#action=com.ibm.team.workitem.viewWorkItem&id=722338)
- 1050: https://github.com/ibm-openbmc/bmcweb/pull/1336
- 1060: https://github.com/ibm-openbmc/bmcweb/pull/1335
- 1110: https://github.com/ibm-openbmc/bmcweb/pull/1333
- 1120: https://github.com/ibm-openbmc/bmcweb/pull/1332

redfish-core/include/utils/error_log_utils.hpp
jeaaustx: I agree ERROR is better than CRITICAL for these cases.

gtmills (approved):

PR #1335, opened 2025-07-18
1060: Skip log entry if Hidden dbus property is missing

baemyung:
There may be a timing window that a PEL dbus entry may not have a field `Hidden`, as a PEL dbus entry is handled in a 2 steps
 - a dbus log entry without Hidden field is created first
 - PEL interface reads a dbus log entry and adds `Hidden` field.

During the period, a dbus log entry could have no `Hidden` field.

This commit is to skip those intermediate dbus log entries for LogEntry GET collection handling.

For example,

```
curl -k -X GET https://${bmc}/redfish/v1/Systems/system/LogServices/CELog/Entries/
```

Then, bmc journal may show like this
```
Jul 17 11:12:54 p10bmc bmcwebd[1168]: [log_services.hpp:1756] Skip the log entry /xyz/openbmc_project/logging/entry/2 as it has no Hidden property
```

Tested:
- Redfish Service Validator passes

PR #1336, opened 2025-07-18
1050: Skip log entry if Hidden dbus property is missing

baemyung:
There may be a timing window that a PEL dbus entry may not have a field `Hidden`, as a PEL dbus entry is handled in a 2 steps
 - a dbus log entry without Hidden field is created first
 - PEL interface reads a dbus log entry and adds `Hidden` field.

During the period, a dbus log entry could have no `Hidden` field.

This commit is to skip those intermediate dbus log entries for LogEntry GET collection handling.

For example,

```
curl -k -X GET https://${bmc}/redfish/v1/Systems/system/LogServices/CELog/Entries/
```

Then, bmc journal may show like this
```
Jul 17 11:12:54 p10bmc bmcwebd[1168]: [log_services.hpp:1756] Skip the log entry /xyz/openbmc_project/logging/entry/2 as it has no Hidden property
```

Tested:

### DIFF
--- a/redfish-core/lib/log_services.hpp
+++ b/redfish-core/lib/log_services.hpp
@@ -1911,11 +1911,19 @@ template <typename Callback>
 void getHiddenPropertyValue(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
                             const std::string& entryId, Callback&& callback)
 {
-    auto respHandler = [callback{std::forward<Callback>(callback)},
-                        asyncResp](const boost::system::error_code ec,
-                                   const bool& hiddenProperty) {
+    auto respHandler =
+        [callback{std::forward<Callback>(callback)}, asyncResp,
+         entryId](const boost::system::error_code ec, bool hiddenProperty) {
         if (ec)
         {
+            if (ec.value() == EBADR)
+            {
+                // Put this log trace to journal by default.
+                BMCWEB_LOG_ERROR << "DBUS property 'Hidden' for entry "
+                                 << entryId << " does not exist";
+                callback(std::nullopt);
+                return;
+            }
             BMCWEB_LOG_ERROR << "DBUS response error: " << ec;
             messages::internalError(asyncResp->res);
             return;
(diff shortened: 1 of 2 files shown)


baemyung:
Related links are:
- STG defect [722338](https://jazz07.rchland.ibm.com:13443/jazz/web/projects/CSSD#action=com.ibm.team.workitem.viewWorkItem&id=722338)
- 1050: https://github.com/ibm-openbmc/bmcweb/pull/1336
- 1060: https://github.com/ibm-openbmc/bmcweb/pull/1335
- 1110: https://github.com/ibm-openbmc/bmcweb/pull/1333
- 1120: https://github.com/ibm-openbmc/bmcweb/pull/1332

jeaaustx (approved):

gtmills (approved):

redfish-core/lib/log_services.hpp
```diff
@@ -2599,11 +2607,17 @@ inline void requestRoutesDBusEventLogEntry(App& app)
                 messages::internalError(asyncResp->res);
                 return;
             }
+            if (hidden == nullptr)
+            {
+                // Skip log entry if Hidden dbus property is missing
+                messages::resourceNotFound(asyncResp->res, "EventLogEntry",
```
gtmills, line 2615: Not sure resourceNotFound is correct here.. InteneralError ? Oh well

redfish-core/lib/log_services.hpp
```diff
@@ -2836,10 +2852,17 @@ inline void requestRoutesDBusCELogEntry(App& app)
                 return;
             }
 
+            if (hidden == nullptr)
+            {
+                // Skip log entry if Hidden dbus property is missing
+                messages::resourceNotFound(asyncResp->res, "LogEntry",
```
gtmills, line 2862: Again.. Not sure resourceNotFound is correct here.. InteneralError ? Oh well.. 
Also we could trace it..

gtmills (approved):

redfish-core/lib/log_services.hpp
baemyung: > Not sure resourceNotFound is correct here.. InteneralError ? Oh well

Previously, it was treated as internalError if Hidden field does not exist.  (L2605 - on previous code).
This may be okay if Hidden field always is expected to be exist.

However, this commit is to handle the case that Hidden field may be added later. If then, the previous code cause the error and we'd like to skip/ignore those entries during the period.  
To address, we'd like to treat it as non-existent (rather than internalError) so that GET is to be returned just like as to be an GET with non-existent ID.

PR #1338, opened 2025-07-29
1120 Fix 204/304 incorrectly tracing CRITICAL message

abiasojo:
Fix PATCH cases to return 200, 204 was previously being returned

curl -k -H "X-Auth-Token: $bmc_token" -H "Content-Type: application/json" -X PATCH https://${bmc}/redfish/v1/Managers/bmc -d '{"Oem":{"IBM":{"USBCodeUpdateEnabled": false}}}' {
  "@Message.ExtendedInfo": [
    {
      "Message": "The request completed successfully.",
      "MessageArgs": [],
      "MessageId": "Base.1.19.Success",
      "MessageSeverity": "OK",
      "Resolution": "None."
    }
  ]
}
curl -k -H "Content-Type: application/json" -d '{"PowerRestorePolicy":"LastState"}' -X PATCH https://${bmc}/redfish/v1/Systems/system {
  "@Message.ExtendedInfo": [
    {
      "Message": "The request completed successfully.",
      "MessageArgs": [],
      "MessageId": "Base.1.19.Success",
      "MessageSeverity": "OK",
      "Resolution": "None."
    }
  ]
}
curl -k -H "Content-Type: application/json" -H "X-Auth-Token: $bmc_token" -X PATCH -d '{"LocationIndicatorActive":true}' https://${bmc}/redfish/v1/Managers/bmc {
  "@Message.ExtendedInfo": [
    {
      "Message": "The request completed successfully.",
      "MessageArgs": [],
      "MessageId": "Base.1.19.Success",
      "MessageSeverity": "OK",
      "Resolution": "None."
    }
  ]
}

### DIFF
--- a/http/http_response.hpp
+++ b/http/http_response.hpp
@@ -206,8 +206,8 @@ struct Response
         }
         response.content_length(*pSize);
 
-        if (is1XXReturn || result() == status::no_content ||
-            result() == status::not_modified)
+        if ((*pSize > 0) && (is1XXReturn || result() == status::no_content ||
+                             result() == status::not_modified))
         {
             BMCWEB_LOG_CRITICAL("{} Response content provided but code was "
                                 "no-content or not_modified, which aren't "
--- a/redfish-core/lib/systems.hpp
+++ b/redfish-core/lib/systems.hpp
@@ -36,7 +36,6 @@
 
 #include <boost/asio/error.hpp>
 #include <boost/beast/http/field.hpp>
-#include <boost/beast/http/status.hpp>
 #include <boost/beast/http/verb.hpp>
 #include <boost/system/error_code.hpp>
 #include <boost/system/linux_error.hpp>
@@ -2970,8 +2969,6 @@ inline void handleComputerSystemPatch(
             return;
         }
     }
-
-    asyncResp->res.result(boost::beast::http::status::no_content);
 
     if (assetTag)
     {


redfish-core/src/utils/dbus_utils.cpp
```diff
@@ -86,10 +86,10 @@ void afterSetProperty(
         messages::internalError(asyncResp->res);
         return;
     }
-    // Only set 204 if another error hasn't already happened.
+    // Only set success if another error hasn't already happened.
     if (asyncResp->res.result() == boost::beast::http::status::ok)
     {
-        asyncResp->res.result(boost::beast::http::status::no_content);
+        messages::success(asyncResp->res);
```
baemyung: I don't think this is the place to change.
This logic is to detect the situation that would be actually "no-content".

I think - the places to fix would be at the originator side to actually call `message::success()` after the successful patch in `handleComputerSystemPatch()`.

There would be more places  like what's done - https://gerrit.openbmc.org/c/openbmc/bmcweb/+/80049

jeaaustx: I agree that this change masks the places that are returning no_content doing a setProperty(). I'm not sure if that is a bad thing though. It does set the expected status for those in a common place. Is there any risk here that a response is being cleared by this?

Are there other places also returning no_content, i.e. not calling setProperty()?

gtmills: Didn't Ed try this at https://github.com/ibm-openbmc/bmcweb/commit/cdf25ffb6b2d99c829094c9a4c4907aec46e3a2e and have to revert at https://github.com/ibm-openbmc/bmcweb/commit/247ae89bc26024d335f1317566bebeee52c75089

gtmills: Section 7.5.2 on https://www.dmtf.org/sites/default/files/standards/documents/DSP0266_1.22.1.pdf

```
For PATCH (update), PUT (replace), and DELETE (delete) operations, the response from the service, after
successful modification, should be one of the following responses:
• The HTTP 200 OK status code with a body that contains the JSON representation of the targeted resource after
the modification has been applied, or, for the delete operation, a representation of the deleted resource.
• The HTTP 202 Accepted status code with a Location header set to the URI of a task monitor when the
processing of the modification requires additional time.
◦ After processing of the task is complete, the modified resource may be returned in response to a request to
the task monitor URI with the HTTP 200 OK status code.
• The HTTP 204 No Content status code with no response body
```

gtmills: > There would be more places like what's done - https://gerrit.openbmc.org/c/openbmc/bmcweb/+/80049

This works for POST because 7.11 POST (action) from https://www.dmtf.org/sites/default/files/standards/documents/DSP0266_1.22.1.pdf says 200 with a OK message is accepted.. This doesn't work for Put, Patch, Delete because of the above

gtmills: https://github.com/openbmc/bmcweb/blob/8d9cf72d00ccbbdd070849601ce8f37bc01cb8c8/redfish-core/src/error_messages.cpp#L69 success message isn't setting an okay

baemyung: DMTF spec says:
> For PATCH (update), PUT (replace), and DELETE (delete) operations, the response from the service, after
successful modification, should be one of the following responses:
• The HTTP 200 OK status code with a body that contains the JSON representation of the targeted resource after
the modification has been applied, or, for the delete operation, a representation of the deleted resource.
• The HTTP 202 Accepted status code with a Location header set to the URI of a task monitor when the
processing of the modification requires additional time.
◦ After processing of the task is complete, the modified resource may be returned in response to a request to
the task monitor URI with the HTTP 200 OK status code.
• The HTTP 204 No Content status code with no response body.
For details on successful responses to action requests, see POST (action)

Then, 
we can modify dbus_utils.hpp and the caller - so that no_content won't print (for POST/PATCH/DELETE)?

redfish-core/src/utils/dbus_utils.cpp
```diff
@@ -87,7 +87,8 @@ void afterSetProperty(
         return;
     }
     // Only set 204 if another error hasn't already happened.
-    if (asyncResp->res.result() == boost::beast::http::status::ok)
+    if (asyncResp->res.result() == boost::beast::http::status::ok &&
+        !asyncResp->res.body()->size())
```
gtmills: Why is this needed?

http/http_response.hpp
```diff
@@ -206,8 +206,8 @@ struct Response
         }
         response.content_length(*pSize);
 
-        if (is1XXReturn || result() == status::no_content ||
-            result() == status::not_modified)
+        if ((*pSize > 0) && (is1XXReturn || result() == status::no_content ||
```
gtmills, line 209: Can you explain more in the commit why this is needed ?

gtmills:
> curl -k -H "X-Auth-Token: $bmc_token" -H "Content-Type: application/json" -X PATCH https://${bmc}/redfish/v1/Managers/bmc -d '{"Oem":{"IBM":{"USBCodeUpdateEnabled": false}}}'

Let's not include when we go upstream since downstream OEM 

Can you add a TESTED: section in the commit msg?

redfish-core/src/utils/dbus_utils.cpp
abiasojo: It's not needed. I have removed it.

http/http_response.hpp
baemyung: I don't think `(*pSize > 0)` is applied for all cases, but rather it is only for `status::no_content'.

```
        if (is1XXReturn || 
             (*pSize > 0 &&  result() == status::no_content) ||
            result() == status::not_modified)
```

gtmills: I this code is good. Due to  https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/304 "The response must not contain a body". I assume 1xx codes are the same.

baemyung (commented):
I think the commit message (and PR description) can be done

1)  use backticks (```) for some programming/command lines like

- 204 no-content
   ```
    curl -k -H "X-Auth-Token: $bmc_token" -H "Content-Type: application/json" -X PATCH https://${bmc}/redfish/v1/Managers/bmc -d '{"Oem":{"IBM":{"USBCodeUpdateEnabled": false}}}'
    ...
   ```

2) 304 no-content ==> 304 not modified

3) not-modified curl command can be written without the unnecessary options (--insecure & --http1.1). 
For example

```
curl  --insecure -i  -H 'If-Modified-Since: Tue, 21 Nov 2050 08:00:00 GMT' https://${bmc_ip}/redfish/ --http1.1
-->
curl   -i  -H 'If-Modified-Since: Tue, 21 Nov 2050 08:00:00 GMT' https://${bmc_ip}/redfish/ 
```

4) (minor) We can also use --etag-save option to save ETAG without reading header

```
curl  -H 'If-Modified-Since: Tue, 21 Nov 2050 08:00:00 GMT' https://${bmc_ip}/redfish/  --etag-save etag.out
ETAG=`cat etag.out`;  echo $ETAG

curl   -H "If-None-Match: ${ETAG}" https://${bmc_ip}/redfish/ 
```

Besides "304 no-content" ==> "304 not modified", it looks good.


Also, can the

baemyung (approved):

gtmills (approved):

abiasojo:
https://gerrit.openbmc.org/c/openbmc/bmcweb/+/82848

PR #1346, opened 2025-08-08
systems: prevent internal error when mode/ips not available

cjcain:
After BMC reboot it can sometimes take a while for the power mode or idle power status parameters to become available. This change will prevent returning an error when there is a timeout trying to read them. This will allow the other systems data to be returned.systems: prevent internal error when mode/ips not available

After BMC reboot it can sometimes take a while for the power mode or idle power status parameters to become available. This change will prevent returning an error when there is a timeout trying to read them. This will allow the other systems data to be returned.

Change-Id: I65f7e2f4ccf946ce108ca10f72cb81dd0e8a7fd2

### DIFF
--- a/redfish-core/lib/systems.hpp
+++ b/redfish-core/lib/systems.hpp
@@ -1548,7 +1548,11 @@ inline void
     if (ec)
     {
         BMCWEB_LOG_ERROR("DBUS response error on PowerMode GetAll: {}", ec);
-        messages::internalError(asyncResp->res);
+        // Don't return an error on timeout (since it should eventually recover)
+        if(ec != boost::system::errc::timed_out)
+        {
+            messages::internalError(asyncResp->res);
+        }
         return;
     }
 
@@ -2064,7 +2068,11 @@ inline void
                 BMCWEB_LOG_ERROR(
                     "DBUS response error on Power.IdlePowerSaver GetSubTree {}",
                     ec);
-                messages::internalError(asyncResp->res);
+                // Don't return an error on timeout (since it should eventually recover)
+                if(ec != boost::system::errc::timed_out)
+                {
+                    messages::internalError(asyncResp->res);
+                }
                 return;
             }
             if (subtree.empty())


redfish-core/lib/systems.hpp
```diff
@@ -1548,7 +1548,11 @@ inline void
     if (ec)
     {
         BMCWEB_LOG_ERROR("DBUS response error on PowerMode GetAll: {}", ec);
-        messages::internalError(asyncResp->res);
+        // Don't return an error on timeout (since it should eventually recover)
+        if (ec.value() != ETIMEDOUT)
```
baemyung: https://www.boost.org/doc/libs/1_86_0/libs/system/doc/html/system.html#ref_boostsystemerrc_hpp

For precise matching, this can be done 
```
if(ec != boost::system::errc::timed_out)
```

redfish-core/lib/systems.hpp
```diff
@@ -2064,7 +2068,11 @@ inline void
                 BMCWEB_LOG_ERROR(
                     "DBUS response error on Power.IdlePowerSaver GetSubTree {}",
                     ec);
-                messages::internalError(asyncResp->res);
+                // Don't return an error on timeout (since it should eventually recover)
+                if (ec.value() != ETIMEDOUT)
```
baemyung: Similarly, 
```
if(ec != boost::system::errc::timed_out)
```

baemyung (approved):
It looks good to me (once it passes CI)

baemyung:
@cjcain
It looks like - CI has failed because the code is not passed with `clang-format-19`.

cjcain:
Decided to abandon this since there was an entity-manager change that would probably resolve it.
https://github.ibm.com/openbmc/openbmc/pull/6449

baemyung:
For the reference, the associated STG defect is https://jazz07.rchland.ibm.com:13443/jazz/web/projects/CSSD#action=com.ibm.team.workitem.viewWorkItem&id=723462

PR #1347, opened 2025-08-10
 Unauthenticated ACF upload failure

abhilashraju:
Unauthenticated ACF upload give invalide request parmameters in 1120 This changeset fixes the issue. There may be some uses cases missed to handle in this patch.

### DIFF
--- a/redfish-core/lib/account_service.hpp
+++ b/redfish-core/lib/account_service.hpp
@@ -1477,121 +1477,6 @@ inline void uploadACF(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
         "InstallACF", decodedAcf);
 }
 
-inline void afterDoUnauthenticatedACFUpload(
-    const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
-    const std::vector<uint8_t>& decodedAcf, const boost::system::error_code& ec,
-    const bool allowUnauthACFUpload)
-{
-    if (ec)
-    {
-        BMCWEB_LOG_ERROR("D-Bus response error reading allow_unauth_upload: {}",
-                         ec.value());
-        messages::internalError(asyncResp->res);
-        return;
-    }
-
-    if (allowUnauthACFUpload)
-    {
-        uploadACF(asyncResp, decodedAcf);
-        return;
-    }
-
-    // Allow ACF upload when D-Bus property ACFWindowActive is true
-    // (aka OpPanel function 74).
-    dbus::utility::getProperty<bool>(
-        "com.ibm.PanelApp", "/com/ibm/panel_app", "com.ibm.panel",
-        "ACFWindowActive",
-        [asyncResp, decodedAcf](const boost::system::error_code& ec1,
-                                const bool isACFWindowActive) {
-            if (ec1)
-            {
-                BMCWEB_LOG_ERROR("Failed to read ACFWindowActive property");
-                // The Panel app doesn't run on simulated systems.
-            }
-
-            if (!isACFWindowActive)
-            {
-                BMCWEB_LOG_ERROR("ACF upload not allowed");
-                messages::insufficientPrivilege(asyncResp->res);
-                return;
-            }
-
-            uploadACF(asyncResp, decodedAcf);
-            return;
-        });
-}
-
-// Allow ACF upload when D-Bus property allow_unauth_upload is true (aka Redfish
-// property AllowUnauthACFUpload).
-inline void doUnauthenticatedACFUpload(
-    const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
-    const std::vector<uint8_t>& decodedAcf)
-{
-    dbus::utility::getProperty<bool>(
-        "xyz.openbmc_project.Settings",
-        "/xyz/openbmc_project/ibmacf/allow_unauth_upload",
-        "xyz.openbmc_project.Object.Enable", "Enabled",
-        std::bind_front(afterDoUnauthenticatedACFUpload, asyncResp,
-                        decodedAcf));
-}
-
-// This is called when someone either is not authenticated or is not
-// authorized to upload an ACF, and they are trying to upload an ACF;
-// in this condition, uploading an ACF is allowed when
-// AllowUnauthACFUpload is true.
-inline void triggerUnauthenticatedACFUpload(
-    const std::shared_ptr<bmcweb::AsyncResp>& asyncResp, nlohmann::json& oem)
-{
-    std::optional<nlohmann::json> ibm;
-    if (!redfish::json_util::readJson(oem, asyncResp->res, "IBM", ibm))
-    {
-        BMCWEB_LOG_WARNING("Illegal Property ");
-        messages::propertyMissing(asyncResp->res, "IBM");
-        return;
-    }
-
-    std::optional<nlohmann::json> acf;
-    std::optional<std::string> acfFile{};
-    if (ibm)
-    {
-        if (!redfish::json_util::readJson(oem, asyncResp->res,   //
-                                          "IBM/ACF", acf,        //
-                                          "IBM/ACFFile", acfFile //
-                                          ))
-        {
-            BMCWEB_LOG_WARNING("Illegal Property ");
-            messages::propertyMissing(asyncResp->res, "ACF");
-        }
-    }
-
-    if (acf && acfFile)
-    {
-        std::vector<uint8_t> decodedAcf;
-        std::string sDecodedAcf;
-        if (!acfFile.has_value() ||
-            !crow::utility::base64Decode(*acfFile, sDecodedAcf))
-        {
-            BMCWEB_LOG_ERROR("base64 decode failure ");
-            messages::internalError(asyncResp->res);
-            return;
-        }
-        try
-        {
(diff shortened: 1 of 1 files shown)


baemyung:
Isn't the issue caused by triggerUnauthenticatedACFUpload() which mistakenly reads `oem` twice?

https://github.com/ibm-openbmc/bmcweb/blob/45e01712d8b3d6266b10a640c63c53773b2b6b1a/redfish-core/lib/account_service.hpp#L1545C1-L1551C1

```
--- i/redfish-core/lib/account_service.hpp
+++ w/redfish-core/lib/account_service.hpp
@@ -1542,14 +1542,6 @@ inline void doUnauthenticatedACFUpload(
 inline void triggerUnauthenticatedACFUpload(
     const std::shared_ptr<bmcweb::AsyncResp>& asyncResp, nlohmann::json& oem)
 {
-    std::optional<nlohmann::json> ibm;
-    if (!redfish::json_util::readJson(oem, asyncResp->res, "IBM", ibm))
-    {
-        BMCWEB_LOG_WARNING("Illegal Property ");
-        messages::propertyMissing(asyncResp->res, "IBM");
-        return;
-    }
-
     std::optional<nlohmann::json> acf;
     std::optional<std::string> acfFile{};
     if (ibm)
```

Besides this, your re-factoring seems good (but I haven't tried).

redfish-core/lib/account_service.hpp
```diff
@@ -2753,7 +2638,65 @@ inline void handleAccountDelete(
...
+        return false;
+    }
+
+    std::optional<std::string> acfFile;
+    std::optional<bool> allowUnauthACFUpload;
+
+    if (!redfish::json_util::readJson(
+            oem, asyncResp->res,                                 //
+            "IBM/ACF/ACFFile", acfFile,                          //
+            "IBM/ACF/AllowUnauthACFUpload", allowUnauthACFUpload //
```
joseph-reynolds: Empty comments?

abhilashraju:
@baemyung there are two checks missing in this patch, 1) ACFWindowActive on com.ibm.PanelApp and 2) allowAcfUpload on Settings. Do you think that was unnecessary ? in this patch the allowAcfUpload value is read from request json instead, that seems to be wrong.

redfish-core/lib/account_service.hpp
abhilashraju: done

baemyung:
> @baemyung there are two checks missing in this patch, 1) ACFWindowActive on com.ibm.PanelApp and 2) allowAcfUpload on Settings. Do you think that was unnecessary ? in this patch the allowAcfUpload value is read from request json instead, that seems to be wrong.


I think the existing code can be fixed by removing the lines L1545-L1462 if we want to have a quick fix.
If your refactoring is preferred, 

>1) ACFWindowActive on com.ibm.PanelApp and 2) allowAcfUpload on Settings. 

Both checks are done when it performs unauthACFLoad - by doUnauthenticatedACFUpload().
- doUnauthenticatedACFUpload() checks "allow_unauth_upload" setting
- afterDoUnauthenticatedACFUpload() checks "ACFWindowActive"

ReadJson of "AllowUnauthACFUpload" is use to set property.
(I think the confusion between "allow_unauth_upload" from setting and "AllowUnauthACFUpload" from json is caused by the same variable name as allowUnauthACFUpload on lambda callback.  Perhaps we can change one of variable name).

redfish-core/lib/account_service.hpp
```diff
@@ -2753,7 +2638,122 @@ inline void handleAccountDelete(
...
+        "xyz.openbmc_project.Settings",
+        "/xyz/openbmc_project/ibmacf/allow_unauth_upload",
+        "xyz.openbmc_project.Object.Enable", "Enabled",
+        std::bind_front(afterDoUnauthenticatedACFUpload, asyncResp,
+                        decodedAcf));
+}
+
+inline bool handleOem(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
+                      nlohmann::json& oem, const crow::Request& req,
+                      const std::string& username, bool unauth = false)
```
baemyung: This 'unauth` would be an option to go thru "unath-ACF-Load" if it is true, after the caller already determines that it would be the code path to do unathACFLoad.

So,
-  I think this variable can be named like `tryUnauthACFLoad` to make it clear.
- Also, I think it is better not use to the default arg value so that the caller must pass the value?
- Also would it be better to name this function like `handleOemPatch()`?

```
inline bool handleOemPatch(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
                      nlohmann::json& oem, const crow::Request& req,
                      const std::string& username, bool tryUnauthACFLoad)
```

redfish-core/lib/account_service.hpp
```diff
@@ -2753,7 +2638,122 @@ inline void handleAccountDelete(
...
+                      std::back_inserter(decodedAcf));
+        }
+        catch (const std::exception& e)
+        {
+            BMCWEB_LOG_ERROR("{}", e.what());
+            messages::internalError(asyncResp->res);
+            return false;
+        }
+        unauth ? doUnauthenticatedACFUpload(asyncResp, decodedAcf)
+               : uploadACF(asyncResp, decodedAcf);
```
baemyung: I think this can be better like

```
        if(unauth)
       {
               doUnauthenticatedACFUpload(asyncResp, decodedAcf);
       } 
       else
       {
               uploadACF(asyncResp, decodedAcf);
        }
```

redfish-core/lib/account_service.hpp
```diff
@@ -2813,13 +2808,9 @@ inline void handleAccountPatch(
...
     {
-        // Irrespective of role can patch ACF if function
-        // 74 is active from panel.
-        if (oem && (username == "service"))
+        if (oem)
         {
-            // allow unauthenticated ACF upload based on panel
-            // function 74 state.
-            triggerUnauthenticatedACFUpload(asyncResp, *oem);
+            handleOem(asyncResp, *oem, req, username);
```
baemyung: I think would need to be "tryUnauthACFLoad" path 
```
   handleOem(asyncResp, *oem, req, username, true);
```

redfish-core/lib/account_service.hpp
```diff
@@ -2861,64 +2851,10 @@ inline void handleAccountPatch(
...
-                return;
-            }
-            uploadACF(asyncResp, decodedAcf);
-        }
-
-        if (allowUnauthACFUpload)
-        {
-            setPropertyAllowUnauthACFUpload(asyncResp, *allowUnauthACFUpload);
-        }
+        return;
```
baemyung: Would it be better to be done like this?

```
    if (oem)
    {
         if( !handleOem(asyncResp, *oem, req, username, false))
        {
              return;
         }
     }
```

abhilashraju:
> > @baemyung there are two checks missing in this patch, 1) ACFWindowActive on com.ibm.PanelApp and 2) allowAcfUpload on Settings. Do you think that was unnecessary ? in this patch the allowAcfUpload value is read from request json instead, that seems to be wrong.
> 
> I think the existing code can be fixed by removing the lines L1545-L1462 if we want to have a quick fix. If your refactoring is preferred,
> 
> > 1. ACFWindowActive on com.ibm.PanelApp and 2) allowAcfUpload on Settings.
> 
> Both checks are done when it performs unauthACFLoad - by doUnauthenticatedACFUpload().
> 
> * doUnauthenticatedACFUpload() checks "allow_unauth_upload" setting
> * afterDoUnauthenticatedACFUpload() checks "ACFWindowActive"
> 
> ReadJson of "AllowUnauthACFUpload" is use to set property. (I think the confusion between "allow_unauth_upload" from setting and "AllowUnauthACFUpload" from json is caused by the same variable name as allowUnauthACFUpload on lambda callback. Perhaps we can change one of variable name).

Yeah I have included the settings and acwindow check later. If you are fine with the changes please do approve

baemyung:
> Yeah I have included the settings and acwindow check later. If you are fine with the changes please do approve

I added a few comments.
- Among them, please look at the comment for L2813 - which I think `auth=true` is missing.
- Please also remove "WIP" from this PR's subject

redfish-core/lib/account_service.hpp
abhilashraju: done

redfish-core/lib/account_service.hpp
abhilashraju: done

redfish-core/lib/account_service.hpp
```diff
@@ -2861,64 +2857,13 @@ inline void handleAccountPatch(
...
-        std::optional<bool> allowUnauthACFUpload;
-
-        if (!redfish::json_util::readJson(
-                *oem, asyncResp->res,                                //
-                "IBM/ACF/ACFFile", acfFile,                          //
-                "IBM/ACF/AllowUnauthACFUpload", allowUnauthACFUpload //
-                ))
+        if (!handleOemPatch(asyncResp, *oem, req, username, false))
         {
-            BMCWEB_LOG_WARNING("Illegal Property");
```
raviteja-b, line 2885: better to log this warning or error?

raviteja-b (approved):

baemyung (approved):
It looks good to me (once it passes CI)

PR #1349, opened 2025-08-13
RBMC: Add compile option

jeaaustx:
Adding compile option for enabling redundant Manager support in bmcweb. This allows staging of the work.

Tested: Built enabled and disabled.

Note: Will create upstream commit to define this option as well once we agree on the naming of it.

Note: Will turn on by default for IBM huygens in the bmcweb_%.bbappend file once the option is merged:
```
EXTRA_OEMESON:append:huygens = " \
    -Dexperimental-redfish-redundant-manager=enabled \
"
```

### DIFF
--- a/config/meson.build
+++ b/config/meson.build
@@ -10,6 +10,7 @@ feature_options = [
     'event-subscription',
     'experimental-redfish-dbus-log-subscription',
     'experimental-redfish-multi-computer-system',
+    'experimental-redfish-redundant-manager',
     'google-api',
     'host-serial-socket',
     'http-zstd',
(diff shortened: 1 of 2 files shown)


baemyung (approved):
It looks good to me

gtmills (approved):
This looks good to me, but let's push upstream with some code so we can show what we mean..  This looks good as a first commit though.

jeaaustx:
> This looks good to me, but let's push upstream with some code so we can show what we mean.. This looks good as a first commit though.

For upstream I will add a dummy callout under the option for filling in the Redundancy part of the Manager.

PR #1350, opened 2025-08-18
RBMC: Get redundancy information for manager

jeaaustx:
TODO: Adding aggregated BMC to RedundancySet.

Tested:
 - Redfish Service Validator - passes checks on Redundancy
 - Redundancy included as part of Manager response when run on both the passive and active manager.

```
    /* Response looks same when run on either active or passive BMC */
    
    curl -s -k -H "X-Auth-Token: $token" -X GET https://${bmc}/redfish/v1/Managers/bmc
    {
      "@odata.id": "/redfish/v1/Managers/bmc",
      "@odata.type": "#Manager.v1_15_0.Manager",
      ...
      "Redundancy": [
        {
          "@odata.id": "/redfish/v1/Managers/bmc#/Redundancy/0",
          "MemberId": "0",
          "MinNumNeeded": 2,
          "Mode": "Failover",
          "Name": "Manager Redundancy",
          "RedundancySet": [
            {
            }
          ],
          "RedundancySet@odata.count": 1,
          "Status": {
            "Health": "OK",
            "State": "Enabled"
          }
        }
      ],
      "Redundancy@odata.count": 1,
      ...
```

    /* Response on both passive and active BMC after disabling redundancy
     * using rbmctool -s
     */
    ```
    curl -s -k -H "X-Auth-Token: $token" -X GET https://${bmc}/redfish/v1/Managers/bmc
    {
      "@odata.id": "/redfish/v1/Managers/bmc",
      "@odata.type": "#Manager.v1_15_0.Manager",
      ...
      "Redundancy": [
        {
          "@odata.id": "/redfish/v1/Managers/bmc#/Redundancy/0",
          "MemberId": "0",
          "MinNumNeeded": 2,
          "Mode": "NotRedundant",
          "Name": "Manager Redundancy",
          "RedundancySet": [
            {
            }
          ],
          "RedundancySet@odata.count": 1,
          "Status": {
            "Health": "OK",
            "State": "Disabled"
          }
        }
      ],
      "Redundancy@odata.count": 1,
      ...
    ```

### DIFF
--- a/redfish-core/include/manager_redundancy.hpp
+++ b/redfish-core/include/manager_redundancy.hpp
@@ -1,0 +1,219 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright OpenBMC Authors
+#pragma once
+
+#include "async_resp.hpp"
+#include "dbus_singleton.hpp"
+#include "dbus_utility.hpp"
+#include "error_messages.hpp"
+#include "generated/enums/redundancy.hpp"
+#include "generated/enums/resource.hpp"
+#include "logging.hpp"
+#include "utils/dbus_utils.hpp"
+
+#include <asm-generic/errno.h>
+
+#include <boost/system/error_code.hpp>
+#include <boost/url/format.hpp>
+#include <nlohmann/json.hpp>
+#include <sdbusplus/message/native_types.hpp>
+#include <sdbusplus/unpack_properties.hpp>
+
+#include <array>
+#include <cstddef>
+#include <format>
+#include <functional>
+#include <map>
+#include <memory>
+#include <ranges>
+#include <string>
+#include <string_view>
+#include <utility>
+
+namespace redfish
+{
+
+inline redundancy::RedundancyMode dBusRedundancyEnabledToRedfish(
+    const bool redundancyEnabled)
+{
+    if (redundancyEnabled)
+    {
+        return redundancy::RedundancyMode::Failover;
+    }
+
+    return redundancy::RedundancyMode::NotRedundant;
+}
+
+inline resource::State getRedundantState(bool failoversAllowed,
+                                         bool redundancyEnabled)
+{
+    if (!redundancyEnabled || !failoversAllowed)
+    {
+        return resource::State::Disabled;
+    }
+
+    return resource::State::Enabled;
+}
+
+inline void getRedundantData(
+    const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
+    const std::string& managerId, nlohmann::json::object_t& redundantObject,
+    const boost::system::error_code& ec,
+    const dbus::utility::DBusPropertiesMap& propertiesList)
+{
+    if (ec)
+    {
+        if (ec.value() != EBADR)
+        {
+            BMCWEB_LOG_ERROR("D-Bus response error {}", ec);
+            messages::internalError(asyncResp->res);
+            return;
+        }
+        BMCWEB_LOG_DEBUG("Can't get Redundancy properties");
+        return;
+    }
+
+    const bool* failoversAllowed = nullptr;
+    const bool* redundancyEnabled = nullptr;
+    const size_t* redundancyMinimum = nullptr;
+
+    const bool success = sdbusplus::unpackPropertiesNoThrow(
+        dbus_utils::UnpackErrorPrinter(), propertiesList, "FailoversAllowed",
+        failoversAllowed, "RedundancyEnabled", redundancyEnabled,
+        "RedundancyMinimum", redundancyMinimum);
+
+    if (!success)
+    {
+        messages::internalError(asyncResp->res);
+        return;
+    }
+
+    auto redundancyIt = asyncResp->res.jsonValue.find("Redundancy");
+
+    if (redundancyIt == asyncResp->res.jsonValue.end())
+    {
+        // Initialize the Redundancy array
+        BMCWEB_LOG_DEBUG("Initializing Redundancy array");
+        asyncResp->res.jsonValue["Redundancy"] = nlohmann::json::array_t();
+        asyncResp->res.jsonValue["Redundancy@odata.count"] = 0;
+        redundancyIt = asyncResp->res.jsonValue.find("Redundancy");
+    }
+
+    auto* redundancy = redundancyIt->get_ptr<nlohmann::json::array_t*>();
+    if (redundancy == nullptr)
+    {
+        messages::internalError(asyncResp->res);
+        return;
+    }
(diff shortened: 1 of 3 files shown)


redfish-core/lib/managers.hpp
```diff
@@ -820,6 +821,133 @@ inline void getManagerObject(
...
+
+inline void getRedundantData(
+    const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
+    nlohmann::json::object_t& redundantObject,
+    const boost::system::error_code& ec,
+    const dbus::utility::DBusPropertiesMap& propertiesList)
+{
+    if (ec)
+    {
+        BMCWEB_LOG_DEBUG("Can't get bmc asset!");
```
gtmills: asset? Is "Can't get bmc redundant data" better?

redfish-core/lib/managers.hpp
```diff
@@ -820,6 +821,133 @@ inline void getManagerObject(
...
+    {
+        redundantObject["Mode"] =
+            dBusRedundancyEnabledToRedfish(*redundancyEnabled);
+    }
+
+    if (redundancyMinimum != nullptr)
+    {
+        redundantObject["MinNumNeeded"] = *redundancyMinimum;
+    }
+
```
gtmills, line 931: Don't we also need the Status or will that come later?

gtmills: Status will look at failover enabled.. "FailoversPaused"

jeaaustx: The PDI for that seems to be in process of changing. (https://gerrit.openbmc.org/c/openbmc/phosphor-dbus-interfaces/+/80750)

I will add initializing good path for Status as hard-coded first. I'll check with Matt on state of the PDI to see direction for basing values off the properties.

jeaaustx: (Looking downstream it has already been changed to FailoversAllowed. So I'll confirm with Matt that is the correct interface I should be using.)
https://github.com/ibm-openbmc/phosphor-dbus-interfaces/blob/9ac3d092d9fdcf72151ac564580d1f58921f1f94/yaml/xyz/openbmc_project/State/BMC/Redundancy.interface.yaml#L36

redfish-core/lib/managers.hpp
jeaaustx: I've updated the logging messages throughout in the latest patch.

redfish-core/lib/managers.hpp
jeaaustx: Matt confirmed the downstream PDI is the correct properties to use. I've adjusted to use these properties and added the good path Status.

For bad path handling I see two other stories to address those, so I will leave that for those stories:

- https://jsw.ibm.com/browse/PFEBMC-4452
- https://jsw.ibm.com/browse/PFEBMC-4453

redfish-core/include/manager_redundancy.hpp
```diff
@@ -0,0 +1,219 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright OpenBMC Authors
+// SPDX-FileCopyrightText: Copyright 2018 Intel Corporation
```
gtmills: let's blast Intel here

jeaaustx: I've removed it.

redfish-core/include/manager_redundancy.hpp
```diff
@@ -0,0 +1,218 @@
...
+inline resource::State getRedundantState(bool failoversAllowed,
+                                         bool redundancyEnabled)
+{
+    resource::State state = resource::State::Enabled;
+
+    if (!redundancyEnabled || !failoversAllowed)
+    {
+        state = resource::State::Disabled;
+    }
+    return state;
```
baemyung: Later, would this function may have more states, or only this Enabled/Disabled?

Perhaps, can this function be like this - simply based on the arguments?
```

inline resource::State getRedundantState(bool failoversAllowed,
                                         bool redundancyEnabled)
{
    if (!redundancyEnabled || !failoversAllowed)
    {
        return resource::State::Disabled;
    }
    return [state](resource::State::Enabled);
}
```

redfish-core/include/manager_redundancy.hpp
```diff
@@ -0,0 +1,218 @@
...
+        BMCWEB_LOG_DEBUG("Iniitalizing Redundancy array");
+        asyncResp->res.jsonValue["Redundancy"] = nlohmann::json::array_t();
+        asyncResp->res.jsonValue["Redundancy@odata.count"] = 0;
+        redundancyIt = asyncResp->res.jsonValue.find("Redundancy");
+    }
+
+    auto* redundancy = redundancyIt->get_ptr<nlohmann::json::array_t*>();
+    if (redundancy == nullptr)
+    {
+        BMCWEB_LOG_DEBUG("Failed to initialize Redundancy array");
```
baemyung: Wouldn't this be ERROR?

redfish-core/include/manager_redundancy.hpp
```diff
@@ -0,0 +1,218 @@
...
+     */
+    for (const auto& [path, services] : subtree)
+    {
+        sdbusplus::message::object_path managerPath(path);
+        if (managerPath.filename() == "bmc0")
+        {
+            // Expect a single service
+            if (services.size() != 1 || services[0].first.empty())
+            {
+                BMCWEB_LOG_DEBUG("Unexpected Redundancy D-Bus subtree");
```
baemyung: Is this ERROR?

redfish-core/include/manager_redundancy.hpp
jeaaustx: I'm not sure if there will be other states, but seems clearer as you suggest so I've changed it.

redfish-core/include/manager_redundancy.hpp
jeaaustx: Originally I was thinking to not fail the entire Manager GET for failing to get the Redundancy state, but this failure would be really unlikely. I've changed it to an internalError.

redfish-core/include/manager_redundancy.hpp
jeaaustx: Originally I was thinking to not fail the entire Manager GET for failing to get the Redundancy state, but this failure would mean the code should be modified to handle multiple services. I've changed this to an internalError and amy displaying the number of services on failure to help with debug. (On debug if the service size was 1 then we would know the failure mapped to an empty service instead so I decided not to add that to the message.)

baemyung (approved):
It looks good to me

PR #1356, opened 2025-09-14
1060: Fix obmc console handling issue

baemyung:
The coredump had happened on session management on obmc_console/websocket because a session object (in an array of sessions) may already become invalid during doRead.
This  can be avoided by moving the sessions into a map of connection and its console socket object is deleted as a part of connection handler object.

The key commit that cherry-picked is:

- Introduce ConsoleHandler class under obmc_console (https://github.com/ibm-openbmc/bmcweb/commit/f948d810b7c8dad91203f050923db51a16cdacc1)

Added new ConsoleHandler class to prepare for the multiple consoles
support. All global fields are moved to the ConsoleHandler class and a
new global map added to remember the ConsoleHandler for each console
path. There is single ConsoleHandler per connection so we don't need
session map per route. There is a limit added for max number of
connection allowed to avoid any service attacks.

Testing:
    - Make sure that single console works fine and data is seen on the
      console.
    - Make sure that multiple consoles of type host console are created
      and data is seen on all consoles. Also using traces made sure
      that new handlers are destroyed.

Change-Id: I77a58a3a186e87611219aed90b221f9b8be7fa2f

### DIFF
--- a/include/obmc_console.hpp
+++ b/include/obmc_console.hpp
@@ -1,13 +1,19 @@
 #pragma once
+#include "app.hpp"
+#include "async_resp.hpp"
+#include "privileges.hpp"
+#include "websocket.hpp"
+
 #include <sys/socket.h>
 
-#include <app.hpp>
-#include <async_resp.hpp>
 #include <boost/asio/local/stream_protocol.hpp>
 #include <boost/container/flat_map.hpp>
-#include <dbus_utility.hpp>
-#include <privileges.hpp>
-#include <websocket.hpp>
+#include <boost/system/error_code.hpp>
+
+#include <array>
+#include <memory>
+#include <string>
+#include <string_view>
 
 namespace crow
 {
--- a/include/obmc_console.hpp
+++ b/include/obmc_console.hpp
@@ -75,13 +75,21 @@ class ConsoleHandler : public std::enable_shared_from_this<ConsoleHandler>
         });
     }
 
+    static void afterSendEx(const std::weak_ptr<ConsoleHandler>& weak)
+    {
+        std::shared_ptr<ConsoleHandler> self = weak.lock();
+        if (self == nullptr)
+        {
+            return;
+        }
+        self->doRead();
+    }
+
     void doRead()
     {
-        std::size_t bytes = outputBuffer.capacity() - outputBuffer.size();
-
         BMCWEB_LOG_DEBUG << "Reading from socket";
         hostSocket.async_read_some(
-            outputBuffer.prepare(bytes),
+            boost::asio::buffer(outputBuffer),
             [this, weakSelf(weak_from_this())](
                 const boost::system::error_code& ec, std::size_t bytesRead) {
             BMCWEB_LOG_DEBUG << "read done.  Read " << bytesRead << " bytes";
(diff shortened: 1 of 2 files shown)


baemyung:
FPS doc: https://jazz07.rchland.ibm.com:13443/jazz/web/projects/FPS%20Collaboration#action=com.ibm.team.workitem.viewWorkItem&id=734712

STG Defect:
https://jazz07.rchland.ibm.com:13443/jazz/web/projects/CSSD#action=com.ibm.team.workitem.viewWorkItem&id=735990

1050 PR: https://github.com/ibm-openbmc/bmcweb/pull/1357
1060 PR: https://github.com/ibm-openbmc/bmcweb/pull/1356

include/obmc_console.hpp
```diff
@@ -1,203 +1,288 @@
...
+        return;
+    }
+    handler->second->inputBuffer += data;
+    handler->second->doWrite();
+}
+
+inline void requestRoutes(App& app)
+{
+    BMCWEB_ROUTE(app, "/console0")
+        .privileges({{"OpenBMCHostConsole"}})
```
jeaaustx: Privileges here are changing from `ConfigureManager` on this level, is that desired?

If change is desired, should it also match check of privileges checked on L242 of the changed code?

include/obmc_console.hpp
```diff
@@ -1,203 +1,288 @@
...
-        sessions.erase(&conn);
-        if (sessions.empty())
+        if (!userPrivileges.isSupersetOf(requiredPrivileges))
         {
-            hostSocket = nullptr;
-            inputBuffer.clear();
-            inputBuffer.shrink_to_fit();
+            BMCWEB_LOG_DEBUG << "User " << conn.getUserName()
+                             << " not authorized for host console connection";
+            conn.close("Unathourized access");
```
jeaaustx: typo: Unauthorized

include/obmc_console.hpp
```diff
@@ -1,203 +1,288 @@
...
+        "GetUserInfo", conn.getUserName());
+}
+
+inline void onMessage(crow::websocket::Connection& conn,
+                      const std::string& data, bool /*isBinary*/)
+{
+    auto handler = getConsoleHandlerMap().find(&conn);
+    if (handler == getConsoleHandlerMap().end())
+    {
+        BMCWEB_LOG_CRITICAL << "Unable to find connection " << &conn;
```
jeaaustx, line 271: Should this be deemed critical?

include/obmc_console.hpp
baemyung: Thanks for finding it.  Fixed it to keep the original ConfigureManager

include/obmc_console.hpp
baemyung: Fixed

include/obmc_console.hpp
baemyung: This is the same as what's in upstream - https://github.com/baemyung/bmcweb/blob/fc1342c5d1ee85c5e2799aee988571274091f854/features/serial/obmc_console.hpp#L330

The reason would be because not having `conn` at the time of `onMessage()` (after `onOpen()`) implies that bmcweb daemon itself would become pretty corrupted.

jeaaustx (approved):

gtmills (approved):

PR #1359, opened 2025-09-26
Change the logging for socket connection retries

asmithakarun:


baemyung (commented):
It would be also better to show max-count on L153.
```
L153:                BMCWEB_LOG_ERROR("Failed to connect, reached max retry count:  {}",
                                 connectRetryCount);
```

baemyung:
Could you add some description (e.g. the example output etc) into the commit message?

asmithakarun:
> It would be also better to show max-count on L153.
> 
> ```
> L153:                BMCWEB_LOG_ERROR("Failed to connect, reached max retry count:  {}",
>                                  connectRetryCount);
> ```

But in this case, connectRetryCount will be at (or would have reached) maxRetryCount at that time of logging right?

baemyung (approved):
It looks good to me

PR #1360, opened 2025-09-30
Add debug for requests and connection count

gtmills:
Up the tracing for requests and ip addresses and the connection count. This will help debug a problem with running out of connections.


baemyung (approved):
It looks good to me

PR #1366, opened 2025-10-09
Audit Log: Expect USER_LOGIN audit entries

jeaaustx:
The phosphor-auditlog backend is parsing USER_LOGIN entries into the same format as USYS_CONFIG entries. [1] Don't warn under DEBUG when these entries are seen.

Tested: Confirmed warning no longer displayed when these entries are included in the GetLatestEntries response.

[1] https://github.com/ibm-openbmc/phosphor-logging/pull/130


gtmills (approved):

baemyung (approved):
It looks good to me

PR #1374, opened 2025-10-20
Audit Log: Add entries for websocket upgrades

jeaaustx:
Record audit entries for websocket upgrades. This includes the paths /bmc-console, /console0 (host), and /console1 (hypervisor).

In order to support the websocket upgrades the existing audit utility function was split into two pieces. auditEntry() continues to be the entry point for Redfish request structures. A new utility function auditConnection() is added for the entry point for the websocket connections. A new writeEvent() function is common between the two which does the actual recording of the audit event.

Note: The audit support is only for upgrades of the websocket. The user access validation is completed before the upgrade. If the user access validation fails no audit log entry will be written.

Note: Direct ssh to the ports used by the consoles are audited through the existing OpenSSH audit entries.

Tested: Open consoles as different users and observed audit entries being recorded.

```
// Service user
type=USYS_CONFIG msg=audit(10/06/25 19:51:35.269:40) : pid=2156 uid=root auid=service ses=8 msg='op=/console0 {"NewSocket":true} acct=service exe=/tmp/bmcwebd hostname=p11bmc addr=10.0.2.1 terminal=? res=success'

type=USYS_CONFIG msg=audit(10/06/25 19:51:41.719:41) : pid=2156 uid=root auid=service ses=8 msg='op=/bmc-console {"NewSocket":true} acct=service exe=/tmp/bmcwebd hostname=p11bmc addr=10.0.2.1 terminal=? res=success'

type=USYS_CONFIG msg=audit(10/06/25 19:51:41.789:42) : pid=2156 uid=root auid=service ses=8 msg='op=/console1 {"NewSocket":true} acct=service exe=/tmp/bmcwebd hostname=p11bmc addr=10.0.2.1 terminal=? res=success'

// Service user doing another hypervisor connection
type=USYS_CONFIG msg=audit(10/06/25 19:51:47.199:44) : pid=2156 uid=root auid=service ses=8 msg='op=/console1 {"NewSocket":false} acct=service exe=/tmp/bmcwebd hostname=p11bmc addr=10.0.2.1 terminal=? res=success'

// Non-service user can only access host console.
type=USYS_CONFIG msg=audit(10/06/25 19:51:59.979:48) : pid=2156 uid=root auid=service ses=8 msg='op=/console0 {"NewSocket":true} acct=janet exe=/tmp/bmcwebd hostname=p11bmc addr=10.0.2.1 terminal=? res=success'

/* Logged entries are returned from Redfish query */
curl -k -H "X-Auth-Token: $token"   https://${BMC_IP}/redfish/v1/Systems/system/LogServices/AuditLog/Entries?\$top=10
{
...
    {
      "@odata.id": "/redfish/v1/Systems/system/LogServices/AuditLog/Entries/1759780319.979:48",
      "@odata.type": "#LogEntry.v1_9_0.LogEntry",
      "EntryType": "Event",
      "EventTimestamp": "2025-10-06T19:51:59+00:00",
      "Id": "1759780319.979:48",
      "Message": "type=\"USYS_CONFIG\" op=\"/console0\" acct=\"janet\" exe=\"/tmp/bmcwebd\" hostname=\"p11bmc\" addr=\"10.0.2.1\" terminal=\"?\" res=\"success\"",
      "MessageArgs": [
        "USYS_CONFIG",
        "/console0",
        "janet",
        "/tmp/bmcwebd",
        "p11bmc",
        "10.0.2.1",
        "?",
        "success"
      ],
      "MessageId": "OpenBMC.0.5.AuditLogUsysConfig",
      "Name": "Audit Log Entry",
      "Oem": {
        "IBM": {
          "@odata.type": "#IBMLogEntryAttachment.v1_0_0.IBM",
          "AdditionalDataFullAuditLogURI": "/redfish/v1/Systems/system/LogServices/AuditLog/Entries/FullAudit/attachment"
        }
      }
    },
...
    {
      "@odata.id": "/redfish/v1/Systems/system/LogServices/AuditLog/Entries/1759780301.789:42",
      "@odata.type": "#LogEntry.v1_9_0.LogEntry",
      "EntryType": "Event",
      "EventTimestamp": "2025-10-06T19:51:41+00:00",
      "Id": "1759780301.789:42",
      "Message": "type=\"USYS_CONFIG\" op=\"/console1\" acct=\"service\" exe=\"/tmp/bmcwebd\" hostname=\"p11bmc\" addr=\"10.0.2.1\" terminal=\"?\" res=\"success\"",
      "MessageArgs": [
        "USYS_CONFIG",
        "/console1",
        "service",
        "/tmp/bmcwebd",
        "p11bmc",
        "10.0.2.1",
        "?",
        "success"
      ],
      "MessageId": "OpenBMC.0.5.AuditLogUsysConfig",
      "Name": "Audit Log Entry",
      "Oem": {
        "IBM": {
          "@odata.type": "#IBMLogEntryAttachment.v1_0_0.IBM",
          "AdditionalDataFullAuditLogURI": "/redfish/v1/Systems/system/LogServices/AuditLog/Entries/FullAudit/attachment"
        }
      }
    },
    {
      "@odata.id": "/redfish/v1/Systems/system/LogServices/AuditLog/Entries/1759780301.719:41",
      "@odata.type": "#LogEntry.v1_9_0.LogEntry",
      "EntryType": "Event",
      "EventTimestamp": "2025-10-06T19:51:41+00:00",
      "Id": "1759780301.719:41",
      "Message": "type=\"USYS_CONFIG\" op=\"/bmc-console\" acct=\"service\" exe=\"/tmp/bmcwebd\" hostname=\"p11bmc\" addr=\"10.0.2.1\" terminal=\"?\" res=\"success\"",
      "MessageArgs": [
        "USYS_CONFIG",
        "/bmc-console",
        "service",
        "/tmp/bmcwebd",
        "p11bmc",
        "10.0.2.1",
        "?",
        "success"
      ],
      "MessageId": "OpenBMC.0.5.AuditLogUsysConfig",
      "Name": "Audit Log Entry",
      "Oem": {
        "IBM": {
          "@odata.type": "#IBMLogEntryAttachment.v1_0_0.IBM",
          "AdditionalDataFullAuditLogURI": "/redfish/v1/Systems/system/LogServices/AuditLog/Entries/FullAudit/attachment"
        }
      }
    },

```

### DIFF
--- a/include/obmc_hypervisor.hpp
+++ b/include/obmc_hypervisor.hpp
@@ -1,6 +1,9 @@
 #pragma once
 
+#include "bmcweb_config.h"
+
 #include "app.hpp"
+#include "audit_events.hpp"
 #include "io_context_singleton.hpp"
 #include "logging.hpp"
 #include "websocket.hpp"
@@ -116,9 +119,19 @@ inline void connectHandler(const boost::system::error_code& ec)
         BMCWEB_LOG_ERROR("Couldn't connect to host serial port: {}", ec);
         for (crow::websocket::Connection* session : sessions)
         {
+            if constexpr (BMCWEB_AUDIT_EVENTS)
+            {
+                audit::auditConnection(session, true, false);
+            }
             session->close("Error in connecting to host port");
         }
         return;
+    }
+
+    // Audit each session. If already audited no new entry will be created
+    for (crow::websocket::Connection* session : sessions)
+    {
+        audit::auditConnection(session, true, true);
     }
 
     doWrite();
@@ -143,6 +156,12 @@ inline void requestRoutes(App& app)
                     boost::asio::local::stream_protocol::socket>(
                     getIoContext());
                 hostSocket->async_connect(ep, connectHandler);
+                return;
+            }
+
+            if constexpr (BMCWEB_AUDIT_EVENTS)
+            {
+                audit::auditConnection(&conn, false, true);
             }
         })
         .onclose([](crow::websocket::Connection& conn,
--- a/include/obmc_shell.hpp
+++ b/include/obmc_shell.hpp
@@ -1,6 +1,9 @@
 #pragma once
 
+#include "bmcweb_config.h"
+
 #include "app.hpp"
+#include "audit_events.hpp"
 #include "io_context_singleton.hpp"
 #include "logging.hpp"
 #include "websocket.hpp"
@@ -70,6 +73,10 @@ class Handler : public std::enable_shared_from_this<Handler>
             // ERROR
             if (session != nullptr)
             {
+                if constexpr (BMCWEB_AUDIT_EVENTS)
+                {
+                    audit::auditConnection(session, true, false);
+                }
                 session->close("Error creating child process for login shell.");
             }
             return;
@@ -99,6 +106,10 @@ class Handler : public std::enable_shared_from_this<Handler>
 
             // for io operation assing file discriptor
             // to boost stream_descriptor
+            if constexpr (BMCWEB_AUDIT_EVENTS)
+            {
+                audit::auditConnection(session, true, true);
+            }
             streamFileDescriptor.assign(ttyFileDescriptor);
             doWrite();
             doRead();
@@ -206,6 +217,12 @@ inline void onOpen(crow::websocket::Connection& conn)
         {
             std::get<0>(insertData)->second->connect();
         }
+        return;
+    }
+
+    if constexpr (BMCWEB_AUDIT_EVENTS)
+    {
+        audit::auditConnection(&conn, false, true);
     }
 }
 
--- a/src/audit_events.cpp
+++ b/src/audit_events.cpp
@@ -2,6 +2,7 @@
 
 #include "http_request.hpp"
 #include "logging.hpp"
+#include "websocket.hpp"
 
 #include <audit-records.h>
 #include <audit_logging.h>
@@ -173,17 +174,14 @@ bool appendItemToBuf(std::string& strBuf, size_t maxBufSize,
     return true;
 }
 
-void auditEvent(const crow::Request& req, const std::string& userName,
+void writeEvent(const std::string& opPath, const std::string& detail,
+                const std::string& userName, const std::string& ipAddress,
                 bool success)
 {
     if (!auditOpen())
     {
         return;
     }
-
-    std::string opPath =
-        std::format("op={}:{} ", std::string(req.methodString()),
-                    std::string(req.target()));
 
     size_t maxBuf = 256; // Limit message to avoid filling log with single entry
     std::string cnfgBuff = opPath.substr(0, maxBuf);
@@ -194,14 +192,6 @@ void auditEvent(const crow::Request& req, const std::string& userName,
         BMCWEB_LOG_WARNING(
             "Audit buffer too small, truncating: cnfgBufLen={} opPathLen={}",
             cnfgBuff.length(), opPath.length());
-    }
-
-    // Determine any additional info for the event
-    std::string detail;
-    if (wantDetail(req))
-    {
-        detail = req.body().substr(0, maxBuf);
-        detail += " ";
     }
 
     if (!detail.empty())
@@ -252,8 +242,6 @@ void auditEvent(const crow::Request& req, const std::string& userName,
         (maxBuf - cnfgBuff.length()), opPath.length(), detail.length(),
         userLen);
 
-    std::string ipAddress = req.ipAddress.to_string();
-
     int rc = audit_log_user_message(
         auditfd, AUDIT_USYS_CONFIG, cnfgBuff.c_str(),
         boost::asio::ip::host_name().c_str(), ipAddress.c_str(), nullptr,
@@ -277,6 +265,54 @@ void auditEvent(const crow::Request& req, const std::string& userName,
             BMCWEB_LOG_ERROR("Error writing audit message: {}", origErrno);
         }
     }
+}
+
+void auditEvent(const crow::Request& req, const std::string& userName,
+                bool success)
+{
+    std::string opPath =
+        std::format("op={}:{} ", std::string(req.methodString()),
+                    std::string(req.target()));
+
+    // Determine any additional info for the event, limit length
+    std::string detail;
+    if (wantDetail(req))
+    {
+        detail = req.body().substr(0, 256);
+        detail += " ";
+    }
+
+    std::string ipAddress = req.ipAddress.to_string();
+
+    writeEvent(opPath, detail, userName, ipAddress, success);
+}
+
+void auditConnection(crow::websocket::Connection* conn, bool isNewSocket,
+                     bool success)
+{
+    if ((conn == nullptr) || !conn->needsAudit())
+    {
+        // Connection does not need auditing
+        return;
+    }
+
+    std::string opPath = std::format("op={} ", conn->url().path());
+    std::string isNewStr;
+    if (isNewSocket)
+    {
+        isNewStr = "true";
+    }
+    else
+    {
+        isNewStr = "false";
+    }
+    std::string detail = std::format("{{\"NewSocket\":{}}} ", isNewStr);
+    std::string userName = conn->getUserName();
+    std::string ipAddress = conn->clientIp();
+
+    BMCWEB_LOG_DEBUG("Auditing connection {} isNew: {}", opPath, isNewStr);
+    writeEvent(opPath, detail, userName, ipAddress, success);
+    conn->clearAudit();
 }
 
 } // namespace audit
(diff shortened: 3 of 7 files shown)


joseph-reynolds (commented):
The code looks good to me.  It seems okay, except for the test results where the audit record for a user not allowed to use /console0 is marked as res=success when it should be res=failed.  But I did not find and changed code where we would get this behavior.

include/obmc_hypervisor.hpp
```diff
@@ -111,6 +114,21 @@ inline void doRead()
...
+            success = false;
+        }
+
+        // Audit each session. If already audited no new entry will be created
+        for (crow::websocket::Connection* session : sessions)
+        {
+            audit::auditConnection(session, true, success);
+        }
+    }
+
```
baemyung: Would it be better to put this into the current logic flow - L132-L141?

```
    if (ec)
    {
        BMCWEB_LOG_ERROR("Couldn't connect to host serial port: {}", ec);
        for (crow::websocket::Connection* session : sessions)
        {
            audit::auditConnection(session, true, false);
            session->close("Error in connecting to host port");
        }
        return;
    }

    //
    for (crow::websocket::Connection* session : sessions)
    {
        audit::auditConnection(session, true, true);
    }

    doWrite();
    doRead();
```

include/obmc_hypervisor.hpp
```diff
@@ -144,6 +162,10 @@ inline void requestRoutes(App& app)
                     getIoContext());
                 hostSocket->async_connect(ep, connectHandler);
             }
+            else if constexpr (BMCWEB_AUDIT_EVENTS)
+            {
+                audit::auditConnection(&conn, false, true);
+            }
```
baemyung, line 165: How about this from L155?

```
            if (hostSocket == nullptr)
            {
                const std::string consoleName("\0obmc-console.hypervisor", 24);
                boost::asio::local::stream_protocol::endpoint ep(consoleName);
                hostSocket = std::make_unique<
                    boost::asio::local::stream_protocol::socket>(
                    getIoContext());
                return;      <=== ADD
            }
            
            if constexpr (BMCWEB_AUDIT_EVENTS) 
            {
                audit::auditConnection(&conn, false, true);
            }
```

include/obmc_shell.hpp
```diff
@@ -207,6 +218,10 @@ inline void onOpen(crow::websocket::Connection& conn)
             std::get<0>(insertData)->second->connect();
         }
     }
+    else if constexpr (BMCWEB_AUDIT_EVENTS)
+    {
+        audit::auditConnection(&conn, false, true);
+    }
```
baemyung, line 226: Similarly,

```
    if (it == mapHandler.end())
    {
        auto insertData =
            mapHandler.emplace(&conn, std::make_shared<Handler>(&conn));
        if (std::get<bool>(insertData))
        {
            std::get<0>(insertData)->second->connect();
        }
        return;        // <== ADD
    }
    
    if constexpr (BMCWEB_AUDIT_EVENTS)
    {
        audit::auditConnection(&conn, false, true);
    }
```

src/audit_events.cpp
```diff
@@ -279,4 +267,49 @@ void auditEvent(const crow::Request& req, const std::string& userName,
...
+
+    std::string ipAddress = req.ipAddress.to_string();
+
+    writeEvent(opPath, detail, userName, ipAddress, success);
+}
+
+void auditConnection(crow::websocket::Connection* conn, bool isNewSocket,
+                     bool success)
+{
+    if ((conn != nullptr) && conn->needsAudit())
```
baemyung: Would it be better to short cut the case first where the audit is not needed like?

```
    if ((conn == nullptr) && !conn->needsAudit())
    {
         return;     // No conn or no need to audit
    }
     ...
```

jeaaustx:
> The code looks good to me. It seems okay, except for the test results where the audit record for a user not allowed to use /console0 is marked as res=success when it should be res=failed. But I did not find and changed code where we would get this behavior.

Oh, my commit message was worded poorly. The test doesn't actually log success on this case. Instead no audit entry is written at all. The privilege check is happening outside and not even getting to the websocket upgrade code paths. (So the commit message was showing the debug log messages showing the privilege check failures outside this path and no audit logging.)

I've looked into how support could be added to add the audit log when the privilege check fails for the user. The change necessary would be to add the console paths to the check for operations doing audit. [1]

Adding these checks would have a performance impact. The websocket upgrades are all GET requests. This would mean the additional check of the string comparison to the three console paths would be required for any GET request completion path. Currently all of the GET requests are ruled out of audit logging by just looking at their verb. To alleviate the performance impact I believe a larger change to the auditing of all entries would have to be made to alter how the determination of the records is determined.

Given this affect is it acceptable to not add the audit log entries for unauthorized users or 1120? Can a story be created to address this in the future?

[1] https://github.com/ibm-openbmc/bmcweb/blob/d37c76feceb63595d3b516e5c53c09714dee092c/include/audit_events.hpp#L41

joseph-reynolds:
> > The code looks good to me. It seems okay, except for the test results where the audit record for a user not allowed to use /console0 is marked as res=success when it should be res=failed. But I did not find and changed code where we would get this behavior.
> 
> ...
> I've looked into how support could be added to add the audit log when the privilege check fails for the user. The change necessary would be to add the console paths to the check for operations doing audit. [1]
> ...
> Given this affect is it acceptable to not add the audit log entries for unauthorized users or 1120? Can a story be created to address this in the future?

Yes.  The hard requirement is to audit new connections.  So I am okay with this.  The BMC **should** audit authorization failures, but that is a different set of requirements.

joseph-reynolds (approved):
The behavior looks okay to me: Audit record are written for all new connections.

include/obmc_hypervisor.hpp
jeaaustx: Yes that is cleaner. This file isn't upstream so trying to keep all the audit stuff together isn't a worry here.

include/obmc_hypervisor.hpp
jeaaustx: Yes that is more typical of code flow of this repo. Thank you for pointing it out. I've made this change also.

include/obmc_shell.hpp
jeaaustx: Yes thank you that is better, I've changed it.

src/audit_events.cpp
jeaaustx: Yes that is cleaner, than kyou I've made that change.

joseph-reynolds (approved):

include/obmc_shell.hpp
```diff
@@ -99,6 +106,10 @@ class Handler : public std::enable_shared_from_this<Handler>
 
             // for io operation assing file discriptor
             // to boost stream_descriptor
+            if constexpr (BMCWEB_AUDIT_EVENTS)
```
gtmills, line 109: You don't need anything in the CHILD ? Line 95/99, a bit of a corner case

include/obmc_shell.hpp
```diff
@@ -70,6 +73,10 @@ class Handler : public std::enable_shared_from_this<Handler>
             // ERROR
             if (session != nullptr)
             {
+                if constexpr (BMCWEB_AUDIT_EVENTS)
```
gtmills, line 76: Why not putting these in the doClose not work ? Does everything not run through the doClose?

jeaaustx: Oh, good idea! I'll take a look. The success case should not wait for the doClose() to log though, only the failure cases. But the boolean indicator in the Connection itself will prevent a double logging on closing the successful open case.

include/obmc_shell.hpp
jeaaustx: Hmm, I was thinking not. I thought might end up with double logging one for the parent and one for child. (Since the forkpty() results in the child and parent running in separate memory spaces, the session boolean wouldn't prevent it.)

I was thinking the socket creation was sufficient, not whether the socket itself ran into problems. I'll experiment adding in both child and parent to see what happens.

include/obmc_shell.hpp
gtmills: Ack. Give it a try, if not, I'm good with this.

gtmills (approved):

include/obmc_shell.hpp
jeaaustx: I tried this but ran into troubles. The doClose is not called immediately on the failure of the websocket upgrade. It waits for a timeout of the doRead(). (See [acceptDone()](https://github.com/ibm-openbmc/bmcweb/blob/d37c76feceb63595d3b516e5c53c09714dee092c/http/websocket_impl.hpp#L206). The openHandler() does not return success/fail so it will always try the doRead(). That seems to wait for the timeout.).

So I'm going to leave this as-is for now.

PR #1377, opened 2025-10-27
Move connection count requests to warning

gtmills:
WIP: A possible solution to easily debug GET requests. Move to WARNING log level 

Up the tracing for requests and ip addresses and the connection count to
warning. This will help debug problems with running out of connections.
Since log levels are runtime now, can move to warning without filling
the journal with a ton of traces.

Also pull in Myung's subscription added/deleted but move it to WARNING.

### DIFF
--- a/http/http_connection.hpp
+++ b/http/http_connection.hpp
@@ -90,7 +90,7 @@ class Connection :
 
         connectionCount++;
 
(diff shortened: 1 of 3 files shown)


baemyung (approved):
Agreed. It looks good

gtmills:
image

jeaaustx (approved):

PR #1385, opened 2025-12-01
WIP:Switch between peer BMC mTLS and https SSL context

priyair:
WIP:Switch between peer BMC mTLS and https SSL context

This commit switches between mTLS and https ssl context, based upon the client cert provided. When client certificate is provided the bmcweb switched to mTLS context, else use https SSL context. Also supports the password fallback mechanism when tlsStrict and TLS is enabled.

Tested By

Provided client cert and bmcweb switched to mTLs context

Provided password or session token the bmcweb switched to https SSL context

### DIFF
--- a/src/ssl_key_handler.cpp
+++ b/src/ssl_key_handler.cpp
@@ -15,6 +15,7 @@
 #include <boost/beast/core/file_posix.hpp>
 #include <boost/system/error_code.hpp>
 #include <fstream>
+
 extern "C"
 {
 #include <nghttp2/nghttp2.h>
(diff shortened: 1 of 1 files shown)


gtmills:
Why do we need this? What does this do? Can we go upstream with this ?

raviteja-b:
> Why do we need this? What does this do? Can we go upstream with this ?

When customer uploads new CA signed https certificate, our mtls certs should not impact https traffic with CA signed certificate, so we are trying ssl context switching.
we are still working on this patch, will propose it upstream once its ready

PR #1386, opened 2025-12-03
1120: Fix CpuCore Health value

baemyung:
Commit 788a0bd4[1] improperly follow CpuCore's state/health like in power supply's state [2].
   - Health=Critical if Available = false or Functional = false.

- However, in the previous releases [3], Health state is based on only Functional state.
  - Health=Critical if Functional = false.

So, this commit is to fix CpuCore's state/health as in the previous release [3].

Tested:

- Under Functional=true but Available=false, GET core.

```
busctl get-property xyz.openbmc_project.PLDM  /xyz/openbmc_project/inventory/system/chassis/motherboard/dcm0/cpu0/core10 xyz.openbmc_project.State.DecoratorAvailability   Available
b false

busctl get-property xyz.openbmc_project.PLDM  /xyz/openbmc_project/inventory/system/chassis/motherboard/dcm0/cpu0/core10 xyz.openbmc_project.State.Decorator.OperationalStatus  Functional
b true
```

- GET /redfish/v1/Systems/system/Processors/<cpu>/SubProcessors/<core>

Before Fix:
```
curl -k -X GET https://${bmc}/redfish/v1/Systems/system/Processors/dcm0-cpu0/SubProcessors/core10
{
  "@odata.id": "/redfish/v1/Systems/system/Processors/dcm0-cpu0/SubProcessors/core10",
  "@odata.type": "#Processor.v1_18_0.Processor",
  "Enabled": false,
  "Id": "core10",
  "Name": "core10",
  "Status": {
    "Health": "Critical",
    "State": "UnavailableOffline"
  }
}
```

After Fix:
```
{
  "@odata.id": "/redfish/v1/Systems/system/Processors/dcm0-cpu0/SubProcessors/core10",
  "@odata.type": "#Processor.v1_18_0.Processor",
  "Enabled": false,
  "Id": "core10",
  "Name": "core10",
  "Status": {
    "Health": "OK",
    "State": "UnavailableOffline"
  }
}
```

[1] https://github.com/ibm-openbmc/bmcweb/commit/788a0bd4
[2] https://github.com/ibm-openbmc/bmcweb/commit/71f0ad1ec1a4db9832ef0ed8f7b629359a488b8a
[3] https://github.com/ibm-openbmc/bmcweb/pull/991

### DIFF
--- a/redfish-core/lib/processor.hpp
+++ b/redfish-core/lib/processor.hpp
@@ -1002,13 +1002,12 @@ inline void getProcessorPaths(
 
 inline void getSubProcessorsCoreHealth(
     const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
-    const std::string& service, const std::string& objPath, bool available)
+    const std::string& service, const std::string& objPath)
 {
     dbus::utility::getProperty<bool>(
         service, objPath,
         "xyz.openbmc_project.State.Decorator.OperationalStatus", "Functional",
-        [asyncResp,
-         available](const boost::system::error_code& ec, bool functional) {
+        [asyncResp](const boost::system::error_code& ec, bool functional) {
             if (ec)
             {
                 if (ec.value() != EBADR)
@@ -1019,7 +1018,7 @@ inline void getSubProcessorsCoreHealth(
                 return;
             }
 
-            if (!functional || !available)
+            if (!functional)
             {
                 asyncResp->res.jsonValue["Status"]["Health"] =
                     resource::Health::Critical;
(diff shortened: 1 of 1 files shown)


baemyung:
Defect: https://jazz07.rchland.ibm.com:13443/jazz/web/projects/CSSD#action=com.ibm.team.workitem.viewWorkItem&id=753561

redfish-core/lib/processor.hpp
```diff
@@ -1019,67 +1018,71 @@ inline void getSubProcessorsCoreHealth(
...
-                    BMCWEB_LOG_ERROR("DBUS response error for Available {}",
-                                     ec.value());
-                    messages::internalError(asyncResp->res);
-                }
-                return;
-            }
-            getSubProcessorsCoreHealth(asyncResp, service, objPath, available);
-            getSubProcessorsCoreState(asyncResp, service, objPath, available);
-        });
+        std::bind_front(afterGetSubProcessorsCoreAvailable, asyncResp, service,
```
jeaaustx, line 1084: nit: For consistency in naming should this be named `afterGetSubProcessorsCoreState` instead?

jeaaustx (approved):
Just one minor nit comment, but otherwise looks good to me.

redfish-core/lib/processor.hpp
baemyung: The code has 2-level depths of lambda calls
```
getSubProcessorsCoreState()
   -> getProperty("Available") [] { afterGetSubProcessorsCoreAvailable(); }
   -> getProperty("Present") [] { afterGetSubProcessorsCorePresent(); }
```

So, if we name afterGetSubProcessorsCoreAvailable() as  afterGetSubProcessorsCoreState(), the second function name may have double `after` like `afterAfterGetSubProcessorsCoreState()`?

Instead, I think it may be better to use `after` of `getProperty(<Name>)` .
What do you think?

jeaaustx: Seems reasonable. It is more complicated since it has the nested getProperty() calls for filling the single field.

(Mainly I was pointing it out because I've had Ed comment on function names like this upstream.)

baemyung: Another way to avoid the nested getProperty may be doing like what's in 1110.

- get ` dbus::utility::getManagedObjects()` instead of `getSubTree()` inside  `getSubProcessorsCoreObject()` or by another function -  https://github.com/ibm-openbmc/bmcweb/blob/1120/redfish-core/lib/processor.hpp#L1205

- parse `ManagedObjectType` for "Available" & "Present" ... similarly in 1110 - https://github.com/ibm-openbmc/bmcweb/blob/1110/redfish-core/lib/processor.hpp#L293

However, I think it can be done later - when we pursue more CpuCore stuffs to upstreaming.

jeaaustx: Agree, I am fine with this as it is now.

gtmills (approved):

PR #1388, opened 2025-12-16
Mfa defect fixes

asmithakarun:
This PR contains the below changes:
1. Restrict secret key generation after MFA setup
2. Disallow RO users from clearing own secretkey

It fixes the following defects:
[756703](https://jazz07.rchland.ibm.com:13443/jazz/web/projects/CSSD#action=com.ibm.team.workitem.viewWorkItem&id=756703)

[756701](https://jazz07.rchland.ibm.com:13443/jazz/web/projects/CSSD#action=com.ibm.team.workitem.viewWorkItem&id=756701)

### DIFF
--- a/redfish-core/lib/account_service_mfa_actions.hpp
+++ b/redfish-core/lib/account_service_mfa_actions.hpp
@@ -27,6 +27,42 @@
 namespace redfish
 {
 
+inline void checkAndCreateSecretKey(
+    const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
+    const std::string& username, const std::string& userPath,
+    const boost::system::error_code& ec, bool isSecretKeySet)
+{
+    if (ec)
+    {
+        BMCWEB_LOG_ERROR("D-Bus response error: {}", ec.value());
+        messages::internalError(asyncResp->res);
+        return;
+    }
+
+    if (isSecretKeySet)
+    {
+        BMCWEB_LOG_WARNING("Secret key is already setup for the user");
+        messages::actionNotSupported(asyncResp->res, "GenerateSecretKey");
+        return;
+    }
+
+    // If secret key is not set, then proceed to create it
+    dbus::utility::async_method_call(
+        [asyncResp, username](const boost::system::error_code& ec1,
+                              const std::string& secretKey) {
+            if (ec1)
+            {
+                BMCWEB_LOG_ERROR("D-Bus response error: {}", ec1.value());
+                messages::internalError(asyncResp->res);
+                return;
+            }
+
+            asyncResp->res.jsonValue["SecretKey"] = secretKey;
+        },
+        "xyz.openbmc_project.User.Manager", userPath,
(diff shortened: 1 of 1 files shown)


raviteja-b (approved):

baemyung:
Should this be upstreamed, as a new commit, or on top of the existing gerrit request commit?
Either way, please add Change-Id or gerrit link.

redfish-core/lib/account_service_mfa_actions.hpp
```diff
@@ -31,19 +31,43 @@ inline void createSecretKeyUtil(
...
 {
-    dbus::utility::async_method_call(
-        [asyncResp, username](const boost::system::error_code& ec,
-                              const std::string& secretKey) {
+    dbus::utility::getProperty<bool>(
+        "xyz.openbmc_project.User.Manager", userPath,
+        "xyz.openbmc_project.User.TOTPAuthenticator", "SecretKeyIsValid",
+        [asyncResp, username, userPath](const boost::system::error_code& ec,
+                                        const bool isSecretKeySet) {
             if (ec)
```
baemyung: It would be better to make this long lambda into a separate function.

baemyung (approved):
With some comments, I agree to include into 1120.

redfish-core/lib/account_service_mfa_actions.hpp
```diff
@@ -323,7 +347,7 @@ inline void requestAccountServiceMFARoutes(App& app)
         // TODO this privilege should be using the generated endpoints, but
         // because of the special handling of ConfigureSelf, it's not able to
         // yet
-        .privileges({{"ConfigureUsers"}, {"ConfigureSelf"}})
```
gtmills, line 326: It isn't valid for a ReadOnly user to generate a new secret key ?

gtmills: If a ReadOnly user wants a new secret key, how do they get it?

asmithakarun: It is not allowed. RO user should contact admin incase of secret key reset, for them generate another secret key.

GenerateSecretKey action is allowed only if: 1. No existing secret key, 2. MFA is not yet verified for the user.

asmithakarun:
> Should this be upstreamed, as a new commit, or on top of the existing gerrit request commit? Either way, please add Change-Id or gerrit link.

There is a commit upstream, and this change was there in previous patchsets.
Got comment from Ed to remove these "SecretKeyIsValid" checks from bmcweb while generating secret key, and have logic in the backend. So the latest patchset will have this change removed.

This PR matches with whats is there in 1110 for now, and once we have ported those checks to the backend, I will revert this commit.

redfish-core/lib/account_service_mfa_actions.hpp
asmithakarun: done

PR #1395, opened 2026-02-10
Dynamic SSL/TLS context switching to support aggregation between peer bmc

priyair:
The implementation maintains two separate SSL contexts:
    HTTPS context — standard TLS with optional client certificate.
    mTLS context — mutual TLS requiring mandatory client certificate authentication.
    The server selects the appropriate context based on the
    client’s Server Name Indication (SNI).
    This enables secure peer‑to‑peer Redfish aggregation
    while continuing to support standard HTTPS clients.

    Tested By:
        '''

            Configured peer BMC aggregation using JSON configuration files.
            Verified requests from client BMCs with TLS strict mode both enabled and disabled.
            Tested with client certificates and authentication tokens.
            HTTPS and mTLS connections are correctly established and switched based on SNI.

            curl -k -H "X-Auth-Token: $bmc_token24" -X GET https://<bmc>/redfish/v1/Systems
            tlsStrict false

            curl -k -H "X-Auth-Token: $bmc_token26"   -X \
                    POST https://<bmc26>/redfish/v1/bmc/Truststore/Certificates \
                      -H "Content-Type: application/json"  \
                     -d '{"CertificateType": "PEM", "CertificateString": "'"$CERT_STRING"'"}'

            curl -k -H "X-Auth-Token: $bmc_token26"   -X \
                    POST https://<bmc26>/redfish/v1/bmc/CertificateService.ReplaceCertificate \
                      -H "Content-Type: application/json"   -d '{
                        "CertificateType": "PEM",
                        "CertificateString": "'"$CERT_STRING"'",
                        "CertificateUri": {
                         "@odata.id": "/redfish/v1/HTTPS/Certificates/1"
                            }
                    }'

            curl -v --cert client.pem --cacert ../serverCA/ca.crt https://<bmc>/redfish/v1/Systems
            tlsStrict false

            curl -k -H "X-Auth-Token: $bmc_token24" -X GET https://<bmc>/redfish/v1/Systems
            tlsStrict true

            curl -v --cert client.pem --cacert ../serverCA/ca.crt https://<bmc>/redfish/v1/Systems
            tlsStrict true
            '''
        '''

Change-Id: I49c8f3d69d4b91a3a1ebe9b5f001943d5f1f47f8

### DIFF
--- a/redfish-core/lib/ethernet.hpp
+++ b/redfish-core/lib/ethernet.hpp
@@ -1050,7 +1050,11 @@ inline void handleIPv6DefaultGateway(
                 return;
             }
             deleteIPv6Gateway(ifaceId, staticGatewayEntry->id, asyncResp);
-            return;
+            /*Fix: When request body contains null, 
+	     * skip and continue to next parameter*/
+            staticGatewayEntry++;
+            entryIdx++;
+            continue;
         }
         if (obj->empty())
         {


src/ssl_key_handler.cpp
```diff
@@ -459,6 +466,25 @@ static std::string ensureCertificate()
     return ensuressl::ensureOpensslKeyPresentAndValid(sslPemFile);
 }
 
+static std::string ensureMtlsCertificate()
```
abhilashraju: why would we create certificate internally? The certificates are supposed to be created by other app(probably spdm) in the first boot . The location of the certificate will be a TPM. For time being we can use directory used by provisining server_cert.pem and server_key.pem

gtmills:
Can you add the upstream review? There is more experts here upstream

src/ssl_key_handler.cpp
```diff
@@ -48,6 +49,12 @@ extern "C"
 namespace ensuressl
 {
 
+namespace crow
+{
+
+std::shared_ptr<boost::asio::ssl::context> g_httpsCtx = nullptr;
```
abhilashraju: can we try not using globals?

src/ssl_key_handler.cpp
```diff
@@ -533,25 +582,187 @@ static bool getSslContext(boost::asio::ssl::context& mSslContext,
...
+
+    const unsigned char *sni = nullptr;
+    size_t sniLen = 0;
+    if (!ssl){
+        BMCWEB_LOG_ERROR("no ssl return");
+        return SSL_CLIENT_HELLO_ERROR;
+    }
+
+    // Determine if this should be mTLS based on SNI
+    if (SSL_client_hello_get0_ext(ssl, TLSEXT_TYPE_server_name,
```
abhilashraju: SSL_get_servername(s, TLSEXT_NAMETYPE_host_name) will not simplify the code?

src/ssl_key_handler.cpp
```diff
@@ -533,25 +582,187 @@ static bool getSslContext(boost::asio::ssl::context& mSslContext,
...
+
 std::shared_ptr<boost::asio::ssl::context> getSslServerContext()
 {
-    boost::asio::ssl::context sslCtx(boost::asio::ssl::context::tls_server);
+    using namespace boost::asio::ssl;
+    namespace fs = std::filesystem;
+
+    // Allocate HTTPS and MTLS context as shared_ptr
+    auto httpsCtx = std::make_shared<context>(context::tls_server);
+    auto mtlsCtx = std::make_shared<context>(context::tls_server);
```
abhilashraju: why do we create this object always?

src/ssl_key_handler.cpp
```diff
@@ -564,41 +775,84 @@ std::shared_ptr<boost::asio::ssl::context> getSslServerContext()
...
-    }
-
-    boost::system::error_code ec;
-    sslCtx.set_verify_mode(mode, ec);
-    if (ec)
-    {
-        BMCWEB_LOG_DEBUG("Failed to set verify mode {}", ec.message());
-        return nullptr;
-    }
+        BMCWEB_LOG_DEBUG("Setting verify peer only forward_unauthorized::hasWebuiRoute");
```
abhilashraju: why do we need these changes?

src/ssl_key_handler.cpp
priyair: SSL_get_servername() returns NULL in ClientHello callback. 
 
we are handling SNI early to select a context before handshake, hence SSL_client_hello_get0_ext() is used inside the ClientHello callback.
SSL_client_hello_get0_ext() works inside the ClientHello callback (before parsing) and thus returning the valid SNI.

src/ssl_key_handler.cpp
priyair: Added structure to hold both SSL contexts

manojkiraneda:
@priyair @raviteja-b  - is there an upstream commit for this ?`host-interface` squad is working on adding the tpm support to bmcweb based on a design documented here at https://gerrit.openbmc.org/c/openbmc/docs/+/77874. I guess we should probably build on top of your patch to avoid any reworks.

raviteja-b:
> @priyair @raviteja-b - is there an upstream commit for this ?`host-interface` squad is working on adding the tpm support to bmcweb based on a design documented here at https://gerrit.openbmc.org/c/openbmc/docs/+/77874. I guess we should probably build on top of your patch to avoid any reworks.

@manojkiraneda we will push these changes upstream in sometime, @priyair is working on tls-strict handling changes  needed for upstream. will update upstream commit in this PR once changes ready  for upstream

priyair (commented):
closing the PR as branch name is similar and file got overwritten, hence Dynamic SSL/TLS context switching for Mtls context aggregation between peer bmc #1401

priyair:
> Can you add the upstream review? There is more experts here upstream

https://gerrit.openbmc.org/c/openbmc/bmcweb/+/87711

PR #1405, opened 2026-02-25
Redfish: Refactor StaticNameServers to use setDbusProperty helper

abhilashraju:
Replace direct sdbusplus::asio::setProperty call with the setDbusProperty helper function in handleStaticNameServersPatch. This simplifies error handling and improves code consistency by using the common utility function.

Changes:
- Use setDbusProperty helper instead of raw sdbusplus call
- Remove explicit error handling callback (now handled by helper)
- Construct object_path using operator/ for better readability

Tested: Verified StaticNameServers PATCH operations work correctly

### DIFF
--- a/redfish-core/lib/ethernet.hpp
+++ b/redfish-core/lib/ethernet.hpp
@@ -1731,18 +1731,12 @@ inline void handleStaticNameServersPatch(
     const std::vector<std::string>& updatedStaticNameServers,
     const std::shared_ptr<bmcweb::AsyncResp>& asyncResp)
 {
-    sdbusplus::asio::setProperty(
-        *crow::connections::systemBus, "xyz.openbmc_project.Network",
-        "/xyz/openbmc_project/network/" + ifaceId,
+    setDbusProperty(
+        asyncResp, "xyz.openbmc_project.Network",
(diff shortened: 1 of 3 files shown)


gtmills:
Cool! Why do we need this for 1110 ? 
Does this fix an error ? What is upsteam?

abhilashraju:
> Cool! Why do we need this for 1110 ? Does this fix an error ? What is upsteam?

This is already fixed in upstream and 1210. Just that we need this fix in 1110 as per testers comment

raviteja-b (approved):

baemyung:
Maybe better to be done with stacking up 2 commits ?
- python format issue
- Refactor this code

gtmills:
> Maybe better to be done with stacking up 2 commits ?
>
>    python format issue
>    Refactor this code

+1. We do this because every commit should do exactly what it says it does and not include hidden changes like Python formatting. More importantly for downstream, we don't want Python formatting mixed with a commit because we might want to cherry-pick that commit into another release, or when we rebase, we might want that commit. In this case, I don't think either is needed, so approving, but for next time, please 1 commit for each feature/thing.

gtmills (approved):

PR #1407, opened 2026-02-28
fix: Allow only deletion of manual guards

shwetha-bhat:
In the bmcweb hardware isolation handler, add a check on the Severity property of HW isolation D-Bus records before allowing deletion. Only records with a Manual severity type are permitted to be deleted via the REST API.

Previously, this check was absent, allowing sys-guarded (predictive or fatal) hardware records to be deleted through the hardware isolation endpoint, which could result in firmware-isolated hardware being inadvertently reconfigured.

Refactored the large lambda in deisolateResource to call smaller functions.

Tested: Verified that deletion of manual guard records succeeds and deletion of sys (predictive/fatal) guard records is rejected with an appropriate error response.

Test-results:
```
Feb 28 08:01:31 p11bmc bmcwebd[1467]: [systems_logservices_hwisolation.hpp:1157] DBUS response error [5 : Input/output error] when tried to delete the given object path:
Feb 28 08:01:31 p11bmc bmcwebd[1467]: [systems_logservices_hwisolation.hpp:1168] DBus ErrorName: xyz.openbmc_project.Common.Error.InsufficientPermission ErrorMsg: Insufficient permission to perform operation
```

1. Test core reconfiguration:
image

2. Test dimm reconfiguration:
image

### DIFF
--- a/redfish-core/include/utils/hw_isolation.hpp
+++ b/redfish-core/include/utils/hw_isolation.hpp
@@ -187,6 +187,141 @@ inline void isolateResource(
 }
 
 /**
+ * @brief Helper function to call D-Bus Delete and handle errors
+ *
+ * @param[in] asyncResp - The redfish response to return to the caller.
+ * @param[in] resourceIsolatedHwEntry - The hardware isolation entry path.
+ * @param[in] resourceObjPath - The redfish resource dbus object path.
+ * @param[in] hwIsolationDbusName - The HardwareIsolation dbus name.
+ *
+ * @return The redfish response in given response buffer.
+ */
+inline void callDeleteAndHandleErrors(
+    const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
+    const std::string& resourceIsolatedHwEntry,
+    const sdbusplus::message::object_path& resourceObjPath,
+    const std::string& hwIsolationDbusName)
+{
+    dbus::utility::async_method_call(
+        [asyncResp, resourceIsolatedHwEntry,
+         resourceObjPath](const boost::system::error_code& ec1,
+                          const sdbusplus::message::message& msg) {
+            if (!ec1)
+            {
+                BMCWEB_LOG_INFO("Successfully de-isolated hardware: {}",
+                                resourceIsolatedHwEntry);
+                messages::success(asyncResp->res);
+                return;
+            }
+
+            const sd_bus_error* dbusError = msg.get_error();
+
+            if (dbusError == nullptr)
+            {
+                BMCWEB_LOG_ERROR(
+                    "DBUS response error [{} : {}] when tried to de-isolate the given resource: {}",
+                    ec1.value(), ec1.message(), resourceIsolatedHwEntry);
+                messages::internalError(asyncResp->res);
+                return;
+            }
+
+            BMCWEB_LOG_ERROR(
+                "DBUS response error [{} : {}] when tried to de-isolate the given resource: {}, DBus ErrorName: {} ErrorMsg: {}",
+                ec1.value(), ec1.message(), resourceIsolatedHwEntry,
+                dbusError->name, dbusError->message);
+
+            if (std::string_view(
+                    "xyz.openbmc_project.Common.Error.NotAllowed") ==
+                dbusError->name)
+            {
+                retChassisPowerStateOffRequiredError(asyncResp,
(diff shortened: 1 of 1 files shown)


redfish-core/include/utils/hw_isolation.hpp
```diff
@@ -236,56 +236,102 @@ inline void deisolateResource(
...
-                        ec1.value(), ec1.message(), resourceIsolatedHwEntry);
+                    // Log the severity for debugging and to prevent compiler optimization issues
+                    BMCWEB_LOG_DEBUG(
+                        "Hardware isolation entry Severity check for {}: {}",
+                        resourceIsolatedHwEntry, severity);
 
-                    const sd_bus_error* dbusError = msg.get_error();
+                    // Store result explicitly to prevent compiler optimization
+                    const size_t findPos = severity.find("Manual");
+                    const bool isManual = (findPos != std::string::npos);
```
deepakala-k: const bool isManual = (severity == "xyz.openbmc_project.HardwareIsolation.Entry.Type.Manual");

redfish-core/include/utils/hw_isolation.hpp
```diff
@@ -236,56 +236,102 @@ inline void deisolateResource(
...
 
-                    if (dbusError == nullptr)
+                    // Only allow deletion if Severity is Manual
+                    // Block deletion for Critical, Warning, and Spare
+                    if (!isManual)
                     {
-                        messages::internalError(asyncResp->res);
+                        BMCWEB_LOG_CRITICAL(
```
deepakala-k: BMCWEB_LOG_CRITICAL implies a system-level critical failure. This is a normal, expected user-facing rejection (user tried to delete a sys-guard). It should be BMCWEB_LOG_WARNING or BMCWEB_LOG_INFO at most.

deepakala-k (commented):
When the system is powered on and if the user tries to delete a system generated guard record, we might first give an error that it is a system generated guard and not give the system is powered on error. We just need to confirm if that is alright

deepakala-k (approved):

redfish-core/include/utils/hw_isolation.hpp
```diff
@@ -186,6 +186,142 @@ inline void isolateResource(
...
+            const sd_bus_error* dbusError = msg.get_error();
+
+            if (dbusError == nullptr)
+            {
+                messages::internalError(asyncResp->res);
+                return;
+            }
+
+            BMCWEB_LOG_ERROR("DBus ErrorName: {} ErrorMsg: {}",
+                             dbusError->name, dbusError->message);
```
baemyung: Would it be better to combine ERROR entries like?
```
      const sd_bus_error* dbusError = msg.get_error();
        if (dbusError == nullptr)
        {
            BMCWEB_LOG_ERROR(
                "DBUS response error [{} : {}] when tried to de-isolate the given resource: {}",
                ec1.value(), ec1.message(), resourceIsolatedHwEntry);
            return;
        }

        BMCWEB_LOG_ERROR(
            "DBUS response error [{} : {}] when tried to de-isolate the given resource: {}, DBus ErrorName: {} ErrorMsg: {}",
            ec1.value(), ec1.message(), resourceIsolatedHwEntry,
            dbusError->name, dbusError->message);
      ...
```

redfish-core/include/utils/hw_isolation.hpp
```diff
@@ -186,6 +186,142 @@ inline void isolateResource(
...
+            BMCWEB_LOG_DEBUG(
+                "Hardware isolation entry Severity check for {}: {}",
+                resourceIsolatedHwEntry, severity);
+
+            // Check if Severity is Manual (exact match with D-Bus enum value)
+            const bool isManual = (severity == "xyz.openbmc_project.HardwareIsolation.Entry.Type.Manual");
+
+            // Only allow deletion if Severity is Manual
+            // Block deletion for Critical, Warning, and Spare
+            if (!isManual)
```
baemyung: It may be clearer to do 
```
            // Check if Severity is Manual (exact match with D-Bus enum value)
            // Only allow deletion if Severity is Manual
            // Block deletion for Critical, Warning, and Spare
            if (severity != "xyz.openbmc_project.HardwareIsolation.Entry.Type.Manual")
            {
```

baemyung (approved):
It seems good to me

PR #1409, opened 2026-03-03
fix: Allow only deletion of manual guards

shwetha-bhat:
In the bmcweb hardware isolation handler, add a check on the Severity property of HW isolation D-Bus records before allowing deletion. Only records with a Manual severity type are permitted to be deleted via the REST API.

Previously, this check was absent, allowing sys-guarded (predictive or fatal) hardware records to be deleted through the hardware isolation endpoint, which could result in firmware-isolated hardware being inadvertently reconfigured.

Refactored the large lambda in deisolateResource to call smaller functions.

Below is the screenshot of test results:

1. Test core reconfiguration:
image

2. Test dimm reconfiguration:
image

### DIFF
--- a/redfish-core/include/utils/hw_isolation.hpp
+++ b/redfish-core/include/utils/hw_isolation.hpp
@@ -168,6 +168,141 @@ inline void isolateResource(
 }
 
 /**
+ * @brief Helper function to call D-Bus Delete and handle errors
+ *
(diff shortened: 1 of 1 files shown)


deepakala-k (approved):

baemyung (approved):
As this code is synced with https://github.com/ibm-openbmc/bmcweb/pull/1407, it looks good to me

PR #1415, opened 2026-03-05
Send Exit Utilization before Enter to prevent conflict

cjcain:
The Idle Power Saver parameters have an enter and exit utilization that have dependencies on each other. The enter must be less than or equal to the exit utilization.

DBUS properties are set individually which can lead to a condition where the new enter utilization could be larger than the current exit utilization. When that condition is encountered, the enter utilization is updated to the current exit utilization.

This fix will reverse the order (exit before enter) preventing requested utilization values from being lost.

Tested on Everest hardware and validator was run.

Change-Id: Id8a68650b1c0dfb825d0df35a742ffd0fb2d8a4a

### DIFF
--- a/redfish-core/lib/systems.hpp
+++ b/redfish-core/lib/systems.hpp
@@ -2424,6 +2424,22 @@ inline void setIdlePowerSaver(
                     "xyz.openbmc_project.Control.Power.IdlePowerSaver",
                     "Enabled", *ipsEnable);
             }
+            if (ipsExitUtil)
+            {
+                setDbusProperty(
+                    asyncResp, "IdlePowerSaver/ExitUtilizationPercent", service,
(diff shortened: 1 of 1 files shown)


baemyung:
It looks like - 1110 is also the same - https://github.com/ibm-openbmc/bmcweb/blob/1110/redfish-core/lib/systems.hpp#L2476
Should 1110 be changed as well?

baemyung:
Upstream: https://gerrit.openbmc.org/c/openbmc/bmcweb/+/88068

baemyung:
I also think it may be better to add the example like https://gerrit.openbmc.org/c/openbmc/bmcweb/+/88068/comment/0a45b47e_31851943/

baemyung (approved):
In terms of code based on the problem statement (order of dbus set of Exit/Enter), it looks good to me.

PR #1427, opened 2026-03-30
bmcweb: Return HTTP 409 for resource conflict errors

gtmills:
Upstream merged at https://github.com/openbmc/bmcweb/commit/793bce1895f015ae27a0ce5fa4c75e38c2dadfd6

Change ResourceAlreadyExists from HTTP 400 Bad Request to HTTP 409 Conflict. While the Redfish specification (DSP0266 v1.23.1) does not explicitly mandate to return 409 it makes more sense because:

- HTTP 400 implies a malformed request which this is clearly not
- HTTP 409 is defined in Table 13 of DSP0266 as Creation or update request could not be completed because it causes a conflict in the current state of the resource which is what happens.

Also mapping ResourceCreationConflict to 409 for the same reason even even though its currently not used in the codebase.

Tested: POST to create a duplicate user account now returns HTTP 409 with ResourceAlreadyExists message instead of HTTP 400. The ResourceCreationConflict message is not used in the codebase and can not be tested.

Change-Id: Id482e5e4f7b3d6ca56228f9d763c95aa50c9856f


raviteja-b (approved):

abiasojo (approved):
Looks okay to me.

PR #1429, opened 2026-04-01
1210 : Aggregation: Add basic authentication support

abiasojo:
Aggregation: Add basic authentication support

Add support for basic authentication when connecting to aggregation
sources. This allows satellite BMCs to be authenticated using
username and password credentials.

The implementation:
- Stores credentials alongside URLs in AggregationSource struct
- Validates credentials: no colons, max 40 chars, not empty strings
- Creates Basic Auth headers using base64 encoding
- Only sends Authorization header when both username and password exist
- Adds PATCH handler for updating credentials independently
- Prevents duplicate aggregation sources with same hostname
- Cleans up credentials when aggregation sources are deleted

Tested: Manual testing with authenticated aggregation sources

Change-Id: Ide17a3c08a4a8f6b90a2ffcd2c798cbbec578db8

### DIFF
--- a/redfish-core/include/query.hpp
+++ b/redfish-core/include/query.hpp
@@ -167,7 +167,7 @@ inline bool handleIfMatch(crow::App& app, const crow::Request& req,
     if constexpr (BMCWEB_REDFISH_AGGREGATION)
     {
         needToCallHandlers =
-            RedfishAggregator::beginAggregation(req, asyncResp) ==
+            RedfishAggregator::getInstance().beginAggregation(req, asyncResp) ==
             Result::LocalHandle;
 
         // If the request should be forwarded to a satellite BMC then we don't
--- a/redfish-core/include/redfish_aggregator.hpp
+++ b/redfish-core/include/redfish_aggregator.hpp
@@ -413,6 +413,13 @@ inline crow::ConnectionPolicy getAggregationPolicy()
             .invalidResp = aggregationRetryHandler};
 }
 
+struct AggregationSource
+{
+    boost::urls::url url;
+    std::string username;
+    std::string password;
+};
+
 class RedfishAggregator
 {
   private:
@@ -587,9 +594,9 @@ class RedfishAggregator
         Resource,
     };
 
-    static void startAggregation(
+    void startAggregation(
         AggregationType aggType, const crow::Request& thisReq,
-        const std::shared_ptr<bmcweb::AsyncResp>& asyncResp)
+        const std::shared_ptr<bmcweb::AsyncResp>& asyncResp) const
     {
         if (thisReq.method() != boost::beast::http::verb::get)
         {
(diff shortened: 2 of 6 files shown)


abiasojo:
jenkins run tests please

gtmills:
Can you add the upstream commit, https://gerrit.openbmc.org/c/openbmc/bmcweb/+/84143 ?

redfish-core/include/query.hpp
```diff
@@ -167,7 +167,7 @@ inline bool handleIfMatch(crow::App& app, const crow::Request& req,
     if constexpr (BMCWEB_REDFISH_AGGREGATION)
     {
         needToCallHandlers =
-            RedfishAggregator::beginAggregation(req, asyncResp) ==
+            RedfishAggregator::getInstance().beginAggregation(req, asyncResp) ==
```
gtmills, line 170: Why did this change, I don't see it in the upstream commit ?

redfish-core/include/redfish_aggregator.hpp
```diff
@@ -587,9 +594,9 @@ class RedfishAggregator
         Resource,
     };
 
-    static void startAggregation(
+    void startAggregation(
```
gtmills, line 597: why did this change?

gtmills:
This looks way bigger than the changes upstream? 
Did you pull more than 1 commit in ?
https://gerrit.openbmc.org/c/openbmc/bmcweb/+/84143/17/redfish-core/include/redfish_aggregator.hpp

redfish-core/include/query.hpp
abiasojo: It's needed to get the code to compile.

redfish-core/include/redfish_aggregator.hpp
abiasojo: It was needed to resolve compile errors

abiasojo:
> This looks way bigger than the changes upstream? Did you pull more than 1 commit in ? https://gerrit.openbmc.org/c/openbmc/bmcweb/+/84143/17/redfish-core/include/redfish_aggregator.hpp

The changes were needed to get the code to compile.

abiasojo:
> Can you add the upstream commit, https://gerrit.openbmc.org/c/openbmc/bmcweb/+/84143 ?

Done.

redfish-core/include/redfish_aggregator.hpp
gtmills: Can you add what commit this is from? 
Should we pull in another commit ?

gtmills: I don't love that there is changes under this commit that weren't part of the upstream commit 
If we need additional changes can we pull in that commit? 
If it looks like 
commit 1 
commit 2 
commit 3 

I think that is cleaner

abiasojo: > I don't love that there is changes under this commit that weren't part of the upstream commit If we need additional changes can we pull in that commit? If it looks like commit 1 commit 2 commit 3
> 
> I think that is cleaner

This is the commit that introduced those changes https://gerrit.openbmc.org/c/openbmc/bmcweb/+/71840

abiasojo: > Can you add what commit this is from? Should we pull in another commit ?

Commit https://gerrit.openbmc.org/c/openbmc/bmcweb/+/71840 has now been pulled in.

PR #1431, opened 2026-04-02
Fix: Redfish Service Validator 3.1.0 Conformance

jnin-dev:
IBMComputerSystems Schema was reporting fails on:

- Actions/Oem/#IBMComputerSystem.v1_0_0.ExecutePanelFunction/target
```
ERROR - FAIL - /Actions/Oem/#IBMComputerSystem.v1_0_0.ExecutePanelFunction/target (/redfish/v1/Systems/system/Actions/Oem/IBM/IBMComputerSystem.ExecutePanelFunction):
    Action URI Error: The target URI for the action is expected to be '/redfish/v1/Systems/system/Actions/Oem/IBMComputerSystem.v1_0_0.ExecutePanelFunction'.
```

- /Oem/IBM/ChapData/@odata.type
```
ERROR - FAIL - /Oem/IBM/ChapData/@odata.type ([Not Present]): Required Property Error: The property '@odata.type' is mandatory, but not present in the payload.
```

Tested:
- Redfish Service Validator 3.1.0 passes on those URIs.

### DIFF
--- a/redfish-core/lib/systems.hpp
+++ b/redfish-core/lib/systems.hpp
@@ -2297,6 +2297,7 @@ inline void getChapData(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp)
             oemIBM["@odata.type"] = "#IBMComputerSystem.v1_0_0.IBM";
 
             nlohmann::json& chapData = oemIBM["ChapData"];
+            chapData["@odata.type"] = "#IBMComputerSystem.v1_0_0.ChapData";
             chapData["ChapName"] = *chapName;
             chapData["ChapSecret"] = *chapSecret;
         });
@@ -2829,8 +2830,8 @@ inline void handleComputerSystemGet(
     getEnabledPanelFunctions(asyncResp);
 
     nlohmann::json& actionOem = asyncResp->res.jsonValue["Actions"]["Oem"];
-    actionOem["#IBMComputerSystem.v1_0_0.ExecutePanelFunction"]["target"] =
-        "/redfish/v1/Systems/system/Actions/Oem/IBM/IBMComputerSystem.ExecutePanelFunction";
+    actionOem["#IBMComputerSystem.ExecutePanelFunction"]["target"] =
+        "/redfish/v1/Systems/system/Actions/Oem/IBMComputerSystem.ExecutePanelFunction";
 
     // ChapData
     getChapData(asyncResp);
@@ -3283,7 +3284,7 @@ inline void requestRoutesSystemActionsOemExecutePanelFunction(App& app)
 {
     BMCWEB_ROUTE(
         app,
-        "/redfish/v1/Systems/system/Actions/Oem/IBM/IBMComputerSystem.ExecutePanelFunction/")
+        "/redfish/v1/Systems/system/Actions/Oem/IBMComputerSystem.ExecutePanelFunction/")
         .privileges(redfish::privileges::postComputerSystem)
         .methods(boost::beast::http::verb::post)(std::bind_front(
             handleSystemActionsOemExecutePanelFunctionPost, std::ref(app)));
--- a/redfish-core/schema/oem/ibm/csdl/IBMComputerSystem_v1.xml
+++ b/redfish-core/schema/oem/ibm/csdl/IBMComputerSystem_v1.xml
@@ -59,8 +59,21 @@
           <Annotation Term="OData.LongDescription" String="The value of this property shall indicate if safe mode is active."/>
         </Property>
       </ComplexType>
+      <Action Name="ExecutePanelFunction" IsBound="true">
+        <Annotation Term="OData.Description" String="This action executes a panel function."/>
+        <Annotation Term="OData.LongDescription" String="This action executes a panel function if the function is enabled."/>
+        <Parameter Name="FuncNo" Type="IBMComputerSystem.v1_0_0.OemActions" Nullable="false">
+          <Annotation Term="OData.Description" String="Panel function number."/>
+          <Annotation Term="OData.LongDescription" String="This parameter shall contain a  panel function number to be executed."/>
+        </Parameter>
+      </Action>
     </Schema>
     <Schema xmlns="http://docs.oasis-open.org/odata/ns/edm" Namespace="IBMComputerSystem.v1_0_0">
+      <ComplexType Name="OemActions">
+        <Annotation Term="OData.AdditionalProperties" Bool="true"/>
+        <Annotation Term="OData.Description" String="The available OEM-specific actions for this resource."/>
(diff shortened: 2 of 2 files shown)


jeaaustx:
jenkins run tests please

redfish-core/lib/systems.hpp
```diff
@@ -2830,7 +2831,7 @@ inline void handleComputerSystemGet(
 
     nlohmann::json& actionOem = asyncResp->res.jsonValue["Actions"]["Oem"];
     actionOem["#IBMComputerSystem.v1_0_0.ExecutePanelFunction"]["target"] =
-        "/redfish/v1/Systems/system/Actions/Oem/IBM/IBMComputerSystem.ExecutePanelFunction";
+        "/redfish/v1/Systems/system/Actions/Oem/IBMComputerSystem.v1_0_0.ExecutePanelFunction";
```
jeaaustx: It looks weird to me for the version to be part of the target. (Means the client is going to have to adjust based on a version change. It is discoverable but I'm wondering if it is hard-coded anywhere.)

Can it be done in a way that passes the validator without the version here?

redfish-core/lib/systems.hpp
```diff
@@ -3283,7 +3284,7 @@ inline void requestRoutesSystemActionsOemExecutePanelFunction(App& app)
 {
     BMCWEB_ROUTE(
         app,
-        "/redfish/v1/Systems/system/Actions/Oem/IBM/IBMComputerSystem.ExecutePanelFunction/")
+        "/redfish/v1/Systems/system/Actions/Oem/IBMComputerSystem.v1_0_0.ExecutePanelFunction/")
```
jeaaustx: Do we need to coordinate this with any client? I'm wondering if anything has this route hard-coded and will fail with this change.

jeaaustx (commented):
nit: First line of commit message I think it would be better to term it as to what changed with this commit. Then as part of the commit message show the validator failure that is being fixed by the drop.

So maybe: Fixed IBMComputerSystems Schema validator failure

followed by rest of what you already have in the commit message.

gtmills:
add to approvelist

redfish-core/lib/systems.hpp
jnin-dev: It should be possible, I'm looking at the upstream version and more documentation to see how to do so

redfish-core/lib/systems.hpp
jnin-dev: Most likely if it's changed, I'll see if I can change it to keep the same route

gtmills: We will have to get the webui-vue to change at 
https://github.com/ibm-openbmc/webui-vue/blob/68eef6b3653dea6538b9627be191151f297dfa3e/src/store/modules/Logs/IBMiServiceFunctionsStore.js#L26

redfish-core/lib/systems.hpp
gtmills: https://github.com/DMTF/Redfish/blob/819ee8dc8c121935debad2060177381b5dcefe53/mockups/public-oem-examples/AccountService/index.json#L26

or 

https://redfish.dmtf.org/redfish/mockups/v1/1784

gtmills: Based on the mockup 

/redfish/v1/Systems/system/Actions/Oem/IBMComputerSystem.ExecutePanelFunction

jnin-dev: OK, I changed it so that it does not use the version number as part of the target and updated the first line of the commit message so that it says what was actually changed i.e. the IBMComputerSystem schema.

redfish-core/schema/oem/ibm/csdl/IBMComputerSystem_v1.xml
```diff
@@ -59,8 +59,21 @@
           <Annotation Term="OData.LongDescription" String="The value of this property shall indicate if safe mode is active."/>
         </Property>
       </ComplexType>
+      <Action Name="ExecutePanelFunction" IsBound="true">
```
gtmills, line 62: Moved to the unversioned schema

redfish-core/lib/systems.hpp
```diff
@@ -3283,7 +3284,7 @@ inline void requestRoutesSystemActionsOemExecutePanelFunction(App& app)
 {
     BMCWEB_ROUTE(
         app,
-        "/redfish/v1/Systems/system/Actions/Oem/IBM/IBMComputerSystem.ExecutePanelFunction/")
+        "/redfish/v1/Systems/system/Actions/Oem/IBMComputerSystem.ExecutePanelFunction/")
```
gtmills, line 3287: This is an error because of v1_0_0
```
actionOem["#IBMComputerSystem.v1_0_0.ExecutePanelFunction"]["target"] =
```

But we are changing the URL to /Oem/IBMComputerSystem.. to match the mockup

redfish-core/schema/oem/ibm/csdl/IBMComputerSystem_v1.xml
gtmills: Can you look upstream and make sure this matches?

jnin-dev: Although not from the OEM, [this file](https://github.com/openbmc/bmcweb/blob/master/redfish-core/schema/dmtf/csdl/ComputerSystem_v1.xml) seems to a similar format where the action is defined in the unversioned schema and the has a type for `OemActions` within the versioned schema. [This file](https://github.com/openbmc/bmcweb/blob/master/redfish-core/schema/oem/openbmc/csdl/OpenBMCComputerSystem_v1.xml) seems to be similar but does not define any action or OemAction for the OpenBMCComputerSystem schema

redfish-core/lib/systems.hpp
jnin-dev: > This is an error because of v1_0_0

I think more concisely the line on 2832-2833 was the cause of the Action URI error because of the mismatch in the namespace for the target and the action but yes that we should change the URL to match Redfish standards

redfish-core/lib/systems.hpp
jnin-dev: They have been pinged, waiting for a reply

jnin-dev: Ok, a story was made so I'm resolving this.

redfish-core/schema/oem/ibm/csdl/IBMComputerSystem_v1.xml
jnin-dev: I believe that this is following the upstream pattern regarding OemActions. I'm going to mark this resolved. Let me know if you believe otherwise.

gtmills (approved):

PR #1432, opened 2026-04-06
1210: Aggregation: Add basic authentication support

abiasojo:
Add support for basic authentication when connecting to aggregation
sources. This allows satellite BMCs to be authenticated using
username and password credentials.

The implementation:
- Stores credentials alongside URLs in AggregationSource struct
- Validates credentials: no colons, max 40 chars, not empty strings
- Creates Basic Auth headers using base64 encoding
- Only sends Authorization header when both username and password exist
- Adds PATCH handler for updating credentials independently
- Prevents duplicate aggregation sources with same hostname
- Cleans up credentials when aggregation sources are deleted

Tested: Manual testing with authenticated aggregation sources

Example commands succeed, and return the expected results.
```
curl --insecure -H "Content-Type: application/json" -X POST -D headers.txt https://$bmc/redfish/v1/AggregationService/AggregationSources/ -d '{"HostName": "https://9.3.29.130/", "UserName" :"<username>", "Password":  "<password>"}' -H "X-Auth-Token: $token"

{
  "@Message.ExtendedInfo": [
    {
      "Message": "The resource was created successfully.",
      "MessageArgs": [],
      "MessageId": "Base.1.19.Created",
      "MessageSeverity": "OK",
      "Resolution": "None."
    }
  ]
}

$curl -k  -X GET https://${bmc}/redfish/v1/Systems -H "X-Auth-Token: $token"

{
  "@odata.id": "/redfish/v1/Systems",
  "@odata.type": "#ComputerSystemCollection.ComputerSystemCollection",
  "Members": [
    {
    },
    {
      "@odata.id": "/redfish/v1/Systems/hypervisor"
    },
    {
      "@odata.id": "/redfish/v1/Systems/oRQfDGjh_system"
    },
    {
      "@odata.id": "/redfish/v1/Systems/oRQfDGjh_hypervisor"
    }
  ],
  "Members@odata.count": 4,
  "Name": "Computer System Collection"
}

$curl -k  -X GET https://${bmc}/redfish/v1/Managers -H "X-Auth-Token: $token"

{
  "@odata.id": "/redfish/v1/Managers",
  "@odata.type": "#ManagerCollection.ManagerCollection",
  "Members": [
    {
    },
    {
      "@odata.id": "/redfish/v1/Managers/oRQfDGjh_bmc"
    }
  ],
  "Members@odata.count": 2,
  "Name": "Manager Collection"
}
```

https://gerrit.openbmc.org/c/openbmc/bmcweb/+/84143

Change-Id: Ide17a3c08a4a8f6b90a2ffcd2c798cbbec578db8

**Allow POST on aggregation sources**
It would be useful for both testing and production if one could create aggregation sources dynamically, rather than using detected hardware. Redfish sources might exist outside of the physical chassis in a way that's detectable, so giving DC software a way to set up a non-trivial aggregation topology make this more extensible.

Note, that as documented in AGGREGATION.md, only a single satellite is supported by this feature.  This patch does not change that behavior, and implementing an entity-manager satellite will override a POST created AggregationSource.

Tested:

Example commands succeed, and return the expected results.
```
curl -vvvv -k --http1.1 --user "root:0penBmc" --request DELETE https://192.168.7.2/redfish/v1/AggregationService/AggregationSources/
curl -vvvv -k --http1.1 --user "root:0penBmc" https://192.168.7.2/redfish/v1/AggregationService/AggregationSources -H "Content-Type: application/json" --request POST --data '{"HostName":"https://localhost:8000"}'
```

Redfish service validator passes.

https://gerrit.openbmc.org/c/openbmc/bmcweb/+/71840

Change-Id: I15711c3075645291b5f2555eefb80bb80418e7e5

**Enable satellite aggregation over POST**
This patch enables Redfish satellite aggregation through REST API
POST requests to /redfish/v1/AggregationService/AggregationSources.
Tested successfully on ubuntu22 using multiple python redfish mockup
servers as satellite bmcs.

The previous urlHasPrefix function incorrectly assumed it would receive
a full URL path, but it actually receives individual URL segments.
Replaced with segmentHasPrefix that vastly simplifies prefix matching
in the currentAggregationSources map.

Also removed the correct duplicated route in
redfish-core/lib/aggregation_service.hpp that was causing routing
conflicts.

The hardcoded 5B247A_ fallback remains temporarily and will be
addressed in change 83181 which adds D-Bus satellite support.

https://gerrit.openbmc.org/c/openbmc/bmcweb/+/83180

Change-Id: I41920ad270abe4b228b43280ea425b80d3f14b50

### DIFF
--- a/redfish-core/include/redfish_aggregator.hpp
+++ b/redfish-core/include/redfish_aggregator.hpp
@@ -1313,11 +1313,7 @@ class RedfishAggregator
             {
                 // We've matched a resource collection so this current segment
                 // might contain an aggregation prefix
-                // TODO: This needs to be rethought when we can support multiple
-                // satellites due to
-                // /redfish/v1/AggregationService/AggregationSources/5B247A
-                // being a local resource describing the satellite
-                if (collectionItem.starts_with("5B247A_"))
+                if (segmentHasPrefix(collectionItem))
                 {
                     BMCWEB_LOG_DEBUG("Need to forward a request");
 
@@ -1367,6 +1363,32 @@ class RedfishAggregator
 
         BMCWEB_LOG_DEBUG("Aggregation not required for {}", url.buffer());
         return Result::LocalHandle;
+    }
+
+    // Check if the given URL segment matches with any satellite prefix
+    // Assumes the given segment starts with <prefix>_
+    bool segmentHasPrefix(const std::string& urlSegment) const
+    {
+        // TODO: handle this better
+        // For now 5B247A_ wont be in the currentAggregationSources map so
+        // check explicitly for now
+        if (urlSegment.starts_with("5B247A_"))
+        {
+            return true;
+        }
+
+        // Find the first underscore
+        std::size_t underscorePos = urlSegment.find('_');
+        if (underscorePos == std::string::npos)
+        {
+            return false; // No underscore, can't be a satellite prefix
+        }
+
+        // Extract the prefix
+        std::string prefix = urlSegment.substr(0, underscorePos);
+
+        // Check if this prefix exists
+        return currentAggregationSources.contains(prefix);
     }
 };
 
--- a/redfish-core/include/redfish_aggregator.hpp
+++ b/redfish-core/include/redfish_aggregator.hpp
@@ -587,9 +587,9 @@ class RedfishAggregator
         Resource,
     };
 
-    static void startAggregation(
+    void startAggregation(
         AggregationType aggType, const crow::Request& thisReq,
-        const std::shared_ptr<bmcweb::AsyncResp>& asyncResp)
+        const std::shared_ptr<bmcweb::AsyncResp>& asyncResp) const
     {
         if (thisReq.method() != boost::beast::http::verb::get)
         {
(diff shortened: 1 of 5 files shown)


gtmills:
Can you add which upstream commits these are and what is their status? Also if the PR msg could reflect the 3 commits being pulled in, that would help.

abiasojo:
commit 1 : merged at https://gerrit.openbmc.org/c/openbmc/bmcweb/+/71840
commit 2 : merged at https://gerrit.openbmc.org/c/openbmc/bmcweb/+/83180
commit 3 : merged at https://gerrit.openbmc.org/c/openbmc/bmcweb/+/84143

abiasojo:
> Can you add which upstream commits these are and what is their status? Also if the PR msg could reflect the 3 commits being pulled in, that would help.

Done

redfish-core/include/redfish_aggregator.hpp
```diff
@@ -1368,6 +1364,32 @@ class RedfishAggregator
...
     }
+
+    // Check if the given URL segment matches with any satellite prefix
+    // Assumes the given segment starts with <prefix>_
+    bool segmentHasPrefix(const std::string& urlSegment) const
+    {
+        // TODO: handle this better
+        // For now 5B247A_ wont be in the currentAggregationSources map so
+        // check explicitly for now
+        if (urlSegment.starts_with("5B247A_"))
```
jeaaustx, line 1427: Is this hard-coded something we should have downstream? Is this a backdoor we shouldn't have?

Maybe upstream this should be under a compile option?

jeaaustx:
I suggest you also add the manual testing steps to the PR description since that was one of the things which was missing from the upstream commit.

abiasojo:
> I suggest you also add the manual testing steps to the PR description since that was one of the things which was missing from the upstream commit.

I have updated this in the commit message and in the comments.

redfish-core/include/redfish_aggregator.hpp
abiasojo: > Is this hard-coded something we should have downstream? Is this a backdoor we shouldn't have?
> 
> Maybe upstream this should be under a compile option?

Here is the reference to the change in upstream. I cherry-picked this from upstream.
https://gerrit.openbmc.org/c/openbmc/bmcweb/+/83180/comment/83326b9f_d0181b9b/

gtmills (approved):
Pulls in 3 commits from upstream

redfish-core/include/redfish_aggregator.hpp
jeaaustx: Yes I know this came from upstream directly. I'm just wondering if for downstream we shouldn't have this hard-coded here.

I'm not that familiar with the implications of this, maybe it isn't a concern? @gtmills what are your thoughts on this?

PR #1438, opened 2026-04-14
1210: Aggregation: Fix idefinite aggregation loop

abiasojo:
Any bmc should be able to aggregate resources from its peers.
Bmcweb will be compiled with aggregation enabled for both satellite and master BMCs. Bmcweb should be able to distinguish between the requests originated from its peer BMC and external redfish clients. Otherwise aggregation will end up in indefinite aggregation loop.

This commit will add extra header to identify the aggregation request from peers.
This implementation assumes that any type of transitive aggregation is not present( A aggregates B aggregates C etc).
If we have such requirement then we can enhance header property to hold all parent bmc information to filter and avoid any aggregation loop.

This cherry-picks only includes the changes to fix the issue with indefinite aggregation loop. It doesn't include the changes for
MTLS support.

Tested By:
bmc and bmc2 are satellites of each other.
curl -k  -X GET https://${bmc}/redfish/v1/Systems -H "X-Auth-Token: $token" {
  "@odata.id": "/redfish/v1/Systems",
  "@odata.type": "#ComputerSystemCollection.ComputerSystemCollection",
  "Members": [
    {
    },
    {
      "@odata.id": "/redfish/v1/Systems/hypervisor"
    },
    {
      "@odata.id": "/redfish/v1/Systems/9Ilut4vH_system"
    },
    {
      "@odata.id": "/redfish/v1/Systems/9Ilut4vH_hypervisor"
    },
    {
      "@odata.id": "/redfish/v1/Systems/CNaH0ljH_system"
    },
    {
      "@odata.id": "/redfish/v1/Systems/CNaH0ljH_hypervisor"
    }
  ],
  "Members@odata.count": 6,
  "Name": "Computer System Collection"

curl -k  -X GET https://${bmc2}/redfish/v1/Systems -H "X-Auth-Token: $token2"
{
  "@odata.id": "/redfish/v1/Systems",
  "@odata.type": "#ComputerSystemCollection.ComputerSystemCollection",
  "Members": [
    {
    },
    {
      "@odata.id": "/redfish/v1/Systems/hypervisor"
    },
    {
      "@odata.id": "/redfish/v1/Systems/tJmMBoFD_system"
    },
    {
      "@odata.id": "/redfish/v1/Systems/tJmMBoFD_hypervisor"
    },
    {
      "@odata.id": "/redfish/v1/Systems/bI7Gq7VR_system"
    },
    {
      "@odata.id": "/redfish/v1/Systems/bI7Gq7VR_hypervisor"
    }
  ],
  "Members@odata.count": 6,
  "Name": "Computer System Collection"

https://gerrit.openbmc.org/c/openbmc/bmcweb/+/71685

### DIFF
--- a/redfish-core/include/redfish_aggregator.hpp
+++ b/redfish-core/include/redfish_aggregator.hpp
@@ -425,9 +425,10 @@ class RedfishAggregator
   private:
     crow::HttpClient client;
 
-    // Dummy callback used by the Constructor so that it can report the number
+    // callback used by the Constructor to report the number
     // of satellite configs when the class is first created
-    static void constructorCallback(
+    // Fills default aggregationSources with satellite information
+    void constructorCallback(
         const boost::system::error_code& ec,
         const std::unordered_map<std::string, boost::urls::url>& satelliteInfo)
     {
@@ -439,6 +440,18 @@ class RedfishAggregator
 
         BMCWEB_LOG_DEBUG("There were {} satellite configs found at startup",
                          std::to_string(satelliteInfo.size()));
+        // Update aggregationSources with satellite information
+        for (const auto& [name, url] : satelliteInfo)
+        {
+            AggregationSource source;
+            source.url = url;
+            // Enhancement can be added to fetch username and password
+            // from some secure sources. Currently satellite configuration
+            // supports only NONE
+            source.username = "";
+            source.password = "";
+            aggregationSources[name] = std::move(source);
+        }
     }
 
     // Search D-Bus objects for satellite config objects and add their
@@ -456,16 +469,6 @@ class RedfishAggregator
                 {
                     BMCWEB_LOG_DEBUG("Found Satellite Controller at {}",
                                      objectPath.first.str);
-
-                    if (!satelliteInfo.empty())
-                    {
-                        BMCWEB_LOG_ERROR(
-                            "Redfish Aggregation only supports one satellite!");
-                        BMCWEB_LOG_DEBUG("Clearing all satellite data");
-                        satelliteInfo.clear();
-                        return;
-                    }
-
                     addSatelliteConfig(interface.second, satelliteInfo);
                 }
             }
@@ -511,7 +514,6 @@ class RedfishAggregator
                 }
                 url.set_port(std::to_string(static_cast<uint16_t>(*propVal)));
             }
-
             else if (prop.first == "AuthType")
             {
                 const std::string* propVal =
@@ -527,7 +529,7 @@ class RedfishAggregator
                 if (*propVal != "None")
                 {
                     BMCWEB_LOG_ERROR(
-                        "Unsupported AuthType value: {}, only \"none\" is supported",
+                        "Unsupported AuthType value: {}, only \"None\" is supported",
                         *propVal);
                     return;
                 }
@@ -615,19 +617,8 @@ class RedfishAggregator
             }
         }
 
-        // Create a copy of thisReq so we we can still locally process the req
-        std::error_code ec;
         auto localReq = std::make_shared<crow::Request>(thisReq.copy());
-        if (ec)
-        {
-            BMCWEB_LOG_ERROR("Failed to create copy of request");
-            if (aggType == AggregationType::Resource)
-            {
-                messages::internalError(asyncResp->res);
-            }
-            return;
-        }
-
+        localReq->addHeader("X-Forwarded-For", "bmcweb");
         if (aggType == AggregationType::Collection)
         {
             boost::urls::url& urlNew = localReq->url();
@@ -896,7 +887,8 @@ class RedfishAggregator
         client(getIoContext(),
                std::make_shared<crow::ConnectionPolicy>(getAggregationPolicy()))
     {
-        getSatelliteConfigs(constructorCallback);
+        getSatelliteConfigs(
+            std::bind_front(&RedfishAggregator::constructorCallback, this));
     }
     RedfishAggregator(const RedfishAggregator&) = delete;
     RedfishAggregator& operator=(const RedfishAggregator&) = delete;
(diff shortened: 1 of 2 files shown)


redfish-core/include/redfish_aggregator.hpp
```diff
@@ -439,6 +440,18 @@ class RedfishAggregator
...
         BMCWEB_LOG_DEBUG("There were {} satellite configs found at startup",
                          std::to_string(satelliteInfo.size()));
+        // Update aggregationSources with satellite information
+        for (const auto& [name, url] : satelliteInfo)
+        {
+            AggregationSource source;
+            source.url = url;
+            // Enhancement can be added to fetch username and password
+            // from some secure sources. Currently satellite configuration
+            // supports only NONE
```
jeaaustx, line 450: Upstream comment has "and MTLS" here.

redfish-core/include/redfish_aggregator.hpp
```diff
@@ -511,7 +514,6 @@ class RedfishAggregator
                 }
                 url.set_port(std::to_string(static_cast<uint16_t>(*propVal)));
             }
-
```
jeaaustx, line 514: Upstream commit has check and handling here for "Protocol". That is missing in this commit. Is that intentional?

If this is not a complete cherry-pick of the upstream commit please update the PR description and commit message to indicate that.

redfish-core/include/redfish_aggregator.hpp
jeaaustx, line 529: Upstream commit has check here for None and MTLS.

jeaaustx: Check is for "None", so this message seem to be incorrect. It was modified by the upstream commit.

jeaaustx (commented):
In the PR description and the commit message the GET curl command and response are repeated. They both look the same to me. Is one supposed to be from $bmc1 and one from $bmc2?

If so, I think it would be helpful to update to indicate that difference. Otherwise please remove the duplicate.

abiasojo:
> In the PR description and the commit message the GET curl command and response are repeated. They both look the same to me. Is one supposed to be from $bmc1 and one from $bmc2?
> 
> If so, I think it would be helpful to update to indicate that difference. Otherwise please remove the duplicate.

It's probably not obvious, but the first line after "Tested:" says bmc and bmc2 are satellites to each other. I see that the message is duplicated. I accidentally copied the wrong output. I will fix that too.

redfish-core/include/redfish_aggregator.hpp
abiasojo: > Upstream comment has "and MTLS" here.

I am only cherry-picking updates for aggregation circular-loop fixes

redfish-core/include/redfish_aggregator.hpp
abiasojo: > Upstream commit has check and handling here for "Protocol". That is missing in this commit. Is that intentional?
> 
> If this is not a complete cherry-pick of the upstream commit please update the PR description and commit message to indicate that.

I will update the message and description

redfish-core/include/redfish_aggregator.hpp
abiasojo: > Upstream commit has check here for None and MTLS.

I am only cherry-picking updates for aggregation circular-loop fixes

abiasojo: > Check is for "None", so this message seem to be incorrect. It was modified by the upstream commit.

It's ensuring that other unsupported AuthTypes are reported as errors.

redfish-core/include/redfish_aggregator.hpp
abiasojo: I have updated to indicate it is a partial cherry-pick.

gtmills:
@abiasojo And you tested this and this works?

abiasojo:
> @abiasojo And you tested this and this works?

I showed the output in the comments above. I could show you a demo if you want to see :)

redfish-core/include/redfish_aggregator.hpp
baemyung: Does this mean that "satellite" configuration does not support MTLS on 1210?
If it is not causing additional issue, I think it may be better to keep the code as what's in upstream.

redfish-core/include/redfish_aggregator.hpp
baemyung: Similarly, is this "Protocol" causing an additional error unless it is not removed?
Or if it does not cause any other side effect, I think it is better to keep it as upstream?

redfish-core/include/redfish_aggregator.hpp
baemyung: > It's ensuring that other unsupported AuthTypes are reported as errors.
Does this mean that  the other than "None"  (i.e. MTLS) is not going to wrk for Mutual aggregation?
Upstream expects "None" or "MTLS" as acceptable types.
Is there a specific reason not to accept "MTLS" type?

If so, I think this PR may need to have 2 commits to clarify the reason.
-  Do not accept MTLS 
-  Cherry-pick upstream (with the above none MTLS type)

redfish-core/include/redfish_aggregator.hpp
abiasojo: > Does this mean that "satellite" configuration does not support MTLS on 1210? If it is not causing additional issue, I think it may be better to keep the code as what's in upstream.

MTLS will be supported. I am currently testing with Basic Authentication. The MTLS support will be added later. The code changes were broken into two.

redfish-core/include/redfish_aggregator.hpp
abiasojo: > Similarly, is this "Protocol" causing an additional error unless it is not removed? Or if it does not cause any other side effect, I think it is better to keep it as upstream?

I am only cherry-picking updates for aggregation circular-loop fixes

redfish-core/include/redfish_aggregator.hpp
abiasojo: > > It's ensuring that other unsupported AuthTypes are reported as errors.
> > Does this mean that  the other than "None"  (i.e. MTLS) is not going to wrk for Mutual aggregation?
> > Upstream expects "None" or "MTLS" as acceptable types.
> > Is there a specific reason not to accept "MTLS" type?
> 
> If so, I think this PR may need to have 2 commits to clarify the reason.
> 
>     * Do not accept MTLS
> 
>     * Cherry-pick upstream (with the above none MTLS type)

I am only cherry-picking updates for aggregation circular-loop fixes.
I updated the commit message to reflect that.
MTLS support will be done in a different PR

baemyung: If so, I think this PR should contain 2 commits, so that MTLS part can be done later.

- Do not accept MTLS first,
- cherry-pick upstream to deal with circular-loop fix for None.

If not, this PR seems confusing as MTLS part is removed without the appropriate explanation/commit.

abiasojo: I have updated the commit message to not have the upstream change id

baemyung (approved):
It looks good as we've discussed/agreed

redfish-core/include/redfish_aggregator.hpp
jeaaustx: My comment here was that this error message is incorrect in that it is not capitalizing "none" in its output. But the code is checking for "None". So if the user sees this error message and puts "none" into their request they will again see the error?

I saw in the upstream commit the error message was altered. (It included the MTLS stuff also, but the none->None part of the change didn't seem related to that.)

abiasojo: > My comment here was that this error message is incorrect in that it is not capitalizing "none" in its output. But the code is checking for "None". So if the user sees this error message and puts "none" into their request they will again see the error?
> 
> I saw in the upstream commit the error message was altered. (It included the MTLS stuff also, but the none->None part of the change didn't seem related to that.)

I have update "none" to "None"

gtmills (approved):

PR #1439, opened 2026-04-21
Move body limit enforcement

gtmills:
https://gerrit.openbmc.org/c/openbmc/bmcweb/+/87688 is upstream 

Make sure the body limit enforcement happens before the Expect: 100-continue check.

Tested: Verified the body limit check happens first, even when Expect:
        100-continue is used. Now rejects with 413 Payload Too Large.
        Normal operations such as the Redfish Validator still work.

Change-Id: I185dab82c75b3c44d89102750fd03666659a57a6

### DIFF
--- a/http/http_connection.hpp
+++ b/http/http_connection.hpp
@@ -698,6 +698,13 @@ class Connection :
                 ip, res, method, value.base(), mtlsSession);
         }
 
+        if (!handleContentLengthError())
+        {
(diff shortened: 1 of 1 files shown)


abiasojo (approved):
Look okay to me

joseph-reynolds (approved):
Looks okay to me.

PR #1440, opened 2026-04-22
1210: Move body limit enforcement (#1439)

rfrandse:
#### Move body limit enforcement (#1439)
```
Make sure the body limit enforcement happens before the Expect:
100-continue check.

Tested: Verified the body limit check happens first, even when Expect:
        100-continue is used. Now rejects with 413 Payload Too Large.
        Normal operations such as the Redfish Validator still work.

Change-Id: I185dab82c75b3c44d89102750fd03666659a57a6

Signed-off-by: Gunnar Mills <gmills@us.ibm.com>```

### DIFF
--- a/http/http_connection.hpp
+++ b/http/http_connection.hpp
@@ -698,6 +698,13 @@ class Connection :
                 ip, res, method, value.base(), mtlsSession);
         }
 
+        if (!handleContentLengthError())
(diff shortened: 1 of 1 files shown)


baemyung (approved):
It looks good to me

PR #1450, opened 2026-04-29
Add BMC Dump when timeout

gtmills:
Have this defect 777850 where we are seeing the code update timeout. Have no idea what is slowing things down so this adds code to get a dump after it times out.

Tested: Lowered this timeout to 2 seconds on 1110 and see the dump generated. Did not test on 1120.

### DIFF
--- a/redfish-core/lib/update_service.hpp
+++ b/redfish-core/lib/update_service.hpp
@@ -430,6 +430,31 @@ inline void softwareInterfaceAdded(
     }
 }
 
+// TODO(Gunnar): Remove this 6/15/26
+inline void createBMCDump()
+{
+    std::vector<std::pair<std::string, std::variant<std::string, uint64_t>>>
+        createDumpParamVec;
+
+    createDumpParamVec.emplace_back(
+        "xyz.openbmc_project.Dump.Create.CreateParameters.OriginatorId",
+        "bmcweb-internal");
+    createDumpParamVec.emplace_back(
+        "xyz.openbmc_project.Dump.Create.CreateParameters.OriginatorType",
+        "xyz.openbmc_project.Common.OriginatedBy.OriginatorTypes.Internal");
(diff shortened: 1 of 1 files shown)


baemyung (approved):
It looks good to me

jnin-dev (approved):
This looks good to me as well

abiasojo (approved):
looks okay to me as it is only temporary

redfish-core/lib/update_service.hpp
```diff
@@ -430,6 +430,31 @@ inline void softwareInterfaceAdded(
     }
 }
 
+// TODO(Gunnar): Remove this 6/15/26
```
jeaaustx, line 433: Is there a story or something to remind to remove this for 1120 before shipping?

redfish-core/lib/update_service.hpp
```diff
@@ -430,6 +430,31 @@ inline void softwareInterfaceAdded(
...
+    crow::connections::systemBus->async_method_call(
+        [](const boost::system::error_code& ec) {
+            if (ec)
+            {
+                BMCWEB_LOG_ERROR("Failed to do a dump:{}", ec.value());
+                return;
+            }
+        },
+        "xyz.openbmc_project.Dump.Manager", "/xyz/openbmc_project/dump/bmc",
+        "xyz.openbmc_project.Dump.Create", "CreateDump", createDumpParamVec);
```
jeaaustx, line 455: Should DumpType be one of the params specified? From the PDI it is not clear what the default will be if none is specified.

jeaaustx (approved):

redfish-core/lib/update_service.hpp
gtmills: https://jsw.ibm.com/browse/PFEBMC-5647

PR #1453, opened 2026-05-02
1120: Prevent ReadOnly user to escate privilege to Admin

baemyung:

Tested:

```

echo ${testuser_token}

```

```

{
  "error": {
    "@Message.ExtendedInfo": [
      {
        "MessageArgs": [],
      }
    ],
  }
}
```

### DIFF
--- a/redfish-core/lib/account_service.hpp
+++ b/redfish-core/lib/account_service.hpp
@@ -2754,6 +2754,19 @@ inline bool handleOemPatch(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
     }
     return true;
 }
+
+inline bool ensurePrivilegedProperties(
+    const std::optional<std::string>& newUserName,
+    const std::optional<std::string>& password,
+    const std::optional<bool>& enabled,
+    const std::optional<std::string>& roleId, const std::optional<bool>& locked,
+    const std::optional<std::vector<std::string>>& accountTypes,
+    const std::optional<std::vector<std::string>>& mfaBypass)
+{
+    return !(newUserName || password || enabled || roleId || locked ||
+             accountTypes || mfaBypass);
+}
+
 inline void handleAccountPatch(
     App& app, const crow::Request& req,
     const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
@@ -2792,6 +2805,15 @@ inline void handleAccountPatch(
         // If user is service
         if (oem)
         {
+            if (!ensurePrivilegedProperties(newUserName, password, enabled,
+                                            roleId, locked, accountTypes,
+                                            mfaBypass))
+            {
+                BMCWEB_LOG_WARNING(
+                    "Unauthenticated users are never allowed to PATCH anything other than the \"Oem\" property");
+                messages::insufficientPrivilege(asyncResp->res);
+                return;
+            }
             handleOemPatch(asyncResp, *oem, req, username, true);
             return;
         }
@@ -2810,6 +2832,15 @@ inline void handleAccountPatch(
     {
         if (oem)
         {
+            if (!ensurePrivilegedProperties(newUserName, password, enabled,
+                                            roleId, locked, accountTypes,
+                                            mfaBypass))
+            {
+                BMCWEB_LOG_WARNING(
+                    "Unauthenticated users are never allowed to PATCH anything other than the \"Oem\" property");
+                messages::insufficientPrivilege(asyncResp->res);
+                return;
+            }
             handleOemPatch(asyncResp, *oem, req, username, true);
             return;
         }
@@ -2820,16 +2851,17 @@ inline void handleAccountPatch(
             messages::insufficientPrivilege(asyncResp->res);
             return;
         }
-        // Read only users should not be allowed to bypass self
-        if (mfaBypass)
+        // Read only users should not be allowed to bypass self, except the
+        // password change
+        if (!ensurePrivilegedProperties(newUserName, std::nullopt, enabled,
+                                        roleId, locked, accountTypes,
+                                        mfaBypass))
         {
             BMCWEB_LOG_WARNING(
-                "User has insufficient privilege to bypass self");
+                "User does not have authority to PATCH anything other than Password");
             messages::insufficientPrivilege(asyncResp->res);
             return;
         }
-        // NOTE: password was already obtained from the previous
-        // readJsonPatch().
     }
 
     // For accounts which have a Restricted Role, restrict which properties


redfish-core/lib/account_service.hpp
```diff
@@ -2810,6 +2818,14 @@ inline void handleAccountPatch(
     {
         if (oem)
```
gtmills, line 2833: Why is this under oem ?

gtmills: I guess you have a 2nd check below ?

baemyung: > Why is this under oem ?

Similar to L2795, in case ReadOnly user tries to patch `UnauthenticatedACFUpload`  and the other fields (e.g. `RoleId`) together, it would need to be prevented.  

Only `UnauthenticatedACFUpload` can be allowed (if dbus `allow_unauth_upload` is set).

This logic is to check that only Oem/ACF can be specified in this condition.

baemyung:

redfish-core/lib/account_service.hpp
abhilashraju, line 3002: Can we address this issue? Keeping properties with known security vulnerabilities under the same privilege level is effectively a Trojan horse. Splitting the path into two Redfish APIs with different privilege levels would significantly improve security.

baemyung: I think this may not be a scope of this issue.
I'm also not sure how to do and whether the upstream community is ready for addressing it or someone is ready for contributing it.

redfish-core/lib/account_service.hpp
```diff
@@ -2792,6 +2792,14 @@ inline void handleAccountPatch(
         // If user is service
         if (oem)
         {
+            if (newUserName || password || enabled || roleId || locked ||
+                accountTypes || mfaBypass)
+            {
+                BMCWEB_LOG_WARNING(
+                    "User has insufficient privilege to bypass self");
+                messages::insufficientPrivilege(asyncResp->res);
```
joseph-reynolds, line 2814: This log message should be more like:
Unauthenticated users are never allowed to PATCH anything other than the "Oem" property.

redfish-core/lib/account_service.hpp
```diff
@@ -2810,6 +2818,14 @@ inline void handleAccountPatch(
...
+                return;
+            }
```
joseph-reynolds, line 2843: This whole "if" block (8 lines) should be moved up, so the test logic is:
if (!userHasConfigureUsers)
{
    if (oem) ...
    if (newUserName || password || ... || mfaBypass) ...
}

redfish-core/lib/account_service.hpp
```diff
@@ -2810,6 +2818,14 @@ inline void handleAccountPatch(
     {
         if (oem)
         {
+            if (newUserName || password || enabled || roleId || locked ||
+                accountTypes || mfaBypass)
+            {
+                BMCWEB_LOG_WARNING(
+                    "User has insufficient privilege to bypass self");
+                messages::insufficientPrivilege(asyncResp->res);
```
joseph-reynolds, line 2841: This log message should be more like:
User does not have authority to PATCH anything other than Oem or Password.

redfish-core/lib/account_service.hpp
```diff
@@ -2810,6 +2818,14 @@ inline void handleAccountPatch(
     {
         if (oem)
         {
+            if (newUserName || password || enabled || roleId || locked ||
```
joseph-reynolds: Remove "Password" from this test.  Why?  if (!userHasConfigureUsers) then that user must be allowed to PATCH their own password (but no other account's password).

redfish-core/lib/account_service.hpp
```diff
@@ -2820,16 +2836,16 @@ inline void handleAccountPatch(
...
-        // Read only users should not be allowed to bypass self
-        if (mfaBypass)
+        // Read only users should not be allowed to bypass self, except the
+        // password change
+        if (newUserName || enabled || roleId || locked || accountTypes ||
+            mfaBypass)
         {
             BMCWEB_LOG_WARNING(
                 "User has insufficient privilege to bypass self");
             messages::insufficientPrivilege(asyncResp->res);
```
joseph-reynolds, line 2862: This log message should be more like:
User does not have authority to PATCH anything other than Password.

joseph-reynolds (changes requested):
I think there needs to be a logic change, for a ReadOnly user trying to change their own password.

redfish-core/lib/account_service.hpp
joseph-reynolds: > Can we address this issue? Keeping properties with known security vulnerabilities under the same privilege level is effectively a Trojan horse. Splitting the path into two Redfish APIs with different privilege levels would significantly improve security.

Having "Oem.IBM.ACF" is adding a lot of complexity (and bugs!) in this area.  I agree with Abhilash's idea.  Can we move the ACF properties out of here, and add it somewhere else, maybe under /ibm/v1.  Can we do this for P12?

redfish-core/lib/account_service.hpp
joseph-reynolds: Oh, I now see what this code was doing.  If "Oem" is PATCHed, then you can't PATCH anything else.  That works, I guess.

redfish-core/lib/account_service.hpp
joseph-reynolds: Oh, I now see what this code was doing. If "Oem" is PATCHed, then you can't PATCH anything else. Please disregard my comment on this.

redfish-core/lib/account_service.hpp
baemyung: Where to?
The current logic is first handling the case "session==nullptr", and the case of !userHasConfigUsers (which uses session).

```

    // Unauthenticated user
    if (req.session == nullptr)
    {
==> Part A
        // If user is service
        if (oem)  
        {
            if (newUserName || password || enabled || roleId || locked ||
                accountTypes || mfaBypass)
        .....
        }
        messages::insufficientPrivilege(asyncResp->res);
        return;
    }

    bool userSelf = (username == req.session->username);
    Privileges effectiveUserPrivileges =
        redfish::getUserPrivileges(*req.session);
    Privileges configureUsers = {"ConfigureUsers"};
    bool userHasConfigureUsers =
        effectiveUserPrivileges.isSupersetOf(configureUsers) &&
        !req.session->isConfigureSelfOnly;
    if (!userHasConfigureUsers)
    {
==> Part B
        if (oem)
        {
            if (newUserName || password || enabled || roleId || locked ||
                accountTypes || mfaBypass)
        .....
        }
```

I'm not sure whether we can combine parts A & B together?

baemyung:
> I think there needs to be a logic change, for a ReadOnly user trying to change their own password.

ReadOnly user would be allowed to change only Password (not together with Oem).
In that case, it will go thru

```
L2817: if (!userHasConfigureUsers)
{
      if(oem) {
          // This path will do only "Oem" change (not password)
         ...
         return;
      }
==> This path will do only "Password" change
        // ConfigureSelf accounts can only modify their own account
        if (!userSelf)
        {
            messages::insufficientPrivilege(asyncResp->res);
            return;
        }
        // Read only users should not be allowed to bypass self, except the
        // password change
        if (newUserName || enabled || roleId || locked || accountTypes ||
            mfaBypass)
        {
            BMCWEB_LOG_WARNING(
                "User does not have authority to PATCH anything other than Password");
            messages::insufficientPrivilege(asyncResp->res);
            return;
        }

==> Fall thru if "Password" is specified ==> Go thru "Password" change
    }
}
```

Anyway, I'll test it again

baemyung:
> Anyway, I'll test it again

Oem/ACF Patch with non-service user is rejected inside `handleOemPatch()` (with `Oem` field).  This part is not changed.

What's changed is that to verified the parameters - whether Oem & Password (for non-service user) are together.  If so, the new code is rejected (with `Forbidden`*) before going into `handleOemPatch()`. 
Previously it is rejected normally as `Unauthorized`.

Both codes (before and after) are rejecting , but the msg is different.

Before FIX -- rejected as `Unauthorized`

```
curl -k -i -H "Content-Type: application/json" -H "X-Auth-Token: ${testuser_token}" -X PATCH "https://${bmc}/redfish/v1/AccountService/Accounts/testuser" -d '{"Password":"TestPd123", "Oem":{"ACF":{"AllowUnauthACFUpload":true}}}'
HTTP/1.1 401 Unauthorized
...
{
  "error": {
    "@Message.ExtendedInfo": [
      {
        "Message": "While accessing the resource at '/redfish/v1/AccountService/Accounts/testuser', the service received an authorization error 'Oem properties access not allowed by non service user'.",
        "MessageArgs": [
          "/redfish/v1/AccountService/Accounts/testuser",
          "Oem properties access not allowed by non service user"
        ],
        "MessageId": "Base.1.19.ResourceAtUriUnauthorized",
        "Resolution": "Ensure that the appropriate access is provided for the service in order for it to access the URI."
      }
    ],
    "code": "Base.1.19.ResourceAtUriUnauthorized",
    "message": "While accessing the resource at '/redfish/v1/AccountService/Accounts/testuser', the service received an authorization error 'Oem properties access not allowed by non service user'."
  }
}root@p11bmc:/tmp#


```

After FIX - -- rejected as `Forbidden` (the parameters are not acceptable with ReadOnly).

```
curl -k -i -H "Content-Type: application/json" -H "X-Auth-Token: ${testuser_token}" -X PATCH "https://${bmc}/redfish/v1/AccountService/Accounts/testuser" -d '{"Password":"TestPwd123", "Oem":{"ACF":{"AllowUnauthACFUpload":true}}}'
HTTP/1.1 403 Forbidden
...
{
  "error": {
    "@Message.ExtendedInfo": [
      {
        "MessageArgs": [],
      }
    ],
  }
```

redfish-core/lib/account_service.hpp
abhilashraju: Isn't it a security defect in redfish API? Or it is just an OEM property defect where we use wrong redfish API for un-secure properties? Ideally all properties in a single patch request should be applied with same security privilege.

joseph-reynolds: > it is just an OEM property defect

Redfish has the concept of a Property Override so one property can require a different authority thaan the others.  Redfish also is struggling with the "Password" property which only the user itself should be allowed to change.  In this way, properties in the same PATCH request can require different authorities.

In general the "ManagerAccount" type (which represents a BMC local user account) is really very complicated.  And we are abusing it a bit by adding even more Oem properties.  So that's why I thnk we should move our Oem.IBM properties out of this.  Moving them within a release family may lead to more bugs, so I vote for doing this in P12.

redfish-core/lib/account_service.hpp
joseph-reynolds: Thanks for explaining.  The code looks correct as it is.

joseph-reynolds:
I got it now.  I was confused by the multiple paths testing overlapping things.
The code looks correct.

joseph-reynolds (approved):
I approve this very complicated set of code.

redfish-core/lib/account_service.hpp
```diff
@@ -2792,6 +2792,14 @@ inline void handleAccountPatch(
         // If user is service
         if (oem)
         {
+            if (newUserName || password || enabled || roleId || locked ||
```
abhilashraju: can we take it out in a function ? the readability will improve

baemyung: I added a function `checkIfParamHasValue()` to enforce the list of arguments.

baemyung:

redfish-core/lib/account_service.hpp
jeaaustx, line 2801: My only concern is that any new PATCH properties get captured in the check for future rebases.

Suggestion is to call ensurePrivilegeProperties() here and capture return value in a local variable. Then check the local variable in the locations below.

This would make it easeir and more likely to confirm all the privileged properties are checked if a new property is added.

jeaaustx (approved):

joseph-reynolds (approved):

jnin-dev (approved):

PR #1456, opened 2026-05-03
1060: Change a few message responses as errors

baemyung:
Some of messages are not currently reported as errors, although they are the results as Redfish request errors. Those include

- ResourceAlreadyExists
- PropertyValueFormatError

For example, PropertyValueFormatError is currently not treated as an error. It shows like

```

curl -k -X PATCH "${bmc}/redfish/v1/AccountService/Accounts/admin" \
  -H "Content-Type: application/json" \
  -d '{ "Password": "" }'

  "Password@Message.ExtendedInfo": [
    {
      "Message": "The value 'null' for the property Password is not a format that the property can accept.",
      "MessageArgs": [
        "null",
        "Password"
      ],
      "MessageId": "Base.1.19.PropertyValueFormatError",
      "MessageSeverity": "Warning",
      "Resolution": "Correct the value for the property in the request body and resubmit the request if the operation failed."
    }
  ]
```

After making it as an error, it will be like
```
{
  "error": {
    "@Message.ExtendedInfo": [
      {
        "Message": "The value 'null' for the property Password is not a format that the property can accept.",
        "MessageArgs": [
          "null",
          "Password"
        ],
        "MessageId": "Base.1.19.PropertyValueFormatError",
        "MessageSeverity": "Warning",
        "Resolution": "Correct the value for the property in the request body and resubmit the request if the operation failed."
      }
    ],
    "code": "Base.1.19.PropertyValueFormatError",
    "message": "The value 'null' for the property Password is not a format that the property can accept."
  }
}
```

Tested:
- Check those response messages

Change-Id: I746c55e42e8f0cde205a3800e3da17cd78cf7e34

### DIFF
--- a/redfish-core/src/error_messages.cpp
+++ b/redfish-core/src/error_messages.cpp
@@ -402,8 +402,8 @@ void resourceAlreadyExists(crow::Response& res, std::string_view arg1,
                            std::string_view arg2, std::string_view arg3)
 {
     res.result(boost::beast::http::status::bad_request);
-    addMessageToJson(res.jsonValue, resourceAlreadyExists(arg1, arg2, arg3),
-                     arg2);
+    addMessageToErrorJson(res.jsonValue,
+                          resourceAlreadyExists(arg1, arg2, arg3));
 }
 
 /**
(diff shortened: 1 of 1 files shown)


gtmills:
What is the value of backporting to 1060, 1110  ?

baemyung:
> What is the value of backporting to 1060, 1110 ?

I thought this may may also be an issue for 1060, but I found that the slack communication suggests to be done on 1120 and onward.
So, I'll cancel 1060/1110 changes.

PR #1462, opened 2026-05-06
Audit Log: Harden addition of detail data

jeaaustx:
Exclude additional routes which might contain secret data.

Tested:
 - Confirmed audit log entries for these paths do not contain the detail data:
```
   - PATCH /redfish/v1/Systems/system
   - PATCH /redfish/v1/EventService/Subscriptions/{}
   - PATCH /redfish/v1/AccountService
   - POST /redfish/v1/EventService/Subscriptions
   - POST /redfish/v1/CertificateService/Actions/CertificateService.GenerateCSR
   - POST /redfish/v1/Managers/bmc/NetworkProtocol/HTTPS/Certificates
   - POST /redfish/v1/AccountService/LDAP/Certificates
   - POST /redfish/v1/Managers/bmc/Truststore/Certificates
   - POST /redfish/v1/LicenseService/Licenses
```

Fixes STG Defect 780305

### DIFF
--- a/http/http_request.hpp
+++ b/http/http_request.hpp
@@ -81,6 +81,17 @@ struct Request : std::enable_shared_from_this<Request>
         ipAddress = boost::asio::ip::address();
         session = nullptr;
         userRole = "";
+        skipAuditDetailFlag = false;
+    }
+
+    void setSkipAuditDetail(bool skip) const
+    {
+        skipAuditDetailFlag = skip;
+    }
+
+    bool skipAuditDetail() const
+    {
+        return skipAuditDetailFlag;
     }
 
     void clearBody()
@@ -171,6 +182,8 @@ struct Request : std::enable_shared_from_this<Request>
     }
 
   private:
+    mutable bool skipAuditDetailFlag = false;
+
     bool setUrlInfo()
     {
         auto result = boost::urls::parse_relative_ref(target());
--- a/src/audit_events.cpp
+++ b/src/audit_events.cpp
@@ -1,7 +1,5 @@
 #include "audit_events.hpp"
 
-#include "bmcweb_config.h"
-
 #include "http_request.hpp"
 #include "logging.hpp"
 #include "websocket.hpp"
@@ -119,10 +117,6 @@ void auditSetState(bool enable)
  */
 inline bool checkSkipDetail(const crow::Request& req)
 {
-    const std::string systemPatch =
-        std::format("/redfish/v1/Systems/{}", BMCWEB_REDFISH_SYSTEM_URI_NAME);
-    const std::string systemPatchTrailing = std::format("{}/", systemPatch);
-
     return req.target().starts_with("/redfish/v1/AccountService") ||
            req.target().starts_with("/redfish/v1/CertificateService") ||
            req.target().starts_with("/redfish/v1/EventService/Subscriptions") ||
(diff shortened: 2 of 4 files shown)


http/http_request.hpp
```diff
@@ -81,6 +81,17 @@ struct Request : std::enable_shared_from_this<Request>
...
         userRole = "";
+        skipAuditDetailFlag = false;
+    }
+
+    void setSkipAuditDetail(bool skip) const
+    {
+        skipAuditDetailFlag = skip;
+    }
+
+    bool skipAuditDetail() const
```
baemyung, line 92: What if named like `getSkipAuditDetail()` [as a pair to `setSkipAuditDetail()`]?

src/audit_events.cpp
```diff
@@ -117,11 +117,14 @@ void auditSetState(bool enable)
...
+    return req.target().starts_with("/redfish/v1/AccountService") ||
+           req.target().starts_with("/redfish/v1/CertificateService") ||
+           req.target().starts_with("/redfish/v1/EventService/Subscriptions") ||
+           req.target().starts_with("/redfish/v1/LicenseService/Licenses") ||
            req.target().starts_with("/redfish/v1/UpdateService") ||
            req.target().starts_with("/ibm/v1") ||
            ((req.method() == boost::beast::http::verb::post) &&
-            (checkPostUser(req) ||
+            (checkPostUser(req) || req.target().contains("Certificates") ||
              req.target().contains("LogService.CollectDiagnosticData")));
```
baemyung, line 128: Similar to the case  of `setChapData()` [in `Systems/system'], perhaps can we do the same way for the other routes? Maybe too many entry points?

Or, can we break this part into 2 if-statements for better readability like?

```
inline bool checkSkipDetail(const crow::Request& req)
{
    if (req.target().starts_with("/redfish/v1/AccountService") ||
        req.target().starts_with("/redfish/v1/CertificateService") ||
        req.target().starts_with("/redfish/v1/EventService/Subscriptions") ||
        req.target().starts_with("/redfish/v1/LicenseService/Licenses") ||
        req.target().starts_with("/redfish/v1/UpdateService") ||
        req.target().starts_with("/ibm/v1"))
    {
        return true;
    }

    if (req.method() == boost::beast::http::verb::post)
    {
        return checkPostUser(req) || req.target().contains("Certificates") ||
               req.target().contains("LogService.CollectDiagnosticData");
    }
    return false;
}
```

jeaaustx: If audit log were upstream this is how I would want to handle all of these routes. Having to carry the downstream patches I was trying to keep the audit log hooks in the mainstream path to a minimum to make rebasing easier and less error prone (with merge conflicts, etc.).

For the setChapData() that is already a downstream patch, so carrying the audit log hook in there seems not very burdensome for rebasing.

I'll take a look at the other ones to see if there are more that would make sense to handle in this manner as well.

Big question back to you is whether it seems long-term problematic to have this second method of specifying to skip the detail information? I don't want to introduce a maintenance nightmare.

baemyung: My suggested the code is just a rearrangement of `checkSkipDetail()` body.
Either way would be fine with me

baemyung (approved):
It looks good with this

gtmills (approved):

jeaaustx:
I used the [new Redfish performance analysis tool](https://github.ibm.com/openbmc/openbmc-tools/pull/59) to see if adding these additional checks would have a detrimental performance affect. I did not obseve any degradation in performance. So I am going to merge this as-is.

http/http_request.hpp
jeaaustx: I had started with the parallel function names but changed the get version since it "read" better when being used. I'm not really seeing other use cases than just within and if() check so I'd rather leave as-is.

PR #1463, opened 2026-05-07
1210: Prevent ReadOnly user to escate privilege to Admin

baemyung:

Tested:

```

echo ${testuser_token}

```

```

{
  "error": {
    "@Message.ExtendedInfo": [
      {
        "MessageArgs": [],
      }
    ],
  }
}
```

### DIFF
--- a/redfish-core/lib/account_service.hpp
+++ b/redfish-core/lib/account_service.hpp
@@ -2754,6 +2754,19 @@ inline bool handleOemPatch(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
     }
     return true;
 }
+
+inline bool ensurePrivilegedProperties(
+    const std::optional<std::string>& newUserName,
+    const std::optional<std::string>& password,
+    const std::optional<bool>& enabled,
+    const std::optional<std::string>& roleId, const std::optional<bool>& locked,
+    const std::optional<std::vector<std::string>>& accountTypes,
+    const std::optional<std::vector<std::string>>& mfaBypass)
+{
+    return !(newUserName || password || enabled || roleId || locked ||
+             accountTypes || mfaBypass);
+}
+
 inline void handleAccountPatch(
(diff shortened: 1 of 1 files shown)


baemyung:

redfish-core/lib/account_service.hpp
```diff
@@ -2754,6 +2754,19 @@ inline bool handleOemPatch(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
...
     return true;
 }
+
+inline bool ensurePrivilegedProperties(
+    const std::optional<std::string>& newUserName,
+    const std::optional<std::string>& password,
+    const std::optional<bool>& enabled,
+    const std::optional<std::string>& roleId, const std::optional<bool>& locked,
+    const std::optional<std::vector<std::string>>& accountTypes,
+    const std::optional<std::vector<std::string>>& mfaBypass)
```
jeaaustx, line 2764: My only concern here is if a new property is added to the patch that this function gets updated to check for it as well.

I don't have a good solution and I don't expect a new property to be very likely. Maybe just add a comment or something where the readJsonPatch() call is made?

jeaaustx (approved):
Looks good, just one comment on maintaining the full list of options for the new function.

redfish-core/lib/account_service.hpp
jeaaustx, line 2801: I have a new idea for trying to prevent an additional property being missed in a future rebase.

What if for 1210 you put a call here to the ensurePrivilegedProperties function and save the result in a local variable.  Then below just check the value of the local variable.

That way the checking of the properties is right next to where the properties are extracted and makes it more likely that any new property would be checked for as well.

jnin-dev (approved):
Looks good to me, not seeing anything wrong

joseph-reynolds (approved):

gtmills (approved):
Defaulting to others here

PR #1464, opened 2026-05-08
1110: Prevent ReadOnly user to escate privilege to Admin

baemyung:

Tested:

```

echo ${testuser_token}

```

```

{
  "error": {
    "@Message.ExtendedInfo": [
      {
        "MessageArgs": [],
      }
    ],
  }
}
```

### DIFF
--- a/redfish-core/lib/account_service.hpp
+++ b/redfish-core/lib/account_service.hpp
@@ -2512,6 +2512,18 @@ inline void
         "xyz.openbmc_project.Object.Delete", "Delete");
 }
 
+inline bool ensurePrivilegedProperties(
+    const std::optional<std::string>& newUserName,
+    const std::optional<std::string>& password,
+    const std::optional<bool>& enabled,
+    const std::optional<std::string>& roleId, const std::optional<bool>& locked,
+    const std::optional<std::vector<std::string>>& accountTypes,
+    const std::optional<std::vector<std::string>>& mfaBypass)
+{
+    return !(newUserName || password || enabled || roleId || locked ||
+             accountTypes || mfaBypass);
+}
+
 inline void
     handleAccountPatch(App& app, const crow::Request& req,
                        const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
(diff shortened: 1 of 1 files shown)


baemyung:

redfish-core/lib/account_service.hpp
```diff
@@ -2559,6 +2570,15 @@ inline void
         {
             if (oem)
             {
+                if (checkIfParamHasValue(newUserName, password, enabled, roleId,
```
abhilashraju: why don't we include the warning and messages too inside the function and rename function accordingly. something like 
if (ensurePrivilegedProp()){
    // proceedWith operation.
}

baemyung: I can change the function name, but I think the message may be better to be outside as the msgs are different -- L2633.

redfish-core/lib/account_service.hpp
```diff
@@ -2512,6 +2512,18 @@ inline void
...
 
+inline bool ensurePrivilegedProperties(
+    const std::optional<std::string>& newUserName,
+    const std::optional<std::string>& password,
+    const std::optional<bool>& enabled,
+    const std::optional<std::string>& roleId, const std::optional<bool>& locked,
+    const std::optional<std::vector<std::string>>& accountTypes,
+    const std::optional<std::vector<std::string>>& mfaBypass)
+{
+    return !(newUserName || password || enabled || roleId || locked ||
```
jeaaustx, line 2523: My only concern here is if a new property is added to the patch that this function gets updated to check for it as well. (Biggest risk is if the property is added upstream and then inherited by a rebase.)

I don't have a good solution and I don't expect a new property to be very likely. Maybe just add a comment or something where the readJsonPatch() call is made? Or maybe for this function itself so that on rebase it is considered and confirmed no new properties need to be included?

jeaaustx (approved):

joseph-reynolds (approved):

jnin-dev (approved):

PR #1465, opened 2026-05-08
1060: Prevent ReadOnly user to escate privilege to Admin

baemyung:

Tested:

```

echo ${testuser_token}

```

```

{
  "error": {
    "@Message.ExtendedInfo": [
      {
        "MessageArgs": [],
      }
    ],
  }
}
```

### DIFF
--- a/redfish-core/lib/account_service.hpp
+++ b/redfish-core/lib/account_service.hpp
@@ -2463,6 +2463,17 @@ inline void
         "xyz.openbmc_project.Object.Delete", "Delete");
 }
 
+inline bool ensurePrivilegedProperties(
+    const std::optional<std::string>& newUserName,
+    const std::optional<std::string>& password,
+    const std::optional<bool>& enabled,
+    const std::optional<std::string>& roleId, const std::optional<bool>& locked,
+    const std::optional<std::vector<std::string>>& accountTypes)
+{
+    return !(newUserName || password || enabled || roleId || locked ||
+             accountTypes);
+}
+
 inline void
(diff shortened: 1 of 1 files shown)


baemyung:

redfish-core/lib/account_service.hpp
```diff
@@ -2503,6 +2514,15 @@ inline void
         {
             if (oem)
             {
+                if (!ensurePrivilegedProperties(newUserName, password, enabled,
+                                                roleId, locked, accountType))
+
+                {
+                    BMCWEB_LOG_WARNING(
```
abhilashraju: still, these are repeated code which can go inside the ensurePrivilegedProperties right?

abhilashraju: did not see the message difference

abhilashraju (approved):

redfish-core/lib/account_service.hpp
baemyung: There are 2 different types of messages.
1) L2517 & L2554 - Oem only for service user - "Unauth users not allowed other than Oem field"
2) L2578 - ReadOnly user can only change its Password - "User does not have authority to change other than Password".

If we put the messages into `ensure()` function, there may be 2 diff functions?

I think - for this case, the main advantage of having a function is to enforce the parameter checks (instead of using `||`). 

If it is still better to put the messages into the function, I can do that - perhaps making 2 ensure functions.

jeaaustx (approved):

jnin-dev (approved):

joseph-reynolds (approved):

PR #1469, opened 2026-05-12
http: Refactor mTLS verification to post-handshake

abhilashraju:
Move mTLS user verification from tlsVerifyCallback to after SSL handshake completion. This change improves the verification flow by:

- Moving verifyMtlsUser call from tlsVerifyCallback to after handshake
- Changing verifyMtlsUser to accept SSL* instead of verify_context
- Using SSL_get_verify_result() and SSL_get1_peer_certificate() APIs
- Adding proper certificate memory management with X509_free()
- Implementing case-insensitive DNS domain comparison per RFC 4343
- Removing premature verification during handshake callback
- Changing handshake failure log from ERROR to WARNING level

This approach ensures certificate verification is complete before extracting user information and properly manages certificate lifecycle.

These changes are cherry-picked from the latest upstream bmcweb to improve mTLS authentication handling(ce3522aab3337e9a9c6ee5b439417317faa91b4b).

Tested: Verified mTLS authentication works correctly with client certificates and session establishment succeeds.

Change-Id: Ib4a7d299c9b542d4b21e70ba00ea0219

### DIFF
--- a/http/http_connection.hpp
+++ b/http/http_connection.hpp
@@ -110,19 +110,10 @@ class Connection :
     Connection& operator=(Connection&&) = delete;
 
     bool tlsVerifyCallback(bool preverified,
-                           boost::asio::ssl::verify_context& ctx)
+                           boost::asio::ssl::verify_context& /*ctx*/)
     {
         BMCWEB_LOG_DEBUG("{} tlsVerifyCallback called with preverified {}",
                          logPtr(this), preverified);
-        if (preverified)
-        {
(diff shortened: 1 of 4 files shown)


baemyung:
As this PR seems cherry-picking the upstream commits, I would be better to show the list of upstream commit links.

abhilashraju:
> As this PR seems cherry-picking the upstream commits, I would be better to show the list of upstream commit links.

done

baemyung (approved):
It looks good to me (on cherry-picking from upstream)

PR #1470, opened 2026-05-12
redfish: Add Protocol and MTLS support to aggregator

abhilashraju:
Add support for configurable Protocol property in satellite BMC configuration. This allows explicit protocol specification instead of deriving it solely from AuthType.

Enhance AuthType handling to support MTLS authentication:
- Set scheme to 'http' for AuthType 'None'
- Set scheme to 'https' for AuthType 'MTLS'
- Maintain error handling for unsupported AuthType values

Remove redundant debug logging and comments in aggregation check to improve code clarity.
Upstream : https://gerrit.openbmc.org/c/openbmc/bmcweb/+/71685 
Tested: Verified satellite BMC connections with both None and MTLS authentication types work correctly with appropriate protocol schemes.

Change-Id: I5AE8D5A488794AF8A461582B9B69CA02

### DIFF
--- a/redfish-core/include/redfish_aggregator.hpp
+++ b/redfish-core/include/redfish_aggregator.hpp
@@ -514,6 +514,17 @@ class RedfishAggregator
                 }
                 url.set_port(std::to_string(static_cast<uint16_t>(*propVal)));
             }
+            else if (prop.first == "Protocol")
+            {
+                const std::string* propVal =
+                    std::get_if<std::string>(&prop.second);
+                if (propVal == nullptr)
+                {
+                    BMCWEB_LOG_ERROR("Invalid Protocol value");
+                    return;
+                }
+                url.set_scheme(*propVal);
+            }
             else if (prop.first == "AuthType")
             {
                 const std::string* propVal =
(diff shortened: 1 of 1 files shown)


baemyung:
Would this be upstreamed?  
If so, I think it would be better to submit a upstream commit and also put into the downstream.

abhilashraju:
> Would this be upstreamed? If so, I think it would be better to submit a upstream commit and also put into the downstream.

yes there is one https://gerrit.openbmc.org/c/openbmc/bmcweb/+/71685 . Updated the commit summary with the same

baemyung (approved):
It looks good to me

redfish-core/include/redfish_aggregator.hpp
```diff
@@ -526,14 +537,21 @@ class RedfishAggregator
...
+                    url.set_scheme("http");
+                }
+                else if (*propVal == "MTLS")
+                {
+                    url.set_scheme("https");
+                }
+                else
                 {
                     BMCWEB_LOG_ERROR(
                         "Unsupported AuthType value: {}, only \"None\" is supported",
```
abiasojo, line 551: Does this error message need to include MTLS?

PR #1472, opened 2026-05-20
Add Redfish API support for SendServiceAlerts setting

devenrao:
This commit adds support for controlling the hardware isolation service alerts setting through the Redfish API under the IBM OEM extension.

Changes:
- Created service_alerts.hpp with GET/SET functions for the SendServiceAlerts property
- Modified systems.hpp to expose SendServiceAlerts in the IBM OEM section of ComputerSystem resource
- Added error handling for missing D-Bus object (EBADR) to prevent redfish validator GET failure 500

The SendServiceAlerts property allows users to enable or disable service alerts for hardware isolation events via Redfish API:
- GET /redfish/v1/Systems/system returns current state
- PATCH /redfish/v1/Systems/system with Oem.IBM.SendServiceAlerts updates the setting

The implementation integrates with the existing D-Bus setting at /xyz/openbmc_project/hardware_isolation/send_service_alerts which is monitored by the faultlog service.

Tested:
- GET request returns current SendServiceAlerts state
- PATCH request successfully updates the D-Bus property
- Setting persists across BMC reboots via phosphor-settings-manager
- Redfish validator passes without 500 errors

Change-Id: Ib6a488f1697cbdc28e3361f92ead70f38d61ba41

### DIFF
--- a/redfish-core/lib/oem/ibm/service_alerts.hpp
+++ b/redfish-core/lib/oem/ibm/service_alerts.hpp
@@ -1,0 +1,92 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright OpenBMC Authors
+#pragma once
+
+#include "async_resp.hpp"
+#include "dbus_singleton.hpp"
+#include "dbus_utility.hpp"
+#include "error_messages.hpp"
+#include "logging.hpp"
+
+#include <asm-generic/errno.h>
+
+#include <boost/system/error_code.hpp>
+#include <sdbusplus/asio/property.hpp>
+
+#include <memory>
+
+namespace redfish
+{
+
+/**
+ * @brief Get service alerts enabled state
+ *
+ * @param[in] asyncResp     Shared pointer for generating response message.
+ *
+ * @return None.
+ */
+inline void
+    getServiceAlertsEnabled(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp)
+{
+    BMCWEB_LOG_DEBUG << "Get service alerts enabled state";
+
+    sdbusplus::asio::getProperty<bool>(
+        *crow::connections::systemBus, "xyz.openbmc_project.Settings",
+        "/xyz/openbmc_project/hardware_isolation/send_service_alerts",
+        "xyz.openbmc_project.Object.Enable", "Enabled",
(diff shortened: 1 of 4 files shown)


redfish-core/lib/oem/ibm/service_alerts.hpp
```diff
@@ -0,0 +1,91 @@
...
+    BMCWEB_LOG_DEBUG << "Set service alerts enabled state to: " << enabled;
+    sdbusplus::asio::setProperty(
+        *crow::connections::systemBus, "xyz.openbmc_project.Settings",
+        "/xyz/openbmc_project/hardware_isolation/send_service_alerts",
+        "xyz.openbmc_project.Object.Enable", "Enabled", enabled,
+        [asyncResp](const boost::system::error_code& ec) {
+        if (ec)
+        {
+            if (ec.value() == EBADR)
+            {
```
jeaaustx, line 79: For PATCH if the object does not exist it should be treated as an error. Can use a more specific error here than internalError.

redfish-core/lib/oem/ibm/service_alerts.hpp
```diff
@@ -0,0 +1,91 @@
...
+                return;
+            }
+
+            BMCWEB_LOG_ERROR << "DBUS response error: " << ec.message();
+            messages::internalError(asyncResp->res);
+            return;
+        }
+
+        asyncResp->res.jsonValue["Oem"]["IBM"]["@odata.type"] =
+            "#IBMComputerSystem.v1_0_0.IBM";
```
jeaaustx: Is this correct for @odata.type for 1060?

Also, arent' the json and xml changes needed for 1060 as well. Seems like it is under a different path, maybe: https://github.com/ibm-openbmc/bmcweb/blob/1060/static/redfish/v1/JsonSchemas/OemComputerSystem/OemComputerSystem.json

jeaaustx (changes requested):

redfish-core/lib/oem/ibm/service_alerts.hpp
devenrao: done

redfish-core/lib/oem/ibm/service_alerts.hpp
devenrao: thank you fixed

jeaaustx (approved):

PR #1478, opened 2026-05-26
Revert dump on timeout and bump to 100 secs

gtmills:
1) Revert https://github.com/ibm-openbmc/bmcweb/pull/1450 as planned
2) Bump the timeout to 100 seconds for 1120 only since we have this hit the timeout twice. Haven't seen since I added that dump so :/ For 1120 going to rely on better hardware so 1120 only and not upstream.


baemyung (approved):
It looks good to me

abiasojo (approved):
It looks okay to me

jeaaustx (approved):

jnin-dev (approved):

PR #1479, opened 2026-05-27
1210: Print IP address on max connections

jnin-dev:
Have seen this a few times where the BMC is spammed by some client and bmcweb hits its max connections. Have also gotten feedback that the current "Max connection count exceeded." isn't very useful since it doesn't tell you who is spamming the BMC. These cases have not been malicious, instead an automated malfunctioning network agent.

Even if they were malicious recording/printing the IP address of the attacker is referenced by several documents on detecting and responding to a DoS attack. "see a surge in web traffic, seemingly out of nowhere, that’s coming from the same IP address or range." [1] "An IP address makes x requests over y seconds" [2]

OpenBMC's security considerations document discusses DoS.[3] bmcweb doesn't claim to have DoS protection and a best practice is for the BMC to be on a dedicated management network.

Network Analysis could be used to identify the offending client(s) but specially for after the fact, having the ip address recorded is more reliable.

The downside of this change is now when bmcweb is at the max connections, it is reading the IP Address and keeping the connection open slightly longer. In practice, did not see a measurable difference.

[1]: https://www.microsoft.com/en-us/security/business/security-101/what-is-a-ddos-attack
[2]: https://www.loggly.com/blog/ddos-monitoring-how-to-know-youre-under-attack/
[3]: https://github.com/openbmc/docs/blob/643e73605907251d80166d3d3c984456fc7fd599/security/network-security-considerations.md

Tested:
200+ concurrent connections with keep-alive true to leave connections. Sent
```
curl -v -k https://${bmc}/redfish/v1
```
Saw in logs
```
[http_connection.hpp:214] 0x1ab4c70 Max connection count exceeded. Request <IP>
```

Change-Id: Ib9225e0ee67caafe90b958eb89534adfb0831646

### DIFF
--- a/http/http_connection.hpp
+++ b/http/http_connection.hpp
@@ -204,10 +204,12 @@ class Connection :
     {
         BMCWEB_LOG_DEBUG("{} Connection started, total {}", logPtr(this),
                          connectionCount);
+        readClientIp();
         if (connectionCount >= 200)
         {
-            BMCWEB_LOG_CRITICAL("{} Max connection count exceeded.",
-                                logPtr(this));
+            BMCWEB_LOG_CRITICAL("{} Max connection count exceeded. Request {}",
+                                logPtr(this), ip.to_string());
+
             return;
         }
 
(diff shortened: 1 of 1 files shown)


jeaaustx (approved):

gtmills:
https://github.com/ibm-openbmc/bmcweb/pull/1483 is the same. One can go away

gtmills (approved):

PR #1480, opened 2026-05-27
1060: Print IP address on max connections

jnin-dev:
Have seen this a few times where the BMC is spammed by some client and bmcweb hits its max connections. Have also gotten feedback that the current "Max connection count exceeded." isn't very useful since it doesn't tell you who is spamming the BMC. These cases have not been malicious, instead an automated malfunctioning network agent.

Even if they were malicious recording/printing the IP address of the attacker is referenced by several documents on detecting and responding to a DoS attack. "see a surge in web traffic, seemingly out of nowhere, that’s coming from the same IP address or range." [1] "An IP address makes x requests over y seconds" [2]

OpenBMC's security considerations document discusses DoS.[3] bmcweb doesn't claim to have DoS protection and a best practice is for the BMC to be on a dedicated management network.

Network Analysis could be used to identify the offending client(s) but specially for after the fact, having the ip address recorded is more reliable.

The downside of this change is now when bmcweb is at the max connections, it is reading the IP Address and keeping the connection open slightly longer. In practice, did not see a measurable difference.

[1]: https://www.microsoft.com/en-us/security/business/security-101/what-is-a-ddos-attack
[2]: https://www.loggly.com/blog/ddos-monitoring-how-to-know-youre-under-attack/
[3]: https://github.com/openbmc/docs/blob/643e73605907251d80166d3d3c984456fc7fd599/security/network-security-considerations.md

Tested:
200+ concurrent connections with keep-alive true to leave connections. Sent
```
curl -v -k https://${bmc}/redfish/v1
```
Saw in logs
```
[http_connection.hpp:214] 0x1ab4c70 Max connection count exceeded. Request <IP>
```

Change-Id: Ib9225e0ee67caafe90b958eb89534adfb0831646

### DIFF
--- a/http/http_connection.hpp
+++ b/http/http_connection.hpp
@@ -293,7 +293,14 @@ class Connection :
     {
         if (connectionCount >= 200)
         {
-            BMCWEB_LOG_CRITICAL << this << "Max connection count exceeded.";
+            boost::asio::ip::address ip;
+            boost::system::error_code ec = getClientIp(ip);
+            if (ec)
+            {
+                return;
+            }
+            BMCWEB_LOG_CRITICAL << this << "Max connection count exceeded."
+                                << " Request " << ip.to_string();
             return;
         }
 


http/http_connection.hpp
```diff
@@ -293,7 +293,14 @@ class Connection :
     {
         if (connectionCount >= 200)
         {
-            BMCWEB_LOG_CRITICAL << this << "Max connection count exceeded.";
+            boost::asio::ip::address ip;
+            boost::system::error_code ec = getClientIp(ip);
```
jeaaustx, line 297: I'm wondering why not use the readClientIp() helper here? (I'm not clear that that would be preferred, maybe someone more familiar with this code can weigh in.)

http/http_connection.hpp
```diff
@@ -293,7 +293,14 @@ class Connection :
...
+            if (ec)
+            {
+                return;
+            }
+            BMCWEB_LOG_CRITICAL << this << "Max connection count exceeded."
+                                << " Request " << ip.to_string();
```
jeaaustx, line 303: If readClientIp() helper is used above, then ipAddress instead of ip would be used here.

http/http_connection.hpp
jnin-dev: I did it this way because in the other versions e.g. 1120 the `ip` variable was a private but here it was only in the local scope inside `readClientIp`. Should I change it to be similar?

http/http_connection.hpp
jnin-dev: I don't think `req` is initialized until `handle()` would that cause a problem?

baemyung (approved):
It looks good to me

baemyung:
> It looks good to me

I will take it back my previous approval.

When can the source IP address be obtained?
Is it after `Connection` is created, or  after reading header?
If it is after reading header, it may defeat the purpose to denying the max connection?

I think we may need to confirm whether the source ip can be obtained after making a Connection - https://github.com/ibm-openbmc/bmcweb/blob/1060/http/http_server.hpp#L188?

Looking at 1120 code, it looks like `connection->start()` happens after Accept (not needing the header read).
If so, getClientIp() before checking connectionCount seems good.

jnin-dev:
> When can the source IP address be obtained?
> Is it after `Connection` is created, or after reading header?

The source IP address can be obtained after the `Connection` is created, the socket is passed through as the adaptor in the `Connection` so we're able to grab the IP address without reading the header of the request. So `getClientIp()` should work since it's grabbing the IP from the socket. Although I have some questions about if we should get the IP address from the socket since if the BMC is behind a reverse proxy we should probably get it from the header?

baemyung:
> > When can the source IP address be obtained?
> > Is it after `Connection` is created, or after reading header?
> 
> The source IP address can be obtained after the `Connection` is created, the socket is passed through as the adaptor in the `Connection` so we're able to grab the IP address without reading the header of the request. So `getClientIp()` should work since it's grabbing the IP from the socket. Although I have some questions about if we should get the IP address from the socket since if the BMC is behind a reverse proxy we should probably get it from the header?

I think getting it from the connection is good. No need to get it from header.

baemyung (approved):
It looks good to me

PR #1486, opened 2026-06-15
RBMC: Implement ActiveRedundancySet for active manager

jnin-dev:
BMC has D-Bus interface `xyz.open_bmc.State.BMC.Redundancy` has the property `Role`[1] which defines the active/passive RBMC

This commit implements `ActiveRedundancySet`[2] for the active manager for now.

Tested:
- Redfish service validator passes
- Verified response on active BMC
```
curl -H "X-Auth-Token: $token" https://${bmc}/redfish/v1/Managers/bmc
```
```
{
  "@odata.id": "/redfish/v1/Managers/bmc",
  "@odata.type": "#Manager.v1_15_0.Manager",

  ...

  "Redundancy": [
    {
      "@odata.id": "/redfish/v1/Managers/bmc#/Redundancy/0",
      "ActiveRedundancySet": [
        {
        }
      ],
      "ActiveRedundancySet@odata.count": 1,
      "MemberId": "0",
      "MinNumNeeded": 1,
      "MinNumNeededForFaultTolerance": 2,
      "Mode": "NotRedundant",
      "Name": "Manager Redundancy",
      "RedundancySet": [
        {
        }
      ],
      "RedundancySet@odata.count": 1,
      "Status": {
        "Health": "OK",
        "State": "Disabled"
      }
    }
  ],
  "Redundancy@odata.count": 1,

  ...

}
```

[1]: https://github.com/openbmc/phosphor-dbus-interfaces/blob/928df5e0ef56613520ad17e16924b524b69408cb/yaml/xyz/openbmc_project/State/BMC/Redundancy.interface.yaml#L109
[2]: https://www.dmtf.org/sites/default/files/standards/documents/DSP2046_2026.1.pdf

### DIFF
--- a/redfish-core/include/manager_redundancy.hpp
+++ b/redfish-core/include/manager_redundancy.hpp
@@ -79,12 +79,14 @@ inline void getRedundantData(
     const size_t* redundancyMinimum = nullptr;
     const size_t* redundancyMaximum = nullptr;
     const size_t* functionalMinimum = nullptr;
+    const std::string* role = nullptr;
 
     const bool success = sdbusplus::unpackPropertiesNoThrow(
         dbus_utils::UnpackErrorPrinter(), propertiesList, "FailoversAllowed",
         failoversAllowed, "RedundancyEnabled", redundancyEnabled,
         "RedundancyMinimum", redundancyMinimum, "RedundancyMaximum",
-        redundancyMaximum, "FunctionalMinimum", functionalMinimum);
+        redundancyMaximum, "FunctionalMinimum", functionalMinimum, "Role",
+        role);
 
     if (!success)
     {
@@ -145,6 +147,23 @@ inline void getRedundantData(
     if (functionalMinimum != nullptr)
     {
         redundantObject["MinNumNeeded"] = *functionalMinimum;
+    }
+
+    if (role != nullptr)
+    {
+        BMCWEB_LOG_DEBUG("Initializing ActiveRedundancySet array");
+        nlohmann::json::array_t activeSet;
+
+        if (*role == "xyz.openbmc_project.State.BMC.Redundancy.Role.Active")
+        {
+            nlohmann::json::object_t activeItem;
+            activeItem["@odata.id"] =
+                boost::urls::format("/redfish/v1/Managers/{}", managerId);
+            activeSet.emplace_back(std::move(activeItem));
+        }
+
+        redundantObject["ActiveRedundancySet@odata.count"] = activeSet.size();
+        redundantObject["ActiveRedundancySet"] = std::move(activeSet);
     }
 
     redundancy->emplace_back(std::move(redundantObject));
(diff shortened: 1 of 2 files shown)


gtmills:
What does this look like on the passive ?

jnin-dev:
> What does this look like on the passive ?
```
"ActiveRedundancySet": [],
"ActiveRedundancySet@odata.count": 0
```

redfish-core/include/manager_redundancy.hpp
```diff
@@ -147,6 +149,23 @@ inline void getRedundantData(
...
 
+    if (role != nullptr)
+    {
+        BMCWEB_LOG_DEBUG("Initializing ActiveRedundancySet array");
+        nlohmann::json::array_t activeSet;
+
+        nlohmann::json::object_t activeItem;
+        if (*role == "xyz.openbmc_project.State.BMC.Redundancy.Role.Active")
+        {
+            activeItem["@odata.id"] =
```
baemyung, line 160: It looks like `activeItem` is needed only inside `if(){}` clause. So move L157 to here.

baemyung:
Question. 
Can this `RedundancySet` have more than one item?

Also, will this be upstreamed?

jnin-dev:
>  Can this RedundancySet have more than one item?

I believe it can have more than one item. 

> Also, will this be upstreamed?

Should be, we'll need the rest of the other redundancy work to be put upstream as well, think there are stories for that

redfish-core/include/manager_redundancy.hpp
jnin-dev: Done, moved it into the branch

gtmills (approved):

baemyung (approved):
It looks good to me

jeaaustx (commented):
Changes look correct based on the plan, but I'm actually wondering about the description of ActiveRedundancySet in the Redundancy schema.

I know we intended this for the active BMC only, but my reading of the schema definition [1] makes me wonder if that is actually correct for the definition. Particularly:

> When `Mode` contains `Failover`, the failure of an active device shall cause a member of this redundancy set to take over its function.

This wording to me seems to say that our Passive should also be listed here when we have redundancy enabled.

[1] https://redfish.dmtf.org/schemas/v1/Redundancy.v1_7_0.json

jnin-dev:
> > When Mode contains Failover, the failure of an active device shall cause a member of this redundancy set to take over its function.

> This wording to me seems to say that our Passive should also be listed here when we have redundancy enabled.

Are we interpreting it as causing a member of the `ActiveRedundancySet` to take over or is it causing a member of `RedundancySet` to take over? I feel like it is the latter

gtmills:
Based on the mockups, https://github.com/DMTF/Redfish/pull/6478/changes#diff-e35249be0d0d3f7ca03892d4fbe2561c9dd541d28ecda7e72cc3a885004c03d3R15-R23 we did it right, but the description could be better

PR #1499, opened 2026-07-08
1210: Fix AdditionalData field after its type change.

baemyung:
Code update fails on reading and parsing of PEL record and the journal shows
```
Fri Jun 26 11:27:18 2026 p11bmc bmcwebd: [update_service.hpp:164] Log property has unexpected value
Fri Jun 26 11:27:18 2026 p11bmc bmcwebd: [dbus_utils.hpp:33] DBUS property error in property: AdditionalData, reason: 1
```

It is because the DBUS `AdditionalData` type has been changed from `as` to `a{ss}`.

For example, the output is changed like
```
$ busctl introspect xyz.openbmc_project.Logging /xyz/openbmc_project/logging/entry/9
.AdditionalData  property  as    6 "DESCRIPTION=Mismatch in IM value f…
==>
.AdditionalData  property  a{ss}  6 "DESCRIPTION" "Mismatch in IM value f…
```

Tested:
- Test on parsing of PEL record with the old and old logic.

### DIFF
--- a/include/dbus_utility.hpp
+++ b/include/dbus_utility.hpp
@@ -15,6 +15,7 @@
 #include <cstddef>
 #include <cstdint>
 #include <functional>
+#include <map>
 #include <memory>
 #include <span>
 #include <string>
@@ -58,7 +59,8 @@ using DbusVariantType = std::variant<
       std::string, std::string, uint64_t>>,
     std::vector<std::pair<sdbusplus::message::object_path, std::string>>,
     std::vector<std::tuple<std::string, uint64_t, std::string, double>>,
-    std::vector<std::tuple<std::string, std::string, uint64_t, std::string>>
+    std::vector<std::tuple<std::string, std::string, uint64_t, std::string>>,
+    std::map<std::string, std::string>
  >;
 
 // clang-format on
--- a/redfish-core/lib/update_service.hpp
+++ b/redfish-core/lib/update_service.hpp
@@ -62,6 +62,7 @@
 #include <format>
 #include <fstream>
 #include <functional>
+#include <map>
 #include <memory>
 #include <optional>
 #include <string>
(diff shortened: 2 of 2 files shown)


baemyung:
Defect: https://jazz07.rchland.ibm.com:13443/jazz/web/projects/CSSD#action=com.ibm.team.workitem.viewWorkItem&id=793400
1120: https://github.com/ibm-openbmc/bmcweb/pull/1498
1210: https://github.com/ibm-openbmc/bmcweb/pull/1499

spinler (approved):

jnin-dev (approved):
Looks good to me

jeaaustx (approved):

jeaaustx:
Same comment I put on 1120:

The change looks good. I'm wondering if the [unit test](https://github.com/ibm-openbmc/bmcweb/blob/1b196a8819e3821f8c63a6683e7f8dc1ff1983fc/test/redfish-core/include/dbus_log_watcher_test.cpp#L20) should be changed also?

[Edited]: Looks like the AddtionalData is going to be ignored by that particular unit test. So explains why unit tests would pass without an update. Just seems like for completeness it should be changed as well.

baemyung:
> Same comment I put on 1120:
> 
> The change looks good. I'm wondering if the [unit test](https://github.com/ibm-openbmc/bmcweb/blob/1b196a8819e3821f8c63a6683e7f8dc1ff1983fc/test/redfish-core/include/dbus_log_watcher_test.cpp#L20) should be changed also?
> 
> [Edited]: Looks like the AddtionalData is going to be ignored by that particular unit test. So explains why unit tests would pass without an update. Just seems like for completeness it should be changed as well.

The unit-test "dbus_log_watcher_test.cpp" is not affected as it does not read/parse the field `AdditionalData`

`AdditionalData` is used only for the specific case - when code update is failed with "Failed" or "Invalid" by the interface `xyz.openbmc_project.Software.Activation`.   I can't generally reproduce it.

So, instead, what I've done is to modify "LogEntries/../Entries/<str>" to read `AddtionalData` and parse it to verify whether the old / new parse logic succeeds.

PR #1500, opened 2026-07-10
Handle GET and PATCH for ReadyToRemove on assemblies

rahulds98:
1. Add GET handling for Assembly ReadyToRemove by reading the xyz.openbmc_project.State.ReadyToRemove interface from the inventory object and exposing it under the assembly

2. Update PATCH handling so non-special assemblies set the D-Bus ReadyToRemove property directly instead of rejecting the request. Keep existing special handling for tod_battery and panel FRUs.

Tested-By:
1. GET returns the correct property under the inventory d-bus
2. PATCH sets the readytoremove on the inventory dbus

GET:
```
curl -k -H https://${bmc}/redfish/v1/Chassis/chassis0/Assembly | jq -r '.Assemblies[] | select(.MemberId=="16")

    {
      "@odata.id": "/redfish/v1/Chassis/chassis0/Assembly#/Assemblies/16",
      "@odata.type": "#Assembly.v1_6_0.AssemblyData",
      "Location": {
        "PartLocation": {
          "ServiceLabel": "Ufcs-SC0-LCD0"
        }
      },
      "MemberId": "16",
      "Model": "6B86",
      "Name": "LCD Operator Panel",
      "PartNumber": "PN12345",
      "ReadyToRemove": true,
      "SerialNumber": "YL6B86010000",
      "SparePartNumber": "F191014",
      "Status": {
        "Health": "OK",
        "State": "Enabled"
      }
    }
```
returns the ReadyToRemove on lcd-op-panel
PATCH:
```
curl -k -X PATCH -H "Content-Type: application/json" https://${bmc}/redfish/v1/Chassis/chassis0/Assembly -d '{"Assemblies":[{"MemberId":"0","ReadyToRemove":false}]}'
```
sets the readytoremove property value on the bmc inventory object

### DIFF
--- a/redfish-core/lib/assembly.hpp
+++ b/redfish-core/lib/assembly.hpp
@@ -152,6 +152,79 @@ void getAssemblyHealth(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
         });
 }
 
+/**
+ * @brief Populate ReadyToRemove for assemblies that implement
+ *        xyz.openbmc_project.State.ReadyToRemove.
+ */
+inline void afterGetAssemblyReadyToRemove(
+    const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
+    const nlohmann::json::json_pointer& assemblyJsonPtr,
+    const boost::system::error_code& ec, bool value)
+{
+    if (ec)
+    {
+        if (ec.value() != EBADR)
+        {
+            BMCWEB_LOG_ERROR("DBUS response error: {}", ec.value());
+            messages::internalError(asyncResp->res);
+        }
+        return;
+    }
+    asyncResp->res.jsonValue[assemblyJsonPtr]["ReadyToRemove"] = value;
+}
+
+inline void getAssemblyReadyToRemove(
+    const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
+    const std::string& serviceName, const std::string& assembly,
+    const nlohmann::json::json_pointer& assemblyJsonPtr)
+{
+    dbus::utility::getProperty<bool>(
+        serviceName, assembly, "xyz.openbmc_project.State.ReadyToRemove",
+        "ReadyToRemove",
+        std::bind_front(afterGetAssemblyReadyToRemove, asyncResp,
+                        assemblyJsonPtr));
+}
+
+inline void afterSetAssemblyReadyToRemove(
+    const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
+    const std::string& assembly, bool value,
+    const boost::system::error_code& ec,
+    const dbus::utility::MapperGetObject& object)
+{
+    if (ec)
+    {
+        if (ec.value() == EBADR)
+        {
+            messages::propertyUnknown(asyncResp->res, "ReadyToRemove");
+            return;
+        }
+        BMCWEB_LOG_ERROR("getDbusObject failed for ReadyToRemove on {}: {}",
+                         assembly, ec.message());
+        messages::internalError(asyncResp->res);
+        return;
+    }
+    if (object.empty())
+    {
+        messages::propertyUnknown(asyncResp->res, "ReadyToRemove");
+        return;
+    }
+    const std::string& service = object.begin()->first;
+    setDbusProperty(asyncResp, "ReadyToRemove", service, assembly,
+                    "xyz.openbmc_project.State.ReadyToRemove", "ReadyToRemove",
+                    value);
+}
+
+inline void setAssemblyReadyToRemove(
+    const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
+    const std::string& assembly, bool value)
+{
+    constexpr std::array<std::string_view, 1> readyToRemoveIface = {
+        "xyz.openbmc_project.State.ReadyToRemove"};
+    dbus::utility::getDbusObject(assembly, readyToRemoveIface,
+                                 std::bind_front(afterSetAssemblyReadyToRemove,
+                                                 asyncResp, assembly, value));
+}
+
 inline void afterGetDbusObject(
     const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
     const std::string& assembly,
@@ -197,6 +270,11 @@ inline void afterGetDbusObject(
                 getAssemblyHealth(asyncResp, serviceName, assembly,
                                   assemblyJsonPtr);
             }
+            else if (interface == "xyz.openbmc_project.State.ReadyToRemove")
+            {
+                getAssemblyReadyToRemove(asyncResp, serviceName, assembly,
+                                         assemblyJsonPtr);
+            }
         }
     }
 }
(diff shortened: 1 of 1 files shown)


rfrandse:
add to approvelist

asmithakarun:
```<Schema xmlns="http://docs.oasis-open.org/odata/ns/edm" Namespace="Assembly.v1_6_0">
<Annotation Term="Redfish.OwningEntity" String="DMTF"/>
<Annotation Term="Redfish.Release" String="2025.3"/>
<EntityType Name="Assembly" BaseType="Assembly.v1_5_1.Assembly"/>
<EntityType Name="AssemblyData" BaseType="Assembly.v1_5_1.AssemblyData">
<Property Name="ReadyToRemove" Type="Edm.Boolean">
<Annotation Term="OData.Permissions" EnumMember="OData.Permission/ReadWrite"/>
<Annotation Term="OData.Description" String="An indication of whether the assembly is prepared by the system for removal."/>
<Annotation Term="OData.LongDescription" String="This property shall indicate whether the assembly is ready for removal. Setting the value to `true` shall cause the service to perform appropriate actions to quiesce the device. A task may spawn while the device is quiescing."/>
</Property>
</EntityType>
</Schema>```

ReadyToRemove property is already in the standard Redfish schema v1_6_0, so you can remove from Oem

File: redfish-core/lib/assembly.hpp [hunk 1]
```diff
@@ -152,6 +154,37 @@ void getAssemblyHealth(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
...
+            if (ec)
+            {
+                if (ec.value() != EBADR)
+                {
+                    BMCWEB_LOG_ERROR("DBUS response error: {}", ec.value());
+                    messages::internalError(asyncResp->res);
+                }
+                return;
+            }
+            asyncResp->res
```
asmithakarun: As mentioned, remove "Oem" as it is in standard Redfish schema

redfish-core/lib/assembly.hpp
```diff
@@ -197,6 +230,12 @@ inline void afterGetDbusObject(
                 getAssemblyHealth(asyncResp, serviceName, assembly,
                                   assemblyJsonPtr);
             }
+            else if (interface ==
+                     "xyz.openbmc_project.State.Decorator.ReadyToRemove")
```
asmithakarun: string literal duplication.
else if (interface == readyToRemoveIface)?

redfish-core/lib/assembly.hpp
```diff
@@ -40,6 +40,8 @@
 #include <utility>
 #include <vector>
 
+static constexpr std::string_view readyToRemoveIface =
```
asmithakarun: static constexpr const char* readyToRemoveIface =
    "xyz.openbmc_project.State.Decorator.ReadyToRemove";

If you change it to the above, then you can directly use in line 167 (you can skip std::string(..)) and line 234

redfish-core/lib/assembly.hpp
rahulds98: done

redfish-core/lib/assembly.hpp
rahulds98: Changed to display ReadyToRemove directly under Assembly properties

redfish-core/lib/assembly.hpp
rahulds98: done

redfish-core/lib/assembly.hpp
```diff
@@ -358,16 +391,17 @@ inline void afterHandleChassisAssemblyPatch(
     }
 
     std::map<std::string, bool> locationIndicatorActiveMap;
-    std::map<std::string, nlohmann::json> oemIndicatorMap;
```
asmithakarun, line 361: As discussed offline, you can keep `OemIndicatorMap` as is and, just add the `readyToRemoveMap` to use it for our usecase (line 394)

We can move from the existing Oem property to use the standard redfish for batteryCM in another PR.
@gtmills ^

redfish-core/lib/assembly.hpp
```diff
@@ -386,16 +420,16 @@ inline void afterHandleChassisAssemblyPatch(
                 return;
             }
         }
-        if (oem)
+        if (readyToRemove)
         {
             if (memberId)
             {
-                oemIndicatorMap[*memberId] = *oem;
```
asmithakarun, line 393: Please have the existing code as it is, in this PR

redfish-core/lib/assembly.hpp
rahulds98: retained handling of oem/openbmc/readytoremove for tod battery and panels, added separate handling for readytoremove

redfish-core/lib/assembly.hpp
rahulds98: done

manojkiraneda (approved):

redfish-core/lib/assembly.hpp
```diff
@@ -152,6 +152,69 @@ void getAssemblyHealth(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
...
+                        assemblyJsonPtr));
+}
+
+inline void afterSetAssemblyReadyToRemove(
+    const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
+    const std::string& assembly, bool value,
+    const boost::system::error_code& ec,
+    const dbus::utility::MapperGetObject& object)
+{
+    if (ec || object.empty())
```
baemyung: I think `ec` may be better handled more granularly

if( ec.value() == EBADR) ==> PropertyUnknown?
if( object.empty()) ==> PropertyUnknown??

baemyung:
Should we attempt to upstream, and add/keep its Change-Id for tracking?

rahulds98:
> Should we attempt to upstream, and add/keep its Change-Id for tracking?

@baemyung Upstream: https://gerrit.openbmc.org/c/openbmc/bmcweb/+/92624
It's WIP for now because the PDI is not merged yet.
I can reference the change ID for now, but PATCH on ReadyToRemove will be return a task(which is uder implementation), which we cannot upstream, since Concurrent Maintenance is all downstream.

redfish-core/lib/assembly.hpp
rahulds98: done

redfish-core/lib/assembly.hpp
```diff
@@ -492,6 +588,15 @@ inline void afterHandleChassisAssemblyPatch(
                 return;
             }
         }
+
+        // Handle top-level ReadyToRemove (standard Assembly schema v1_6_0)
```
baemyung: I think this comment can be removed as the logic is clear enough.
(Also, I don't think it is good to put the schema version number, as schema may be changed separately)

baemyung (approved):
It looks good to me

redfish-core/lib/assembly.hpp
rahulds98: done

PR #1501, opened 2026-07-10
Add UnavailableOffline state mapping and add getResourceStatus helper

jnin-dev:
1. Added a utility function to reduce code duplication for retrieving
resource's Status.State and Status.Health
2. Mapped D-Bus `Available` to Redfish `UnavailableOffline`

Tested:
- Build successfully
- Unit tests passed
- Sent requests and received Status.State and Status.Health as expected
  with the new mapping
```
curl -k -v https://${bmc}/redfish/v1/Systems/system/Processors/mcm/SubProcessors/core0
```
Results with
```
{
  "@odata.id": "/redfish/v1/Systems/system/Processors/mcm/SubProcessors/core0",
  "@odata.type": "#Processor.v1_18_0.Processor",
  "Enabled": true,
  "Id": "core0",
  "Name": "Node 0 processor module Core 0",
  "Status": {
    "Health": "OK",
    "State": "UnavailableOffline"
 }
```
Upstream commits are:
utils: https://gerrit.openbmc.org/c/openbmc/bmcweb/+/92845
cable: https://gerrit.openbmc.org/c/openbmc/bmcweb/+/78685
fan: https://gerrit.openbmc.org/c/openbmc/bmcweb/+/92846
pcie: https://gerrit.openbmc.org/c/openbmc/bmcweb/+/92847
fabric adapters: https://gerrit.openbmc.org/c/openbmc/bmcweb/+/93368
fabric ports: https://gerrit.openbmc.org/c/openbmc/bmcweb/+/93369
assembly: https://gerrit.openbmc.org/c/openbmc/bmcweb/+/93370
memory: https://gerrit.openbmc.org/c/openbmc/bmcweb/+/93371
processors and accelerators: https://gerrit.openbmc.org/c/openbmc/bmcweb/+/93372
subprocessor: https://gerrit.openbmc.org/c/openbmc/bmcweb/+/92439?tab=comments
power supply: https://gerrit.openbmc.org/c/openbmc/bmcweb/+/93373

### DIFF
--- a/redfish-core/include/utils/resource_utils.hpp
+++ b/redfish-core/include/utils/resource_utils.hpp
@@ -1,0 +1,165 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright OpenBMC Authors
+#pragma once
+
+#include "async_resp.hpp"
+#include "dbus_utility.hpp"
+#include "error_messages.hpp"
+#include "generated/enums/resource.hpp"
+#include "logging.hpp"
+
+#include <asm-generic/errno.h>
+
+#include <boost/system/error_code.hpp>
+#include <nlohmann/json.hpp>
+
+#include <functional>
+#include <memory>
+#include <string>
+
+namespace redfish
+{
+namespace resource_utils
+{
+
+/**
+ * @brief Fetches resource status.health from DBus interfaces
+ *
+ */
+inline void determineResourceHealth(
+    const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
+    const nlohmann::json::json_pointer& jsonPtr, bool functional)
+{
+    BMCWEB_LOG_DEBUG("determineResourceHealth");
+
+    if (!functional)
+    {
+        asyncResp->res.jsonValue[jsonPtr]["Status"]["Health"] =
+            resource::Health::Critical;
+    }
+    else
+    {
+        asyncResp->res.jsonValue[jsonPtr]["Status"]["Health"] =
+            resource::Health::OK;
+    }
+}
+
+inline void determineResourceState(
+    const std::shared_ptr<bmcweb::AsyncResp>& asyncResp, bool present,
+    bool available, const nlohmann::json::json_pointer& jsonPtr)
+{
+    BMCWEB_LOG_DEBUG("determineResourceState");
+
+    // Absent takes priority over unavailable
+    if (!present)
+    {
+        asyncResp->res.jsonValue[jsonPtr]["Status"]["State"] =
+            resource::State::Absent;
+    }
+    else if (!available)
+    {
+        asyncResp->res.jsonValue[jsonPtr]["Status"]["State"] =
+            resource::State::UnavailableOffline;
+    }
+    else
+    {
+        asyncResp->res.jsonValue[jsonPtr]["Status"]["State"] =
+            resource::State::Enabled;
+    }
+}
+
+inline void getStatusAvailableState(
+    const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
+    const nlohmann::json::json_pointer& jsonPtr, bool present,
+    const boost::system::error_code& ec, bool available)
+{
+    if (ec)
+    {
+        if (ec.value() != EBADR)
+        {
+            BMCWEB_LOG_ERROR("DBUS response error for {}, ec {}", "Available",
+                             ec.value());
+            messages::internalError(asyncResp->res);
+            return;
+        }
+        available = true;
+    }
+    determineResourceState(asyncResp, present, available, jsonPtr);
+}
+
+inline void getStatusPresentState(
+    const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
+    const std::string& service, const std::string& path,
+    const nlohmann::json::json_pointer& jsonPtr,
+    const boost::system::error_code& ec, bool present)
+{
+    if (ec)
+    {
+        if (ec.value() != EBADR)
+        {
+            BMCWEB_LOG_ERROR("DBUS response error for {}, ec {}", "Present",
+                             ec.value());
+            messages::internalError(asyncResp->res);
+            return;
+        }
+        // Interface is missing, default to true for Enabled
+        present = true;
+    }
+    dbus::utility::getProperty<bool>(
+        *crow::connections::systemBus, service, path,
+        "xyz.openbmc_project.State.Decorator.Availability", "Available",
+        std::bind_front(getStatusAvailableState, asyncResp, jsonPtr, present));
+}
+
+inline void getResourceState(
+    const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
+    const std::string& service, const std::string& path,
+    const nlohmann::json::json_pointer& jsonPtr)
+{
+    BMCWEB_LOG_DEBUG("getResourceStatus");
+    dbus::utility::getProperty<bool>(
+        *crow::connections::systemBus, service, path,
+        "xyz.openbmc_project.Inventory.Item", "Present",
+        std::bind_front(getStatusPresentState, asyncResp, service, path,
+                        jsonPtr));
+}
+
+inline void afterGetResourceHealth(
+    const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
+    const nlohmann::json::json_pointer& jsonPtr,
+    const boost::system::error_code& ec, bool functional)
+{
+    if (ec)
+    {
+        if (ec.value() != EBADR)
+        {
+            BMCWEB_LOG_ERROR("DBUS response error for {}, ec {}", "Functional",
+                             ec.value());
+            messages::internalError(asyncResp->res);
+            return;
+        }
+        // Interface missing, default to OK
+        functional = true;
+    }
+    determineResourceHealth(asyncResp, jsonPtr, functional);
+}
+
+/*
+ * @brief Retrieves the status of the resource's status.health
+ *
+ * Queries interface:
+ * - xyz.openbmc_project.State.Decorator.OperationalStatus::Functional
+ */
+inline void getResourceHealth(
+    const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
+    const std::string& service, const std::string& path,
+    const nlohmann::json::json_pointer& jsonPtr)
+{
+    dbus::utility::getProperty<bool>(
+        *crow::connections::systemBus, service, path,
+        "xyz.openbmc_project.State.Decorator.OperationalStatus", "Functional",
+        std::bind_front(afterGetResourceHealth, asyncResp, jsonPtr));
+}
+
+} // namespace resource_utils
+} // namespace redfish
--- a/redfish-core/lib/assembly.hpp
+++ b/redfish-core/lib/assembly.hpp
@@ -19,6 +19,7 @@
 #include "utils/asset_utils.hpp"
 #include "utils/json_utils.hpp"
 #include "utils/name_utils.hpp"
+#include "utils/resource_utils.hpp"
 
 #include <asm-generic/errno.h>
 
@@ -77,39 +78,30 @@ inline void getAssemblyLocationCode(
         });
 }
 
-inline void getAssemblyState(
+inline void getAssemblyReadyToRemove(
     const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
     const auto& serviceName, const auto& assembly,
     const nlohmann::json::json_pointer& assemblyJsonPtr)
 {
-    asyncResp->res.jsonValue[assemblyJsonPtr]["Status"]["State"] =
-        resource::State::Enabled;
-
-    dbus::utility::getProperty<bool>(
-        serviceName, assembly, "xyz.openbmc_project.Inventory.Item", "Present",
-        [asyncResp, assemblyJsonPtr,
-         assembly](const boost::system::error_code& ec, const bool value) {
-            if (ec)
-            {
-                if (ec.value() != EBADR)
+    std::string fru = sdbusplus::message::object_path(assembly).filename();
+    if (fru == "panel0" || fru == "panel1")
+    {
+        dbus::utility::getProperty<bool>(
+            serviceName, assembly, "xyz.openbmc_project.Inventory.Item",
+            "Present",
+            [asyncResp, assemblyJsonPtr,
+             assembly](const boost::system::error_code& ec, const bool value) {
+                if (ec)
                 {
-                    BMCWEB_LOG_ERROR("DBUS response error: {}", ec.value());
-                    messages::internalError(asyncResp->res);
+                    if (ec.value() != EBADR)
+                    {
+                        BMCWEB_LOG_ERROR("DBUS response error: {}", ec.value());
+                        messages::internalError(asyncResp->res);
+                    }
+                    return;
                 }
-                return;
-            }
 
-            if (!value)
-            {
-                asyncResp->res.jsonValue[assemblyJsonPtr]["Status"]["State"] =
-                    resource::State::Absent;
-            }
-
-            std::string fru =
-                sdbusplus::message::object_path(assembly).filename();
-            // Special handling for LCD and base panel CM.
-            if (fru == "panel0" || fru == "panel1")
-            {
+                // Special handling for LCD and base panel CM.
                 asyncResp->res.jsonValue[assemblyJsonPtr]["Oem"]["OpenBMC"]
                                         ["@odata.type"] =
                     "#OpenBMCAssembly.v1_0_0.OpenBMC";
@@ -118,38 +110,8 @@ inline void getAssemblyState(
                 // can be placed.
                 asyncResp->res.jsonValue[assemblyJsonPtr]["Oem"]["OpenBMC"]
                                         ["ReadyToRemove"] = !value;
-            }
-        });
-}
-
-void getAssemblyHealth(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
-                       const auto& serviceName, const auto& assembly,
-                       const nlohmann::json::json_pointer& assemblyJsonPtr)
-{
-    asyncResp->res.jsonValue[assemblyJsonPtr]["Status"]["Health"] =
-        resource::Health::OK;
-
-    dbus::utility::getProperty<bool>(
-        serviceName, assembly,
-        "xyz.openbmc_project.State.Decorator.OperationalStatus", "Functional",
-        [asyncResp, assemblyJsonPtr](const boost::system::error_code& ec,
-                                     bool functional) {
-            if (ec)
-            {
-                if (ec.value() != EBADR)
-                {
-                    BMCWEB_LOG_ERROR("DBUS response error {}", ec.value());
-                    messages::internalError(asyncResp->res);
-                }
-                return;
-            }
-
-            if (!functional)
-            {
-                asyncResp->res.jsonValue[assemblyJsonPtr]["Status"]["Health"] =
-                    resource::Health::Critical;
-            }
-        });
+            });
+    }
 }
 
 /**
@@ -261,14 +223,17 @@ inline void afterGetDbusObject(
             }
             else if (interface == "xyz.openbmc_project.Inventory.Item")
             {
-                getAssemblyState(asyncResp, serviceName, assembly,
-                                 assemblyJsonPtr);
+                resource_utils::getResourceState(asyncResp, serviceName,
+                                                 assembly, assemblyJsonPtr);
+
+                getAssemblyReadyToRemove(asyncResp, serviceName, assembly,
+                                         assemblyJsonPtr);
             }
             else if (interface ==
                      "xyz.openbmc_project.State.Decorator.OperationalStatus")
             {
-                getAssemblyHealth(asyncResp, serviceName, assembly,
-                                  assemblyJsonPtr);
+                resource_utils::getResourceHealth(asyncResp, serviceName,
+                                                  assembly, assemblyJsonPtr);
             }
             else if (interface == "xyz.openbmc_project.State.ReadyToRemove")
             {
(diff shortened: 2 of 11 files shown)


redfish-core/include/utils/resource_utils.hpp
```diff
@@ -0,0 +1,153 @@
...
+ *
+ */
+inline void getStatusProperty(
+    const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
+    const std::shared_ptr<ResourceStatus>& status, const std::string& service,
+    const std::string& path, const std::string& interface,
+    const std::string& property, const nlohmann::json::json_pointer& jsonPtr,
+    std::function<void(ResourceStatus&, bool)>&& callback)
+{
+    sdbusplus::asio::getProperty<bool>(
```
jeaaustx: Should be using the dbus::utility version of this function.

redfish-core/include/utils/resource_utils.hpp
```diff
@@ -0,0 +1,153 @@
...
+        *crow::connections::systemBus, service, path, interface, property,
+        [status, asyncResp, property, jsonPtr, callback{std::move(callback)}](
```
jeaaustx: nit: Common convention for function parameters is to put the asyncResp first.

redfish-core/include/utils/resource_utils.hpp
```diff
@@ -0,0 +1,153 @@
...
+    else if (!status->available)
+    {
+        statusJson["State"] = resource::State::UnavailableOffline;
+    }
+    else
+    {
+        statusJson["State"] = resource::State::Enabled;
+    }
+
+    if (!status->functional)
```
jeaaustx: So the functional value is not affected and does not affect either of the other values. I'm wondering if it should just be handled separately? 

Cons of that: Additional callback function for that one.
Pros of that: Removes dependance on other values being detected in order to the Health.

I don't have a strong opinion here, just something that crossed my mind on review.

redfish-core/include/utils/resource_utils.hpp
```diff
@@ -0,0 +1,153 @@
...
+    const std::shared_ptr<ResourceStatus>& status,
+    const nlohmann::json::json_pointer& jsonPtr)
+{
+    BMCWEB_LOG_DEBUG("determineResourceState");
+    if (--status->pending != 0)
+    {
+        return;
+    }
+
+    nlohmann::json& statusJson = asyncResp->res.jsonValue[jsonPtr]["Status"];
```
jeaaustx: Should be using a json::object_t here. See: https://github.com/openbmc/bmcweb/blob/master/docs/COMMON_ERRORS.md#special-note-json

redfish-core/include/utils/resource_utils.hpp
```diff
@@ -0,0 +1,153 @@
...
+ * - xyz.openbmc_project.State.Decorator.Availability::Available
+ * - xyz.openbmc_project.State.Decorator.OperationalStatus::Functional
+ */
+inline void getResourceStatus(
+    const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
+    const std::string& service, const std::string& path,
+    const nlohmann::json::json_pointer& jsonPtr)
+{
+    // Default values
+    asyncResp->res.jsonValue[jsonPtr]["Status"]["State"] =
```
jeaaustx: Looks to me like if the D-Bus properties don't exist the determineResourceState() function will still be called to set the State and Health. It will be based on the default values for ResourceStatus. So will set to Enabled and Ok.

So I'm wondering if this initialization is needed? In the case of an internalError() is the response going to see Eanbled/OK and the error?

redfish-core/lib/cable.hpp
```diff
@@ -490,7 +462,8 @@ inline void getCableProperties(
             }
             else if (interface == "xyz.openbmc_project.Inventory.Item")
             {
-                fillCableHealthState(asyncResp, cableObjectPath, service);
+                resource_utils::getResourceStatus(asyncResp, service,
```
jeaaustx: Looks like will now be filling in Health also for a cable where before that was not part of the response, correct?

Probably should call this out in the commit message.

redfish-core/lib/assembly.hpp
```diff
@@ -99,12 +97,6 @@ inline void getAssemblyState(
                 return;
             }
 
-            if (!value)
-            {
-                asyncResp->res.jsonValue[assemblyJsonPtr]["Status"]["State"] =
-                    resource::State::Absent;
-            }
-
             std::string fru =
```
jeaaustx: Should this check move up above the getProperty() now? (Since if not one of these panels we won't be doing anything with the property value here?

jeaaustx (commented):
Overall the change looks reasonable to me. I have a few comments in the code and a couple of comments on the commit messages:

- The utils function commit message talks about UnavailableOffline but the output example in the commit message only shows Absent. For clarity it would be good to show the different cases in the output in the commit message.
- I'm wondering about carrying these patches and the ease of handling rebasing in the future. I suggest considering breaking these up to a single commit that adds the utility function and then separate commits for each of the places changing to use the utility function. (This would help with chicken/egg problem on rebase of needing support in place that pulls in the utility function for any other commits dependant on the function.) But...
- Is this going upstream? If so then above concern about rebase is less important. However for upstream it is still easiest to get the utility function commit in by itself.
- Probably should add some unit tests for the utiilty function.

redfish-core/include/utils/resource_utils.hpp
jnin-dev: Fixed

redfish-core/lib/assembly.hpp
jnin-dev: It looks like the ReadyToRemove is dependent on the value from Present so I think it should stay there. Let me know if you think otherwise

redfish-core/lib/cable.hpp
jnin-dev: Done, I've added that to the commit message

redfish-core/include/utils/resource_utils.hpp
jnin-dev: Done, I've changed it

redfish-core/include/utils/resource_utils.hpp
jnin-dev: Fixed, changed it to nlohmann:json::object_t

jnin-dev:
> The utils function commit message talks about UnavailableOffline but the output example in the commit message only shows Absent. For clarity it would be good to show the different cases in the output in the commit message.

I added to the commit message about the difference cases for Status.State

>I'm wondering about carrying these patches and the ease of handling rebasing in the future. I suggest considering breaking these up to a single commit that adds the utility function and then separate commits for each of the places changing to use the utility function. 

I believe the utility function will go upstream. I've separated them into different commits for each resource since it'll be cleaner

>Probably should add some unit tests for the utiilty function.

Done.

redfish-core/include/utils/resource_utils.hpp
```diff
@@ -0,0 +1,149 @@
...
+ * @brief Fetches resource status from DBus interfaces
+ *
+ */
+inline void determineResourceState(
+    const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
+    const std::shared_ptr<ResourceStatus>& status,
+    const nlohmann::json::json_pointer& jsonPtr)
+{
+    BMCWEB_LOG_DEBUG("determineResourceState");
+    if (--status->pending != 0)
```
baemyung: `pending` is uint_8. (Also it is for tracking how many fields are needed).

What if determineResourceState() is invoked before setting `pending=1` (with the default value of `pending=0`)?

Would it need to be like?
```
if( status->pending && (--status->pending) != 0)
{
    return;
}
```

redfish-core/include/utils/resource_utils.hpp
```diff
@@ -0,0 +1,149 @@
...
+inline void getResourceStatus(
+    const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
+    const std::string& service, const std::string& path,
+    const nlohmann::json::json_pointer& jsonPtr)
+{
+    BMCWEB_LOG_DEBUG("getResourceStatus");
+    auto status = std::make_shared<ResourceStatus>();
+    status->pending = 3;
+
+    getStatusProperty(asyncResp, status, service, path,
```
baemyung, line 148: I'm not sure whether all status related fields (present, available, operationalStatus) are provided by the same service.
I think it is possible that they may be on the different services.

redfish-core/include/utils/resource_utils.hpp
baemyung: Instead of having `pending`, I think it can be tracked by using `std::optional` like

```
struct ResourceStatus
{
   std::optional<bool> present;
    std::optional<bool> available;
    std::optional<bool> functional;
};
```

In this way, we can track which fields are collected, and derive the status from them.

redfish-core/include/utils/resource_utils.hpp
jnin-dev: Present, available are most likely provided by the same interface. I've separated the Health per @jeaaustx suggestion

redfish-core/include/utils/resource_utils.hpp
jnin-dev: I've updated it from `pending` to use `std::optional`

jnin-dev: Actually, would `std::optional` be better? With `std::optional` we won't be able to know if the property is still being queried or if the interface doesn't exist (ec == EBADR) so we can't track completion e.g.
```
if (status->present.has_value() &&
    status->available.has_value())
```
 As a result, the Status.State is recomputed and written to multiple times depending on how many interfaces there will be

redfish-core/include/utils/resource_utils.hpp
```diff
@@ -0,0 +1,182 @@
...
+};
+
+/**
+ * @brief Fetches resource status.health from DBus interfaces
+ *
+ */
+inline void determineResourceHealth(
+    const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
+    const nlohmann::json::json_pointer& jsonPtr,
+    const std::optional<bool> available)
```
baemyung: This would be -- to avoid the unnecessary copy argument.
```
const std::optional<bool>& available
``

File: redfish-core/include/utils/resource_utils.hpp [hunk 11]
```diff
@@ -0,0 +1,182 @@
...
+            const boost::system::error_code& ec, bool value) {
+            if (ec)
+            {
+                if (ec.value() != EBADR)
+                {
+                    BMCWEB_LOG_ERROR("DBUS response error for {}, ec {}",
+                                     property, ec.value());
+                    messages::internalError(asyncResp->res);
+                    return;
+                }
```
baemyung: Would `return` be missed here?

redfish-core/include/utils/resource_utils.hpp
```diff
@@ -0,0 +1,182 @@
...
+inline void getResourceHealth(
+    const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
+    const std::string& service, const std::string& path,
+    const nlohmann::json::json_pointer& jsonPtr)
+{
+    dbus::utility::getProperty<bool>(
+        *crow::connections::systemBus, service, path,
+        "xyz.openbmc_project.State.Decorator.Availability", "Available",
+        [asyncResp, jsonPtr](const boost::system::error_code& ec,
+                             const std::optional<bool> value) {
```
baemyung:
```
const std::optional<bool>& value) {
```

redfish-core/include/utils/resource_utils.hpp
baemyung: >(ec == EBADR)
In that case, the associated variable should still be not set ( !has_value()).
If you need to set a specific value for the case, you may set it there
```
if(ec.value() == EBADR)
{
       ...
     status->variable = <some-value>;
     return;
}
```

jnin-dev: OK, deciding to stay with std::optional per our discussion

test/redfish-core/include/utils/resource_utils_test.cpp
```diff
@@ -0,0 +1,255 @@
...
+{
+
+std::shared_ptr<bmcweb::AsyncResp> createAsyncResp()
+{
+    return std::make_shared<bmcweb::AsyncResp>();
+}
+
+// Tests for determineResourceState
+
+TEST(DetermineResourceState, Absent)
```
baemyung, line 26: Shouldn't we have another test case - which does not have the values for the attributes?
For example, 
```
    auto status = std::make_shared<ResourceStatus>();
   // <-- NO value set to present or available.
    determineResourceState(asyncResp, status, ""_json_pointer);
```

test/redfish-core/include/utils/resource_utils_test.cpp
```diff
@@ -0,0 +1,255 @@
...
+    EXPECT_EQ(asyncResp->res.jsonValue["Resource1"]["Status"]["Health"],
+              resource::Health::Critical);
+}
+
+// Tests for ResourceStatus struct
+
+TEST(ResourceStatus, DefaultValues)
+{
+    ResourceStatus status;
+
```
baemyung, line 190: What would be the expected result with this defect case?

```
auto asyncResp = createAsyncResp();
ResourceStatus status;
std::optional<bool> available;   // NO value set
determineResourceHealth(asyncResp, ""_json_pointer, available);
```

redfish-core/lib/processor.hpp
baemyung, line 150: I think we can first refactor this code line L150 - L295 first.

```
  const bool* cpuPresent = nullptr;
  const bool* cpuAvailable  = nullptr;
  const uint16_t* coresCount = nullptr;

  const bool success = sdbusplus::unpackPropertiesNoThrow(
                dbus_utils::UnpackErrorPrinter(), cpuInterfacesProperties, "Present",
                cpuPresent, "Available", cpuAvailable, "CoreCount", coresCount, ....);

  if (!success)
  {
                return;
  }
  ...
  if (coresCount != nullptr)
  {
                asyncResp->res.jsonValue["TotalCores"] = *coresCount;
  }

  ....

...

File: redfish-core/include/utils/resource_utils.hpp [hunk 15]
```diff
@@ -0,0 +1,182 @@
...
+                callback(*status, value);
+            }
+            determineResourceState(asyncResp, status, jsonPtr);
+        });
+}
+
+/*
+ * @brief Retrieves the status of the resource's status.state
+ *
+ * Queries three interfaces:
```
jeaaustx: nit: two interfaces are now queried

redfish-core/include/utils/resource_utils.hpp
jnin-dev: I've removed it, determineResourceState() should set the State and Health once all the calls are finished and default to Enabled/OK. In the case of an internal error, it shouldn't see it now.

redfish-core/include/utils/resource_utils.hpp
jnin-dev: Done, I've decoupled the `functional` property from the Status.State.

redfish-core/include/utils/resource_utils.hpp
jnin-dev: Done.

redfish-core/include/utils/resource_utils.hpp
jnin-dev: I don't think so, I've changed it to have it set the std::optional to a default of true (for Enabled/OK since the resources originally set it to that if EBADR was met). 

This allows me to check if all the async calls are finished so the state is only computed once on lines 67-70. Let me know if this is anti-pattern for bmcweb

redfish-core/include/utils/resource_utils.hpp
jnin-dev: Done, added the reference

redfish-core/include/utils/resource_utils.hpp
```diff
@@ -0,0 +1,182 @@
...
+                                     property, ec.value());
+                    messages::internalError(asyncResp->res);
+                    return;
+                }
+            }
+            else
+            {
+                callback(*status, value);
+            }
+            determineResourceState(asyncResp, status, jsonPtr);
```
jeaaustx: Confirming the new flow:
With the removal of the pending count it looks like the new flow is that the determineResourceState() will be called for both "Present" and "Available" retrieval.

The first call will set a state, but the second call might alter it again. But should set the correct state based on the full information.

test/redfish-core/include/utils/resource_utils_test.cpp
jnin-dev: Done, I've added it on line 107-115 with the name OptionalNotSetReturnsEarly

test/redfish-core/include/utils/resource_utils_test.cpp
jnin-dev: It would be similar to `ec == EBADR` or the interface doesn't exist so it'll default to `Status.Health = OK` which should be the same behavior before the refactors

redfish-core/include/utils/resource_utils.hpp
```diff
@@ -0,0 +1,194 @@
...
+}
+
+/*
+ * @brief Retrieves the status of the resource's status.state
+ *
+ * Queries three interfaces:
+ * - xyz.openbmc_project.Inventory.Item::Present
+ * - xyz.openbmc_project.State.Decorator.Availability::Available
+ */
+inline void getResourceStatus(
```
jeaaustx: Should this be named getResourceState() instead? (to better align with the Health version.)

redfish-core/lib/assembly.hpp
jeaaustx: Yes it does use the value. What I meant is should the check 
```if (fru == "panel0" || fru == "panel1")```
be done before the getProperty() call. 

Since the setting of Status/State has been decoupled I don't see the `value` being used except if the fru is one of those panels? (So we avoid the getProperty() call at all if the fru is not one of the panels.)

Am I missing where `value` is being used elsewhere?

So,
```
if (fru == "panel0" || fru == "panel1")
{
dbus::utility::getProperty<bool>(
...
/* same lambda except no need to check fru again since it was already checked.
*/
}
```

redfish-core/include/utils/resource_utils.hpp
jnin-dev: The check on line 67 should prevent any subsequent call to `determineResourceState()` from altering the state until all the calls are set I believe. 

I added the callback on line 115 to default the `std::optional` to true in order to know when the call was finished by setting the `std::optional`

redfish-core/include/utils/resource_utils.hpp
jnin-dev: Done.

redfish-core/include/utils/resource_utils.hpp
jnin-dev: Done, updated to `getResourceState`

redfish-core/lib/assembly.hpp
jnin-dev: Done, this makes sense to avoid unnecessary `getProperty()` calls

redfish-core/lib/processor.hpp
jnin-dev: `sdbusplus::unpackPropertiesNoThrow()` appears to take a `dbus::utility::DBusPropertiesMap&`. I can get this from iterating through the `dbus::utility::DBusInterfacesMap&` and calling `sdbusplus::unpackPropertiesNoThrow()` on each interface. But I don't see a pattern of doing it that way but rather the current way like https://github.com/ibm-openbmc/bmcweb/blob/f96fef531a6b73d86f5134ca8ce2a218b5403000/redfish-core/lib/virtual_media.hpp#L156-L164
Beyond:  https://github.com/ibm-openbmc/bmcweb/blob/f96fef531a6b73d86f5134ca8ce2a218b5403000/redfish-core/lib/sensors.hpp#L523-L536 
Where `sensor_utils::objectPropertiesToJson` calls https://github.com/ibm-openbmc/bmcweb/blob/f96fef531a6b73d86f5134ca8ce2a218b5403000/redfish-core/include/utils/sensor_utils.hpp#L477-L479 
Let me know what you think

redfish-core/include/utils/resource_utils.hpp
jeaaustx: Ahh, I missed the L115 check. That makes sense to me now.

redfish-core/lib/processor.hpp
jeaaustx: I think this change should be separate commit/PR. This refactoring doesn't seem specifically related to the Status work. (I think including it as part of this work will make rebasing harder for this set of commits.)

jeaaustx (commented):
Changes look good to me. I have just the one comment about the refactoring.

redfish-core/lib/processor.hpp
jnin-dev: Sounds good, I'll go ahead and put these changes in a separate PR

redfish-core/include/utils/resource_utils.hpp
jnin-dev: Sounds good, I'll resolve this now

redfish-core/include/utils/resource_utils.hpp
jnin-dev: I will mark this as resolved, please reopen if you have anything you would like to add

test/redfish-core/include/utils/resource_utils_test.cpp
jnin-dev: I will mark this as resolved, reopen if there's anything you would like to add, thanks

baemyung:
@jnin-dev 
Please add the list of upstream commits here, or on PR description section.

baemyung (approved):
It looks good to me

gtmills (approved):
Defaulting mostly to upstream and Myung here

jeaaustx (approved):

PR #1502, opened 2026-07-13
Add TPM support for mTLS key/certs

Jino26:
Allow mTLS identity used for BMC-to-BMC Redfish aggregation - both the private key and the certificate - to live in the TPM instead of on the filesystem, loaded through the OpenSSL OSSL_STORE API via a provider (e.g. tpm2-openssl).

Why: mutual aggregation authenticates each BMC to its peer with a client certificate and with a filesystem-resident key that identity is only as strong as the readability of a PEM file, anyone who can read the rootfs can extract the key and impersonate the BMC. Moving the key into TPM makes it non-exportable; the private key never exists in bmcweb's memory or on flash, signing happens inside TPM and the identity is bound to that specific board. Storing the certificate in a TPM NV index as well lets the whole mTLS identity be provisioned into the hardware at manufacturing, with nothing to deploy onto the filesystem.

Key changes:
- meson: add uri-key (single key for both aggregation roles), uri-cert, and openssl-config-path options, resolved into bmcweb_config.h. A handle:<persistent-handle> URI selects a TPM object; file:// or a bare path preserves the existing filesystem behavior.
- config: ship openssl.cnf (activates the default + tpm2 providers) and wire OPENSSL_CONF into bmcweb.service via openssl-config-path so the provider is scoped to bmcweb.

Validated on hardware with both the key (persistent handle) and the certificate (NV index) resident in each BMC's TPM.

### DIFF
--- a/config/bmcweb.service.in
+++ b/config/bmcweb.service.in
@@ -18,6 +18,7 @@ Type=simple
 WorkingDirectory=/home/root
 SyslogLevelPrefix=true
 WatchdogSec=@BMCWEB_WATCHDOG_TIMEOUT_SECONDS@s
+@BMCWEB_OPENSSL_ENV@
 
 [Install]
 WantedBy=network.target
--- a/config/meson.build
+++ b/config/meson.build
@@ -57,8 +57,11 @@ feature_options = [
 string_options = [
     'dns-resolver',
     'mutual-tls-common-name-parsing-default',
+    'openssl-config-path',
     'redfish-manager-uri-name',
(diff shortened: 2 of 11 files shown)


Jino26:
Recipe meson options:

```
-Duri-key=handle:0x81000000 \
-Duri-cert=handle:0x1500010 \
-Dopenssl-config-path=/etc/ssl/openssl_bmcweb.cnf \
```

rfrandse:
add to approvelist

PR #1504, opened 2026-07-15
redfish: aggregator: cache EM satellites via D-Bus signals

abhilashraju:
Replace per-request D-Bus round-trips to Entity Manager with an in-memory cache (emSatelliteSources) that is populated at startup and kept current by subscribing to InterfacesAdded / InterfacesRemoved signals on the /xyz/openbmc_project/inventory ObjectManager path.

Changes:
- Add emSatelliteSources map to RedfishAggregator to hold Entity-Manager-managed satellite URLs separately from REST- created aggregationSources.
- Subscribe to InterfacesAdded / InterfacesRemoved D-Bus signals in the constructor; handleSatelliteAdded and handleSatelliteRemoved update emSatelliteSources accordingly.
- Populate emSatelliteSources from the initial getSatelliteConfigs call and pass the error_code through to constructorCallback.
- Update getSatelliteConfigs handler signature to include an error_code argument; call handler on both success and error paths.
- Replace all getSatelliteConfigs async callbacks in the Aggregation Service GET / PATCH / Collection handlers with direct reads from the two in-memory maps — no D-Bus round-trip on each request.
- Forward requests to REST-created sources from findSatellite when the EM prefix lookup fails, enabling correct routing for both source types.
- Duplicate-hostname check on POST now covers both REST-created and EM-managed sources.
- Add printAggregationSources debug helper and a constexpr alias for the SatelliteController interface string.
- Templatise addSatelliteConfig to accept both DBusPropertiesMap and a signal-derived interface property map.

Tested: verified aggregation routes resolve without D-Bus calls after startup; added / removed satellite signals update the cache.

Upstream https://gerrit.openbmc.org/c/openbmc/bmcweb/+/92257

Change-Id: I684b775c1e0833dff1397eacdc58683da72c4c3b

### DIFF
--- a/redfish-core/include/redfish_aggregator.hpp
+++ b/redfish-core/include/redfish_aggregator.hpp
@@ -27,6 +27,7 @@
 #include <boost/url/url.hpp>
 #include <boost/url/url_view.hpp>
 #include <nlohmann/json.hpp>
+#include <sdbusplus/bus/match.hpp>
 #include <sdbusplus/message/native_types.hpp>
 
 #include <algorithm>
@@ -36,6 +37,7 @@
 #include <cstdint>
 #include <functional>
 #include <limits>
+#include <map>
 #include <memory>
 #include <ranges>
 #include <string>
@@ -44,6 +46,7 @@
 #include <unordered_map>
 #include <utility>
 #include <variant>
+#include <vector>
 
 namespace redfish
 {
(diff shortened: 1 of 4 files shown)


gtmills:
Why do we need this change ?

gtmills (changes requested):
IMO this isn't worth carrying this patch downstream.. Let's drive this upstream and then we can pull in downstream

redfish-core/include/redfish_aggregator.hpp
```diff
@@ -420,38 +423,110 @@ struct AggregationSource
     std::string password;
 };
 
+constexpr auto satelliteControllerIface =
+    "xyz.openbmc_project.Configuration.SatelliteController";
```
asmithakarun, line 427: Should there be a dbus interfaces change merged for this?

redfish-core/include/redfish_aggregator.hpp
```diff
@@ -920,9 +1037,12 @@ class RedfishAggregator
         return handler;
     }
 
-    // Aggregation sources with their URLs and optional credentials
+    // REST-created aggregation sources with their URLs and optional credentials
     std::unordered_map<std::string, AggregationSource> aggregationSources;
 
+    // Cache of satellite sources managed by Entity Manager (read-only via REST)
+    std::unordered_map<std::string, boost::urls::url> emSatelliteSources;
```
asmithakarun, line 1044: Any reason to have it as a public data member?

PR #1505, opened 2026-07-21
1210: Generate 64 bit serial numbers

baemyung:
Even though the certificate is self signed, we should pass as many certificate tests as possible.  testssl.sh prints
```
 Serial 4B32D4F0   NOT ok: length should be >= 64 bits entropy (is: 4 bytes)
```

On our default certificate.  This is relatively easy to fix.

Change-Id: Ib1eb07b637ebf49ecf954b3d98ced9e9ef0f5a34

### DIFF
--- a/src/ssl_key_handler.cpp
+++ b/src/ssl_key_handler.cpp
@@ -286,9 +286,10 @@ std::string generateSslCertificate(const std::string& cn)
             // number If this is not random, regenerating certs throws browser
             // errors
             bmcweb::OpenSSLGenerator gen;
-            std::uniform_int_distribution<int> dis(
-                1, std::numeric_limits<int>::max());
-            int serial = dis(gen);
+            std::uniform_int_distribution<uint64_t> dis(
+                std::numeric_limits<uint64_t>::min(),
+                std::numeric_limits<uint64_t>::max());
+            int64_t serial = static_cast<int64_t>(dis(gen));
 
             ASN1_INTEGER_set(X509_get_serialNumber(x509), serial);
 


baemyung:
The reason that uses `static_cast<int64_t>()` is due to the build failure on 1210 (unlike upstream).
```
./../../../../../../../../ibm-bmcweb/src/ssl_key_handler.cpp
| ../../../../../../../../../ibm-bmcweb/src/ssl_key_handler.cpp: In function 'std::string ensuressl::generateSslCertificate(const std::string&)':
| ../../../../../../../../../ibm-bmcweb/src/ssl_key_handler.cpp:294:59: error: conversion to 'long int' from 'uint64_t' {aka 'long unsigned int'} may change the sign of the result [-Werror=sign-conversion]
|   294 |             ASN1_INTEGER_set(X509_get_serialNumber(x509), serial);
|       |                                                           ^~~~~~
| cc1plus: all warnings being treated as errors
````

baemyung:
Should this be applied to 1120 too? 
Probably not, as it is unlikely that the multiple systems generate the self-signed certificate at the same time.

gtmills:
Can you link upstream?

gtmills:
> Should this be applied to 1120 too?
Probably not, as it is unlikely that the multiple systems generate the self-signed certificate at the same time.

Agreed.

gtmills:
https://gerrit.openbmc.org/c/openbmc/bmcweb/+/90552 is upstream and merged

gtmills (approved):

baemyung:
STG defect: https://jazz07.rchland.ibm.com:13443/jazz/web/projects/CSSD#action=com.ibm.team.workitem.viewWorkItem&id=797132

PR #1507, opened 2026-07-23
redfish: Sort EventLog entries by Created timestamp

ishaandesai23:
EventLog entries were sorted by "Id" with alphanumeric comparison. While this avoids the multi-digit lexicographic sorting, "Id" is an opaque identifier with no guaranteed relationship to event time.

"Created" carries a timestamp that is guaranteed to match chronological order. This makes the sort both correct by construction and intuitive to people expecting entries in event-time sequence.

Change sortJsonArrayByKey to use "Created" instead of "Id" as the sort key for the EventLog Entries collection.

Tested: Verified that EventLog entries sorted by Created timestamp match the expected chronological order. Confirmed via D-Bus that Timestamp values increase monotonically with Id, producing identical ordering.

### DIFF
--- a/redfish-core/lib/log_services.hpp
+++ b/redfish-core/lib/log_services.hpp
@@ -1784,7 +1784,7 @@ inline void afterLogEntriesGetManagedObjects(
                                              entriesArray.emplace_back());
     }
 
-    redfish::json_util::sortJsonArrayByKey(entriesArray, "Id");
+    redfish::json_util::sortJsonArrayByKey(entriesArray, "Created");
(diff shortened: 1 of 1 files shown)


jnin-dev (approved):
It looks good to me but could your commit message reflect on why we want it to order by Created? Would be helpful in case in the future someone needs to know why it was chosen to order by Created

jnin-dev:
https://gerrit.openbmc.org/c/openbmc/phosphor-logging/+/84722

PR #1467, opened 2026-05-12
Adding context switching framework

abhilashraju:
This PR enables ssl contexts changes required to support aggregation through MTLS channel. Upstream gerrit review is already raised for it 
https://gerrit.openbmc.org/c/openbmc/bmcweb/+/89586
https://gerrit.openbmc.org/c/openbmc/bmcweb/+/89586

PR #1508, opened 2026-07-30
http: Add SIGUSR1 to signal_set

gtmills:
https://github.com/ibm-openbmc/bmcweb/pull/1467 broke this

This function is downstream only. 

We should talk about this function.. Ideally something upstream.. A true flight recorder..

The boost::asio::signal_set in Server's constructor was registered with only SIGINT, SIGTERM, and SIGHUP. SIGUSR1 was handled in the startAsyncWaitForSignal() callback to write the current session snapshot to disk for BMC dumps, but was never added to the set.

Because SIGUSR1 was not intercepted by Boost.Asio, the OS applied its default disposition for the signal: terminate the process. This caused bmcweb to be killed (status=10/USR1) every time a BMC dump was initiated, as dreport sends SIGUSR1 to collect the session data file.

Add SIGUSR1 to the signal_set constructor so Boost.Asio intercepts it, allowing the existing handler to call writeCurrentSessionData() and re-arm the signal wait. bmcweb now survives the dump collection.

### DIFF
--- a/http/http_server.hpp
+++ b/http/http_server.hpp
@@ -55,7 +55,10 @@ class Server
         // NOLINTNEXTLINE(misc-include-cleaner)
         signals(getIoContext(), SIGINT, SIGTERM, SIGHUP), handler(handlerIn),
         sslContextFactory(std::move(contextFactory))
-    {}
+    {
+        // NOLINTNEXTLINE(misc-include-cleaner)
+        signals.add(SIGUSR1);
+    }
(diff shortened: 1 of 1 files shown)


spinler (approved):

baemyung (approved):
It looks good to me

gtmills:
jenkins run tests please

jeaaustx (approved):

PR #1510, opened 2026-08-10
assembly: Return Task for ReadyToRemove PATCH

rahulds98:
This commit adds on top of setting readyToRemove on inventory D-Bus object and returns a task to monitor the progress interface on the CM object

Major changes:

* Add startCmTask() — registers a PropertiesChanged D-Bus match on the CM object path (/com/ibm/ConcurrentMaintenance/{add,remove}) before writing ReadyToRemove so no signal is missed, then calls task::TaskData::createTask() and returns 202 Accepted with a Location header pointing at the new task resource.

* Add mapCmStatus() — translates xyz.openbmc_project.Common.Progress. OperationStatus values (Completed / Failed / Aborted / InProgress) to Redfish TaskState/TaskStatus and emits the appropriate task message (taskCompletedOK or taskAborted).

* The task uses a 30-minute timer as a safety net for the case where the CM daemon never emits a terminal status.

* Remove the now-superseded setAssemblyReadyToRemove() and afterSetAssemblyReadyToRemove() helpers.

Tested:
1. PATCH now returns a /redfish/v1/TaskService/Tasks/{}

2. The task tracks the states of the CM object's progress interface. The task is returned with a Running state and it reaches terminal state for failed/aborted/completed.


redfish-core/lib/assembly.hpp
```diff
@@ -43,6 +46,165 @@
...
+        std::array<std::string_view, 1>{
+            "xyz.openbmc_project.State.ReadyToRemove"},
+        [inventoryPath, value](const boost::system::error_code& ec,
+                               const dbus::utility::MapperGetObject& object) {
+            if (ec || object.empty())
+            {
+                BMCWEB_LOG_ERROR(
+                    "getDbusObject failed for ReadyToRemove on {}: {}",
+                    inventoryPath, ec.message());
+                return;
```
baemyung, line 167: Should this result in as an error - internalError()?

redfish-core/lib/assembly.hpp
```diff
@@ -43,6 +46,165 @@
...
+            // has already been committed as 202 Accepted; errors are logged.
+            sdbusplus::asio::setProperty(
+                *crow::connections::systemBus, service, inventoryPath,
+                "xyz.openbmc_project.State.ReadyToRemove", "ReadyToRemove",
+                value, [inventoryPath](const boost::system::error_code& ec2) {
+                    if (ec2)
+                    {
+                        BMCWEB_LOG_ERROR(
+                            "Failed to set ReadyToRemove on {}: {}",
+                            inventoryPath, ec2.message());
```
baemyung: Should this result in as an error - internalError()?

redfish-core/lib/assembly.hpp
```diff
@@ -43,6 +46,165 @@
...
+            {
+                BMCWEB_LOG_ERROR(
+                    "getDbusObject failed for ReadyToRemove on {}: {}",
+                    inventoryPath, ec.message());
+                return;
+            }
+            const std::string& service = object.begin()->first;
+            // setDbusProperty cannot be used here because the HTTP response
+            // has already been committed as 202 Accepted; errors are logged.
+            sdbusplus::asio::setProperty(
```
baemyung, line 228: Use `setDbusProperty()`

redfish-core/lib/assembly.hpp
```diff
@@ -43,6 +46,165 @@
...
+    const std::string cmObjPath(readyToRemove ? removePath : addPath);
+
+    // Match registered before setReadyToRemove so no PropertiesChanged signal
+    // from the CM object can be missed.
+    const std::string matchStr =
+        "type='signal',"
+        "interface='org.freedesktop.DBus.Properties',"
+        "member='PropertiesChanged',"
+        "path='" +
+        cmObjPath + "'";
```
baemyung: I think this may be better as 
```
const std::string matchStr = std::format(
    "type='signal', interface='org.freedesktop.DBus.Properties', member='PropertiesChanged', path='{}'",
    cmObjPath);
```

redfish-core/lib/assembly.hpp
```diff
@@ -424,7 +586,7 @@ inline void handleChassisAssemblyHead(
 inline void afterHandleChassisAssemblyPatch(
     const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
     const std::string& chassisID,
-    std::vector<nlohmann::json::object_t>& assemblyData,
+    std::vector<nlohmann::json::object_t>& assemblyData, task::Payload& payload,
```
baemyung: It looks like `payload` is intended to move and then I think it can be
```
task::Payload&& payload
```
Then, caller may use `afterHandleChassisAssemblyPatch(..., std::move(payload))`

redfish-core/lib/assembly.hpp
```diff
@@ -43,6 +46,165 @@
...
+                {
+                    return task::completed;
+                }
+            }
+            return !task::completed;
+        },
+        matchStr);
+
+    taskHandle->state = "Running";
+    taskHandle->startTimer(std::chrono::minutes(30));
```
baemyung: Is this 30 minutes pre-defined or configurable?
Perhaps, can it be define as a constant above like

```
namespace redfish
{
constexpr CMTimeoutMinutes = 30;

....

taskHandle->startTimer(std::chrono::minutes(CMTimeoutMinutes));
```

redfish-core/lib/assembly.hpp
asmithakarun: Agree. +1

redfish-core/lib/assembly.hpp
```diff
@@ -43,6 +46,165 @@
...
+            const std::string& service = object.begin()->first;
+            // setDbusProperty cannot be used here because the HTTP response
+            // has already been committed as 202 Accepted; errors are logged.
+            sdbusplus::asio::setProperty(
+                *crow::connections::systemBus, service, inventoryPath,
+                "xyz.openbmc_project.State.ReadyToRemove", "ReadyToRemove",
+                value, [inventoryPath](const boost::system::error_code& ec2) {
+                    if (ec2)
+                    {
+                        BMCWEB_LOG_ERROR(
```
asmithakarun: We should also be setting the task state here right?
If it fails here, after setReadyToRemove, the task just sits in state "Running" forever until it is expired?

redfish-core/lib/assembly.hpp
rahulds98: done

redfish-core/lib/assembly.hpp
rahulds98: Agree, the task is now returned only when setproperty succeeds.

redfish-core/lib/assembly.hpp
rahulds98: done

redfish-core/lib/assembly.hpp
rahulds98: done

redfish-core/lib/assembly.hpp
rahulds98: done

redfish-core/lib/assembly.hpp
rahulds98: we can't really use setDbusProperty() here because we have an action on success of setdbus property, keeping  sdbusplus::asio::setProperty(.

redfish-core/lib/assembly.hpp
rahulds98: agreed, done

redfish-core/lib/assembly.hpp
```diff
@@ -43,6 +47,235 @@
...
+ * @param asyncResp     Shared response, used only to set the error if busy.
+ * @param inventoryPath Inventory path logged in the error message.
+ * @param onClear       Callback invoked when CM is confirmed not busy.
+ */
+inline void checkCmNotBusy(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
+                           const std::string& inventoryPath,
+                           std::function<void()> onClear)
+{
+    // Check /remove first.
+    dbus::utility::getDbusObject(
```
baemyung: This part of the code is 2-level depth chains of callbacks.
I think it is appropriate to flatten them using std::bind_front().
Esp, after the future rebase from the upstream, 10-line of lambda body will be enforced.

```

inline void afterCmAddPathGetDbusObject(...)
{
...
}

inline void afterCmRemovePathGetDbusObject(...)
{
...
 if (!ecRemove && !objRemove.empty())
....

dbus::utility::getDbusObject(
        std::string(cmdAddPath),
        std::array<std::string_view, 1>{cmProgressIface},
       std::bind_front(afterCmAddPathGetDbusObject, asyncResp, inventoryPath, onClear));
...
}

inline void checkCmNotBusy(...)
{
...
dbus::utility::getDbusObject(
        std::string(cmRemovePath),
        std::array<std::string_view, 1>{cmProgressIface},
       std::bind_front(afterCmRemovePathGetDbusObject, asyncResp, inventoryPath, onClear));

````

File: redfish-core/lib/assembly.hpp [hunk 9]
```diff
@@ -43,6 +47,235 @@
...
+                        task::Payload&& payload,
+                        const std::string& inventoryPath, bool readyToRemove)
+{
+    const std::string cmObjPath(readyToRemove ? cmRemovePath : cmAddPath);
+
+    checkCmNotBusy(
+        asyncResp, inventoryPath,
+        [asyncResp, payload = std::move(payload), inventoryPath, readyToRemove,
+         cmObjPath]() mutable {
+            dbus::utility::getDbusObject(
```
baemyung: Can this body be moved as a separate function?

rahulds98: Moved this logic to a new function afterCheckCmNotBusy()

redfish-core/lib/assembly.hpp
rahulds98: done

redfish-core/lib/assembly.hpp
```diff
@@ -44,6 +48,254 @@
...
+                    {
+                        BMCWEB_LOG_ERROR("Failed to set ReadyToRemove: {}",
+                                         ec2.message());
+                        messages::internalError(asyncResp->res);
+                        return;
+                    }
+
+                    // property write succeeded, register the match,
+                    //  create the task, and commit 202 Accepted.
+                    const std::string matchStr = std::format(
```
asmithakarun: The match is registered inside the setProperty success callback here. The match is not registered before the write as mentioned in the commit msg. This needs an update.

We need the match to be registered before? afterCheckCmNotBusy() -> getDbusObject(inventoryPath) -> Register the match and create TaskData with that match -> setProperty -> On setProperty success, populateResp()

Thoughts?

asmithakarun: Because if the CM daemon creates the CM object and emit Status: InProgress (or even Status: Completed) before bmcweb's setProperty callback fires and registers the match, then those signals are lost and the task hangs there till the 30-minute timer expires.
This is a race condition.

redfish-core/lib/assembly.hpp
```diff
@@ -44,6 +48,254 @@
...
+ *        an existing Progress object at /remove then /add.
+ *
+ * Calls onClear() when neither object exists.  Returns 503 resourceInUse if
+ * either object is found.
+ *
+ * @param asyncResp     Shared response, used only to set the error if busy.
+ * @param inventoryPath Inventory path logged in the error message.
+ * @param onClear       Callback invoked when CM is confirmed not busy.
+ */
+inline void checkCmNotBusy(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
```
asmithakarun, line 111: The backend service will have only one child object /add or /remove at a time. why cant we just do a single ObjectMapper GetSubTreePaths call on /com/ibm/ConcurrentMaintenance with depth 1, see if there is any created, if yes, cm busy, else not busy - one shot.
shouldnt this suffice?

```
dbus::utility::getSubTreePaths(
    "/com/ibm/ConcurrentMaintenance", /*depth=*/1,
    std::array<std::string_view, 1>{cmProgressIface},
    [asyncResp, inventoryPath, onClear](
        const std::vector<std::string>& paths) {
        if (!ec && !paths.empty())
        {
            messages::resourceInUse(asyncResp->res);
            return;
        }
        onClear();   ----> path empty, so proceed
    });
```

redfish-core/lib/assembly.hpp
rahulds98: Yes, makes sense, thanks for pointing this out, new sequence:
checkCmNotBusy → 503 if busy → afterCheckCmNotBusy → getDbusObject → derive cmObjPath from readyToRemove → register PropertiesChanged match → createTask → setProperty ReadyToRemove → 202 Accepted

redfish-core/lib/assembly.hpp
rahulds98: Yes, this makes sense, thanks, it is now changed to look for objects under /com/ibm/ConcurrentMaintenance with depth 1

redfish-core/lib/assembly.hpp
```diff
@@ -555,10 +774,10 @@ inline void afterHandleChassisAssemblyPatch(
         }
 
         auto iter3 = readyToRemoveMap.find(std::to_string(assemblyIndex));
-
         if (iter3 != readyToRemoveMap.end())
         {
-            setAssemblyReadyToRemove(asyncResp, assembly, iter3->second);
+            startCmTask(asyncResp, std::move(payload), assembly, iter3->second);
+            return; // payload moved; stop processing further assemblies
```
baemyung: I think this part is not correct. `assembly` should not be moved to `startCmTask()` and continue (not return here).

What if `LocationIndicatorActive` and `ReadyToRemove` are patched together in one PATCH like this?

```
PATCH 
-d '{"Assemblies":[{"MemberId":"0","ReadyToRemove":false, "LocationIndicatorActive": true}]}'
```

baemyung (changes requested):
I think the code needs to be modified.
Please see my comment - https://github.com/ibm-openbmc/bmcweb/pull/1510#discussion_r3958223470

redfish-core/lib/assembly.hpp
jnin-dev: +1 looks like if there are multiple assemblies and the first has ReadyToRemove, it will return before other assemblies are processed

redfish-core/lib/assembly.hpp
```diff
@@ -44,6 +48,221 @@
...
+            const std::string matchStr = std::format(
+                "type='signal',"
+                "interface='org.freedesktop.DBus.Properties',"
+                "member='PropertiesChanged',"
+                "path='{}'",
+                cmObjPath);
+
+            std::shared_ptr<task::TaskData> taskHandle =
+                task::TaskData::createTask(
+                    [](const boost::system::error_code& ec2,
```
jnin-dev, line 183: Should we put this in a callback to be more readable? Similar to https://github.com/ibm-openbmc/bmcweb/blob/4a632edc4bd3ef2ffcdc4a2a5d5460fe53cded80/redfish-core/lib/update_service.hpp#L353-L357

redfish-core/lib/assembly.hpp
```diff
@@ -44,6 +48,221 @@
...
+inline void checkCmNotBusy(const std::shared_ptr<bmcweb::AsyncResp>& asyncResp,
+                           const std::string& inventoryPath,
+                           std::function<void()> onClear)
+{
+    dbus::utility::getSubTreePaths(
+        std::string(cmBasePath), 1, std::array<std::string_view, 0>{},
+        [asyncResp, inventoryPath, onClear = std::move(onClear)](
+            const boost::system::error_code& ec,
+            const dbus::utility::MapperGetSubTreePathsResponse& paths) mutable {
+            if (!ec && !paths.empty())
```
jnin-dev: Should this check for ec?

redfish-core/lib/assembly.hpp
```diff
@@ -44,6 +48,221 @@
...
+                        {
+                            BMCWEB_LOG_ERROR(
+                                "CM task timer expired or bus error: {}",
+                                ec2.message());
+                            taskData->messages.emplace_back(
+                                messages::internalError());
+                            taskData->state = "Exception";
+                            taskData->status = "Warning";
+                            return task::completed;
+                        }
```
jnin-dev, line 194: On timer expiration this overwrites the `taskAborted` message, state, and status. Did we want to do this? 
Should we instead check for 
```
if (ec2)
{
    if (ec2 != operation_aborted)
    {
        BMCWEB_LOG_ERROR(
               "CM task timer expired or bus error: {}",
               ec2.message());
        taskData->messages.emplace_back(
              messages::internalError());
         taskData->state = "Exception";
         taskData->status = "Warning";
    }
    return task::completed
}
```
Since I don't think a timer expiring is an internal error.

redfish-core/lib/assembly.hpp
rahulds98: populateResp is this case is defered until the setproperty suceeds, If we make a callback here, we would send a 202 before setproperty, doesn't make sense in case property write fails.

redfish-core/lib/assembly.hpp
rahulds98: added check for ec2(which is only set for timeout), only one trace added here.
Reverted overwrite to state, message and status

redfish-core/lib/assembly.hpp
rahulds98: added error log for ec. Only set for mapper failure.

redfish-core/lib/assembly.hpp
rahulds98: Fixed

redfish-core/lib/assembly.hpp
```diff
@@ -584,11 +824,13 @@ inline void handleChassisAssemblyPatch(
...
-            const std::vector<std::string>& assemblyList) mutable {
+        [asyncResp, chassisID, assemblyData = std::move(assemblyData),
+         payload = task::Payload(
+             req)](const boost::system::error_code& ec,
+                   const std::vector<std::string>& assemblyList) mutable {
             afterHandleChassisAssemblyPatch(asyncResp, chassisID, assemblyData,
-                                            ec, assemblyList);
+                                            std::move(payload), ec,
+                                            assemblyList);
         });
```
baemyung, line 836: 1) It looks like `payload` is being created in the lamba capture and moved it to callback.
Instead, it seems better to be done by creating the `payload` (i.e task) inside `afterHandleChassisAssemblyPatch()` by passing `req`?

```
inline void afterHandleChassisAssemblyPatch(
   const crow::Request& req,
    const std::string& chassisID,
    std::vector<nlohmann::json::object_t>& assemblyData,
    const std::vector<std::string>& assemblyList)
{
....
   auto payload = task::Payload(req);
   ...
}

handleChassisAssemblyPatch()
{
....
   assembly_utils::getChassisAssembly(
        asyncResp, chassisID,
        [asyncResp, chassisID, assemblyData = std::move(assemblyData), req](const boost::system::error_code& ec,
                   const std::vector<std::string>& assemblyList) mutable {
            afterHandleChassisAssemblyPatch(asyncResp, req, chassisID, assemblyData, ec,
                                            assemblyList);
        });
}
```

OR 
2) for the quick/simple change and for better reading, create it before getChassisAssembly() and pass it to lambda?

```
   auto payload = task::Payload(req);
   assembly_utils::getChassisAssembly(
        asyncResp, chassisID,
        [asyncResp, chassisID, assemblyData = std::move(assemblyData), 
         payload = std::move(payload)] (const boost::system::error_code& ec,
                   const std::vector<std::string>& assemblyList) mutable {
            afterHandleChassisAssemblyPatch(asyncResp, chassisID, assemblyData,
                                            std::move(payload), ec,
                                            assemblyList);
        });
```

rahulds98: Went with option 2 for better readability and similarity in other places in the repo

redfish-core/lib/assembly.hpp
```diff
@@ -44,6 +48,235 @@
 namespace redfish
 {
 
+constexpr std::string_view cmAddPath = "/com/ibm/ConcurrentMaintenance/add";
+constexpr std::string_view cmRemovePath =
+    "/com/ibm/ConcurrentMaintenance/remove";
+constexpr std::string_view cmProgressIface =
+    "xyz.openbmc_project.Common.Progress";
```
baemyung, line 57: Since assembly.hpp may be included from many other source files, these would need to be set as `static`.
```
static constexpr std::string_view cmAddPath = "/com/ibm/ConcurrentMaintenance/add";
static constexpr std::string_view cmRemovePath =
    "/com/ibm/ConcurrentMaintenance/remove";
static constexpr std::string_view cmProgressIface =
    "xyz.openbmc_project.Common.Progress";
```

redfish-core/lib/assembly.hpp
baemyung: I meant the long lambda body is better to be a separate function.

```
inline void afterCheckCmNotBudy(...)
{
              dbus::utility::getDbusObject(
                inventoryPath,
                std::array<std::string_view, 1>{
                    "xyz.openbmc_project.State.ReadyToRemove"},
             ...
}

inline void startCmTask(...)
{
    ....
    checkCmNotBusy(
        asyncResp, inventoryPath,
        std::bind_front( afterCheckCmNotBudy, ...));
}
```

redfish-core/lib/assembly.hpp
```diff
@@ -44,6 +48,235 @@
...
+
+                    const std::string service = object.begin()->first;
+
+                    // write the property to trigger the CM app.
+                    sdbusplus::asio::setProperty(
+                        *crow::connections::systemBus, service, inventoryPath,
+                        "xyz.openbmc_project.State.ReadyToRemove",
+                        "ReadyToRemove", readyToRemove,
+                        [asyncResp, payload = std::move(payload), cmObjPath](
+                            const boost::system::error_code& ec2,
```
baemyung: Similarly, this long lambda body is better to be a separate function.

baemyung (commented):
Besides the long lambda body as separate functions, the code seems looking good.

redfish-core/lib/assembly.hpp
rahulds98: done

redfish-core/lib/assembly.hpp
rahulds98: oops, I must have missed pushing this, I have updated now.

redfish-core/lib/assembly.hpp
rahulds98: done

redfish-core/lib/assembly.hpp
```diff
@@ -44,6 +48,227 @@
...
+    bool readyToRemove)
+{
+    dbus::utility::getDbusObject(
+        inventoryPath,
+        std::array<std::string_view, 1>{
+            "xyz.openbmc_project.State.ReadyToRemove"},
+        [asyncResp, payload = std::move(payload), inventoryPath,
+         readyToRemove](const boost::system::error_code& ec,
+                        const dbus::utility::MapperGetObject& object) mutable {
+            if (ec || object.empty())
```
baemyung, line 161: L161 - L249: This long lamda body can also be moved out as a separate function.
And doing the similar moveout as a function can be done for the inner long lambda body.

baemyung (approved):
There may be more refactoring (to move out the long lambda body into a separate function), but the current code seems much more simpler than before.

jnin-dev (approved):

redfish-core/lib/assembly.hpp
rahulds98: done

PR #1512, opened 2026-08-17
1210: Support Eventing for Redundant Manager

baemyung:
Register a new D-Bus PropertiesChanged signal match for the xyz.openbmc_project.State.BMC.Redundancy interface on bmc0. The handler bmcRedundancyPropertyChange() dispatches three event types:

```
  FailoversAllowed=true  -> ResourceStatusChangedOK (health OK)
  FailoversAllowed=false -> ResourceStatusChangedWarning (health Warning)
  Role change            -> ResourceStateChanged (Standby/Active)
  any other property     -> ResourceChanged (fallback)
```

Tested:
- Run Redfish Event Listener against the active and the passive bmc.
- Perform the failover - e.g. rbmctool -f
- Monitor and check the above events.

For example, in the case of the failover done (e.g. Role change), the event listener may see like this.

```
bodydata: {'@odata.type': '#Event.v1_4_0.Event', 'Events': [{'@odata.type': '#Message.v1_1_1.Message',
   'EventId': 4, 'EventTimestamp': '2026-08-13T17:35:50+00:00', 'MemberId': '0',
   'Message': 'One or more resource properties have changed.', 'MessageArgs': [],
   'MessageId': 'ResourceEvent.1.3.ResourceChanged', 'MessageSeverity': 'OK',
   'OriginOfCondition': '/redfish/v1/Managers/bmc', 'Resolution': 'None.'}],
   'Id': 4, 'Name': 'Event Log'}

```

### DIFF
--- a/include/event_dbus_monitor.hpp
+++ b/include/event_dbus_monitor.hpp
@@ -1,3 +1,5 @@
+#include "bmcweb_config.h"
+
 #include "dbus_singleton.hpp"
 #include "logging.hpp"
 
@@ -37,6 +39,7 @@ static std::shared_ptr<sdbusplus::bus::match::match> matchEventLogCreated;
 static std::shared_ptr<sdbusplus::bus::match::match> matchPostCodeChange;
 static std::shared_ptr<sdbusplus::bus::match::match> matchPlatformSAIChange;
 static std::shared_ptr<sdbusplus::bus::match::match> matchPartitionSAIChange;
+static std::shared_ptr<sdbusplus::bus::match::match> matchBMCRedundancyChange;
 
 static uint64_t postCodeCounter = 0;
 // NOLINTEND(cppcoreguidelines-avoid-non-const-global-variables)
@@ -56,6 +59,7 @@ void registerSAIStateChangeSignal();
 void registerDumpUpdateSignal();
 void registerStateChangeSignal();
 void registerVMIConfigChangeSignal();
+void registerBMCRedundancyChangeSignal();
 
 inline void sendEventOnEthIntf(const std::string& origin)
 {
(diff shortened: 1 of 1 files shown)


jnin-dev (approved):
Looks good to me, don't see anything out of the obvious

include/event_dbus_monitor.hpp
```diff
@@ -381,11 +385,107 @@ inline void registerEventLogCreatedSignal()
...
+
+    // Role change (Standby ↔ Active): emit ResourceStateChanged so
+    // subscribers know the ActiveRedundancySet has changed.
+    auto roleIt = values.find("Role");
+    if (roleIt != values.end())
+    {
+        const std::string* role = std::get_if<std::string>(&roleIt->second);
+        if (role != nullptr)
+        {
+            std::string_view roleLabel = "Standby";
```
jeaaustx: The PDI uses Passive, should this be using that also?

include/event_dbus_monitor.hpp
```diff
@@ -381,11 +385,107 @@ inline void registerEventLogCreatedSignal()
...
+                                             BMCWEB_REDFISH_MANAGER_URI_NAME);
+            redfish::EventServiceManager::getInstance().sendEvent(
+                redfish::messages::resourceStateChanged(
+                    BMCWEB_REDFISH_MANAGER_URI_NAME, roleLabel),
+                origin, "Manager");
+            return;
+        }
+    }
+
+    // Any other property change (RedundancyEnabled, counts, etc.):
```
jeaaustx, line 457: Would it make sense to have unique event messages for FailoverImminent or FailoverInProgress changes?

include/event_dbus_monitor.hpp
```diff
@@ -381,11 +385,107 @@ inline void registerEventLogCreatedSignal()
...
+    {
+        const bool* allowed = std::get_if<bool>(&failoverIt->second);
+        if (allowed != nullptr)
+        {
+            std::string origin = std::format("/redfish/v1/Managers/{}",
+                                             BMCWEB_REDFISH_MANAGER_URI_NAME);
+            if (*allowed)
+            {
+                redfish::EventServiceManager::getInstance().sendEvent(
+                    redfish::messages::resourceStatusChangedOK(
```
jeaaustx, line 419: This seems like it hides the actual change. Would it be better to differntiate with a string? (Or is the new redundancy message support in Redfish going to add messages that could be used here instead?)

baemyung: > This seems like it hides the actual change. Would it be better to differntiate with a string? (Or is the new redundancy message support in Redfish going to add messages that could be used here instead?)

This is to notify the resource state change condition (e.g. Warning -> Ok).

Do you mean that - it may be better to notify saying "FailoverAllowed" is changed and thus ResourceState becomes OK?

Redfish does not have such message but only supports the following messages - the section 2.1.6 ResourceEvent - https://www.dmtf.org/sites/default/files/standards/documents/DSP2065_2026.1.pdf.

- ResourceStatusChangedOK
- ResourceStatusChangedCritical
- ResourceStatusChangedWarning

Not only that, in this particularly case, "FailoversAllowed" is not a redfish attribute but an internal dbus attribute and thus it will not be exposed.

In case the other resource attributes are changed other than the attributes like "State"/"Role", `resourceChanged()` notification is generated (so the receiver can query more of the resource attributes (e.g. RedundancyEnabled).

baemyung: > This seems like it hides the actual change. Would it be better to differntiate with a string? (Or is the new redundancy message support in Redfish going to add messages that could be used here instead?)

This is to notify the resource state change condition (e.g. Warning -> Ok).

Do you mean that - it may be better to notify saying "FailoverAllowed" is changed and thus ResourceState becomes OK?

Redfish does not have such message but only supports the following messages - the section 2.1.6 ResourceEvent - https://www.dmtf.org/sites/default/files/standards/documents/DSP2065_2026.1.pdf.

- ResourceStatusChangedOK
- ResourceStatusChangedCritical
- ResourceStatusChangedWarning

include/event_dbus_monitor.hpp
baemyung: > Would it make sense to have unique event messages for FailoverImminent or FailoverInProgress changes?

DMTF has only a few predefined resource event notifications - https://www.dmtf.org/sites/default/files/standards/documents/DSP2065_2026.1.pdf (section 2.16 ResourceEvent).
- ResourceCreated
- ResourceChanged
- ResourceRemoved
- ....

Like RedundancyEnabled attr changes (which may cause  FailoverImminent & FailoverInProgress), it would need to be done by
- Set/update the attribute
- Generate `ResourceChanged` event.  The receiver can query the attributes.

include/event_dbus_monitor.hpp
baemyung: > The PDI uses Passive, should this be using that also?

Thanks for finding it.  Changed to use "Passive" & "Active" in Role.

include/event_dbus_monitor.hpp
jeaaustx: I was wondering if the upcoming 9/2026 release of Redfish from DMTF will include new messages for these? It is supposed to have some for the reasons why failover is not allowed etc. I haven't looked in detail at it yet though.

Without something more specific I'm okay with ResourceStatusChangedOK signifying failovers are allowed and ResourceStatusChangedWarning signifying failovers are no longer allowed.

include/event_dbus_monitor.hpp
jeaaustx: I was wondering if these should be handled like the Role version using redfish::messages::resourceStateChanged?

baemyung: > I was wondering if these should be handled like the Role version using redfish::messages::resourceStateChanged?

Do you mean to treat "Role" as a part of "State"?

Although Role change may go thru the process of the manager state changes, but I don't think State can be derived from Role information.

Instead, I think this Role would cause the properties of ActiveRedundancySet.
If so, I think it is more like "ResourceChanged".

Any opinions? @gtmills @jnin-dev

include/event_dbus_monitor.hpp
baemyung: If there is a new msg related to Failover in the coming or future DMTF spec, it can be changed at the time.

include/event_dbus_monitor.hpp
jnin-dev: Are we thinking that `resourceStateChanged` to represent a change in `FailoverInProgress` or `FailoverImminent` similar to how `Role` is being treated?

`resourceChanged` seems like a generic catch all for the other properties that change the configuration. I think would be good for the subscriber to know that a significant state transition is going to happen, is happening, or is finished?

baemyung: > Are we thinking that `resourceStateChanged` to represent a change in `FailoverInProgress` or `FailoverImminent` similar to how `Role` is being treated?
> 
> `resourceChanged` seems like a generic catch all for the other properties that change the configuration. I think would be good for the subscriber to know that a significant state transition is going to happen, is happening, or is finished?

FailoverInProgress or FailoverImminent are not Redfish State. They are internal DBUS attributes for Failover.  It may be shown via rbmctool but not via Redfish.

Here the state is like "OK", "Warning" or "Critical".

ResourceChanged is not a catchall but it is needed because there is no associated Redfish message to represent a specific Redfish resource attribute change.  
DMTF defines the messages for "State" via "ResourceStateChanged" etc.

For example, if Redfish attribute (e.g. ActiveRedundancySet) is changed, it will be notified via `ResourceChanged` notification (because there is no such ResourceActiveRedundancySetChanged message).

Dbus internal "Role" attribute is related to Redfish "ActiveRedundancySet" attribute, not necessary Redfish "State" attribute.

jeaaustx: I see, I hadn't looked at the Redfish spec for the Events, I was just basing my comments on what might be good to distinguish. So looks good as is based on this.

jeaaustx (approved):

gtmills (approved):
Defaulting to others here

baemyung:
Jira story: https://jsw.ibm.com/browse/PFEBMC-4899

gtmills:
Request from @myungbae
Pull Request: https://github.com/ibm-openbmc/bmcweb/pull/1512
Upstream: Downstream now, Upstream Later
Fixes: https://jsw.ibm.com/browse/PFEBMC-4899
Releases: 1210
Release List comments: 
Release Note:

PR #1513, opened 2026-08-18
dump_offload: prevent connection rejection from terminating active of…

abhilashraju:
…fload

When a dump offload is already in progress, any subsequent offload request is rejected with a "503 Service Unavailable" error and the rejected connection is closed.

However, bmcweb uses a shared static `systemHandlers` pointer, the `.onclose` callback (which is registered for all connections on the dump attachment route) gets triggered when the newly rejected connection is closed. Since `systemHandlers` is not null, the callback would execute the cleanup logic, cancel active async socket reads, remove socket files, and nullify `systemHandlers`. This abruptly terminates the ongoing (first) offload transfer and leaves subsequent attempts wedged.

This commit resolves the issue by verifying if the connection being closed (`conn`) matches the active connection registered in the active handler (`systemHandlers->connection`). If the closed connection is a non-active connection (such as a rejected parallel request), the cleanup logic is skipped and the handler remains active.

Tested:
1. Initiated a non-disruptive system dump offload from ASMI.
2. Attempted a second offload from a new ASMI tab/browser.
3. Verified the second request is gracefully rejected with a 503 Service Unavailable without interrupting the first active offload which continues to completion.
4. Subsequent offloads can be successfully initiated and completed.

### DIFF
--- a/include/dump_offload.hpp
+++ b/include/dump_offload.hpp
@@ -511,6 +511,12 @@ inline void requestRoutesDumpOffload(App& app)
                 BMCWEB_LOG_DEBUG("No handler to cleanup");
                 return;
             }
+            if (systemHandlers->connection.get() != &conn)
+            {
+                BMCWEB_LOG_DEBUG(
+                    "Close event for non-active connection, ignoring cleanup");
+                return;
+            }
             BMCWEB_LOG_DEBUG("cancelling AsyncOperations");
             // Cancel all async operations to release shared_ptr references
             systemHandlers->cancelAsyncOperations();


asmithakarun:
Changes looks good.
This might be needed for 1210 as well, as it shares the same code!

asmithakarun:
https://jazz07.rchland.ibm.com:13443/jazz/web/projects/CSSD#action=com.ibm.team.workitem.viewWorkItem&id=807208 is the defect this fix is for.

asmithakarun:
@abhilashraju there seems to be clang formatting issues that CI is reporting. Can you check and resolve?

asmithakarun:
jenkins run tests please

asmithakarun:
jenkins run tests please

asmithakarun (approved):
Looks good to me

baemyung (approved):
It looks good to me

gtmills (approved):

PR #1517, opened 2026-08-31
1060: Fix too many unspecified connection log entries

baemyung:
The client connections may not always be cleanly done and this may potentially cause too many undesired log entries like
```
Wed Aug 26 05:38:03 2026 p10bmc bmcweb: (2026-08-26 05:38:03) [ERROR "http_connection.hpp":648] 0x1495548 Error while reading: unspecified system error
Wed Aug 26 05:38:43 2026 p10bmc bmcweb: (2026-08-26 05:38:43) [ERROR "http_connection.hpp":648] 0x1b909d0 Error while reading: unspecified system error
Wed Aug 26 05:39:23 2026 p10bmc bmcweb: (2026-08-26 05:39:23) [ERROR "http_connection.hpp":648] 0x1495548 Error while reading: unspecified system error
```

This commit is to shutdown the socket hard in cases where the protocol may be in error, and suppress the unnecessary log entries.

Tested:

- Redfish Service Validator run and passes as it was before

- Run the event listener and disconnect the connection (unexpectedly
  not going thru the graceful close.  Before the fix, this may
  produce the messages like
```
Aug 31 18:39:03 p10bmc bmcweb[891]: (2026-08-31 18:39:03) [ERROR "http_connection.hpp":655] 0x2344ec8 Error while reading: stream truncated
```
  After the fix, this kind of the messages will be suppressed.

### DIFF
--- a/http/http_connection.hpp
+++ b/http/http_connection.hpp
@@ -472,25 +472,43 @@ class Connection :
             return adaptor.is_open();
         }
     }
-    void close()
+
+    void hardClose()
     {
+        BMCWEB_LOG_DEBUG << this << " Closing socket";
+        boost::beast::get_lowest_layer(adaptor).close();
+    }
+
+    void tlsShutdownComplete(const boost::system::error_code& ec)
+    {
+        if (ec)
+        {
+            BMCWEB_LOG_WARNING << this << " Failed to shut down TLS cleanly "
+                               << ec;
+        }
+        hardClose();
+    }
+
+    void gracefulClose()
+    {
+        if (sessionIsFromTransport && userSession != nullptr)
+        {
+            BMCWEB_LOG_DEBUG
+                << this << " Removing TLS session: " << userSession->uniqueId;
+            persistent_data::SessionStore::getInstance().removeSession(
+                userSession);
+        }
+
         if constexpr (std::is_same_v<Adaptor,
                                      boost::beast::ssl_stream<
                                          boost::asio::ip::tcp::socket>>)
         {
-            adaptor.next_layer().close();
-            if (sessionIsFromTransport && userSession != nullptr)
-            {
-                BMCWEB_LOG_DEBUG
-                    << this
-                    << " Removing TLS session: " << userSession->uniqueId;
-                persistent_data::SessionStore::getInstance().removeSession(
-                    userSession);
-            }
+            adaptor.async_shutdown(std::bind_front(
+                &Connection::tlsShutdownComplete, shared_from_this()));
         }
         else
         {
-            adaptor.close();
+            hardClose();
         }
     }
 
@@ -641,38 +659,27 @@ class Connection :
                                        std::size_t bytesTransferred) {
             BMCWEB_LOG_DEBUG << this << " async_read_header "
                              << bytesTransferred << " Bytes";
-            bool errorWhileReading = false;
+
             if (ec)
             {
-                errorWhileReading = true;
-                if (ec == boost::beast::http::error::end_of_stream)
-                {
-                    BMCWEB_LOG_WARNING
-                        << this << " Error while reading: " << ec.message();
-                }
-                else
+                cancelDeadlineTimer();
+
+                if (ec == boost::beast::http::error::header_limit)
                 {
                     BMCWEB_LOG_ERROR
-                        << this << " Error while reading: " << ec.message();
-                }
-            }
-            else
-            {
-                // if the adaptor isn't open anymore, and wasn't handed to a
-                // websocket, treat as an error
-                if (!isAlive() &&
-                    !boost::beast::websocket::is_upgrade(parser->get()))
-                {
-                    errorWhileReading = true;
-                }
-            }
+                        << this
+                        << " Header field too large, closing: " << ec.message();
 
-            cancelDeadlineTimer();
+                    res.result(boost::beast::http::status::
+                                   request_header_fields_too_large);
+                    keepAlive = false;
+                    doWrite(res);
+                    return;
+                }
 
-            if (errorWhileReading)
-            {
-                close();
-                BMCWEB_LOG_DEBUG << this << " from read(1)";
+                BMCWEB_LOG_WARNING << this << " End of stream, closing: " << ec;
+                hardClose();
+
                 return;
             }
 
@@ -685,7 +692,7 @@ class Connection :
                 BMCWEB_LOG_DEBUG << "Unable to get client IP";
                 if (errCode == boost::asio::error::not_connected)
                 {
-                    close();
+                    hardClose();
                     BMCWEB_LOG_ERROR << this << " socket not connected";
                     return;
                 }
@@ -705,7 +712,7 @@ class Connection :
                 {
                     BMCWEB_LOG_DEBUG << "Content length greater than limit "
                                      << *contentLength;
-                    close();
+                    hardClose();
                     return;
                 }
 
@@ -732,7 +739,7 @@ class Connection :
             {
                 BMCWEB_LOG_ERROR << this
                                  << " Error while reading: " << ec.message();
-                close();
+                hardClose();
                 BMCWEB_LOG_DEBUG << this << " from read(1)";
                 return;
             }
@@ -761,7 +768,7 @@ class Connection :
                     ec == boost::asio::ssl::error::stream_truncated)
                 {
                     BMCWEB_LOG_WARNING << this << " end of stream: " << ec;
-                    close();
+                    hardClose();
                     return;
                 }
                 BMCWEB_LOG_DEBUG << this << " from write(2)";
@@ -769,7 +776,7 @@ class Connection :
             }
             if (!keepAlive)
             {
-                close();
+                gracefulClose();
                 BMCWEB_LOG_DEBUG << this << " from write(1)";
                 return;
             }
@@ -841,7 +848,7 @@ class Connection :
 
             BMCWEB_LOG_WARNING << self << "Connection timed out, closing";
 
-            self->close();
+            self->hardClose();
         });
 
         BMCWEB_LOG_DEBUG << this << " timer started";


baemyung:
1060 defect: https://jazz07.rchland.ibm.com:13443/jazz/web/projects/CSSD#action=com.ibm.team.workitem.viewWorkItem&id=810822

gtmills:
How did you test ? This is one since going into a service pack isn't going to get as much testing..

baemyung:
jenkins run tests please

http/http_connection.hpp
```diff
@@ -761,15 +769,15 @@ class Connection :
                     ec == boost::asio::ssl::error::stream_truncated)
                 {
                     BMCWEB_LOG_WARNING << this << " end of stream: " << ec;
-                    close();
+                    hardClose();
                     return;
                 }
                 BMCWEB_LOG_DEBUG << this << " from write(2)";
```
jeaaustx, line 774: Should there be a hardClose() here?

http/http_connection.hpp
```diff
@@ -641,38 +660,27 @@ class Connection :
...
-            else
-            {
-                // if the adaptor isn't open anymore, and wasn't handed to a
-                // websocket, treat as an error
-                if (!isAlive() &&
-                    !boost::beast::websocket::is_upgrade(parser->get()))
-                {
-                    errorWhileReading = true;
+                        << this
+                        << " Header field too large, closing:" << ec.message();
```
jeaaustx: nit: missing space after ':'

http/http_connection.hpp
```diff
@@ -472,25 +472,44 @@ class Connection :
...
+        {
+            BMCWEB_LOG_WARNING << this << " Failed to shut down TLS cleanly "
+                               << ec;
+        }
+        self->hardClose();
+    }
+
+    void gracefulClose()
     {
+        if (sessionIsFromTransport && userSession != nullptr)
```
jeaaustx, line 494: I threw Bob at this, and he came up with this problem:

In gracefulClose(), removeSession() is called synchronously before adaptor.async_shutdown() is invoked. This means there is a window between session revocation and actual TLS channel closure during which a client could re-authenticate using the same transport credentials while the old TLS channel is still technically open.

In the original close(), session removal and socket closure happened in the same synchronous call — the ordering was consistent. The move to async_shutdown introduced this gap.

Suggested fix is to handle the session close separately for the TLS and non-TLS paths. For TLS fold it into tlsShutdownComplete().

http/http_connection.hpp
```diff
@@ -472,25 +472,44 @@ class Connection :
...
-            adaptor.next_layer().close();
-            if (sessionIsFromTransport && userSession != nullptr)
-            {
-                BMCWEB_LOG_DEBUG
-                    << this
-                    << " Removing TLS session: " << userSession->uniqueId;
-                persistent_data::SessionStore::getInstance().removeSession(
-                    userSession);
-            }
+            adaptor.async_shutdown(std::bind_front(
```
jeaaustx, line 506: I threw Bob at this, and he came up with the problem:

std::bind_front with a member-function pointer stores a raw this as the implicit first argument. Even though a shared_ptr is also captured as an explicit argument, the raw this is used directly to dispatch the member function when the completion handler fires. If the Connection object is destroyed before async_shutdown completes (e.g. during io_context teardown or service restart while a TLS shutdown is in flight), the raw this dereference is undefined behaviour / crash.

He suggests using a lambda instead:
adaptor.async_shutdown(
    [self = shared_from_this()](const boost::system::error_code& ec) {
        self->tlsShutdownComplete(self, ec);
    });

jeaaustx (commented):
I don't know this code that well so I asked Bob to also review it. I've added a couple of things he found that I thought might actually be a problem.

http/http_connection.hpp
baemyung: >In gracefulClose(), removeSession() is called synchronously  ...

Bob would assume that the code may be multi-threaded but bmcweb is currently assumed/written as single-threaded.

So, the current bmcweb code can't call them synchronously.

This part of the code also comes from the upstream - https://github.com/openbmc/bmcweb/blob/43d0cb81b162832ada3969f34409ca11d6e7ba26/http/http_connection.hpp#L493

http/http_connection.hpp
baemyung: I have similar experiences with Bob about this kind of bind issue with `this` and `shared_from_this()`, and I've always suspected that the current code style is incorrect as you pointed out.

In upstream code, there are a few places that may be in problem
- https://github.com/openbmc/bmcweb/blob/43d0cb81b162832ada3969f34409ca11d6e7ba26/http/http_connection.hpp#L497
- https://github.com/openbmc/bmcweb/blob/43d0cb81b162832ada3969f34409ca11d6e7ba26/http/websocket_impl.hpp#L323
- ...several others.

I think the simpler fix would be not to pass `this`, as the smart pointer will directly be worked like `this`.

```
    void tlsShutdownComplete(const std::shared_ptr<self_type>& self,
                             const boost::system::error_code& ec)
{
    self->hardClose();
}

            adaptor.async_shutdown(std::bind_front(
                &self_type::tlsShutdownComplete, this, shared_from_this()));
```

==>

```
    void tlsShutdownComplete(const boost::system::error_code& ec)
{
        hardClose();
}


            adaptor.async_shutdown(std::bind_front(
                &self_type::tlsShutdownComplete, tshared_from_this()));
```

http/http_connection.hpp
baemyung: This path is not necessarily caused by the end of stream, but a temporary error (e.g. temporary io layer issue) and may be resumed.
Upstream is also doing so - https://github.com/openbmc/bmcweb/blob/43d0cb81b162832ada3969f34409ca11d6e7ba26/http/http_connection.hpp#L813

http/http_connection.hpp
baemyung: I'll fix this only in this PR, but will pursue the upstream code issues separately later by further investigating/showing the reasonings.

http/http_connection.hpp
baemyung: Done

http/http_connection.hpp
jnin-dev: I think Bob was thinking that since we clear the user session then call async_shutdown() that there's a possibility another request over the same channel where the connection is reused before the TLS stream is cleaned up and was suggesting something like:
```
void tlsShutdownComplete(ec)
{
    if (sessionIsFromTransport && userSession != nullptr)
    {
        BMCWEB_LOG_DEBUG
            << this << " Removing TLS session: " << userSession->uniqueId;
        persistent_data::SessionStore::getInstance().removeSession(
            userSession);
    }
    if (ec)
        {
        BMCWEB_LOG_WARNING << this << " Failed to shut down TLS cleanly "
                           << ec;
    }
    hardClose();
}
```
So that there is no gap between shutdown and session cleanup similar to how it was before?
```
adaptor.next_layer().close();
if (sessionIsFromTransport && userSession != nullptr)
{
    BMCWEB_LOG_DEBUG
        << this
        << " Removing TLS session: " << userSession->uniqueId;
    persistent_data::SessionStore::getInstance().removeSession(
        userSession);
}
```
Although if upstream is using this not sure it's a problem?

baemyung: > ... that there's a possibility another request over the same channel where the connection is reused before the TLS stream is cleaned up 

I'm not sure whether I understand this.
The question is really about the lifetime of the object and pointer. 
As long as they are in-sync, it is okay.
The previous code may have the possibility of the object removal while the pointer is still active.

http/http_connection.hpp
jeaaustx: [Deleted incorrect reply] Sorry, the Bob re-review had a stale file. Got him to update and he likes this fix. Giving you his statement here in case it helps you with upstream resaoning:

std::bind_front(&Connection::tlsShutdownComplete, shared_from_this()) binds shared_from_this() as the implicit object — meaning this is never stored raw anywhere. The shared_ptr is the object, so lifetime is fully managed. This is the correct, idiomatic pattern. H1 and L3 are both cleanly resolved in the current head.

http/http_connection.hpp
jeaaustx: I gave Bob your comments on re-review. He has clarified the issue here is a "logical ordering issue: the session token is revoked before the TLS close_notify handshake completes, which is the reverse of the intended invariant"

Given the short window in execution "this has no practical security or correctness impact".

He still suggest the change for design correctness, but given lack of real impact the change doesn't seem necessary.

http/http_connection.hpp
jeaaustx: Bob still thinks this is a problem:

boost::beast::http::async_write drives a response_serializer to completion by calling the underlying stream's async_write_some in a loop. If any intermediate async_write_some returns an error, Beast surfaces it immediately — but the serializer is left in a partially-consumed, indeterminate state. There is no mechanism in Beast's HTTP layer to checkpoint a partial write and safely resume from the same offset. Resuming doWrite with the same serializer after an error would re-send from an unknown byte position, producing a garbled response on the wire.

More concretely: after the non-EOS error return at line 775, the Connection object is left with:

The socket open but in an unknown state.
The deadline timer already cancelled (line 763) and not re-armed — no new startDeadline() is called on the error path.
No further async_read_header, async_write, or timer posted.
The connection is therefore permanently inert — it will never make forward progress again and will not be collected until the io_context shuts down. This is a leak, not a retry.

The upstream reference establishes precedent but not correctness; the same bug exists there.

jeaaustx: unresolving so this comment will show up

http/http_connection.hpp
```diff
@@ -761,15 +768,15 @@ class Connection :
                     ec == boost::asio::ssl::error::stream_truncated)
                 {
                     BMCWEB_LOG_WARNING << this << " end of stream: " << ec;
-                    close();
+                    hardClose();
                     return;
                 }
                 BMCWEB_LOG_DEBUG << this << " from write(2)";
```
jeaaustx, line 774: [Github is not letting me change the old comment to unresolved. So creating a new one.]

Bob still thinks this is a problem:

boost::beast::http::async_write drives a response_serializer to completion by calling the underlying stream's async_write_some in a loop. If any intermediate async_write_some returns an error, Beast surfaces it immediately — but the serializer is left in a partially-consumed, indeterminate state. There is no mechanism in Beast's HTTP layer to checkpoint a partial write and safely resume from the same offset. Resuming doWrite with the same serializer after an error would re-send from an unknown byte position, producing a garbled response on the wire.

More concretely: after the non-EOS error return at line 775, the Connection object is left with:

The socket open but in an unknown state.
The deadline timer already cancelled (line 763) and not re-armed — no new startDeadline() is called on the error path.
No further async_read_header, async_write, or timer posted.
The connection is therefore permanently inert — it will never make forward progress again and will not be collected until the io_context shuts down. This is a leak, not a retry.

The upstream reference establishes precedent but not correctness; the same bug exists there.

http/http_connection.hpp
baemyung: I think what Bob suggests is that tlsShutdownComplete() is to run first, and then removeSession() inside shutdown-callback, for the graceful sequence point of view.

I think it is unnecessary. 
- Doing removeSession() first is to make sure the tracking of session token is removed.  After this, the session (with such token) will not be reused (as it is basically forgotten from bmcweb point of view).
- Thus anything after (e.g.TLS close_notify) will be ignored.

In this way (current code - here and upstream), the flow can be simplified by removing the session first.

baemyung: Or, 
Perhaps if we want, the correct ordering (shutdown & token erase) may be pursued as a part of upstream.

http/http_connection.hpp
baemyung: As Bob tells, there may be some weakness here and upstream too.
However, this part is beyond of this PR's intention and the code changes.
If needed, a separate effort (upstream) may need to be done to cover those weakness.

jeaaustx (approved):

http/http_connection.hpp
jnin-dev: Although 1060 and upstream are different, I'm fine with leaving it as is since upstream already does this approach and there seems to be no issue found yet.

jnin-dev (approved):

gtmills (approved):

gtmills:
Always risk with backporting something like this.. But seems reasonable to me

